Issue #24, opened 2016-02-02
Add a button to deploy a User's hub modules


paxtonhare:
This is different than the Install hub feature. This is to install the user's custom modules.

This is no backend call for this feature just yet.

paxtonhare:
This depends on UI mockups.

PR #25, opened 2016-02-03
#24 - adding the backend code to support deploying a user's modules

paxtonhare:
- also added domains to the Java API and xquery

### DIFF
--- a/data-hub/src/hub-in-a-box/root/com.marklogic.hub/lib/flow-lib.xqy
+++ b/data-hub/src/hub-in-a-box/root/com.marklogic.hub/lib/flow-lib.xqy
@@ -25,14 +25,18 @@ import module namespace hul = "http://marklogic.com/hub-in-a-box/hub-utils-lib"
 
 declare namespace hub = "http://marklogic.com/hub-in-a-box";
 
+declare namespace envelope = "http://marklogic.com/hub-in-a-box/envelope";
+
 declare option xdmp:mapping "false";
 
+(: xml describing the default writer :)
 declare variable $DEFAULT-WRITER :=
   element hub:writer {
     attribute type { "xquery" },
     attribute module { "/com.marklogic.hub/writers/default.xqy" }
   };
 
+(: xml describing the default content plugin :)
 declare variable $DEFAULT-CONTENT-PLUGIN :=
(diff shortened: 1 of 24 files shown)


data-hub/src/hub-in-a-box/root/com.marklogic.hub/lib/flow-lib.xqy
```diff
@@ -274,7 +409,14 @@ declare function flow:run-flow(
     ()
 };
 
+(:~
+ : Construct an envelope
+ :
+ : @param $map - a map with all the stuff in it
+ : @return - the newly constructed envelope
+ :)
 declare function flow:make-envelope($map as map:map)
```
dmcassel: Envelope is one of your transforms, right? I get that you're providing a built-in one, but would it be helpful to put this in a separate file to illustrate how to build other ones?

data-hub/src/hub-in-a-box/services/metadata/domain.xml
```diff
@@ -0,0 +1,12 @@
+<!-- Copyright 2016 MarkLogic Corporation.  All Rights Reserved. -->
```
dmcassel: Would be cool if (at some point) you had a scriptable test to compare this to what the code provides, to make sure they stay in sync. Might not be easy, though.

data-hub/src/main/java/com/marklogic/hub/flow/SimpleFlow.java
```diff
@@ -26,8 +26,8 @@
  */
 public class SimpleFlow extends AbstractFlow {
 
-    public SimpleFlow(String name) {
-        super(name, "simple");
+    public SimpleFlow(String domainName, String name) {
```
dmcassel: would be better to use "flowName", I think -- also consistent with FlowManagerService.

data-hub/src/hub-in-a-box/root/com.marklogic.hub/lib/flow-lib.xqy
paxtonhare: Actually it's not a built in transform. that concept is so last week :)

we are not allowing the user to override the envelope.

PR #36, opened 2016-02-15
Quick Start Flows tree directory and form validations

maeisabelle:
I implemented the flows directory as a tree (using TreeData and removed DirectoryModel) and added form validations for creating new domains and flows. To follow is the deployment to server and making sure deployed are marked as synched and those are not are marked as not synched.

Kindly review.


quick-start/src/main/java/com/marklogic/hub/web/form/DomainForm.java
```diff
@@ -1,29 +1,58 @@
...
+
+    public String getConformFlowName() {
+        return conformFlowName;
+    }
+
+    public void setConformFlowName(String conformFlowName) {
+        this.conformFlowName = conformFlowName;
+    }
+
+    public void validate(List<DomainModel> domainList) {
```
paxtonhare: See ticket #35 

We actually don't want to require both of these.

maeisabelle:
No need to merge this PR. We can use https://github.com/marklogic/data-hub-in-a-box/pull/37.

PR #37, opened 2016-02-16
Quick Start Input flow and synching

maeisabelle:
Glenn and I already made the change to run the flow and fix the synching of the user modules. Kindly review.

FYI: This is based on a previous PR https://github.com/marklogic/data-hub-in-a-box/pull/36.


data-hub/src/main/java/com/marklogic/hub/Mlcp.java
```diff
@@ -0,0 +1,190 @@
...
+            Thread errorThread = null;
+            try {
+                List<String> arguments = new ArrayList<>();
+                
+                arguments.add(mlcpPath);
+                arguments.add("import");
+                arguments.add("-mode");
+                arguments.add("local");
+                arguments.add("-host");
+                arguments.add("localhost");
```
paxtonhare: Should these values be hard coded? host, port, user, password?

data-hub/src/main/resources/ml-modules/root/com.marklogic.hub/lib/flow-lib.xqy
```diff
@@ -456,12 +458,17 @@ declare function flow:run-plugin(
   $options as map:map)
 {
   let $module-uri := $plugin/@module
-  let $destination := $plugin/@dest
-  let $module-name := hul:get-module-name($module-uri)
-  let $ns := $PLUGIN-NS || fn:lower-case($module-name)
-  let $func := xdmp:function(fn:QName($ns, "create-" || $destination), $module-uri)
   return
-    $func($identifier, $content, $headers, $triples, $options)
+    if (fn:empty($module-uri))
```
paxtonhare: Rather than this fn:empty check and the one above on line 408, I'd rather you check in the caller:

line 396:

```xquery
    xdmp:invoke-function(function() {
      for $plugin in $flow/hub:plugins/hub:plugin[fn:not(hub:type = 'null')]
      let $destination := $plugin/@dest
```

Note the change:
```xquery
[fn:not(hub:type = 'null')]
```

data-hub/src/main/resources/ml-modules/root/com.marklogic.hub/mlcp-flow-transform.xqy
```diff
@@ -0,0 +1,32 @@
...
+
+declare function mlcpFlow:transform(
+  $content as map:map,
+  $context as map:map
+) as map:map*
+{
+  let $uri := map:get($content, "uri")
+  let $_ := xdmp:log('mlcp-flow-transform received: ' || $uri)
+  
+  let $paramNodes := xdmp:unquote(map:get($context, 'transform_param'))/node()/*
```
paxtonhare: This is a lot of work to get the transform_param into a map. If you pass it in as json instead you can do this:

```xquery
xdmp:from-json-string(map:get($context, 'transform_param'))
```

That should give you a map

paxtonhare:
Normally we don't send in 2 separate pull requests with a lot of duplicate changes. The typical flow is something like this:

**Feature 1**
- from master create a feature branch
- make a bunch of changes
- submit pull

**Feature 2**
- from master create a separate feature branch
- make a bunch of other changes
- submit a pull

It looks like you did feature 1 then you created the feature2 branch off of the feature 1 branch.

PR #63, opened 2016-02-25
5 spec for plugins

paxtonhare:


data-hub/src/main/resources/ml-modules/root/com.marklogic.hub/lib/flow-lib.xqy
```diff
@@ -369,7 +383,12 @@ declare function flow:run-collector(
   $options as map:map) as xs:string*
 {
   let $module-name := hul:get-module-name($module-uri)
-  let $ns := $PLUGIN-NS || fn:lower-case($module-name)
+  let $filename as xs:string := hul:get-file-from-uri($module-uri)
+  let $type := flow:get-type($filename)
+  let $ns :=
+    if ($type eq "sjs") then ()
```
dmcassel: You're using enums in other places; make sense to use them here too for consistency and avoiding magic strings? I've seen one case of someone updating the mime-type list so that ".js" would be treated the same as ".sjs".

data-hub/src/main/resources/ml-modules/root/com.marklogic.hub/lib/flow-lib.xqy
```diff
@@ -474,24 +497,41 @@ declare function flow:make-envelope($map as map:map)
...
+    if (fn:empty($module-uri)) then ()
     else
       let $destination := $plugin/@dest
       let $module-name := hul:get-module-name($module-uri)
-      let $ns := $PLUGIN-NS || fn:lower-case($module-name)
-      let $func := xdmp:function(fn:QName($ns, "create-" || $destination), $module-uri)
+      let $filename as xs:string := hul:get-file-from-uri($module-uri)
+      let $type := flow:get-type($filename)
+      let $ns :=
+        if ($type eq "sjs") then ()
```
dmcassel: As noted above -- use an enum here?

quick-start/src/main/webapp/WEB-INF/static/top/modal/flowModal.html
```diff
@@ -8,26 +8,41 @@ <h4 class="modal-title" id="myModalLabel">New Flow</h4>
...
+             <input name="flowType" type="hidden" ng-model="flowForm.flowType"/>
+            <div class="form-group">
+              <label for="flowName">Flow Name</label>
+              <input class="form-control" name="flowName" type="text" ng-model="flowForm.flowName" placeHolder="Enter Flow Name" required="required"/>
+            </div>
+            <div class="form-group">
+              <label for="pluginFormat">Plugin Type</label>
+              <select name="pluginFormat" class="form-control" ng-model="flowForm.pluginFormat">
+                <option value="sjs">Javascript</option>
+                <option value="xqy">XQuery</option>
```
dmcassel: There are multiple extensions that map to XQuery. Better to use mime-types here? Probably not if this is something that users would create manually.

data-hub/src/main/resources/ml-modules/root/com.marklogic.hub/lib/flow-lib.xqy
paxtonhare: flow:get-type() actually uses the registered mime types. I have it returning "sjs" which, to your point, should be an enum. It should probably be:

```xquery
declare variable $TYPE-JAVASCRIPT := "type-javascript";
```

I'll make the change. Thx.

PR #83, opened 2016-03-03
fixed #56 - created a staging and final database

### LINKED ISSUE #56: Create a staging and tranformed database
Right now everything is going into 1 database. We need this to be 2 databases. One for input flow docs. One for the conformed docs.

Input flow => staging db
Conformance Flow => tranformed db

paxtonhare:


data-hub/src/main/java/com/marklogic/hub/DataHub.java
```diff
@@ -55,52 +61,91 @@
...
 import com.marklogic.mgmt.appservers.ServerManager;
+import com.marklogic.mgmt.databases.DatabaseManager;
+import com.marklogic.mgmt.restapis.RestApiManager;
+import com.marklogic.rest.util.Fragment;
 
 public class DataHub {
 
     static final private Logger LOGGER = LoggerFactory.getLogger(DataHub.class);
-
+    static final private String STAGING_NAME = "data-hub-in-a-box-STAGING";
```
dmcassel: Are these default names that can get overridden by configuration files?

quick-start/src/main/resources/application.properties
```diff
@@ -1,10 +1,11 @@
...
-server.port=8080
+server.port=8081
 server.contextPath=/
 spring.thymeleaf.cache=true
 logging.level.com.marklogic.hub: DEBUG
 mlHost.default=localhost
-mlRestPort.default=8010
+mlStagingRestPort.default=8010
+mlFinalRestPort.default=8011
 mlUsername.default=admin
```
dmcassel: Looks like this username is used to configure the hub. If it's also going to be used for application access, are there situations in which staging and final (different app servers) should use different credentials?

data-hub/src/main/java/com/marklogic/hub/DataHub.java
paxtonhare: Not at this time.

quick-start/src/main/resources/application.properties
paxtonhare: We decided to keep it crazy simple. It's Quick Start and thus it's more of a demo of the hub.

data-hub/src/main/java/com/marklogic/hub/DataHub.java
dmcassel: Okay. Probably worth making a note (or issue) to come to that.

PR #90, opened 2016-03-08
72 define search options

paxtonhare:


quick-start/src/main/java/com/marklogic/hub/factory/EntityModelFactory.java
```diff
@@ -95,6 +95,9 @@ public EntityModel createEntity(String entityName, String entityFilePath) {
                 + flowType.toString();
         List<String> flowNames = FileUtil.listDirectFolders(flowsFilePath);
         for (String flowName : flowNames) {
+            if (flowName.equals("REST")) {
```
dmcassel: Not clear from this snippet why you'd skip a REST flow. A comment would help, unless it's clear from broader context.

quick-start/src/main/java/com/marklogic/hub/service/FileSystemWatcherService.java
```diff
@@ -131,28 +131,31 @@ private void registerAll(final Path start) throws IOException {
             public FileVisitResult preVisitDirectory(Path dir, BasicFileAttributes attrs)
                 throws IOException
             {
+                if (dir.endsWith("REST")) {
```
dmcassel: Not clear why REST is skipped from this snippet. A quick comment would be useful here.

PR #107, opened 2016-03-11
79 - Run-flow transform

maeisabelle:
Implemented a run-flow transform that can be invoked from PUT /v1/documents, java code inside a Java container, and REST extension. To verify if the transform is already installed, run http://localhost:8010/LATEST/config/transforms where localhost is the mlHost and 8010 is the port (mlStagingRestPort or mlFinalRestPort). I also updated README.md to guide the users how to use it. 

Let me know if I also need to create a JAVA method that inserts multiple documents using the run-flow transform.

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/transforms/run-flow.xqy
+++ b/marklogic-data-hub/src/main/resources/ml-modules/transforms/run-flow.xqy
@@ -13,9 +13,8 @@ declare function runFlow:transform(
 {
   let $entityName := map:get($params, 'entity-name')
   let $flowName := map:get($params, 'flow-name')
-  let $flowType := map:get($params, 'flow-type')
-
-  let $flow := flow:get-flow($entityName,$flowName,$flowType)
+  
+  let $flow := flow:get-flow($entityName,$flowName,())
   
   let $uri := map:get($context, 'uri')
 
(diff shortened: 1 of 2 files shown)


marklogic-data-hub/src/main/resources/ml-modules/transforms/run-flow.xqy
```diff
@@ -0,0 +1,26 @@
...
+
+declare function runFlow:transform(
+  $context as map:map,
+  $params as map:map,
+  $content as document-node()
+  ) as document-node()
+{
+  let $entityName := map:get($params, 'entity-name')
+  let $flowName := map:get($params, 'flow-name')
+  let $flowType := map:get($params, 'flow-type')
```
paxtonhare: this transform is only for input flows. There is no need for a flow-type param.

marklogic-data-hub/src/main/resources/ml-modules/transforms/run-flow.xqy
```diff
@@ -0,0 +1,25 @@
...
+declare function runFlow:transform(
+  $context as map:map,
+  $params as map:map,
+  $content as document-node()
+  ) as document-node()
+{
+  let $entityName := map:get($params, 'entity-name')
+  let $flowName := map:get($params, 'flow-name')
+  
+  let $flow := flow:get-flow($entityName,$flowName,())
```
paxtonhare: close. you still need to pass in input here. Just don't make it a user supplied param.

PR #110, opened 2016-03-11
100 new flow and new entity button fix

gmarintes:


paxtonhare:
did you happen to run `./gradlew jshint`

I'm thinking not. Run that and see if there are any errors to fix.

PR #132, opened 2016-03-17
fixed #91 - check plugins during install for errors

### LINKED ISSUE #91: Detect (vet) bad user plugins
Similar to how app services detects when a transform is bad, we should static check user plugins before or during installing.

paxtonhare:


marklogic-data-hub/src/main/resources/ml-modules/root/com.marklogic.hub/lib/flow-lib.xqy
```diff
@@ -574,3 +574,69 @@ declare function flow:run-writer(
...
+  ))
+};
+
+declare function flow:validate-entities()
+{
+  let $errors := json:array()
+  let $identifier := "123"
+  let $content :=
+    map:new((
+      map:entry("identifier", $identifier),
```
dmcassel: doesn't look like the identifier entry gets used. Remove?

PR #134, opened 2016-03-18
102 - Scaffolding for Rest

maeisabelle:
Created scaffolding for rest service extensions and transforms. Kindly review.


marklogic-data-hub/src/main/resources/scaffolding/rest/services/sjs/template.sjs
```diff
@@ -0,0 +1,20 @@
+function get(context, params) {
+  return null;
+};
```
paxtonhare: You don't need semicolons at the end of these functions.

PR #141, opened 2016-03-21
fixed #140 - vet plugins not working correctly

### LINKED ISSUE #140: Vet plugins isn't quite right
It should static-check vs running

paxtonhare:


marklogic-data-hub/src/main/resources/ml-modules/root/com.marklogic.hub/lib/flow-lib.xqy
```diff
@@ -612,16 +612,32 @@ declare function flow:validate-entities()
...
         try {
-          flow:run-plugin(
-            $plugin,
-            $data-format,
-            "123",
-            (),
-            (),
-            (),
-            $options)
+          if ($type eq $flow:TYPE-XQUERY) then
```
dmcassel: I think this section needs a comment or two for the developer who looks at this somewhere down the road. State what kind of errors are being looked for, note that the JS check won't find much, and mention that the open nature of the input prevents more interesting checks.

PR #147, opened 2016-03-22
101 - Handle detection of REST directory

maeisabelle:
Detection if modules are synched are now done solely in the factories of entity, flow and rest models using SyncStatusService. Kindly review.


paxtonhare:
I just tested this out.

It looks like the file watcher isn't picking up the changes. If I manually refresh it will tell me that files need synching, but not automatically.

maeisabelle:
This is what I found out on my end. The synch status is working on my end when adding a new REST or flow then deploying it to the server. But deleting it (after it was deployed) won’t. Is that what happened to you? When you also said you made changes, is it changes on the actual file or addition of a xqy / sjs file?

maeisabelle:
I made changes in the watcher to not skip the REST directory and also removed the unnecessary getEntities call on onWatchEvent so the synch status will immediately reflect. The only issue I'm seeing is when we remove a directory/file as it is not tracked. Deploying to the server also doesn't deleted remove directories and files. Let me know if we still need to handle this.

PR #167, opened 2016-03-30
Add hr hub

paxtonhare:


examples/hr-hub/plugins/entities/Employee/input/load-global-corp/content/content.sjs
```diff
@@ -0,0 +1,18 @@
+/*
+ * Create Content Plugin
+ *
+ * @param id       - the identifier returned by the collector
+ * @param content  - your final content
```
dmcassel: isn't the parameter the inbound content, which may be changed by this function?

Issue #200, opened 2016-04-14
Create a non-admin user for doing hub stuff


paxtonhare:
This task is to create a non-admin user whenever the hub gets installed.

Right now we ask for an admin user/password to install the hub. We should also be allowing the user to specify a non-admin user/password.

Let's do it like this:
- Change up quickstart
  We should have 2 users:
  1. **admin user**
     mlAdminUsername
     mlAdminPassword
  2. **non-admin user**
     mlUsername
     mlPassword
- Don't ask for the non-admin stuff via the Login screen. We will do that later via a separate ticket if necessary.
- If no mlUsername and mlPassword are provided, then default them to the same as mlAdminUsername and mlAdminPassword. This way an advanced user can edit the environment.properties file with the non-admin user if they want to. But by default we just use the admin user.
- Modify the installUserModules() code from DataHub to use the non-admin credentials.
- Modify the examples/gradle-advanced example to also support non-admin users. This should be super simple. See https://github.com/rjrudin/ml-gradle/blob/master/examples/sample-project/gradle.properties for how Rob is adding the different users in the build.properties file.

maeisabelle:
Hi @paxtonhare,

How do you want to be added? Will this be a static username and password defined in the properties file just like the admin username and password or would you want it specified in the login/config page?

paxtonhare:
It would be for the advanced gradle approach only. So in build.gradle and
properties.

maeisabelle:
Hi @paxtonhare,

When you said 

> It would be for the advanced gradle approach only. 

Does it mean, it will not be part of the Datahub.install? Thus, it will be in the discretion of the user if he/she wants to create it? If yes, is it a gradle task that the user can run?

maeisabelle:
Hi @paxtonhare,

Kindly review my pull request above that implements the create user as a Gradle task. Let me know if that's what is needed for this issue. If no, please further explain the requirement. Thank you! :)

maeisabelle:
As discussed last night, we will also add the input fields for the non-admin username and password.

maeisabelle:
Hi @paxtonhare,

Kindly review my pull request if it already addresses the requirement of this ticket. I also noticed that the minimum requirement of the user account to use is to have the role 'rest-admin', 'xa-admin' and 'manage-admin-internal'. Would you want this specified in the README.md for quick-start as well? It took me a lot of time to figure that out and the developers using quick start may benefit if we have that one in the README.md.

PR #209, opened 2016-04-18
Non-admin user for doing hub stuff

maeisabelle:
I already implemented a Gradle task (gradle createHubUser) that creates a hub user (hubuser/hubpassword). Let me know if this already addresses #200.


paxtonhare:
this PR isn't what I was looking for. I'll go update the ticket to better reflect what is needed.

PR #211, opened 2016-04-20
200 - Use non-admin users for other hub operations

maeisabelle:
Kindly review if I already addressed all use cases for #200.


paxtonhare:
@maeisabelle I'd like to see a unit test where we create a non-admin user and make sure that loading modules works w/o being admin.

a typical non-admin user should have the rest-reader and rest-writer roles.

I don't care how you create the non-admin users since it's just for testing.

maeisabelle:
Hi @paxtonhare,

I already implemented the unit test for installing user modules. Note that for creating users, I encountered several errors when using a user with only rest-writer and/or rest-reader as roles. What I did instead is to create a user with the following roles: ('rest-admin','xa-admin','manage-admin-internal') and permission (xdmp:permission('rest-admin', 'update')). Doing the former allows me to get the entities, deploy and load data without error. Doing the later allowed me to run flow in the Harmonize flows. 

Kindly review.

marklogic-data-hub/src/main/java/com/marklogic/hub/DataHubUser.java
```diff
@@ -0,0 +1,114 @@
+package com.marklogic.hub;
```
paxtonhare: It looks like this file is only used by the unit tests. I'd rather it not be part of the publicly exposed Hub Jar since it's only purpose is to support a test case. Also, it might confuse Hub users into thinking it's a class they can/should use.

marklogic-data-hub/src/main/java/com/marklogic/hub/DefaultHubConfigFactory.java
```diff
@@ -34,6 +34,9 @@ public HubConfig newHubConfig() {
         String prop = null;
         String mlUsername = getProperty("mlUsername");
         String mlPassword = getProperty("mlPassword");
+
```
paxtonhare: All of the other properties do something like:

```java
prop = getProperty("...");
if (prop != null) {
  c.whatever = prop;
}
```

I would rather this be consistent with the other code.

marklogic-data-hub/src/main/java/com/marklogic/hub/HubConfig.java
```diff
@@ -69,6 +75,9 @@
     public String authMethod = DEFAULT_AUTH_METHOD;
 
     public String modulesPath;
+    
+    public String hubUsername;
```
paxtonhare: Can you explain the difference between:

```java
public String username = DEFAULT_USERNAME;
public String password = DEFAULT_PASSWORD;
```

and

```java
public String hubUsername;
public String hubPassword;
```

Thx.

marklogic-data-hub/src/test/java/com/marklogic/hub/DataHubTest.java
```diff
@@ -42,4 +46,14 @@ public void testValidateInvalidServer() throws ServerValidationException {
         exception.expect(ServerValidationException.class);
         dh.validateServer();
     }
+    
+    @Test
```
paxtonhare: This test looks like it doesn't verify that the modules actually get installed. Ideally I'd like to see you: 
- install the user modules
- run a flow
- verify that the flow ran by checking for staged or conformed docs

Then you can be fairly certain that the install worked.

quick-start/src/main/java/com/marklogic/hub/web/form/LoginForm.java
```diff
@@ -13,6 +13,8 @@
 	private String mlTracePort;
 	private String mlUsername;
 	private String mlPassword;
+	private String mlAdminUsername;
+    private String mlAdminPassword;
```
paxtonhare: The whitespace seems to be off in a bunch of places. This ^ is one of them. Please take a look at whitespace to make sure your .editorconfig is working and that you are using 4 spaces in Java files.

marklogic-data-hub/src/main/java/com/marklogic/hub/HubConfig.java
maeisabelle: That's a duplicate. I'm modifying the code accordingly.

PR #213, opened 2016-04-21
fixed #197

### LINKED ISSUE #197: Refactor to Auto deploy code
Rather than updating the UI every time a file changes we want to start automatically deploying code changes.

- Update the FileWatcher to automatically deploy user modules when a file changes.
- Remove all the code that updates the UI with file change status
- Clean up the UI js code that updates the Icon state.
- Basically, clean up the code as much as possible
- Keep the UI button to deploy modules. But change it to always deploy when pressed. This means we should clear out the timestamp file (there is a function for that) and then deploy.

_After discussion with Damon (see below) I'm adding a few extras_
- when running flows we must wait until any current load is finished. could be a thread wait. could be whatever. just so long as when I poke the run flow button it won't run until after deploy is finished.
- For the Deploy modules button, also make it clear out the modules first. This means we need an undeployUserModules() method

divino:
fixed #197


divino:
fixed #197

divino:
fixed #213

PR #220, opened 2016-04-25
219 - Fix Bugs in Load Data Feature

### LINKED ISSUE #219: Fix bugs in load data feature
1. No validation for empty input file path or default input file path causing NullPointerException when loading data if it is not set
2. Other options in the groups are not passed in mlcp during loading (as well as during saving)

maeisabelle:
This will fix #219 
- Fixed issue in saving options and loading of previous options
- Also added validation for missing input file path

Kindly review.


quick-start/src/main/resources/static/app/mlcp/mlcp.groups.js
```diff
@@ -237,6 +237,16 @@
             ]
           }
         ];
+        angular.forEach(previousOptions, function(value, key) {
+          $.each(groups, function(i, group) {
```
paxtonhare: please use angular.forEach... just in case we remove jQuery later.

quick-start/src/main/resources/static/app/mlcp/mlcp.controller.js
```diff
@@ -157,7 +177,9 @@
     }
 
     $scope.$watch('mlcp.input_file_path', function(value) {
-      searchPath(value);
+      if(isValidInputFilePath(value)) {
```
paxtonhare: The first time you open this dialog ever you get the error right off the bat. We need to set a default value for input_file_path or something.

Issue #193, opened 2016-04-08
Performance tracing


damonfeldman:
We need a perf trace of every action in both marklogic and tomcat. This will tell us how long each task is taking, how much of that time is in tomcat, how much in ml.

To do this we need elapsed-time logging every time we cross a boundary 
- from the browser to QuickStart
- from QuickStart to the code in the core DataHub libraries 
- from QuickStart or core DataHub code to MarkLogic

That is, every REST call and also every call between QuickStart and the hub library will be logged.

Possible tech for each area:
- QuickStart may be able to use some Spring interceptor or annotation to log all reqeusts.
- core Data Hub library
   - get System.currentTimeMillis() before and after the operation, and log the difference to compute elapsed time.
- MarkLogic
   - have a standard function to compute elapsed time and log. (See code below in a comment)

The point of this is that when something takes a long time (pushing code, running a flow, etc) we will know which portion took how much time and can optimize as needed.

damonfeldman:
Here is code for a utility function to wrap XQuery code in some logging:
```
declare function perflog:logit($what-is-it, $x)
{
 let $before := xdmp:elapsed-time()
 let $resp := $x()
 let $time := xdmp:elapsed-time() - $before
 let $_ := xdmp:log($what-is-it || " took " || $time)
 return
  $resp
};
```

To call this, import the common module and invoke like so:
```
perflog:logit("my code that does something", function() {
  let $x := 5
  return "hello"
   (: your stuff here :)
 })
 ```

maeisabelle:
Hi @damonfeldman and @paxtonhare,

I think we cannot get the elapsed time from the browser to QuickStart as one resides in the client side while the other resides in the server side (also we are not guaranteed that the timestamp in the client side is the same as the server side). I also think it will just complicate things if we add it. What we can get is the total elapsed time when a post/get request is triggered in the browser and the result is returned.

For getting the elapsed time for the quick start to the code in Datahub libraries, I'm thinking of using AOP in spring (as doing the logging manually will make the codes too coupled). I'm still figuring out how it can be done but I need to confirm first if I can assume that we will only support one user using the application. If yes, I can declare and use an instance variable in the LoggingService (the Aspect) to track the timestamp when the user first enters QuickStart then compute the difference when s/he enters the data hub core libraries. Knowing where the code is actually at will be done via the annotation  @Before with the execution of the package of QuickStart API controller and the Data hub core libraries. 

[...]

PR #224, opened 2016-05-02
193 - Performance trace

maeisabelle:
Added performance trace in quick-start (using aop), DataHub core libraries called in quick start (using a utility), and xquery files inside ml-modules/(services|transforms) (using a module). Kindly review. This is to address #193.


marklogic-data-hub/src/main/resources/ml-modules/root/com.marklogic.hub/lib/perflog-lib.xqy
```diff
@@ -0,0 +1,33 @@
...
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+:)
+xquery version "1.0-ml";
+
+module namespace perflog = "http://marklogic.com/data-hub/perflog-lib";
+
+(:~
```
paxtonhare: Your comment doesn't quite match up to the function

marklogic-data-hub/src/main/resources/ml-modules/root/com.marklogic.hub/lib/perflog-lib.xqy
```diff
@@ -0,0 +1,33 @@
...
+ : determine if a module exists or not
+ :
+ : @param $uri - the uri of the module to check
+ : @return - true if exists, false otherwise
+ :)
+declare function perflog:logit($what-is-it, $x)
```
paxtonhare: You weren't on the call when I talked about this code. I mentioned that it was an example and that you all shouldn't use the exact names as I wrote it in about 45 seconds. I'd like to see the variable names make a little more sense.

Also I'd prefer to see some sort of greppable thing in the log message. Maybe something like:

xdmp:log("PERFORMANCE: " || $what-is-it || " took " || $time)

You should also look at the log file output to make sure it makes sense. I'm pretty sure $time is something like "PT0.00001S" and doesn't need the ms at the end.

marklogic-data-hub/src/main/resources/ml-modules/services/collector.xqy
```diff
@@ -31,9 +34,11 @@ declare option xdmp:mapping "false";
 declare function service:build-response($items as item()*)
   as document-node()
 {
-  document {
-    json:to-array($items) ! xdmp:to-json(.)
-  }
+  perflog:logit('Collector.buildResponse',function() {
```
paxtonhare: I think wrapping this function in the perflog is a bit overkill. It's already covered by the functions below. The idea is to get a sense of how long a transaction is taking. Thus, I'd prefer to see 1 log per transaction.

marklogic-data-hub/src/main/resources/ml-modules/services/collector.xqy
```diff
@@ -46,12 +51,14 @@ declare function post(
   $input   as document-node()*
   ) as document-node()*
 {
-  debug:dump-env(),
+  perflog:logit('Collector.post',function() {
```
paxtonhare: I'd like to see a better name logged. at the minimum use the : vs the .

perflog:logit('collector:post')

but maybe

perflog:logit('GET => /v1/resources/collector')

I'm not sure what works best. I just want it to be pretty easy to track down the offending code if I find a performance issue.

marklogic-data-hub/src/main/resources/ml-modules/services/debug.xqy
```diff
@@ -19,6 +19,9 @@ module namespace service = "http://marklogic.com/rest-api/resource/debug";
 
 import module namespace debug = "http://marklogic.com/data-hub/debug-lib"
   at "/com.marklogic.hub/lib/debug-lib.xqy";
+
```
paxtonhare: I don't think it makes sense to perf-log this file. It's so basic. Please back this file out of the PR.

marklogic-data-hub/src/main/resources/ml-modules/services/tracing.xqy
```diff
@@ -23,6 +23,9 @@ import module namespace debug = "http://marklogic.com/data-hub/debug-lib"
 import module namespace trace = "http://marklogic.com/data-hub/trace"
   at "/com.marklogic.hub/lib/trace-lib.xqy";
 
+import module namespace perflog = "http://marklogic.com/data-hub/perflog-lib"
```
paxtonhare: This module doesn't need performance tracing either. Please back it out of the PR.

quick-start/build.gradle
```diff
@@ -42,6 +42,7 @@ dependencies {
         exclude group: 'com.google.guava', module: 'guava'
     }
     compile 'com.google.guava:guava:19.0'
+    compile 'org.springframework.boot:spring-boot-starter-aop:1.3.3.RELEASE'
```
paxtonhare: Any clue how much larger this makes the quick-start.jar?

maeisabelle: Hi @paxtonhare,

I compared the jar file created in master and in my branch, from 67.134229 MB, it became 69.004115 MB (so almost 2 MB). 

By the way, I noticed you merged my branch to your branch 'develop' and saw your changes. Would I still continue with the changes you requested?

paxtonhare:
I manually merged this.

PR #457, opened 2017-07-20
Collection Name as input option for harmonize flows

bhandaru:
Please review. The features implemented in this commit are as below.

1) Modified the Harmonize Flow Input UI to accept an additional parameter(collectionName) to be passed to backend harmonize flow.

2) Modified the MLCPRunner.java to write a .done file in the input_file_path location when it has successfully completed the input flow. This helps in subscribing to the input flow complete event for a client application to trigger harmonize flow upon the completion of the input flow. Helps in tying the triggering of input flow and harmonize flow together.

### DIFF
--- a/quick-start/src/main/java/com/marklogic/quickstart/service/FlowManagerService.java
+++ b/quick-start/src/main/java/com/marklogic/quickstart/service/FlowManagerService.java
@@ -38,6 +38,7 @@ import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.HashMap;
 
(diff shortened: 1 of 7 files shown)


paxtonhare:
@bhandaru Rather than add a one-off option to the Quickstart UI, I'd rather just expose the ability to pass in an arbitrary map of key/value pairs much like we did for gradle: https://github.com/marklogic-community/marklogic-data-hub/issues/416

I think we should generate 2 issues from this:
1. Add the ability to pass key/value pairs to Harmonize or MLCP flows via Quickstart
2. Add the ability to remember Run Flow options in the QuickStart UI.

PR #559, opened 2017-10-31
Issue413

tcfenstermaker:
Basic functionality of displaying files in a selected directory, and being able to select one of those files and have it set as the input_file_path of MLCP. The URI replacement is still set to just the path name, minus the file name. The current folder display is set to just the folder, and does not display the file name. 

However, on a redisplay of the Flow, it will display the filename with its path in the Current Folder display. It will also not show the peer files in the file list, and instead just display the .. relative path as the only available option. It is also possible type in the file name in the Current Folder display, which works in setting the MLCP input_file_path, but is inconsistent with what I have it behaving as now.

The event model, and the relationship between the path displayed and the input_file_path, make this more complex than it seems at first. (I can go into detail elsewhere.) So before I "fix" these things, however, I'd like to discuss more what desired behavior we'd like. Do we allow multiple files? If so, how does that look? Do we want a separate display of the file name? Etc.

Meanwhile I believe this pull request contains the basic functionality requested, so I think it serves as a start at least for a candidate to be merged into the current development baseline.

### DIFF
--- a/quick-start/src/main/java/com/marklogic/quickstart/model/SearchPathModel.java
+++ b/quick-start/src/main/java/com/marklogic/quickstart/model/SearchPathModel.java
@@ -23,8 +23,8 @@ public class SearchPathModel {
 
 	public SearchPathModel(String name, String relativePath, String absolutePath) {
         this.name = name;
-		this.relativePath = relativePath;
-		this.absolutePath = absolutePath;
+        this.relativePath = relativePath;
+        this.absolutePath = absolutePath;
 	}
 
 }
--- a/quick-start/src/main/java/com/marklogic/quickstart/model/SearchPathModel.java
+++ b/quick-start/src/main/java/com/marklogic/quickstart/model/SearchPathModel.java
@@ -22,8 +22,9 @@ public class SearchPathModel {
 	public String absolutePath;
 
 	public SearchPathModel(String name, String relativePath, String absolutePath) {
+        this.name = name;
 		this.relativePath = relativePath;
 		this.absolutePath = absolutePath;
-		this.name = name;
 	}
+
 }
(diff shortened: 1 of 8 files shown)


quick-start/src/main/java/com/marklogic/quickstart/model/SearchPathModel.java
```diff
@@ -22,8 +22,9 @@
 	public String absolutePath;
 
 	public SearchPathModel(String name, String relativePath, String absolutePath) {
+        this.name = name;
```
paxtonhare: I hate to nitpick... but are you using editorconfig to get your spacing right?

quick-start/src/main/resources/application.properties
```diff
@@ -3,6 +3,7 @@ server.port=${port:8080}
 server.contextPath=/
 # server.error.whitelabel.enabled=false
 logging.level.com.marklogic.hub=INFO
+logging.level.com.marklogic.quickstart.web=DEBUG
```
paxtonhare: Can you take this out?

paxtonhare:
@tcfenstermaker this looks awesome. I only had 2 comments about the code.

tcfenstermaker:
I am not sure about .editorconfig, as I do have that installed and configured in intelliJ; I'm looking now to see if it's applying in this project. I will remove the debug setting in the properties file.

quick-start/src/main/java/com/marklogic/quickstart/model/SearchPathModel.java
tcfenstermaker: No idea why this is appearing this way here. It looks correct in my editor. Editorconfig is installed and applied by default to the entire project, and the project's checked-in .editorconfig is available, which sets the indent on Java files to 4 spaces. Perhaps this is just a formatting issue with the pull request viewer.

tcfenstermaker:
@paxtonhare Not sure what was up with the formatting in that `SearchPathModel.java` file; the existing two lines had tabs. So I fixed all three to be four-space indents. I also removed the debug setting from `application.properties`. Take one more look.

PR #598, opened 2017-12-01
Feature/bug 558

dmcassel:


quick-start/src/main/java/com/marklogic/quickstart/service/EntityManagerService.java
```diff
@@ -156,18 +161,50 @@ public EntityModel createEntity(String projectDir, EntityModel newEntity) throws
...
+                if (!origDirectory.renameTo(newDirectory)) {
+                    throw new IOException("Unable to rename " + origDirectory.getAbsolutePath() + " to " +
+                        newDirectory.getAbsolutePath());
+                }
+
+                fullpath = newDirectory.getAbsolutePath() + File.separator + title + ENTITY_FILE_EXTENSION;
+                entity.setFilename(fullpath);
+
+                // Redeploy the flows
+                dataHubService.reinstallUserModules(envConfig().getMlSettings(), this, this);
```
paxtonhare: can't you pass null, null instead of this, this?

dmcassel: I'll check and change, assuming that works. (Seems like it should, just didn't think of it)

PR #603, opened 2017-12-05
Feature/delete hub

dmcassel:
On running a unit test that didn't clean up after itself, running "gradlew mlUndeploy" would fail, but 1) unit tests could clean up after themselves, and 2) "gradlew mlDeploy" followed by "gradlew mlUndeploy" worked fine. This turned out to be a difference in configuration. marklogic-data-hub/gradle.properties used "data-hub-JOB" as the name for the app serve and the database, but everywhere else they were "data-hub-JOBS".


paxtonhare:
I bet this one was fun to track down.

PR #627, opened 2017-12-15
Fixed #616

### LINKED ISSUE #616: deleting an entity property causes quickstart to forget the existing primary key/range index/required field settings
#### The issue

Deleting an entity property causes quickstart to forget the existing primary key/range index/required field settings. Create a new entity with 2 properties, make the first one a primary key, range index and required field. save. edit the entity again and delete the second property. All the settings for the first  (all other properties if you had them) disappear.

What behavior are you expecting?

Remove properties from entity definitions without any changes to the other properties. 

#### Tech details

Which Operating System are you using?

windows 7

Which version of MarkLogic are you using?  9.0-3.1

Which version of the Data Hub Framework are you using? 2.0.2

Is this a QuickStart UI bug? If so, which browser are you using? dunno, chrome Version 61.0.3163.100

aebadirad:
Added fix for #616 so that only the selected properties are wiped of their index/pkey/etc settings.

Also included E2E tests for properties and a check for this condition, plus tweaked settings to get the E2E settings to run.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -289,6 +289,7 @@ $ ./gradlew -Dtest.single=FlowRunner test
 
 If you want to run just the Quick-Start UI End to End tests, you will need nodejs 8.9.1 or later installed:
 ```jshelllanguage
+gradlew bootrun
 cd quick-start
 npm install
 npm install -g protractor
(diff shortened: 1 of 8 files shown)


aebadirad:
I've made the E2E tests for properties more reliable by replacing the coordinate click based system with a lower level implementation that leverages the browser driver and raw window functions to obtain the necessary element to click on, rather than from inside protractor/selenium

PR #645, opened 2018-01-12
Issue 503

RobertSzkutak:
Allows transform_param to become usable and no longer read-only.

Allows you to set a jobId in transform_param.


RobertSzkutak:
@aebadirad Could you be more specific about the testing criteria? Thanks! We can touch base offline sometime if you think it would be easier to discuss.

aebadirad:
I'm going to pull this in without test, because we do not (and shouldn't) test MLCP and this is a UI pass-through to MLCP

PR #646, opened 2018-01-12
#591 - Show 2.0 features for server versions >= 9

### LINKED ISSUE #591: Quickstart runs in 2.0 mode only for version 9, not > 9.x
If you start up the quickstart with a version 10 MarkLogic server, it does not use entity services as it should.

@paxtonhare  has found the location of the bug:

*ngIf="getMarkLogicVersion().startsWith('9')"

I expect the 9.0 interface to entities.

#### Tech details

Windows

Which version of MarkLogic are you using?

10 nightly

Which version of the Data Hub Framework are you using?

2.0.0 

Is this a QuickStart UI bug? If so, which browser are you using?

yes, doesn't matter though.

You're a genius you don't need details

wooldridge:
This includes "Entities" link in header and "Code Generation" section in Create Flow dialog.
Check for version !== 8 rather than == 9.

Fixes #591

### DIFF
--- a/quick-start/src/main/ui/app/header/header.component.html
+++ b/quick-start/src/main/ui/app/header/header.component.html
@@ -12,7 +12,7 @@
     <div class="mdl-layout-spacer"></div>
     <nav class="mdl-navigation mdl-layout--large-screen-only">
       <a id="database-tab" class="mdl-navigation__link" [ngClass]="{'active': isActive('/')}" [routerLink]="['']"><i class="fa fa-dashboard"></i> Dashboard</a>
-      <a id="entities-tab" class="mdl-navigation__link" *ngIf="!getMarkLogicVersion().startsWith('8')" [ngClass]="{'active': isActive('/entities')}" [routerLink]="['entities']"><i class="fa fa-industry"></i> Entities</a>
+      <a id="entities-tab" class="mdl-navigation__link" *ngIf="getMarkLogicVersion() >= 9" [ngClass]="{'active': isActive('/entities')}" [routerLink]="['entities']"><i class="fa fa-industry"></i> Entities</a>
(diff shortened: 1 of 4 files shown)


popzip:
>=9 (ie not 7)

quick-start/src/main/ui/app/header/header.component.html
```diff
@@ -12,7 +12,7 @@
     <div class="mdl-layout-spacer"></div>
     <nav class="mdl-navigation mdl-layout--large-screen-only">
       <a id="database-tab" class="mdl-navigation__link" [ngClass]="{'active': isActive('/')}" [routerLink]="['']"><i class="fa fa-dashboard"></i> Dashboard</a>
-      <a id="entities-tab" class="mdl-navigation__link" *ngIf="getMarkLogicVersion().startsWith('9')" [ngClass]="{'active': isActive('/entities')}" [routerLink]="['entities']"><i class="fa fa-industry"></i> Entities</a>
+      <a id="entities-tab" class="mdl-navigation__link" *ngIf="!getMarkLogicVersion().startsWith('8')" [ngClass]="{'active': isActive('/entities')}" [routerLink]="['entities']"><i class="fa fa-industry"></i> Entities</a>
```
aebadirad: I agree with @popzip that we should be >= to 9, not just looking for 'not 8' for completeness of version checking.

wooldridge: @aebadirad, can you tell me what ML version formats I can expect from the /api/current-project endpoint? I know that value looks like "10.0-20180115" on my local build but I'm not sure how/if it varies with other distributions.

grechaw: i think to get the major version you can always get the value before the first period and parse and an integer.

wooldridge: @aebadirad I addressed the review comments with https://github.com/marklogic-community/marklogic-data-hub/pull/646/commits/05407a3facd4a0cc80a75b0f973ba71469327656. Can we merge this?

PR #649, opened 2018-01-12
Issue 575

RobertSzkutak:
Allows input_file_path to be editable so that it may be modified to be relative.


RobertSzkutak:
@aebadirad When you say "bad" relative path are you referring to illegal filepath characters or a path that doesn't exist on the local filesystem? If the case of the latter, could you provide a reference on how that could be tested?

Issue #504, opened 2017-09-08
Allow specifying flow options for harmonization flows run from quickstart


damonfeldman:
Flows have the ability to take additional parameters, which are included in the transform_param REST parameter when the flow is called. 

This task is to allow the harmonization process to specify additional parameters which will then be included in the options map in a harmonization flow, and may used by plugin logic.

┆Status: To Do

popzip:
whoever is working on this, please touch base with @aajacobs on status and whether doing on a feature branch might be easier for demo-ing at SKO in a few weeks.

PR #654, opened 2018-01-15
#504 send options along with harmonization flows

dmcassel:


quick-start/src/test/java/com/marklogic/quickstart/service/FlowManagerServiceTest.java
```diff
@@ -155,4 +172,99 @@ public void runMlcp() throws IOException, InterruptedException, JSONException {
...
+                if (percentComplete == 100)
+                {
+                    synchronized (monitor) {
+                        monitor.notify();
+                    }
+                }
+            }
+        });
+
+        // Wait for the flow to finish
```
aebadirad: I really like the sync block here for waiting for it to complete. +1

grechaw:
Hi @dmcassel , Mike had told me that this PR implemented the passing of target database... I don't think it does though.  This PR pushes options down into the headers of the harmonized document, but the code that chooses target database doesn't look in the options to do anything.  Is there some oversight?  I'm trying to help Mike connect the dots from the UI to the services, and to understand the codebase better myself.

dmcassel:
@grechaw it doesn't ... it allows users to pass along arbitrary options that will be used in a harmonization flow. The framework itself pays no attention to them. That was the goal of #504.

grechaw:
Thanks Dave for that confirmation.  I think all the folks are on the same page about this request now

PR #689, opened 2018-01-28
Allow key-value options to be passed in for harmonization flows in UI

### LINKED ISSUE #684: QuickStart harmonize flow settings not persisted during browser session
In the harmonize flow settings (which includes batch size and thread count), the values are not persisted when I change the values, leave that page view, and then return. The settings go back to the defaults.

To reproduce:

1. In the Flows view, create a harmonize flow ("harmonize 1") with defaults (batch size = 100, thread count = 4).
2. Change the settings (batch size = 50, thread count = 2).
3. Click "Dashboard" to change the page view, then click "Flows" to return to the Flows view.
4. Select the flow you created in step 1. 
5. The flow settings are back to their defaults (batch size = 100, thread count = 4).

Furthermore, if I have multiple harmonize flows defined, I can't define separate flow settings for the different harmonize flows. Only one set of settings is maintained.

To reproduce:

1. In the Flows view, create two harmonize flows ("harmonize 1", "harmonize 2") with defaults (batch size = 100, thread count = 4).
2. Change the settings for one of the flows, "harmonize 1" (batch size = 50, thread count = 2).
3. Click "harmonize 2" to switch to the other flow.
4. The second flow has the non-default settings  (batch size = 50, thread count = 2).
5. Manually change the settings for the second flow (batch size = 80, thread count = 6).
6. Click "harmonize 1" to switch back to the first flow. The flow settings remain the same (batch size = 80, thread count = 6), they don't return to what we defined in step 2.

I would expect the settings associated with a harmonize flow to be independent of the other flows and to persist as I use other parts of the application.

#### Tech details

Which Operating System are you using? Mac OS

Which version of MarkLogic are you using? trunk

Which version of the Data Hub Framework are you using? develop branch

Is this a QuickStart UI bug? If so, which browser are you using? Chrome

wooldridge:
The features allows the user to add a configurable number of key-value options.
The key-value list is built as a configurable Angular component. 
Harmonization flow options in the UI are persisted in local storage. 
Also fixes #684 .

### DIFF
--- a/quick-start/src/main/ui/app/app.module.ts
+++ b/quick-start/src/main/ui/app/app.module.ts
@@ -60,6 +60,9 @@ import { FacetsComponent } from './facets/facets.component';
 import { ObjectToArrayPipe } from './object-to-array.pipe';
 import { DatePipeModule } from './date-pipe/date-pipe.module';
 
+import { SelectKeyValuesComponent } from './select-key-values/select-key-values.component';
+
+
 @NgModule({
   declarations: [
     AppComponent,
@@ -97,7 +100,8 @@ import { DatePipeModule } from './date-pipe/date-pipe.module';
     FacetsComponent,
     TitlecasePipe,
     TruncateCharactersPipe,
-    ObjectToArrayPipe
+    ObjectToArrayPipe,
+    SelectKeyValuesComponent
   ],
   entryComponents: [
     HasBugsDialogComponent,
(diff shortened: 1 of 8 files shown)


wooldridge:
@aebadirad, OK, looks like this got inadvertently closed when I was doing housekeeping during the rebase of my branch (in the process, I deleted the wooldridge:DHFPROD-167 branch), my bad. I can refile a PR with the updated, rebased branch.

wooldridge:
Opened this feature again as: https://github.com/marklogic-community/marklogic-data-hub/pull/713

PR #715, opened 2018-02-02
Test on other hosts

grechaw:
This is just a change to the test setup for quickstart.  Now, it will honor the property in gradle.properties rather than just running on localhost.


grechaw:
I've just replaced this PR... the previous one had a couple of commits that aren't ready for develop branch.  Let's see how jenkins does with this one.

grechaw:
i think i must have to open a new one to trigger CI

grechaw:
Ah, well it seems that if I use the same branch, it just reopens an existing PR.

PR #718, opened 2018-02-05
Export jobs

dmcassel:


marklogic-data-hub/src/main/java/com/marklogic/hub/job/JobManager.java
```diff
@@ -85,6 +101,153 @@ public JobDeleteResponse deleteJobs(String jobIds) {
...
+     * Export Job documents and their associated Trace documents to a zip file.
+     *
+     * @param exportFilePath specifies where the zip file will be written
+     * @param jobIds a comma-separated list of jobIds; if null, all will be exported
+     */
+    public void exportJobs(Path exportFilePath, String jobIds) {
+        File zipFile = exportFilePath.toFile();
+        WriteToZipConsumer zipConsumer = new WriteToZipConsumer(zipFile);
+
+        String[] jobsArray = jobIds != null ? jobsArray = jobIds.split(",") : null;
```
paxtonhare: does this work?

Shoudn't it be:

`String[] jobsArray = jobIds != null ? jobIds.split(",") : null;`

marklogic-data-hub/src/main/java/com/marklogic/hub/job/JobManager.java
```diff
@@ -85,6 +101,153 @@ public JobDeleteResponse deleteJobs(String jobIds) {
...
+
+    }
+
+    /**
+     * We create this new client instead of using traceClient because the DataMovementManger relies on an
+     * internal-only endpoint, and tracing-rewriter.xml doesn't account for it. As such, we're using the
+     * Jobs app server, but pointing to the Traces database.
+     * @return
+     */
+    private DatabaseClient getTraceDatabaseClient() {
```
paxtonhare: Look at HubConfig.newTraceDbCient()

I put those new*Client() methods there so that we don't duplicate the creation code. Also, it handles SSL and stuff. This approach will likely fail for ssl.

ml-data-hub-plugin/src/main/groovy/com/marklogic/gradle/task/HubTask.groovy
```diff
@@ -38,7 +38,7 @@ abstract class HubTask extends DefaultTask {
 
     @Internal
     JobManager getJobManager() {
-        return new JobManager(getHubConfig().newJobDbClient());
+        return new JobManager(getHubConfig().newJobDbClient(), getHubConfig().newTraceDbClient());
```
paxtonhare: See above comment. This one is correct. it's using newTraceDbClient() from HubConfig

marklogic-data-hub/src/main/java/com/marklogic/hub/job/JobManager.java
dmcassel: oddly, it appears to have worked, but I'll fix it

marklogic-data-hub/src/main/java/com/marklogic/hub/job/JobManager.java
dmcassel: I think I started with that, but the Trace app server uses a non-standard rewriter that doesn't include some hidden URLs. DMSDK failed because of that. We talked about re-copying the rewriter stuff, but that doesn't feel like a good approach long term. 

Won't the `jobClient.getSecurityContext()` figure out the SSL situation?

paxtonhare: that rewriter is moving into the server install dir and will be maintained like the regular rest rewriter. So we can count on it being ok. Erik says there is a magical tool that checks to make sure all the various rewriter*.xml files are in sync when a server build is created.

ml-data-hub-plugin/src/main/groovy/com/marklogic/gradle/task/HubTask.groovy
dmcassel: Okay, cool. I'll update the writer for now and switch to using newTraceDbClient

marklogic-data-hub/src/main/java/com/marklogic/hub/job/JobManager.java
dmcassel: Okay, cool. I'll update the writer for now and switch to using newTraceDbClient

dmcassel:
The two failing tests are TracingTest ones that are passing locally but not in Jenkins.

dmcassel:
@paxtonhare I think I've addressed your review comments. Please re-review

PR #723, opened 2018-02-07
React to changes in the harmonize flow options on blur

wooldridge:
Instead of on update, which is overkill


dmcassel:
Jenkins shows a couple failing tests. Looking into those prior to merging.

dmcassel:
These tests passed locally, not sure why they failed in Jenkins.

PR #728, opened 2018-02-09
DHFPROD-274 put sample data and model into an example project.

grechaw:
Here's the PR for the github side of the world, adding an example project with data for front-end developers.


settings.gradle
```diff
@@ -3,4 +3,5 @@ include ':marklogic-data-hub'
 include ':ml-data-hub-plugin'
 include ':quick-start'
 
+include ':examples:data-integration-tests'
```
paxtonhare: @grechaw what's the advantage to adding this here?

grechaw: It made IDEA recognize that project.  However if it causes any issues at all then of course it should be removed; now that that example works from the command line, I don't think it's needed as an explicit subproject.

PR #730, opened 2018-02-10
Center the actions icons, narrow the space between them

wooldridge:


wooldridge:
@dmcassel, some minor adjustment to improve the alignment and placement of the actions icons.

Issue #711, opened 2018-02-02
One entity's indexes configurations clobber all the others'


mikrovvelle:
#### The issue

In QuickStart (2.0.3), let's say I have two entities, each with a handful of string fields. If I do this:

1. Open the edit dialogue for "Entity-A"
2. Set two fields to be range-indexed (with the ⚡️ icon)
3. Close the edit dialogue.

I get a modal prompt---

> Saved. Update Indexes in MarkLogic?

Click "OK". So far so good. The two indexes are saved in my project code and visible in the ML Admin panel.

*But then*, I do *this*:

1. Open the edit dialogue for "Entity-B"
2. Set two (other) fields to be range-indexed (with the ⚡️ icon)
3. Close the edit dialogue.
4. Respond to the modal prompt about Updating the Indexes with "Yes"

Now there's a problem. The Indexes for "Entity-A" have disappeared from both my code and the ML Admin panel.

This happens both for Staging and for Final.

#### Tech details

- OS X, ML running on Linux (Docker)
- MarkLogic 9.0-3.1
- Data Hub Framework 2.0.3, but the project originated on 2.0.2
- Tested on
    - Chrome 63.0.3239.132
    - Safari 11.0.1 (13604.3.5)

#### The devil is in the details...

> If possible, rerun the command with -d for debugging output and attach the output:

I did; output attached.

┆Status: Done

aebadirad:
Confirmed on 2.0.3 dev. It's a quickstart UI bug - the entity controller is being too greedy when setting the indexes.

popzip:
Should be fixed in next release.

dmcassel:
➤ Kasey Alderete commented:

dupe.

aebadirad:
Fixed with release of 3.0.0

PR #743, opened 2018-02-14
fixing #711

paxtonhare:


aebadirad:
Merging this in a bit early due to needing to refactor the entity manager class

PR #757, opened 2018-02-23
Trying to stabilize e2e tests

ayuwono:
I found out that I have been running using npm run e2e 

The jenkins job is running ./gradlew quick-start:test instead. I found out that maximizing the window on the index might get it stabilize

Can we try to merge these changes to develop so we can see the result on jenkins?


ayuwono:
I've created a jenkins job on my fork for this, so it can wait

PR #767, opened 2018-03-01
adding support document

dmcassel:


dmcassel:
I interpreted the ticket as saying it should be a separate doc, but that's probably because I like it being separately available. However, looking at the other projects on github.com/marklogic, it is part of the readme. I'll move it.

dmcassel:
@popzip I've moved the support statement into the readme

PR #786, opened 2018-03-07
Add telemetetry call via eval + setup test for it

aebadirad:


marklogic-data-hub/src/test/java/com/marklogic/hub/DataHubInstallTest.java
```diff
@@ -21,12 +21,19 @@
 import static org.junit.Assert.assertTrue;
 
 public class DataHubInstallTest extends HubTestBase {
+    private static int afterTelemetryInstallCount = 0;
```
dmcassel: A bunch of our tests call `installHub`. What if one of the other test classes runs before this one? Wouldn't it be better to check `getTelemtryInstallCount` before and after the install and look for the +1?

marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/HubAppDeployer.java
```diff
@@ -75,6 +77,18 @@ public boolean equals(Object obj) {
...
+
+        //Below is telemetry metric code for tracking successful dhf installs
+        //TODO: when more uses of telemetry are defined, change this to a more e-node based method
+        ServerEvaluationCall eval = appConfig.newDatabaseClient().newServerEval();
+        String query = "xdmp:feature-metric-increment(xdmp:feature-metric-register(\"datahub.core.install.count\"))";
+        try {
+            eval.xquery(query).eval().close();
+        }
+        catch(FailedRequestException e) {
+            logger.error("Failed to increment feature metric telementry count: " + query, e);
```
dmcassel: (minor) typo: "telementry" -> "telemetry"

marklogic-data-hub/src/test/java/com/marklogic/hub/DataHubInstallTest.java
aebadirad: There's no +1. It only stores the count until it transmits it, then dumps it, usually within 30 seconds. I'm more concerned the check won't pass if a machine is a bit slow at running the tests - as it'll show null instead of 1.

PR #811, opened 2018-03-16
Fixing hard-coding of admin/admin in project setup

grechaw:
This PR addresses a couple of things in the build.

It fixes an issue with SLF4J bindings by including a bridge jar between slf4j and log4j, upon which mlcp depends.

Secondly it gets rid of a little nefarious hard-coding in test setup that confused the heck out of me for a while.  Now the values in gradle.properties are used for mlUsername and mlPassword rather than code slapping them onto some generated ones.  I think this is an improvement but would like confirmation.

I found that some setup code relied on the value in gradle.properties regardless of the hard-coded bit in project creation.  This approach gets rid of one code path, but not the last.


grechaw:
This PR also comments out some code that @srinath had committed in order to set up SSL testing.  We may want to talk over that before merge.

PR #820, opened 2018-03-19
Pii gen

grechaw:
This PR deserves a walk-through


grechaw:
Here's the comment from Srinath that he put in jira:

"The task "createLocalFile" in build.gradle is the one that I checked in for ssl and cert auth tests and it uses localhost. I see that you had commented that task in your PR. I am ok with it being commented for now as I am going to rewrite the ssl and cert auth tests with self signed certs. I am not aware of any other check in of mine that is using localhost. "

PR #828, opened 2018-03-21
Fix for DHFPROD-412, other failing tests

srinathgit:
Fix for DHFPROD-412. The field cache is invalidated on before every test run


srinathgit:
The tests ran fine when initiated from my local machine with mlHost set to a linux cluster. Obviously, looks like the tests are still failing in jenkins. I will work on debugging the issue.

PR #838, opened 2018-03-27
Refactor of unit tests to make DHFPROD-630 actionable

grechaw:
Yes this is a big change:

Data Hub Changes:
New exceptions for configuration and project exceptions.  project exception is thrown when you try to retrieve non-existent flows and entities (may cause quickstart issues, so revision may be needed)

Testing changes
1. more idempotent setup
2. remove static scoping from HubTestBase
3. various null-swallowing changes
4. fix of junit 4 vs. 5 running styles
5. pull properties from gradle.properties in project, rather than maniupulating generated propertes
6. cleanup of imports


grechaw:
Running locally to check the PR -- entire test suite.  

The core run took 24.5 minutes and has two failures in ScaffoldingTest
The quickstart java tests, in my setup, still have a couple of issues:

service.FlowManagerServiceTest still gets a handle on localhost, which results in three failures.  These issues might not show up in jenkins, which runs on localhost, but srinath will find the issues.

service.EntityManagerServiceTest has one error I don't understand at this time.  There's some handle on the project directory preventing deletion.  This could very well be a Windows-only issue.  Or maybe it's just trying to delete the directory twice.

grechaw:
Latest commit fixes an issue with hubconfig serialization and deserialization, introduced with @aebadirad refactor.

Issue #557, opened 2017-10-30
QuickStart - Browse Data does not display content for certain URIs


bciano:
#### The issue

Short description of the problem:

Quick Start "Browse Data" does not display documents that include an "&" in the document URI

Application is not encoding the ampersand, thus submitting an invalid document URI in the request to the REST API

This works: http://localhost:8080/api/search/doc?database=FINAL&docUri=/XML_polis_3587737-GTA-00001.xml?dekcode=350**%26**objcode=856

What behavior are you expecting?
QS should properly encode the doc URI to be sent to the /doc ML REST api

#### Tech details

Which Operating System are you using?
mac osx

Which version of MarkLogic are you using?
9.0-2

Which version of the Data Hub Framework are you using?

Is this a QuickStart UI bug? If so, which browser are you using?

yes.  Chrome Version 64.0.3253.0

#### The devil is in the details...

If possible, rerun the command with -d for debugging output and attach the output:

05:04:31.367 [http-nio-8080-exec-1] ERROR o.a.c.c.C.[.[.[.[dispatcherServlet] - Servlet.service() for servlet [dispatcherServlet] in context with path [] threw exception [Request processing failed; nested exception is com.marklogic.client.ResourceNotFoundException: Local message: Could not read non-existent document. Server Message: RESTAPI-NODOCUMENT: (err:FOER0000) Resource or document does not exist:  category: content message: /XML_polis_3581860-GTA-00043.xml?dekcode=350] with root cause
com.marklogic.client.ResourceNotFoundException: Local message: Could not read non-existent document. Server Message: RESTAPI-NODOCUMENT: (err:FOER0000) Resource or document does not exist:  category: content message: /XML_polis_3581860-GTA-00043.xml?dekcode=350
	at com.marklogic.client.impl.OkHttpServices.getDocumentImpl(OkHttpServices.java:784)
	at com.marklogic.client.impl.OkHttpServices.getDocument(OkHttpServices.java:656)
	at com.marklogic.client.impl.DocumentManagerImpl.read(DocumentManagerImpl.java:389)
	at com.marklogic.client.impl.DocumentManagerImpl.read(DocumentManagerImpl.java:273)
	at com.marklogic.client.impl.DocumentManagerImpl.read(DocumentManagerImpl.java:239)
	at com.marklogic.client.impl.DocumentManagerImpl.readAs(DocumentManagerImpl.java:191)
[...]

PR #841, opened 2018-03-29
Issue #557 fix

clockworked247:
Add URI encoding before requesting document contents on Browse Data document details pane.


grechaw:
I think that jenkins is probably already running this PR.  We'll get notified when all the tests run.

clockworked247:
@grechaw why the build failure when 0 skipped and 0 failed?

PR #849, opened 2018-04-03
downgrade ml-app-deployer

grechaw:
Experiment to see if jenkins works better with previous app deployer


grechaw:
Note to @aebadirad that review and merge tests are different for new 3.0-develop.  maybe that's ok

grechaw:
Thing is, jenkins has not been failing today with the same errors.  Today's two test failures look like a change in server behavior--- they are key failures related to binaries.
Maybe keys need to be updated though.  I'll take a look locally.

grechaw:
We don't want to downgrade.  That was not the issue causing I/O Exceptions.

PR #856, opened 2018-04-03
Updating tests to handle implicit conversion behavior

grechaw:
This is just a test adjustment from @aebadirad


grechaw:
i went and took the downgrade off of this PR -- didn't seem to be helping anything

PR #875, opened 2018-04-09
DHFPROD-466 Improving tests for quickstart host config

grechaw:
This PR, again, delivers some big changes to unit tests.
It has a small issue, again something with setup and teardown, where tracing counts fail for certain legacy flows.

When I run the tests with develop, then run the e2e tests on this branch, they all pass. mysterious what prevents tracing from working on these handful of tests.

also there's a init() test which fails once during a whole run, then passes.  This looks like an innocuous issue.

I'd like a good analysis from @srinath as first priority, as getting this right is blocking my other fixes for the sprint.

### DIFF
--- a/marklogic-data-hub/build.gradle
+++ b/marklogic-data-hub/build.gradle
@@ -297,9 +297,10 @@ task collectorTest(type: Test, dependsOn: testClasses) {
 //}
 
 test {
-    // we exclude the StreamCollectorTest
-    exclude '**/StreamCollector*'
-	//dependsOn tasks.createLocalFile
-    // depend on the collectorTest task above
-    dependsOn tasks.collectorTest
+    //dependsOn tasks.collectorTest
+
+    // the core test suite runs these tests, so exclude them here.
+    include 'com/marklogic/hub/CoreTestSuite.class'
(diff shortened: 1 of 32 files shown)


grechaw:
The latest push should give us complete e2e clean tests.
I've divided the tests into a few suites and configured gradle to exclude running individual tests in subfolders.
This has enabled me to do a grand setup and teardown for most of the tests, while retaining independence for the e2e tests which were getting some interference.

grechaw:
So there are two solutions for number 2.    We can separate stream collector, as you have, and provide a separate hub install/uninstall, or we could run the whole suite with the memory constraints you need for stream collector test.

grechaw:
Well that single failure is the mlcp one.  Can we merge this and take up the classpath length as a separate issue?

srinathgit:
I think we can do that. I am not sure if the failure in jenkins is also due to classpath length. I have only seen that (classpath length issue) in my local machine.

grechaw:
I've committed something that might help the mlcp classpath issue.  I see what the problem is, and have tried just shrinking the runtime classpath to be smaller for the mlcp call.

Issue #440, opened 2017-06-16
Update wiki for custom gradle environment properties


gduplantier:
Update wiki to include explanation/example of how to add additional gradle environment properties (`mlMyProperty=value`) that will serve as substitutions in the config files (`"%%mlMyProperty%%"`). Reference HubConfig.java:624.  Also, update the wiki, https://github.com/marklogic/marklogic-data-hub/wiki/Customizing-Forest-Configurations. No need to do the gradle extension on mlAppConfig

mhuang-ml:
@popzip, I'll have a PR with an example of using additional gradle environment properties that will serve as substitutions to create a new database and app server.  I'll open the PR against the 2.x-develop branch initially and then I can also create a PR against the develop branch as well.  I just need to polish up the README with the explanation of what the example is providing, which I should have done today or tomorrow at the latest.

mhuang-ml:
@aebadirad and @dmcassel, I've created a PR[907](https://github.com/marklogic/marklogic-data-hub/pull/907) against the 2.x-develop branch.  If this passes the review then I'll create a PR against the develop branch as well.  I can't assign anyone as a reviewer for the PR though and I don't have access to update the wiki either, but I have the change locally done for the wiki update for customizing the forest configurations page.

mhuang-ml:
@aebadirad  and @dmcassel, I've opened an identical PR against the develop branch as well under [PR#](https://github.com/marklogic/marklogic-data-hub/pull/908).  Please feel free to close or decline a branch if you see fit.

dmcassel:
@aebadirad since I've moved on to another project, I've requested your review of these PRs

mhuang-ml:
@aebadirad, I don't have access to update the wiki, but here's the changes that I made to the markdown file locally:

### Using Properties files
Then in your build.properties file you can specify the token like so:

```
CUSTOM_FOREST_DIR_PATH=/my/custom/path
```
For more information regarding this, then you can refer to the [ml-gradle wiki](https://github.com/marklogic-community/ml-gradle/wiki/Configuring-resources)

grechaw:
Hi, we've released a brief property reference for 4.0.0 here : https://marklogic.github.io/marklogic-data-hub/docs/gradleproperties/

other than those properties, ml-gradle contains the full list of properties relevant for a data hub project.

PR #907, opened 2018-04-21
Fixing Issue #440: Adding example w/ gradle props

mhuang-ml:
Adding a new example that utilizes gradle properties within
deployment configuration files for substitutions for variables. This
eliminates the need to utilize the mlAppConfig extension in
build.gradle files


mhuang-ml:
@aebadirad  and @dmcassel, please review this when you get a chance.  There is an identical [PR#908](https://github.com/marklogic/marklogic-data-hub/pull/908) that has the same changes against the develop branch.  If you think that this change should only be made against the develop branch, then feel free to close this PR.

PR #935, opened 2018-04-26
DHFPROD-675 add index confirm for save new entity

wooldridge:
When a new entity is created and saved in the Entities view, display Update Indexes dialog, similar to as is done when editing an entity. 

Update Indexes confirmation dialog feature moved to confirmUpdateIndexes() in entity-modeler.component.ts and now called from both new and edit code paths.


wooldridge:
@aebadirad, yes, this fix should go on the 2.x branch as well. I will make a PR for that.

ayuwono:
@wooldridge & @aebadirad, we need to push this fix to 3.x branch as well

Issue #578, opened 2017-11-17
Deleting an entity deletes associated flows and code without warning or confirmation


damonfeldman:
Deleting an entity should have more warnings or safety. Could be an enhancement but I'll put this in as a bug.

### Hub Version
2.0

┆Status: Done

mhuang-ml:
@dmcassel, @damonfeldman, and @popzip , I've added an additional modal dialog box so that the user will have to press "Yes" twice before we delete the entity and all associated flows and code.  Are there specific checks that we would want to perform?  I've made the changes to the current "develop" branch, should this change be made to the 1.0 and 2.0 branches as well?

Here's the original dialog message that is currently in the develop branch that I'm leaving unchanged:
![initialyesdialog_unchanged](https://user-images.githubusercontent.com/16652599/38780637-e68afeea-40a7-11e8-996a-3530b1f4ed8a.png)

When a user presses the "Yes" button, then the new dialog box will appear like the following:
![additionaldialogmessage](https://user-images.githubusercontent.com/16652599/38780653-25b223c8-40a8-11e8-9b86-1ec5b3096618.png)

If the user presses "Yes" to the new dialog box, then it will proceed with deleting the entity and all associated flows and code.  If the user presses "No" to the second modal window, then it will **NOT** delete anything.

I just wanted to double check this implementation to make sure I'm not missing anything within the scope of this ticket.

mhuang-ml:
@grechaw, thank you for the clarification for the changes coming up in the 3.0 branch.  Is the plan that 1.0 and 2.0 would have the entities and flows separated as well?  I'm trying to wrap my head around if there is any portion of this task that can be done now in terms of a warning/confirmation, or will all of this become moot with the future changes?

grechaw:
We're not likely to go back to 2 or 1 to separate the flows -- all our work is in the 3 branch now.

mhuang-ml:
@grechaw, thanks for the updates.  So it seems like I can scrap the code for the 3.0/develop branch then since these are going to be split out during the summer.  Is there value in me adding the additional dialogs to the 1.0/2.0 branches now since the flows aren't going to be separated in the future?  Or I can scrap this code altogether and leave this ticket in the backlog so that it can be addressed when the entities and flows are split out.

grechaw:
[...]

PR #937, opened 2018-04-26
Updating Issue #578: Update the deletion message

mhuang-ml:
Removing the double confirmation dialog and updating the
deletion message that is displayed to the user to make it
clearer to what they are deleting.

@aebadirad and @ayuwono, synchronizing the changes that were made with the develop 7f02225 and 2.x-develop 28658b9 branches with 3.x-develop.


aebadirad:
thanks :)

PR #938, opened 2018-04-26
First pass on 2.x flow upgrade to 3, and removing unneeded modules

aebadirad:


marklogic-data-hub/src/main/java/com/marklogic/hub/scaffold/impl/ScaffoldingImpl.java
```diff
@@ -309,6 +310,22 @@ public boolean update2xFlow(String entityName, String flowName, FlowType flowTyp
                 throw new RuntimeException(e);
             }
         }
+
+        if (mainPath.toFile().exists() || xqyMainPath.toFile().exists()) {
```
grechaw: What's with the logic here, in relation to the one at line 284?  The first block runs on some set, and then this runs over the same ones, in addition to the xquery main.

aebadirad: First one only looks to update a main.sjs to match our new code sjs code path. If there's not a main.sjs (but a main.xqy) it skips over the block.

Below it's just looking for if you have a main.xqy or main.sjs, and if so, grab that file and replace the paths from com.marklogic.com/ to the server-side provided ones at Marklogic/, hence the OR

grechaw: ok so you meant both to execute for main.sjs.  that's what i was looking for.  I could nitpick about style, but will restrain myself.... this time mwahahah

PR #939, opened 2018-04-26
DHFPROD-502 fix tutorial, primary key does not add element range index

wooldridge:
Remove the information in the tutorial about setting a primary key to a property automatically adding an element range index for that property.

According to @grechaw, you now have to explicitly set an element range index if you want such a thing on a property with a primary key. This is based on how Entity Services now works.


aebadirad:
Makes sense, we want to align as much as we can to entity services so they become "one" with the DHF.

PR #953, opened 2018-05-01
Refactor gradle plugin tests

grechaw:
This PR fixes an annoying logging issue and also refactrors the plugin test suite.
I'll make a crucible review for QA to peruse.

### DIFF
--- a/ml-data-hub-plugin/build.gradle
+++ b/ml-data-hub-plugin/build.gradle
@@ -67,6 +67,8 @@ test {
         // in Gradle output.
         showStandardStreams = true
     }
+    include 'com/marklogic/gradle/fullcycle/**'
+    include 'com/marklogic/gradle/task/**'
 }
 
 task sourcesJar(type: Jar, dependsOn: classes) {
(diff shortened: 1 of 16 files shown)


grechaw:
Hi @aebadirad  this PR is ready to go -- we'll see if jenkins agrees.  It speeds up the plugin tests by separating those that require full setup/teardown from those that do not.

It also rationalizes the settings used for the tests.  Where previously inline properties pointed to localhost/admin/admin, now we use the gradle.properties file to drive settings.

Since all three subprojects now use the same properties, we might be able to simplify further by lifting resources up to the root project, but that's for later.

grechaw:
Hi @aebadirad  this PR is ready to go -- we'll see if jenkins agrees.  It speeds up the plugin tests by separating those that require full setup/teardown from those that do not.

It also rationalizes the settings used for the tests.  Where previously inline properties pointed to localhost/admin/admin, now we use the gradle.properties file to drive settings.

Since all three subprojects now use the same properties, we might be able to simplify further by lifting resources up to the root project, but that's for later.

jenkins is not reporting back success right now maybe.  If you click through to the PR job you can see it passed.

grechaw:
oh it's green now, my browser was stale.

PR #988, opened 2018-05-11
pii e2e test

srinathgit:
Adding pii e2e test

### DIFF
--- a/marklogic-data-hub/src/test/java/com/marklogic/hub/PiiE2E.java
+++ b/marklogic-data-hub/src/test/java/com/marklogic/hub/PiiE2E.java
@@ -1,5 +1,7 @@
 package com.marklogic.hub;
 
+import static org.junit.Assert.fail;
+
 import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.FileReader;
@@ -32,6 +34,7 @@ import com.google.gson.JsonPrimitive;
 import com.google.gson.JsonSyntaxException;
 import com.marklogic.appdeployer.command.Command;
 import com.marklogic.appdeployer.command.security.*;
+
 import com.marklogic.appdeployer.impl.SimpleAppDeployer;
 import com.marklogic.client.DatabaseClient;
 import com.marklogic.client.DatabaseClientFactory;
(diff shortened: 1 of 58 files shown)


marklogic-data-hub/src/test/java/com/marklogic/hub/PiiE2E.java
```diff
@@ -0,0 +1,330 @@
...
+				e.printStackTrace();
+			} catch (JsonSyntaxException e) {
+				// TODO Auto-generated catch block
+				e.printStackTrace();
+			} catch (FileNotFoundException e) {
+				// TODO Auto-generated catch block
+				e.printStackTrace();
+			} catch (URISyntaxException e) {
+				// TODO Auto-generated catch block
+				e.printStackTrace();
```
grechaw: If you really want to just print stack traces for these, then go ahead and remove the TODO comments.  I think you might actually want to throw a RuntimeException like a "YourTestHanessIsScrewedUpException" :)

srinathgit:
Charles,

I have rebased against marklogic:feature/pii .

grechaw:
Great, now I just see your 2 commits instead of 10

marklogic-data-hub/src/test/resources/pii-test/entity-config/final-entity-options.xml
```diff
@@ -0,0 +1,40 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
grechaw: This file is not really needed for this test.  I think you should remove it so as to make what you're testing cleaner.

marklogic-data-hub/src/test/resources/pii-test/entity-config/staging-entity-options.xml
```diff
@@ -0,0 +1,40 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
grechaw: This file is not really needed for this test.  I think you should remove it so as to make what you're testing cleaner.

marklogic-data-hub/src/test/resources/pii-test/hub-internal-config/databases/trace-database.json
```diff
@@ -0,0 +1,53 @@
+{
```
grechaw: Since @akshaysonvane is currently working on merging jobs and tracing, it will be interesting to see whether this test fares ok.  We should retain backward compatibility with his change.

PR #994, opened 2018-05-14
E2E/initial pii

ayuwono:
Modify the existing entity page objects because there is a new column for PII

Also there is an initial test for a creation of a property with PII with the verification


ayuwono:
It seems that it's conflicting with Rahul's work and PR. Closing it

PR #999, opened 2018-05-15
Removed trace references from gradle plugin and tests

akshaysonvane:


grechaw:
Hi @akshaysonvane you can rebase the from the existing branch and just add commits to it -- no need for another PR.  I made one of the changes you have here, so you'll have to resolve those commits.

akshaysonvane:
@grechaw I've rebased the feature branch with your commit. Won't create an issue anymore.

PR #1001, opened 2018-05-15
DHFPROD-401 First pass at rationalizing error calls

grechaw:
I've combed through the server-side errors and made all the calls consistent.    I created anew prefix called "DATAHUB-" for those errors which did not seem to fix into pre-existing categories of RESTAPI errors.  A message file is ready to go into the server too.


grechaw:
I've requested @ehennum as the reviewer of this PR, as he will most readily see issues with my proposal here.

grechaw:
jenkins found a bug in the original implementation.  The most recent commit fixes it.

grechaw:
Tested locally and I want to make room on CI server for more interesting PRs

PR #1010, opened 2018-05-17
Improvement to test infrastructure to capture screenshots with proper filenames

bsrikan:
Added a flag so screenshots on failures will have test name instead of a random number.
Will help debugging failed tests faster and better by knowing which screenshot to look at for a particular test.


bsrikan:
Name will be same as expectation string of a spec or it().
Ex: 
Starts-off-with-the-right-stuff.png for spec
it('Starts off with the right stuff', function() {
....
}

PR #1025, opened 2018-05-23
Check to verify if on windows or not, if so, don't run these specific…

aebadirad:
PR for review for DHFPROD-411


quick-start/src/test/java/com/marklogic/quickstart/web/HubConfigJsonTest.java
```diff
@@ -14,6 +16,7 @@
 import java.io.IOException;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static sun.awt.OSInfo.OSType.WINDOWS;
```
grechaw: maybe this is a relic that was not useful?
take it out if you like, leave it in, approved on your discretion

PR #1038, opened 2018-05-31
DHFPROD-852: support for property validation for any whitespace character

alexsmr:
* added per-name field validation
* added validation message
* added input field highlight
* added disabling of Save button in the case of invalid form

### DIFF
--- a/quick-start/src/main/ui/app/entity-modeler/entity-editor.component.html
+++ b/quick-start/src/main/ui/app/entity-modeler/entity-editor.component.html
@@ -52,7 +52,7 @@
           <td class="col-toggler" (click)="property.required = !property.required"><i [ngClass]="{'active': property.required}" class="fa fa-exclamation"></i></td>
           <td class="col-toggler" (click)="property.pii = !property.pii"><i [ngClass]="{'active': property.pii}" class="fa fa-lock"></i></td>
 
-          <td><input type="text" [(ngModel)]="property.name"></td>
+          <td><input type="text" name="name" [(ngModel)]="property.name" [pattern]="PROPERTY_NAME_PATTERN" required></td>
           <td>
             <select [ngModel]="getTypeForEdit(property)" (ngModelChange)="setType(property, $event)">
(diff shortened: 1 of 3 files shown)


ayuwono:
Hi @alexsmr, I think this is breaking the regression
Issue: https://github.com/marklogic/marklogic-data-hub/issues/1043
It seems that right now we can't save an entity with no properties. We should be able to have an entity without any properties as we should be able to load data as-is.

alexsmr:
Hi @ayuwono. Yes, it is unintended, but does Entity without properties make sense at all? Does it breaks specific test case?

alexsmr:
Talked to @sudo-jmakeig: it does not make much sense but it is not an invalid state. I will commit new fix against the regression.

ayuwono:
@alexsmr, if it doesn't make sense, we need to change the tutorial: https://marklogic.github.io/marklogic-data-hub/tutorial/create-product-entity/

This is what's been using on the workshops too. I think it's fine to have an entity without properties for the purpose of loading data as-is on that entity. Then on the next iteration, user adds properties on that entity for the harmonization flow process.

If this is changing, we need to confirm it and change the tests as well as the documentation + tutorial

grechaw:
Hi @alexsmr as we try to rationalize DHF to entity services, we should really have a common set of validation criteria.  An entity type with no properties is indeed valid.  The rationale that was floating around during ES design was that someone should be able simply to assert the entities and relationships in a model without committing to a single property.

Practically speaking that might not be very helpful, I recognize.  It's hard for me to see actual useful integration without something to look up an entity with, for example.

alexsmr:
Hi @grechaw this is exactly what Justin had explained me. I have fixed the validation. Please see PR https://github.com/marklogic/marklogic-data-hub/pull/1044 . As a side note, we need to address validations more seriously, not just for Entity editor, but for entire product, also, the same problem exist in OpsDirector. I plan to add general validation design guideline. Also it would be great if we can loop in PMs as a standard part of feature definition process.

PR #1064, opened 2018-06-07
Dhfprod 958 scaffold new role

grechaw:
This PR I hope fixes up remaining issues with the hub admin role and jobs + traces.


ayuwono:
@grechaw, just did a sanity test: installation is good now, and the counts on jobs page is good too

PR #1070, opened 2018-06-11
 Fix for DHFPROD-872,DHFPROD-767

srinathgit:
The fix preserves all nodes (processing instruction, comments) when running XML input flow with xqy and retains only root element when running with sjs

### DIFF
--- a/marklogic-data-hub/src/server-side/impl/flow-lib.sjs
+++ b/marklogic-data-hub/src/server-side/impl/flow-lib.sjs
@@ -109,7 +109,9 @@ function cleanData(resp, destination, dataFormat)
 {
   if (resp instanceof Document) {
     if (fn.count(resp.xpath('node()')) > 1) {
-      fn.error(null, "DATAHUB-TOO-MANY-NODES", Sequence.from([400, "Too Many Nodes!. Return just 1 node"]));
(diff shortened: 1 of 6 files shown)


marklogic-data-hub/src/server-side/impl/flow-lib.sjs
```diff
@@ -109,7 +109,9 @@ function cleanData(resp, destination, dataFormat)
 {
   if (resp instanceof Document) {
     if (fn.count(resp.xpath('node()')) > 1) {
-      fn.error(null, "DATAHUB-TOO-MANY-NODES", Sequence.from([400, "Too Many Nodes!. Return just 1 node"]));
+	  if(fn.count(resp.xpath('element()')) > 1) {
```
grechaw: Indentation needs fix here, maybe there's tabs and spaces going on...

marklogic-data-hub/src/server-side/impl/flow-lib.xqy
```diff
@@ -299,7 +299,10 @@ declare function flow:run-flow(
     => rfc:with-id($identifier)
     => rfc:with-content(
         if ($content instance of document-node()) then
-          $content/node()
+		  if(fn:count($content/node()) > 1) then
```
grechaw: same deal here, for some reason.

Issue #1039, opened 2018-05-31
No document counts on collection facets


ayuwono:
QuickStart version 3.0.0
Branch: develop
Browser: Chrome
OS: Mac

On Browse Data, Jobs, and Traces pages; the collection facets used to have document counts. Now there is no count

<img width="1276" alt="screen shot 2018-05-31 at 3 30 09 pm" src="https://user-images.githubusercontent.com/7096668/40812896-0b77a0aa-64ed-11e8-9daa-ba3574c8b350.png">

marklogic-builder:
➤ Scott Brooks commented:

In review here: https://github.com/marklogic/marklogic-data-hub/pull/1071

ayuwono:
Yes, it's on develop branch, I think it should 3.1.0

marklogic-builder:
➤ Aries Yuwono commented:

Verified passed on develop branch. Facet counts are back on browsing data, jobs, and trace page

PR #1071, opened 2018-06-11
Fix for #1039

clockworked247:
MDL component styles require that the span tag have contents, otherwise, style applied sets "display: none;"   Fix adds count value within the badge span tag.


clockworked247:
branch incorrectly named, closing this in favor of: https://github.com/marklogic/marklogic-data-hub/pull/1072

PR #1074, opened 2018-06-12
Dhfprod 500 attachments

grechaw:
This change fixes issues we've seen with $attachments being attached to JSON instances at multiple locations.

I extended FlowRunnerTest.java to cover the four combinations of language and format.  The DHF flow libs were not removing the keys from the json objects.

Some of our flows are not generating map:map from content plugins, which is the trigger on which ES instances are made... 

I'm looking over E2E flows, there are some assertion failures now.  I'll see what has changed and modify this PR accordingly before asking for review.


grechaw:
This PR has been incorporated into #1099

Issue #969, opened 2018-05-03
Improve error stack trace display on QuickStart Jobs


ayuwono:
Currently it's hard to read the error stack trace on Jobs that has errors

To produce the stack trace, just add a line with invalid property, example:

let foo = xs.string(source.foo);

This is on 3.0.0

Currently the stack trace looks like this:
<img width="1426" alt="screen shot 2018-05-03 at 3 09 13 pm" src="https://user-images.githubusercontent.com/7096668/39605263-9e00486a-4ee4-11e8-8813-b0de9bbd3654.png">

┆Status: Closed

ayuwono:
And the corresponding trace doesn't tell anything on the error

<img width="1452" alt="screen shot 2018-05-03 at 3 34 25 pm" src="https://user-images.githubusercontent.com/7096668/39605908-975c3ade-4ee7-11e8-92bf-eecb32125cae.png">

marklogic-builder:
➤ Charles Greer commented:

Needs pretty printing at least.

Highlighting is nice too.

If we can find where in the flow the failure it, even better.

This was direct feedback from the workshop

marklogic-builder:
➤ Scott Brooks commented:

I've sent this bug for review.  The bug fix to handle both issues with the readability of the output & and rendering of the stack within the job details view.

Further enhancements should be tracked as a task as they will require more time due to the need to:

  * Re-write stack output coming from the backend to have structure, rather than leave it as a string, as it is today.
  * Once the stack is structured, we can apply smart formatting and highlighting to further enhance the usability of the output. We might even be able to link directly to where the code needs to be changed.
  * Lastly, we must examine other code paths that could benefit from job output restructuring (ie. hasLiveOutput)

PR #1080, opened 2018-06-14
Issue #969 Pretty printing added to output

clockworked247:
This is a bug fix to handle both issues with the readability of the output & and rendering of the stack within the job details view.


grechaw:
From your change it looks like the indentation was already correct in the output, just the newlines were rendered as \\n.

Do you think the problem perhaps should have been solved upstream somewhere or are you comfortable with fixing it here in the browser?

Issue #1036, opened 2018-05-31
Unable to save map


ayuwono:
QuickStart version 3.0.0
Branch: mapping/menus-source
Browser: Chrome
OS: Mac

Step:
1) Create an entity
2) Create an input flow and load the data (I use products from example online-store)
3) Add some properties to the entity
4) Create a harmonize flow
5) Click on Create Map button
6) Do the mapping on the drop down menus
7) Click on Save Map button

When it goes to harmonize flow info page, the map is not saved. It still shows the "Create Map" button. It should show the map name link with a trash bin button to delete it.

If you click on "Create Map" button again, it will go to the map with the settings that you performed on step 6.

<img width="1299" alt="screen shot 2018-05-31 at 3 10 22 pm" src="https://user-images.githubusercontent.com/7096668/40811334-a6cb0572-64e5-11e8-8c83-de2b10ae723f.png">

┆Status: Closed

marklogic-builder:
➤ Charles Greer commented:

Just validation that problem goes away when re do the saving task.

marklogic-builder:
➤ Scott Brooks commented:

In review:  https://github.com/marklogic/marklogic-data-hub/pull/1081

aebadirad:
Is this still valid?

marklogic-builder:
➤ Scott Brooks commented:

Should no longer be relevant...

PR #1081, opened 2018-06-14
Temporary fix for #1036 - will require rewrite when new backend in place

clockworked247:
Map name now shows up instead of "Create Map" after a map has been created and saved.


grechaw:
Ah, I see Mike requested a change, when that's there we'll move ahead and merge.

clockworked247:
@wooldridge Good catch.  I've made the changes to support the trash can button.   Can you please review again?  Thanks.

Again, this is a temporary fix using localStorage.  Will require changes to use MapService for saving/loading maps, once the new server code is ready to use.

PR #1086, opened 2018-06-19
DHFPROD-983 job db reports no counts

grechaw:
This fix to the angular app keeps the old configuration key from before the regression (job), but displays the database name as label (jobs).

Fixing this any other way would have introduced compatibility issues -- all of the configuration keys start with 'job' but the database name is -JOBS


ayuwono:
@grechaw, ran the sanity test on QuickStart UI for this branch, it's good, the jobs count is displayed now

PR #1089, opened 2018-06-19
E2e/stabilize toast

ayuwono:
Somehow the toast button on QuickStart UI is not stable. Trying to use sleep to see if the tests are stable


ayuwono:
Hi @grechaw, I don't know what I did wrong...I just need to commit and push the last four commits

stabilize toast on flows 5bf18a7
stabilize delete props dialog box 326efe0
undo delete jobs test 06bba48
stabilize toast on flows 509fea8

But some other things are included

PR #1094, opened 2018-06-20
Dhfprod 791 hubinfo

meredithju:


ml-data-hub-plugin/src/main/groovy/com/marklogic/gradle/task/HubInfoTask.groovy
```diff
@@ -23,6 +23,9 @@ class HubInfoTask extends HubTask {
 
     @TaskAction
     void dumpHubInfo() {
-        print(this.hubConfig.toString());
+        //print(this.hubConfig.toString());
+
+        print("test");
+
```
grechaw: looks like this is needing to be reverted to how it was?

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/HubConfigImpl.java
```diff
@@ -1175,4 +1178,20 @@ private boolean getEnvPropBoolean(Properties environmentProperties, String key,
         }
         return res;
     }
+
+    @JsonIgnore
+    public String getInfo()
+    {
+        ObjectMapper objmapper = new ObjectMapper();
```
grechaw: It would be slightly better to pull objmapper out into the class as a private object so it doesn't have to be recreated every time the function is called.

marklogic-data-hub/src/test/java/com/marklogic/hub/core/HubConfigTest.java
```diff
@@ -80,4 +90,20 @@ public void testLoadBalancerProps() {
         assertEquals("host1", config.getLoadBalancerHosts()[0]);
     }
 
+    @Test
+    public void testHubInfo() {
+        HubConfig config = getHubConfig();
+
+        try {
+
+            System.out.println(config.getInfo());
```
grechaw: Do you feel like making any assertions in here?
We don't want to use System.out.println in uni tests.

PR #1102, opened 2018-06-23
e2e using new harmonize content

ayuwono:
This change is to use new custom harmonize content plugin for sku and SKU property. Also using xpath to get the price property

### DIFF
--- a/quick-start/e2e/qa-data/plugins/contentWithOptions.sjs
+++ b/quick-start/e2e/qa-data/plugins/contentWithOptions.sjs
@@ -1,73 +1,73 @@
-     /*
-     * Create Content Plugin
-     *
-     * @param id         - the identifier returned by the collector
-     * @param options    - an object containing options. Options are sent from Java
-     *
-     * @return - your content
-     */
-    function createContent(id, options) {
-      var doc = cts.doc(id);
-      var root = doc.root.toObject();
-    
-      var source;
-      var opts;
-    
-      // for xml we need to use xpath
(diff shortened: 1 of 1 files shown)


quick-start/e2e/qa-data/plugins/contentWithOptions.sjs
```diff
@@ -1,73 +1,73 @@
...
-      
-      // return the instance object
-      return {
-        '$attachments': attachments,
-        '$type': 'Product',
-        '$version': '0.0.1',
-        'sku': sku,
-        'title': title,
-        'price': price,
-        'opt1': value1,
```
aebadirad: are we still using opt1 and opt2 for tests anywhere else?

bsrikan: I do see opt1 in 1 traces test 'should verify the traces viewer page'
and 3 jobs test
'should verify the loaded data'
'should verify the harmonized data with sku as original property'
'should verify the harmonized data with SKU as original property'

Tests dont use opt2 anywhere.

bsrikan: if I am not wrong, I think opts was missed out and adding them to the updated custom content plugin should be good enough. Will confirm with @ayuwono

ayuwono:
Good catch, modifying it right now

ayuwono:
@aebadirad , fixed the content with options values

PR #1108, opened 2018-06-26
Dhfprod 307 mapping qs

aebadirad:


quick-start/src/main/java/com/marklogic/quickstart/model/MappingModel.java
```diff
@@ -22,45 +22,117 @@
...
+        if(getSourceContext() != null) {
+            node.put("sourceContext", getSourceContext());
+        }
+        if(getTargetEntityType() != null) {
+            node.put("targetEntityType", getTargetEntityType());
+        }
+        if(getDescription() != null) {
+            node.put("description", getDescription());
+        }
+        return node;
```
grechaw: normally this would be done simply with a valueToTree() call

JsonNode node = mapper.valueToTree(this)

that's why the annotations are on this class in the first place.  Yeah, a nitpick, and you can do what you want with it.

quick-start/src/main/java/com/marklogic/quickstart/service/MappingManagerService.java
```diff
@@ -141,14 +90,13 @@ public void deleteMapping(String mapping) throws IOException {
...
-        List<MappingModel> mappings = getMappings();
-
-        for (MappingModel mapping : mappings) {
-            if (mapping.getName().equals(mappingName)) {
-                return mapping;
-            }
+        mappingManager = MappingManager.getMappingManager(envConfig().getMlSettings());
+        try{
+           ObjectMapper objectMapper = new ObjectMapper();
+            return MappingModel.fromJson(objectMapper.readTree(mappingManager.getMappingAsJSON(mappingName, -1)));
```
grechaw: heres the other side of my comment above -- mapper.convertValue( ..., MappingModel.class)

also ObjectMapper is best as a singleton object field.

grechaw:
@aebadirad ill see if i can make those changes quick so you can see what i mean, then put the commit up.

grechaw:
I have a commit that makes these changes, tests all still pass but since the serialization is probably changed under the hood ill not put it here.

The commit if you're interested is here: https://github.com/grechaw/marklogic-data-hub/commit/3482ad77a529494e5122d37f87cff39f7fc0c8a1

(also a docs update to plugin sneaked in there)

grechaw:
We'll defer jackson optimization

PR #1109, opened 2018-06-26
DHFPROD-967 change source URI in mapping

wooldridge:
Allow user to change the source document URI in the mapping UI.
Add a copy-to-clipboard icon to result items in Browse Data view.


aebadirad:
Pulling this through to merge with the rest of the code

wooldridge:
I will look at #1081 now.

Issue #1057, opened 2018-06-05
Trace UI is showing nothing on ingested xml document


ayuwono:
QuickStart v 3.0.0
Branch: develop
Chrome browser on Mac OSX

Steps:
1) Create an entity
2) Ingest an xml document (xml doc is attached). Input flow => plugins: sjs, document type: xml
3) Go to Traces tab, and click on the xml trace

There is no trace for raw data and output:
<img width="1486" alt="screen shot 2018-06-04 at 5 29 34 pm" src="https://user-images.githubusercontent.com/7096668/40948480-827fa21e-681d-11e8-8fd3-4bc28be09613.png">

┆Status: In Review

marklogic-builder:
➤ Charles Greer commented:

regression from 3.0

marklogic-builder:
➤ Scott Brooks commented:

[~cgreer] is this something you think I should pick up?  I know you've been working on Tracing UI related regressions, so maybe you have a sense of where the problem is already.  Happy to implement under your direction.

grechaw:
Sure, you can diagnose at least.  I don't know where the issue is yet.  If it's enode we should take care of it as soon as possible

Issue #959, opened 2018-05-03
QuickStart Traces output has bad formatting


wooldridge:
JSON displayed in the output pane does not have the linebreaks that the JSON has in the rawOutput pane.

This is on the 3.x develop branch of DHF QuickStart.

Here is what the output data looks like in the Traces database:
```
"output": "{game_id:\"1000178\", SKU:\"141606253740\", title:\"permanent megalomaniac\", price:\"4.5\", description:\"\", years_active:\"0\", publication_date:\"0\", players:\"3 to 6\", age_range:\"\", setup_time:\"0\", playing_time:\"0\", chance:\"\", category:\"Board Game|Strategy\", has_extensions:\"False\", has_accessories:\"False\", has_apparel:\"False\", popularity_tier:\"3\", probability_apparel:\"0.3\", probability_accessories:\"0.3\", probability_extensions:\"0.3\"}", 
```

Here is what the display looks like in QuickStart:

<img width="1398" alt="trace_output" src="https://user-images.githubusercontent.com/477757/39594129-a46a2c30-4ec0-11e8-8b4e-82d97d20d2f4.png">

┆Status: Closed

marklogic-builder:
➤ Mike Wooldridge commented:

[~ayuwono], can you QA? The formatting for Traces output should now be fixed and consistent with the rawContent outpiut. PR was merged on develop branch.

PR #1132, opened 2018-07-03
DHFPROD 948 trace ui

meredithju:


srinathgit:
@meredithju ,

This fix will resolve DHFPROD-948 (github issue #1057 ) but will reintroduce the original issue it addressed  (QuickStart Traces output has bad formatting #959 ).

srinathgit:
1. Rather than checking the index  of "</",  how about this ?
```
    try {
        return JSON.stringify(JSON.parse(data), null, '  ');       
    } catch(e) {
        return data;
    }
```
JSON.parse throws an exception when input is not json and we can return 'data'  in catch block in case of xml .
2. Also, can you please add a message when you commit a change  ?

srinathgit:
@meredithju ,

Can you squash the last 3 commits to a single commit and add a commit message ? 

```
git reset --soft HEAD~3
git commit -m "<message for the combined commit>"
```

PR #1140, opened 2018-07-06
fix label for triggers database

meredithju:


bsrikan:
Can we pls merge this now that it says Approved!

PR #1156, opened 2018-07-11
DHFPROD-1040 QuickStart mapping UI redesign

wooldridge:
Changes to QuickStart mapping UI as described in DHFPROD-1040. Details:

- Added side nav for managing saved maps
- Use child routing for mapping edit view
- Added reset button and popup at save
- Added messaging for case of no source docs
- Made mapping desc and URI editable


ayuwono:
@wooldridge , I just tested this branch. Looks good. Flows and maps deployment is fine. Running harmonize flows is fine. Performance on multiple maps is good too

Issue #1104, opened 2018-06-25
Some icons on QuickStart UI login page are missing if you start QuickStart war


ayuwono:
QuickStart v3.1.0
Chrome 

Some icons on QuickStart UI login page are missing if you start QuickStart war

1) Build the war: java -jar quick-start/build/libs/quick-start-3.1.0.war
2) Go to localhost:8080
3) On the login page, you can see some icons are missing
<img width="1133" alt="screen shot 2018-06-25 at 10 39 40 am" src="https://user-images.githubusercontent.com/7096668/41869189-0e852a5e-786d-11e8-8932-afce3cc9d73a.png">

If you the QuickStart UI using gradlew bootRun then gradlew runUI, the icons exists

┆Status: Closed

marklogic-builder:
➤ Kim Coleman commented:

Chrome (on Windows) is reporting the failure to load the images on the front page as an auth failure. All of the following requests are failing with a 401:

  * GET http://localhost:8080/api/current-project/  
  * GET [http://localhost:8080/main/ui/assets/img/logo.svg](http://localhost:8080/main/ui/assets/img/logo.svg) 
  * GET http://localhost:8080/main/ui/assets/img/odh.svg

Example error response:

\{"timestamp":"2018-07-11T23:10:29.595+0000",
{{ "status":401,}}
{{ "error":"Unauthorized",}}
{{ "message":"Unauthorized",}}
{{ "path":"/main/ui/assets/img/odh.svg"}}
{{}}}

It's not an issue with filesystem permissions, afaict.

aebadirad:
Part of PR #1161

PR #1161, opened 2018-07-12
DHFPROD 1107 logo

meredithju:


ayuwono:
@meredithju, i believe there are two icons that are broken, you fixed the odh.svg. Can you make sure that the other one is fixed too? http://localhost:8080/main/ui/assets/img/logo.svg

aebadirad:
@ayuwono if it's okay, let's bring this PR in and then issue another PR to fix and close out #1104 ?

ayuwono:
Sure @aebadirad , I'm okay with it

PR #1162, opened 2018-07-12
DHFPROD-1108-Deploy security configs for PII

akshaysonvane:
Add null check on destroy


srinathgit:
installUserModules() method specifically loads modules/ artifacts in entities and mapping dirs and I think deploying security artifacts shouldn't be part of this method.

grechaw:
We decided not to merge this PR so I'll clean it out for the time being.

Issue #1166, opened 2018-07-12
Hub Graphic corrupted / missing


popzip:
the 'what is it' image

https://marklogic.github.io/marklogic-data-hub/understanding/concepts/

also make sure use the latest - please ping PM or Marketing

kcoleman-marklogic:
Seems to have been a victim of yesterday's branch cleanup. Working on replacing it with a more recent version of the architecture diagram.

popzip:
That works! If we dont have it already im going to log something about future opportunity to marry up consistency in language

kcoleman-marklogic:
The new graphic is on live now. 

I'll merge the changes into docs-develop, too.

kcoleman-marklogic:
Bumping down priority because problem on live site has been fixed. Leaving this open, though, because I want to make a further change for 3.1 that enables you to click to see an larger version of the image.

marklogic-builder:
➤ Kim Coleman commented:

This is now on the docs-develop branch, too. See:

[http://pubs:9010/understanding/concepts/](http://pubs:9010/understanding/concepts/)

New arch diagram should be present, and if you click on it, you should get a bigger image (on a different page).

PR #1174, opened 2018-07-13
Issue #1166 : Replace missing ODH architecture diagram with a more current one

kcoleman-marklogic:
Replacing the ODH arch diagram with the one from MLW 2018. PM has reviewed and approved.


wooldridge:
@kcoleman-marklogic, can we go any larger with whatever original you're working from? I'm having to squint to see some of the smaller text labels. Looks like we're at 760px width and have room in the layout to go to, say, 900px. But if what we're working with is already 760px then enlarging might just degrade the quality.

kcoleman-marklogic:
Mike and I batted this back and forth for a bit in chat. I feel like the 900px image is still too big. However, he's right about the readability.

I'm going to push this as-is to get _something_ up on the live site, but also continue to investigate whether or not I can satisfy us both via a lightbox or something similar.

PR #1178, opened 2018-07-16
Integration tests initiative

rahulvudutala:
Quick-Start API End Points Integration tests for ModeltoModelMapping


quick-start/src/test/java/com/marklogic/quickstart/integrationtests/EndToEndAPITest.java
```diff
@@ -0,0 +1,350 @@
...
+		// Tests api's to initialize the project folder and set the environment configuration
+		projectInitResponse = utils.initilizeProjectConfiguration();
+		projectInitJson = projectInitResponse.jsonPath();
+		projectInitStatus =  projectInitJson.getBoolean("initialized");
+		projectPath =  projectInitJson.getString("path");
+		
+		assertEquals(200, projectInitResponse.statusCode());
+		assertTrue(projectInitStatus);
+		assertTrue(projectPath.contains("marklog-data-hub-develop/marklogic-data-hub/quick-start/ye-olde-project"));
+		System.out.println("projectInitResponse: "+ projectInitResponse.asString());
```
grechaw: As we discussed, these should be log statements.  use logger.debug() with a private static slf4j logger.  Understand how logback.xml works to configure the logging (don't just use info() in other words)

quick-start/src/test/java/com/marklogic/quickstart/integrationtests/EndToEndAPITest.java
```diff
@@ -0,0 +1,350 @@
...
+		System.out.println("Project Init Test Passed");
+		
+		// Tests api's to login into the quick-start application. A session cookie is received in
+		// header which has to be attached to the subsequent requests for security
+		loginResponse = utils.doLogin();
+		
+		assertTrue(!loginResponse.isEmpty());
```
grechaw: assertFalse() with a positive assertion might be more readable?

quick-start/src/test/java/com/marklogic/quickstart/integrationtests/EndToEndAPITest.java
```diff
@@ -0,0 +1,350 @@
...
+	String responseEntityTitle;
+	
+	// Flow process variables
+	JsonPath createInputFlowJson;
+	JsonPath createHarmonizeFlowJSON;
+	JsonNode mlcpOptions;
+	JsonNode runHmJsonObj;
+	Map<String, Object> runFlowOptions;
+	Response createInputFlowResponse;
+	Response runIpFlowResponse;
```
grechaw: try to be consistent using abbreviations or not.  best to keep the full names in Java.

quick-start/src/test/java/com/marklogic/quickstart/integrationtests/IntegrationUtils.java
```diff
@@ -0,0 +1,292 @@
...
+		JsonPath projectInitJson = projectInitResponse.jsonPath();
+		loginInfo.projectId = projectInitJson.getInt("id");
+		loginInfo.environment = projectInitJson.getList("environments").get(0).toString();
+		loginInfo.username = "admin";
+		loginInfo.password = "admin";
+	}
+	
+	private FlowModel createFlowModel(String entityName, String flowType, DataFormat dataFormat, 
+			CodeFormat codeFormat, boolean useEsModel, String mappingName) {
+		flowModel = new FlowModel();
```
grechaw: really weird to use an instance variable here.  Why  not just local?

quick-start/src/test/java/com/marklogic/quickstart/integrationtests/IntegrationUtils.java
```diff
@@ -0,0 +1,292 @@
...
+	
+	private String sessionID;
+	private String projectPath;
+	
+	// created this field to enable fetch the flowName that was created during the createFlow process
+	// and verify it.
+	private String flowName;
+	private int hmFlowNameCount = 1;
+	
+	private Cookie requestCookie;
```
grechaw: these all should be fields if this class is being used to encapsulate something.  I think though that as a utility class you probably don't want shared state and an object.  It's possible that these functions could all be static even ( not sure about that)

quick-start/src/test/java/com/marklogic/quickstart/integrationtests/IntegrationUtils.java
```diff
@@ -0,0 +1,292 @@
...
+				given()
+					.body(loginInfo)
+				.when()
+					.post("/api/login");
+		if(loginResponse.statusCode() == 200) {
+			sessionID = StringUtils.substringBetween(loginResponse.getHeader("Set-Cookie"), "JSESSIONID=", ";");
+		}
+		
+		// building cookie to use for subsequent endpoint requests
+		buildCookie();
```
grechaw: so this relies on object state here -- you need that cookie to be around for subsequent things.  Makes me think this is not a utility class but something that holds state for a test session.

quick-start/src/test/java/com/marklogic/quickstart/integrationtests/IntegrationUtils.java
```diff
@@ -252,7 +240,6 @@ public String getFlowName() {
 	
 	public void waitForReloadModules() {
 		try {
-			System.out.println("Waiting to deploy Modules");
```
grechaw: a logger.info message would probably be a good thing here.

quick-start/src/test/java/com/marklogic/quickstart/integrationtests/EndToEndAPITest.java
```diff
@@ -126,22 +126,21 @@ public void modelMappingE2ETests() throws JSONException, IOException, Interrupte
 		
 		assertEquals(200, projectInitResponse.statusCode());
 		assertTrue(projectInitStatus);
-		assertTrue(projectPath.contains("marklog-data-hub-develop/marklogic-data-hub/quick-start/ye-olde-project"));
-		System.out.println("projectInitResponse: "+ projectInitResponse.asString());
-		System.out.println("Project Init Test Passed");
+		assertTrue(projectPath.contains("marklogic-data-hub/quick-start/ye-olde-project"));
```
grechaw: unfortunately this assertion will still fail on windows where the path contains backslashes.
You need a system-independent assertion.  You can do this with a regex, by broadening the assertion, or by replacing backslashes with slashes before comparing.

quick-start/src/test/java/com/marklogic/quickstart/integrationtests/EndToEndAPITest.java
```diff
@@ -0,0 +1,350 @@
...
+		respFlowName = createInputFlowJson.getString("flowName");
+		respEntityName = createInputFlowJson.getString("entityName");
+		respFlowType = createInputFlowJson.getString("flowType");
+		pluginDir = createInputFlowJson.getString("plugins.pluginPath");
+		
+		assertEquals(200, createInputFlowResponse.statusCode());
+		assertTrue(utils.getFlowName().equals(respFlowName));
+		assertTrue(entityName.equals(respEntityName));
+		assertTrue(flowType.equals(respFlowType));
+		assertTrue(pluginDir.contains("/"+respFlowName+"/content.sjs"));
```
grechaw: this also fails on windows

quick-start/src/test/java/com/marklogic/quickstart/integrationtests/EndToEndAPITest.java
```diff
@@ -0,0 +1,350 @@
...
+		
+		// Tests api's to initialize the project folder and set the environment configuration
+		projectInitResponse = utils.initilizeProjectConfiguration();
+		projectInitJson = projectInitResponse.jsonPath();
+		projectInitStatus =  projectInitJson.getBoolean("initialized");
+		projectPath =  projectInitJson.getString("path");
+		
+		assertEquals(200, projectInitResponse.statusCode());
+		assertTrue(projectInitStatus);
+		assertTrue(projectPath.contains("marklogic-data-hub/quick-start/ye-olde-project"));
```
grechaw: something like this can help:

projectPath = projectPath.replaceAll("\\\\","/");
assertTrue(projectPath.contains("marklogic-data-hub/quick-start/ye-olde-project"));

quick-start/build.gradle
```diff
@@ -60,6 +60,7 @@ dependencies {
     compile("org.springframework.boot:spring-boot-starter-thymeleaf")
     compile("org.aspectj:aspectjweaver:1.8.9")
     compile("org.springframework:spring-messaging")
+    compile("io.rest-assured:rest-assured:3.0.3")
```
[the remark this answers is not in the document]
rahulvudutala: Changed compile to testCompile

quick-start/src/test/java/com/marklogic/quickstart/integrationtests/EndToEndAPITest.java
rahulvudutala: Updated to consistent variable names. Using full names instead of abbreviations.

quick-start/src/test/java/com/marklogic/quickstart/integrationtests/EndToEndAPITest.java
```diff
@@ -0,0 +1,350 @@
...
+	}
+
+	@AfterAll
+	public void tearDown() {
+		deleteProjectDir();
+	}
+
+	@Before
+	public void serverSetup() {
+		RestAssured.port = port;
```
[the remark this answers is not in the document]
rahulvudutala: RestAssured class has the port variable to be public and static. So they have not provided any methods like setPort() to set override default port.

quick-start/src/test/java/com/marklogic/quickstart/integrationtests/EndToEndAPITest.java
rahulvudutala: Checking for the existence directory instead of verifying with a partial string

quick-start/src/test/java/com/marklogic/quickstart/integrationtests/EndToEndAPITest.java
rahulvudutala: Changed it to assertFalse() for more readability.

quick-start/src/test/java/com/marklogic/quickstart/integrationtests/EndToEndAPITest.java
rahulvudutala: Removed the assert for verifying the existence of plugins directory as the pluginDir results an array of Strings. The input flow is tested by checking the FlowName and EntityName and the Http Status.

quick-start/src/test/java/com/marklogic/quickstart/integrationtests/IntegrationUtils.java
rahulvudutala: Changed a few other instance variables to local variables where instance variables are not required.

quick-start/src/test/java/com/marklogic/quickstart/integrationtests/IntegrationUtils.java
rahulvudutala: I wanted to maintain the cookie as a private instance variable instead of passing around every time I invoke a method. Same with other instance variables as well. 

If this is not a good coding practice, I will pass the parameters and make the utils class static

rahulvudutala: I have renamed the class to E2ETestsRequestHelper.Java

quick-start/src/test/java/com/marklogic/quickstart/integrationtests/IntegrationUtils.java
rahulvudutala: I have renamed the class to E2ETestsRequestHelper.Java

quick-start/src/test/resources/integration-test-data/create-entity-model.json
```diff
@@ -0,0 +1,86 @@
...
+    "description": null,
+    "primaryKey": null,
+    "required": [],
+    "elementRangeIndex": [],
+    "rangeIndex": [],
+    "wordLexicon": [],
+    "pii": [],
+    "properties": [
+      {
+        "showCollation": false,
```
grechaw: this is a minor nit, but where did these properties come from?  The values in here are not from entity services or the data hub model.

quick-start/src/test/resources/integration-test-data/input/input-es-trips.json
```diff
@@ -0,0 +1,56 @@
...
+      },
+      "info": {
+        "title": "e2eentity",
+        "version": "0.0.1"
+      }
+    },
+    "triples": [
+      {
+        "triple": {
+          "subject": {
```
grechaw: these are not valid RDF triples -- does it matter?

quick-start/src/test/resources/integration-test-data/create-entity-model.json
rahulvudutala: These properties come from UI to the rest endpoint "/api/current-project/entities/{entityName}". It has the Entity UI data along with the entity information. This data is modeled into EntityModel object in quick-start.

I gathered this data by inspecting the endpoint in the browser network tab.

quick-start/src/test/resources/integration-test-data/input/input-es-trips.json
rahulvudutala: It doesn't matter to the current tests. I have copied all the input test resources from datahub core test/resources. Probably might be useful when expanding the tests.

quick-start/src/test/java/com/marklogic/quickstart/integrationtests/E2ETestsRequestHelper.java
```diff
@@ -0,0 +1,318 @@
...
+	private void buildCookie() {
+		requestCookie = new Cookie.Builder("JSESSIONID", sessionID).setSecured(true)
+			      .setComment("session id cookie").build();
+	}
+	
+	private void createLoginCredentials(Response projectInitResponse) {
+		JsonPath projectInitJson = projectInitResponse.jsonPath();
+		loginInfo.projectId = projectInitJson.getInt("id");
+		loginInfo.environment = projectInitJson.getList("environments").get(0).toString();
+		loginInfo.username = "admin";
```
grechaw: how do you test non-admin?

grechaw:
@rahulvudutala There's already a properties file you can use for credentials in the quickstart folder.  The tests, when you create a project dir, copy these properties over into the project dir, and you can leverage other tests to load them.  

admin/admin is OK for some functionality tests, but admin does not set permissions on documents such that you can verify flows.  You'll want to always avoid using admin/admin (though you may also find that your tests don't work unless they're admin/admin in which case we need to fix something)

Issue #1176, opened 2018-07-13
Able to create maps with duplicate names


ayuwono:
QuickStart v3.1.0
develop branch
Chrome on MacOSX

Able to create maps with duplicate names. Doing so will overwrite the existing map with the same name either on the same entity or the other entity

Expected behavior: should not be able to do that, should have a check that will throw an error message

┆Status: To Do

wooldridge:
This problem should be fixed with the merge #1182.

PR #1182, opened 2018-07-17
Do not allow duplicate mapping names

wooldridge:
When duplicate name is entered, the following error text is displayed:

A mapping with this name already exists

And the Create button is disabled.


wooldridge:
Fixes #1176

PR #1185, opened 2018-07-18
DHFPROD-1125 rollback of URI conns

wooldridge:
In the mapping UI, if the user provides a non-existent URI, don’t clear connections, roll them back.

Fixes marklogic/datahub-central#856.


bsrikan:
Verified and found bug to be fixed in feature branch.
Tested around the area also was able to run input/harmonize flow.
Good to merge.

PR #1187, opened 2018-07-18
Mappings page-objects DHFPROD-886

bsrikan:
Commits include:
* Page objects for Mappings page
* Fix for failed regression tests that look for ng-reflect-model attribute


bsrikan:
These changes are for QS and dont affect core tests. Four failed tests in code per PR jenkins job seems to have failed for earlier runs as well.

Good to merge to develop IMO.
@aebadirad can you please help do the needful.

PR #1191, opened 2018-07-19
E2e/change qa input data

ayuwono:
Moving input data to qa-data folder (DHFPROD-1145) so that e2e tests are not depended on examples/online-store. On the future it will be easier for qa to put more input data for test coverage

On redeploy, sometimes it failed as the tooltip message is covering the deploy status. Using browser.sleep() will make it more consistent

Add a verification on triples plugin on document trace


quick-start/e2e/specs/create/create.ts
```diff
@@ -578,8 +578,7 @@ export default function() {
 
     it ('should redeploy modules', function() {
       flowPage.redeployButton.click();
-      browser.wait(EC.textToBePresentInElement(element(by.css('#last-deployed-time')),
-        'Last Deployed: less than a minute ago' || 'Last Deployed: 1 minute ago'));
+      browser.sleep(5000);
```
bsrikan: Hi Aries, Is there any particular reason for commenting this and add sleep in create and run tests.

bsrikan: Can we try clicking on any entity after redeployButton.click() in the flows page so the focus moves away and deploy status text is visible again.

ayuwono: Clicking on the entity will change the state of open/close entity, it might affect the later test cases because they perform opening entities

I tried to do less harmful action: browser.actions().mouseMove(flowPage.flowsTab).perform();
but redeploy still failed, the label is not showing "1 minute or less than a minute". I think there is a bug on after we have auto redeploy.

screen shot 2018-07-20 at 9 10 14 am

We can stay at browser.sleep() for now or maybe try to remove all redeploys since we have auto redeploy

PR #1196, opened 2018-07-21
Dhfprod 434 simplerroles

grechaw:
This PR contains all of the WIP that @aebadirad has bee doing to resolve model-to-model mapping.  It also contains simplifications to the users and roles.

Do not merge -- it's expected that we'll see some failures in scaffolding e2e, but resolve the other outstanding failures in jenkins today.


grechaw:
This PR should now pass.  I've addressed a final lingering issue of having more than one database client.

marklogic-data-hub/src/test/java/com/marklogic/hub/MappingE2E.java
```diff
@@ -135,6 +135,9 @@ public void setupEach() {
             }
             createMappings();
```
srinathgit: This file has been changed in https://github.com/marklogic/marklogic-data-hub/pull/1202 and these changes are included in that PR.

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/HubConfigImpl.java
```diff
@@ -914,16 +899,7 @@ public DatabaseClient newStagingManageClient(String databaseName) {
...
-        DatabaseClientConfig config = new DatabaseClientConfig(appConfig.getHost(), finalPort, appConfig.getRestAdminUsername(), appConfig.getRestAdminPassword());
-        config.setDatabase(finalDbName);
-        config.setSecurityContextType(SecurityContextType.valueOf(finalAuthMethod.toUpperCase()));
-        config.setSslHostnameVerifier(finalSslHostnameVerifier);
-        config.setSslContext(finalSslContext);
-        config.setCertFile(finalCertFile);
-        config.setCertPassword(finalCertPassword);
-        config.setExternalName(finalExternalName);
-        return appConfig.getConfiguredDatabaseClientFactory().newDatabaseClient(config);
+        return newStagingClient(finalDbName);
```
srinathgit: Can the methods public DatabaseClient newStagingManageClient(String databaseName)  and public DatabaseClient newFinalManageClient() be removed as we don't use them ?

srinathgit:
A few fields in HubConfigImpl.java are redundant after the change. Should we retain those ?
For example,
```    
private String finalScheme = DEFAULT_SCHEME;
    private boolean finalSimpleSsl = false;
    private SSLContext finalSslContext;
    private DatabaseClientFactory.SSLHostnameVerifier finalSslHostnameVerifier;
    private String finalCertFile;
    private String finalCertPassword;
    private String finalExternalName;
```

srinathgit:
These two methods use different user credentials ("admin" in case of newModulesDbClient() and "data-hub-user" in case of newStagingClient() ) but use the same "stagingSslContext", "stagingSslHostnameVerifier", "stagingCertFile", "stagingCertPassword" (and possibly "stagingExternalName"  ? ) propoerties . These properties have to be different
```
    private DatabaseClient newStagingClient(String dbName) {
        AppConfig appConfig = getAppConfig();
        DatabaseClientConfig config = new DatabaseClientConfig(appConfig.getHost(), stagingPort, getMlUsername(), getMlPassword());
        config.setDatabase(dbName);
        config.setSecurityContextType(SecurityContextType.valueOf(stagingAuthMethod.toUpperCase()));
        config.setSslHostnameVerifier(stagingSslHostnameVerifier);
        config.setSslContext(stagingSslContext);
        config.setCertFile(stagingCertFile);
        config.setCertPassword(stagingCertPassword);
        config.setExternalName(stagingExternalName);
        return appConfig.getConfiguredDatabaseClientFactory().newDatabaseClient(config);
    }

    public DatabaseClient newModulesDbClient() {
        AppConfig appConfig = getAppConfig();
        DatabaseClientConfig config = new DatabaseClientConfig(appConfig.getHost(), stagingPort, appConfig.getRestAdminUsername(), appConfig.getRestAdminPassword());
        config.setDatabase(appConfig.getModulesDatabaseName());
        config.setSecurityContextType(SecurityContextType.valueOf(stagingAuthMethod.toUpperCase()));
        config.setSslHostnameVerifier(stagingSslHostnameVerifier);
        config.setSslContext(stagingSslContext);
        config.setCertFile(stagingCertFile);
        config.setCertPassword(stagingCertPassword);
        config.setExternalName(stagingExternalName);
        return appConfig.getConfiguredDatabaseClientFactory().newDatabaseClient(config);
    }

```

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/HubConfigImpl.java
grechaw: Yes absolutely.

grechaw:
On your first comment, I didn't actually remove the whole final appserver, just made it so DHF doesn't actually use it.  It would be nice to remove the whole set of properties, but I haven't thought through what that means for compatibility.

On the second comment, you're right, we need new properties.  It was a shortcoming to overload staging for everything.  I'll address that and get back to you

grechaw:
So it turns out I cannot remove all the fields for the final appserver.  Although DHF no longer uses the connection it still is responsible for provisioning the final appserver, and it cannot do that without the various fields and codes.

As far as the modules client does, I think we are ok for now but will need to change something later on.  My rationale is that the modules client uses the staging appserver, and this PR suggests that everything the DHF does should be via a single set of credentials.  For QS scenarios these credentials must be hub-adminny.  For runtime flows, these credentials must be data-hubby.  But in either case, the modules client should be using the same credentials as the staging client.... would that not mean that the block should stay the same as they are now?

aebadirad:
Also consider final app server also needs its own modules/triggers/schemas db set up with it too. These may or may not overlap with customers needs. We may want to address this later as part of an optional 'put your consumption application here', which then serves the data using 'final' tdes/triggers/modules etc that DHF doesn't wipe on redeploy.

Either way I'm willing to bring this in to develop, then branch and fix the remaining tests. We've had this branch out too long and it's pretty large/core.

grechaw:
on final having its own stuff, yes, that's so.  Having DHF not use final appserver is a good first step.  I do need to address what @srinathgit explained to me here a few moments ago -- the modules client should not use alternate credentials.

I'm fine with taking that as a follow-on bug too if you want to merge this now.  What's your feeling about getting this merged first @srinathgit ?

Issue #752, opened 2018-02-20
QuickStart won't connect to HTTPS-enabled App Services


mikrovvelle:
1. The issue

Running QuickStart (2.0.4) on a local instance, I can't connect with HTTPS enabled on App Server ports, and the following `gradle-local.properties`:

```properties
mlAdminScheme=https
mlManageScheme=https
mlAppServicesScheme=https
mlStagingScheme=https
mlFinalScheme=https
mlTraceScheme=https
mlJobScheme=https

mlAdminSimpleSsl=true
mlManageSimpleSsl=true
mlAppServicesSimpleSsl=true

mlSSLHostname=local.local
mlSSLCertificateTemplate=local.local
```

(those last two properties are project-specific)

The config is found by QuickStart, you are prompted to enter a username/password, and then it fails with the following error:

> Cannot connect to MarkLogic at localhost. Are you sure MarkLogic is running?

(yes, I'm sure) If I debug from the top of the stack-trace, I see at this point:

```java
        return this.getAuthenticationManager().authenticate(authRequest);
```

`getAuthenticationManager()` returns a `MarkLogicAuthenticationManager` with a `restConfig` and the following three fields:

* `host` = `localhost`
* `restPort` = 8000
* `scheme` = `http`

It appears that my `mlAppServicesScheme=https` were ignored by QuickStart. Other gradle commands from the command line work fine. Just QuickStart can't connect.

      1. Tech details

        1. Which Operating System are you using?

MacOS (client), CentOS Docker serving MarkLogic

        1. Which version of MarkLogic are you using?

9.0-3.1

        1. Which version of the Data Hub Framework are you using?

2.0.4

        1. Is this a QuickStart UI bug? If so, which browser are you using?

This is a QuickStart UI bug. tested on Safari 11.0.1 and Crome 64.0.3282.167

[https-stacktrace.txt](https://github.com/marklogic-community/marklogic-data-hub/files/1739978/https-stacktrace.txt)

┆Status: New

sashamitrovich:
This is a critical issue for some customers, can confirm.

dmcassel:
should be fixed in develop branch.

dmcassel:
@ayuwono please test with the dev branch

aebadirad:
Fixed with release of 3.0.0

wattsferry:
I'd like to request this bug to be re-opened.  

[...]

PR #1198, opened 2018-07-23
fixing ssl not working from quickstart

paxtonhare:
fixing #752 - the ssl issue with quickstart


grechaw:
Aries can you take a look to see how this PR fares against existing qs test, and whether we will need some more coverage?

ayuwono:
@grechaw I'll look into it, I haven't worked with SSL on QuickStart, will tag along with Srinath to pick his brain

grechaw:
That's fine with me

PR #1203, opened 2018-07-24
E2e/typeahead

ayuwono:
Added e2e tests for typeahead capability on mapping page
- Test the typeahead for partial text on source property names
- Test the typeahead on source types
- Test to map properties
- Test the saved properties are retained on the map

Added a qa data with long properties and various type of data

Some functions on flows are added to accommodate the mappings
- Added a new run input flow function so we can select the folder name and input file type. The old function is hardcoded to select products folder. We will need to change the old function to the new one so testing is more flexible
- Added a new create input flow function. We need to separate input and harmonize flow creation functions as over the time we are introducing new capabilities such as mapping. Separating these functions will make it easier to develop the tests


ayuwono:
1) The reason I have a separate suite is because it needs different input data that has long properties
2) Same as above as we need a new data set. But maybe we can incorporate "Products" entity here?
3) Sure we can add those functions

ayuwono:
@bsrikan, we can change this typeAhead suite as a spec of mappings suite when it's available. We can reorganize that

I don't think right now I can use Order or Product data to test typeahead as they only have string data set

Using Order entity will be buggy as it still has issue on entity type property

I added the verification page objects on mappings

ayuwono:
I changed the suite name to mappings

This spec is named typeAhead because it's focused to test the typeahead capability
Next, we can add more specs related to mappings and run the flow with mapping
```
import runTypeAhead from './typeAhead'
import runSomething from './something'

export default function () {
  describe('mappings', function () {
    runTypeAhead();
    runSomething();
  })
}

```

PR #1209, opened 2018-07-25
DHFPROD-1131 fix to get tests green

aebadirad:
To get tests going green and resolve the single quote string escape issue


aebadirad:
meeerrgggging

PR #1214, opened 2018-07-26
3.x develop

grechaw:
Since the develop branch is finally green, I suggest we put this into 3.x-develop as a clean base from which to prepare ramp toward release candidate.


grechaw:
wrong PR !

PR #1218, opened 2018-07-26
E2e/harmonize mappings

ayuwono:
- Extended mapping typeahead test to run harmonize flow with mapping, along with final data verification
- Removed old create flow functions. Separating this function into createInputFlow() and createHarmonizeFlow(). The new createHarmonizeFlow() can use the mapping
- Removed old run flow function. The new runInputFlow() has some parameters to use different input data folder, input file type, and suffix uri
- Added some page objects for viewer page to verify the mapped harmonized data
- Added a function on flow to recognize if entity disclosure is collapsed or not, so clicking on it will be more consistent

### DIFF
--- a/quick-start/e2e/specs/mappings/typeAhead.ts
+++ b/quick-start/e2e/specs/mappings/typeAhead.ts
@@ -30,6 +30,11 @@ export default function() {
           toBe(true, 'Load TypeAhead' + ' is not present');
       });
 
+      it ('should redeploy modules', function() {
+        flowPage.redeployButton.click();
+        browser.sleep(5000);
+      });
+
       it ('should run Load TypeAhead flow', function() {
         flowPage.clickEntityDisclosure('TypeAhead');
         browser.wait(EC.elementToBeClickable(flowPage.getFlow('TypeAhead', 'Load TypeAhead', 'INPUT')));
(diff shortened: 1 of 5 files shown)


quick-start/e2e/specs/mappings/typeAhead.ts
```diff
@@ -140,7 +140,7 @@ export default function() {
         mappingsPage.saveMapButton().click();
         browser.wait(EC.elementToBeClickable(mappingsPage.entityMapping('MapTypeAhead')));
         //flicker bug, sleep will be removed once it's fixed
```
grechaw: an actual bug in another project?  if so a link will help us maintain in the future

ayuwono:
Please hold merging this, need to commit a fix on verification

quick-start/e2e/specs/mappings/typeAhead.ts
ayuwono: The mapping page is flickering/reloading. Filed a bug: https://project.marklogic.com/jira/browse/DHFPROD-1161

aebadirad: I think this is a result of how the maps service is tied to entities changing, reloading the model. We should fix this.

PR #1227, opened 2018-07-30
DHFPROD-1010 Added error message for invalid properties file

akshaysonvane:
Throws relevant error message.


marklogic-data-hub/src/main/java/com/marklogic/hub/flow/impl/FlowImpl.java
```diff
@@ -208,7 +208,7 @@ public String serialize() {
             return finalWriter.toString().replaceFirst("<!--", "\n<!--").replaceFirst("-->", "-->\n");
         }
         catch(Exception e) {
```
grechaw: Throwing just a string message for any Exception seems very greedy and like it might obscure other issues.  Are you sure about this?  You can always add another block.

akshaysonvane:
The only place I can think of that might cause an exception is the "toString()" method which will cause a NPE. I've added another catch block for NPE.

Issue #1129, opened 2018-07-03
mlDeployRoles does not deploy roles from user-config


sparnell:
Roles defined in user-config/security/roles are not deployed when running mlDeployRoles regardless of where they are stored. The only way to get roles created is to run mlDeploy.

marklogic-builder:
➤ Srinath Sambasubramanian commented:

I can reproduce this issue in the latest 'develop' build.

grechaw:
I have figured out why this happens.  The data hub gets a command list from ml-app-deployer and replaces the role and user deployment with its own implementation.

I'm seeing what can be done to restore the link between the ml gradle tasks and the hub ones.

marklogic-builder:
➤ Rahul Vudutala commented:

Roles defined in user-config/security/roles can now be deployed using "./gradlew mlDeployRoles" command.

Tested the command on DHF 3.1.0 (latest commit: e3942a0e933ca3595caeeab25d3e1f5536ad4ed0) against ML 9.0-6.2

PR #1228, opened 2018-07-30
Dhfprod 434 preinstall security

grechaw:
This PR makes some significant changes to the gradle plugin.

1. it makes mlDeployRoles and mlDeployUsers and mlDeploySecurity work #1129 
2. it upgrades ml-gradle and ml-app-deployer 
3. addresses bootstrapping by installing roles and users before preInstallCheck
4. adjusts plugin tests to all run as hub-admin-user (rather than admin)
5. fixed test cases to adjust for different permissions in assertions.


grechaw:
It looks like #1 is a simple change to getModulesDbClient().  Of course that's failing in my tests, so it's not as simple as i might think.

For number two, quick-start avoids using properties files for config -- the whole thing should be driven by logins in the UI.  I know there is an 'advanced' login but i don't know what it does.  But this should work (don't know whether it does).

1.  if you have to install a data hub, use admin to do it.  the bootstrapping just requires admin.
2.  after that step, you should be able to do everything needed for quickstart with hub-admin-user, including run flows, install code, etc.  just not uninstall hub.

srinathgit:
I am getting compilation error in LoadHubModules.execute() at line 82
        modulesLoader.addFailureListener( (throwable, client) -> {
            // ensure we throw the first exception
            if (caughtException == null) {
                caughtException = throwable;
            }
        });

srinathgit:
Hi @grechaw ,

But hubConfig.newAppServicesClient() uses admin/admin by default unless 'mlAppServicesUsername' and 'mlAppServicesPassword' are not overridden in gradle.properties.

```
public DatabaseClient newAppServicesDatabaseClient(String databaseName) {
DatabaseClientConfig config = new DatabaseClientConfig(getHost(), getAppServicesPort(), getAppServicesUsername(), getAppServicesPassword());
```

grechaw:
I think some other defaulting is going on.   In order to avoid such a bug I've set my admin username and password to non default values these day, and the tests all pass.  So I think its getting the property now ultimately from mlSecurityUsername, which is appropriate.

PR #1231, opened 2018-07-31
Dhfprod 872

aebadirad:
fix handling of xml comments and processing instructions in harmonize flows


grechaw:
So you'll need to add key updates for this change, it seems.

Issue #1144, opened 2018-07-06
Able to save an entity with duplicate property names


bsrikan:
#### The issue
Quickstart allows saving an entity with duplicate properties

What behavior are you expecting?
Quickstart should throw an error while saving entities with duplicate properties

#### Tech details

Which Operating System are you using?
Mac
Which version of MarkLogic are you using?
9.0nightly
Which version of the Data Hub Framework are you using?
3.1
Is this a QuickStart UI bug? If so, which browser are you using?
Chrome 66

marklogic-builder:
➤ Srikanth Balasubramanian commented:

Entities get saved with duplicate properties. There are tests failing in regression because of this issue. Unless page is refreshed, duplicate properties persist and on refresh the last update gets persisted. This is incorrect behavior and we should not overwrite properties when they are duplicate. Indicating user with an error message while trying to save entity with dup props is a proper way to handle this issue. Let the user decide what the value of properties should be.

marklogic-builder:
➤ Scott Brooks commented:

PR in review:  [https://github.com/marklogic/marklogic-data-hub/pull/1238](https://github.com/marklogic/marklogic-data-hub/pull/1238)

marklogic-builder:
➤ Scott Brooks commented:

Approved and merged to develop branch.  Ready for QA.

marklogic-builder:
➤ Srikanth Balasubramanian commented:

Verified with latest develop branch. Working as expected. "must be unique" text is now added to notify users when there is a dup property name and Save button gets disabled.

PR #1238, opened 2018-08-01
Able to save an entity with duplicate property names

clockworked247:
Issue #1144 

Validation now checks to ensure property name is not empty, unique & without spaces.  Also, I updated the e2e test where it checks for error message string.

### DIFF
--- a/quick-start/e2e/specs/create/create.ts
+++ b/quick-start/e2e/specs/create/create.ts
@@ -535,7 +535,7 @@ export default function() {
       entityPage.getPropertyName(lastProperty).sendKeys('test white space');
       // verify the error message on white space in property name
       let errorMessage = entityPage.errorWhiteSpaceMessage;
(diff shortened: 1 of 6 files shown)


ayuwono:
@clockworked247 , thanks for changing the e2e test. Actually there is another area of the test that has duplicate property scenario, I've already addressed in this commit on my PR:
https://github.com/marklogic/marklogic-data-hub/pull/1239/commits/5ea2caf201cfb6c8e40d8fb3c066165b122e31d5

clockworked247:
@ayuwono I just updated the check of the error message but did not add logic to the test to attempt to add a duplicate property.

clockworked247:
Hmmm, build failure...  no info on why either...

ayuwono:
Yes, I have to change the test. Before, you can save the entity even they have duplicate prop, but right now, the save button is disabled.

clockworked247:
@grechaw Thanks!

PR #1249, opened 2018-08-06
DHFPROD-1164 add valMaxLen

wooldridge:
Limit the displayed length of source value in mapping UI so it doesn't overflow source container.

Fixes DHFPROD-1164


wooldridge:
@aebadirad, thanks! Yes, there is a title attribute present that shows the selected value on hover.

wooldridge:
@ayuwono, this is a minor code change, putting a var back that I believe got inadvertently removed. OK to merge from QA's standpoint?

ayuwono:
@wooldridge, testing it right now

PR #1255, opened 2018-08-07
Dhfprod 1095

aebadirad:
This satisfies the bug, however, xml traces that use json sources still show up blank when being returned. They save properly, they just are mistranslated in the json transformation.


marklogic-data-hub/src/main/resources/scaffolding/harmonize/sjs/main.sjs
```diff
@@ -1,6 +1,6 @@
 // dhf.sjs exposes helper functions to make your life easier
 // See documentation at:
-// https://github.com/marklogic/marklogic-data-hub/wiki/dhf-lib
+// https://marklogic.github.io/marklogic-data-hub/docs/server-side/
```
grechaw: oof good catch

marklogic-data-hub/src/server-side/impl/trace-lib.xqy
```diff
@@ -302,7 +302,7 @@ declare %private function trace:write-error-trace(
             "/" || $trace/*:trace/*:traceId || $extension,
             $trace,
             xdmp:default-permissions(),
-            ("trace", $trace/*:trace/*:type)
+            ("trace", $trace/*:trace/*:flow-type, $trace/*:trace/*:flowType)
```
grechaw: This looks like something we missed during a name change or format introduction.  time to introduce some ml-unit-test?

grechaw: Two test failures -- assertions around binary nodes.  this bugfix looks legit, the tests' output has changed in what looks like a non-problematic way.  but they need to change to do the merge

ayuwono:
Application is not hang anymore with invalid credentials

But can we change the error message?
Response with status: 401 OK for URL: http://localhost:8080/api/login

ayuwono:
It's good now with the latest commit

PR #1257, opened 2018-08-07
HubInfo Task to display hub information in string format instead of object hashcode

rahulvudutala:


grechaw:
Ugh, why didn't we just make the HubInfo toString() method return getInfo()?  That would be a better solution... nobody needs the the existing serialization.

grechaw:
What issue number is this tied to anyway?

rahulvudutala:
This is tied to DHFPROD-791. The hubInfo task still displays the HubConifg object hashcode.
I will create a toString() method in HubConfigImpl class and invoke the getInfo() method.

PR #1267, opened 2018-08-13
Work In Progress -- Dhfprod 1192 enode modules

grechaw:
This PR is to give my initial work a whirl with jenkins.  My hunt-and-peek of the tests is basically working in core now, and I'd like to see a complete run and report.


grechaw:
I'm closing this PR.  I don't think it works to put the arch back to the 2.x style, as the SJS/XQY codepaths took advantage of breaking out of that approach

PR #1269, opened 2018-08-14
Running tests with servers set to cert-auth/ssl,flows as 'data-hub-user'

srinathgit:
**This PR is not ready to be merged yet**. 

This PR enables
1. Running tests with all servers set to cert-auth/ssl in marklogic.com domain
2. Running the installation as 'hub-admin-user' and flows as 'data-hub-user'.


marklogic-data-hub/build.gradle
```diff
@@ -247,46 +247,70 @@ ext {
...
-//	   	        "mlManageScheme=https\n" +
-//	    		"mlJobAuth=certificate")
-//	}
-//}
+def propFile = file("gradle.properties")
+String fileContents = propFile.text
+boolean sslRun = Boolean.parseBoolean(System.properties['ssl'])
+boolean certAuth = Boolean.parseBoolean(System.properties['certauth'])
+def props = new Properties()
+propFile.withInputStream { props.load(it) }
```
grechaw: All these properties should just be available to build.gradle without having to read the file.
the variable names should just be `certauth`, `ssl` etc.  Maybe you were having trouble reading them that way?

marklogic-data-hub/src/test/java/com/marklogic/hub/HubTestBase.java
```diff
@@ -340,67 +360,110 @@ protected void disableTracing() {
 
 		});
     }
-
-    protected HubConfig getHubConfig() {
-        return getHubConfig(PROJECT_PATH);
+    //'requireAdmin' indicates whether hubconfig uses 'hub-admin-user' or not ('data-hub-user' is used).
+    protected HubConfig getHubConfig(boolean requireAdmin) {
```
grechaw: Can you think of a different way to expose this than as a boolean flag?  I'm not sure if it's easy, but an overload like getAdminHubConfig() might be cleaner looking.  It should be very clear to the future reader whether any particular connection is using admin or not.

marklogic-data-hub/build.gradle
srinathgit: 1. The values for 'certauth' and 'ssl' are provided in command line . For example, ./gradlew marklogic-data-hub:test -Dcertauth=true (or -Dssl=true). So, I am reading them from System.properties
2. I think I can use other properties like 'mlHost', 'mlSecurityUsername' and 'mlSecurityPassword' directly. I will make that change.

grechaw:
It looks like, from your latest commit, that you didn't need to actually switch on the boolean flag in any of the tests.  I was a little surprised, like in updateIndexes().  The tests will tell tho

srinathgit:
1. I initially thought that I needed the boolean flag but looks like getHubAdminConfig() in cases of installation, scaffolding, creating flows etc and getHubFlowRunnerConfig() for running flows.
2. I am not clear which test you are referring to when you said updateIndexes().

grechaw:
Looks like jenkins is happy.  I'm happy with your changes.

grechaw:
When you say 'servers set to cert-auth/ssl' is that new testing or have you ported all the ssl/tls tests to cert auth?

srinathgit:
This check-in would enable running the dhf core tests(EndToEndFlowTest, CoreTestSuite, ScaffoldingE2E ,MappingE2E) where all the application servers (Manage, Admin, App-Services, data-hub-staging/final/job) are ssl enabled (with digest auth) or are set to certificate based authentication .

grechaw:
would enable == will run them that way?
I'm wondering whether we have a profusion of new tests.  the jenkins job run didn't seem any longer than before though.

srinathgit:
1. In order to run the tests , we would have to modify the PR jenkins job (replace the current ./gradlew marklogic-data-hub:test  to ./gradlew marklogic-data-hub:test  -Dcertauth=true or -Dssl=true). So, current jenkins run is the same as we used to before.
2. I am trying to setup a test jenkins job to run the tests with cert based auth and ssl

grechaw:
excellent, thanks for that clarification

srinathgit:
@grechaw ,

So, the tests have run fine in jenkins. The PR can be merged.

1. Plain text connections (https://jenkins.marklogic.com/job/dhf-core-develop-rh7-PR/1070/)
2. Servers set to cert-based auth (https://jenkins.marklogic.com/view/DHF-9.0-Server-All/job/NO_CI_dhf-core-develop-cert-ssl/4/)
3. Servers set to ssl with digest authentication (https://jenkins.marklogic.com/view/DHF-9.0-Server-All/job/NO_CI_dhf-core-develop-cert-ssl/5/)

PR #1278, opened 2018-08-20
DHFPROD-1202 add new database name settings

akshaysonvane:
Add Staging Modules, Staging Triggers, Staging Schemas, Final Modules, Final Triggers, and Final Schemas database names to advanced settings. Remove Modules, Triggers and Schemas.
Fixed scaffolding for gradle.properties file


wooldridge:
@akshaysonvane, I updated my 9-branch ML server and tried again. It seemed to finish without error but with a couple problems:

- After first running through the 41 steps, it printed "Installation Complete" in the browser and then proceeded to run the 41 steps again, and then finished with "Installation Complete" again. 
- After completing installation, the "FINISHED" button is not active. There is no way for me to proceed. I can click "BACK" to return to the authentication screen, and from there I can move forward and get to the Dashboard, which looks like everything is OK.

image

ayuwono:
Actually, when you wait long enough, it will proceed to Dashboard, it seems to be stuck at 56%. But, regardless it's still a bug, I also thought that installation is not complete.

PR #1284, opened 2018-08-21
Created example of a fully disconnected (offline) zip distribution of a DHF project

derms:
Based on ml-gradle example at

https://github.com/marklogic-community/ml-gradle/tree/master/examples/disconnected-project-using-plugins-and-gradlew


grechaw:
This is awesome @derms !

We are in the process of reworking project structure to be compatible with ml-gradle.  It seems this work may affect how this project needs to be structured for 4.0.  As part  of upgrade testing we can take a look at this example and upgrade it to see where we are.  It will take a few days before we're able to evaluate and get to that point.

derms:
ah ok cool. maybe what I can do is merge this example into the 3.x-develop branch and create another example for the version 4.0 structure. What do you think?

grechaw:
Seems like a reasonable path.  Our branches aren't set up for that right now, but as the 4.0 release is prepared we'll have a spot to merge it as is

PR #1292, opened 2018-08-24
fixed loading of ml-modules to final-modules

akshaysonvane:


grechaw:
this looks like it will fix lingering issue in DataHubInstallTest

PR #1298, opened 2018-08-28
Upgrade protractor to version 5.4.0

ayuwono:
This is to fix the windows issue on 3.x-develop

[10:59:59] I/update - chromedriver: unzipping chromedriver_2.41.zip
path.js:28
    throw new TypeError('Path must be a string. Received ' + inspect(path));
    ^

TypeError: Path must be a string. Received undefined


grechaw:
this needs to be in develop too right?

ayuwono:
No, develop already has the correct version 5.4.0

PR #1306, opened 2018-08-30
Added gradle property reference for DHF Docs

akshaysonvane:


kcoleman-marklogic:
The sidenav config is in _data/sidenav.yml. It's pretty straightforward to add something.  Generally speaking, you're just reusing the title and permalink info from the config info at the top of the markdown.

akshaysonvane:
Okay, got it. I'll create a new PR with changes in the sidenav.
Thanks @kcoleman-marklogic and @grechaw

PR #1317, opened 2018-09-04
WIP - Dhfprod 1192 upgrade

grechaw:
This PR just to get Jenkins working on the upgrade branch.


marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/DeployHubAmpsCommand.java
```diff
@@ -56,7 +57,10 @@ public void execute(CommandContext context) {
         //new AmpsInstaller(securityStagingClient).installAmps(stagingModulesDatabaseName);
         ServerEvaluationCall call = installerClient.newServerEval();
         try (InputStream is = new ClassPathResource("installer-util/install-amps.xqy").getInputStream()) {
-            call.xquery( new InputStreamHandle(is));
+            String ampCall = IOUtils.toString(is, "utf-8");
+            is.close();
```
grechaw: this should actually be auto-closed with the try-with-resources statement.

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/HubConfigImpl.java
```diff
@@ -116,6 +116,9 @@
     private String hubAdminRoleName = DEFAULT_ADMIN_ROLE_NAME;
     private String hubAdminUserName = DEFAULT_ADMIN_USER_NAME;
 
+    //we assume DHF 2.x unless otherwise told, our earliest 'from' version
+    private String DHFVersion = "2.0.0";
```
grechaw: this is repeated code somewhere else.   I guess the DHF Version may very well be different from jarVersion?  but why would that be.  This smells I guess

PR #1334, opened 2018-09-08
Workaround CMA payload issues

ndw:
This cleans up the payloads so that they're idempotent.

### DIFF
--- a/marklogic-data-hub/src/main/resources/hub-internal-config/configurations/04-amps.json
+++ b/marklogic-data-hub/src/main/resources/hub-internal-config/configurations/04-amps.json
@@ -1,6 +1,6 @@
 {
   "config": {
-    "amps": [
+    "amp": [
       {
         "namespace": "",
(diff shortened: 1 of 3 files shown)


marklogic-data-hub/src/main/resources/hub-internal-config/configurations/04-amps.json
```diff
@@ -1,6 +1,6 @@
 {
   "config": {
-    "amps": [
+    "amp": [
```
grechaw: did you try this?  As I recall I the schema validation prevented me from submitting this with 'amp', one of those plural issues

grechaw: works -- i think when i first added the plural, it was because i had not installed the modules db yet.  with sec* builtins that's an error, with CMA it's silent fail.  So thanks for this pR, it's all good.

PR #1335, opened 2018-09-10
Added undeployAll to undeploy both final and staging

akshaysonvane:


ayuwono:
QS sanity check if fine on this branch

PR #1345, opened 2018-09-14
Greedier pattern group match for runWriter signature update

aebadirad:


bsrikan:
This can be merged to develop. I have verified the fix and it is working as expected.

PR #1354, opened 2018-09-18
DHFPROD-1299 server-version dependent amps install

grechaw:
This PR partially reverts the update to use CMA.  Since the CMA method for amps only works with 9.0-6 and later, we check the server version in this patch to switch on server version.
If the server version is earlier than 9.0-5, a configuration error is now thrown.


ayuwono:
Tested on QuickStart UI installation on different server versions:
- 9.0-6: installation & uninstallation succeed
- 9.0-5: installation & uninstallation succeed
- 9.0-4: installation not allowed, which is expected. The UI is displaying this nice error on pre-install check
screen shot 2018-09-18 at 2 51 54 pm

Check is good on QuickStart UI

grechaw:
Thanks aries!

PR #1374, opened 2018-10-02
Fixes for DHFPROD-1384, DHFPROD-1388

srinathgit:
I am deploying staging and final db payloads only when updateIndexes() is called. Not sure if I have to include job db as well . In DHF , the indexes for job db are created during hub installation but it doesn't come out of the box in DHaaS. I can add that as well if required.


grechaw:
Superceded by other PR

PR #1376, opened 2018-10-03
Bugfix/dhfprod 1384

grechaw:
This change helps with DHS deployments by making index-only deployments work.  

I think this PR has a regression for on-premises DHF so this is still a WIP.


marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/DeployHubDatabaseCommand.java
```diff
@@ -79,10 +88,31 @@ public void execute(CommandContext context) {
...
+                        payloadJson = (ObjectNode) mapper.readTree(payload);
+                    } catch (IOException e1) {
+                        throw new DataHubConfigurationException(e);
+                    }
+                    // for DHS we have to remove some keys
+                    logger.warn("Deploying indexes only to a provisioned environment");
+                    payloadJson.remove("schema-database");
+                    payloadJson.remove("triggers-database");
+                    SaveReceipt receipt = dbMgr.save(payload);
+                    // no forest deploy required for DHS
```
srinathgit: I think you meant this line to be SaveReceipt receipt = dbMgr.save(payloadJson.toString());
I made that change and ran  and I still get HTTP 403. As you can see the payload doesn't have reference to schemas and triggers db but still fails

23:04:58.694 [main] ERROR c.m.r.util.MgmtResponseErrorHandler - Logging HTTP response body to assist with debugging: {"errorResponse":{"statusCode":"403", "status":"Forbidden", "messageCode":"", "message":"You do not have credentials to access /manage/v2/databases/data-hub-FINAL/properties ."}}
23:04:58.694 [main] ERROR c.m.m.r.databases.DatabaseManager - Error occurred while sending PUT request to /manage/v2/databases/data-hub-FINAL/properties; logging request body to assist with debugging: 
{"path-namespace":[{"prefix":"es","namespace-uri":"http://marklogic.com/entity-services"}],"range-path-index":[{"collation":"http://marklogic.com/collation/codepoint","invalid-values":"reject","path-expression":"//*:instance/Manager/title","range-value-positions":false,"scalar-type":"string"},{"collation":"http://marklogic.com/collation/codepoint","invalid-values":"reject","path-expression":"//*:instance/Employee/name","range-value-positions":false,"scalar-type":"string"}],"range-element-index":[{"collation":"http://marklogic.com/collation/en","invalid-values":"reject","localname":"salary","namespace-uri":null,"range-value-positions":false,"scalar-type":"decimal"}],"database-name":"data-hub-FINAL","uri-lexicon":true,"triple-index":true,"collection-lexicon":true}

grechaw: FlowDeveloper I believe does not have privileges for final.
So if this behavior is correct, then is the issue one of not catching the error?

ehennum: flowDeveloper has the same indexing privileges on staging, jobs, and final.

ehennum: I notice the "database-name" property in the payload.

grechaw: I'll investigate some more.  the 'database-name' property should be stripped out by the changes I made in ml-app-deployer... I must have some mistake in that part of the code for it to be still leaking through.

grechaw: @srinathgit yes, looks like somethings wrong in there doesn't it.  shame on me for writing without a verifying test.

grechaw:
This PR is now ready for review and merge @aebadirad yay

PR #1389, opened 2018-10-05
MLUI-142 : resizable component story 

clockworked247:
MLUI-142 - Resizable Scoping, Refactoring, Regression & Story


alexsmr:
I do not see resizable.component.scss moved to the same location as the component itself. Have you left it behind?
Just a note: we better prefix PR with story name as well for everyone else on DHF team.

alexsmr:
Please fix the conflics

PR #1395, opened 2018-10-05
MLUI-145 refactored select component and added select story

brucean52:


quick-start/src/main/ui/app/shared/stories/components/select/select.component.stories.ts
```diff
@@ -0,0 +1,99 @@
...
+        }
+    `]
+})
+class StoryCardComponent {
+
+    @Input() width = 500;
+    @Input() height = 200;
+
+    constructor() {}
+}
```
alexsmr: StoryCardComponent exists on ../../utils/ level. Please use the one from there. See in Scott's PR how it is done.

alexsmr:
One more thing, have you run regression after the refactoring (it is one part of the task)? I know it is very unstable now, but if you'd be able to compare the regression result on marklogic/mlui-integration before and after your changes, that will do.

PR #1416, opened 2018-10-09
Hub modules are now loaded before all other modules

rjrudin:
This fixes an issue where, when a DHF4 project depends on a library like smart-mastering-core or marklogic-unit-test via ml-gradle's "mlRestApi" Gradle configuration, the library fails to be loaded. That's due to the library having REST extensions in it which must be loaded via /v1/config/*. But because the hub modules haven't been loaded yet, those calls fail because the staging server can't find its rewriter. 

So this fix simply ensures that the command for deploying hub modules is executed before the command for deploying all other modules.


grechaw:
Something's wrong with the integration with jenkins, as this PR is actually passing fine

kkanthet:
@grechaw we added a PR job for GUI tests on QS that job failed.. i will remove it until QS is stable.

grechaw:
Pulling this in as preparation for @aebadirad fixing some of the other issues with ml-deploy

PR #1424, opened 2018-10-10
MlUI-122: folder-browser component refactoring and story creation

alexsmr:


alexsmr:
Thanks, I missed that.

Issue #1062, opened 2018-06-07
Online shopping tutorial repository code missing an element index in Order entity


danielholgate:
DHF Quickstart 3.0.0

Tutorial code at https://github.com/marklogic/marklogic-data-hub/tree/develop/examples/online-store
is slightly incorrect and is missing an index.

Problem is that running the tutorial from the existing "Finished Online Shopping Hub Example" tutorial code fails at the Harmonize Orders stage (the error shown in Job is http 400. MarkLogic server log error is here:
[ml_server_log.txt](https://github.com/marklogic/marklogic-data-hub/files/2080977/ml_server_log.txt)

This can be fixed in the Quickstart UI by simply clicking the lightening bolt but it should be fixed in the repository code.  The entity should look like this ( "elementRangeIndex" : [ "id" ] is currently missing)

{
  "info" : {
    "title" : "Order",
    "version" : "0.0.1"
  },
  "definitions" : {
    "Order" : {
      "primaryKey" : "id",
      "required" : [ ],
      "elementRangeIndex" : [ "id" ],
      "rangeIndex" : [ "id" ],
      "wordLexicon" : [ ],
      "properties" : {
        "id" : {
          "datatype" : "string",
          "collation" : "http://marklogic.com/collation/codepoint"
        },
        "price" : {
          "datatype" : "decimal"
        },
        "products" : {
          "datatype" : "array",
          "items" : {
            "$ref" : "#/definitions/Product"
          }
        }
      }
    }
  }
}

grechaw:
Thanks @danielholgate , well reported.

popzip:
does anyone know if this is fixed in 4.0? does QA test tutorial? @kkanthet ?

grechaw:
@rahulvudutala did a spot check of examples, not a full run-through.

rahulvudutala:
I didn't make any change for elementRangeIndex in the Order Entity for 4.0.1 release. I will test it and update the example if there are any issues.

grechaw:
thanks @rahulvudutala

rahulvudutala:
@popzip The issue persists. This is not fixed in 4.0.1. I have made the change and will create a PR

grechaw:
The issue is a bug/gap/enhancement though.   Although quick-start deploys changes to indexes there's no way to do it from gradle without redeploying more than needed.

aebadirad:
Should be fixed now with the release of 4.0.2

[...]

PR #1427, opened 2018-10-11
DHFPROD-1401: mlUpdateIndexes gradle task

rahulvudutala:
This PR has commits for
* DHFPROD-1401: mlUpdateIndexes gradle task
* Added range-path index for Order entity in online-store example (Issue #1062)
* gradle-tasks-demo example


rahulvudutala:
Looks like None of the tests are probably using the Order entity. But its good to have a unique name for the entity so that the tests doesn't fail

PR #1431, opened 2018-10-11
stabilize entity disclosure

ayuwono:
add additional check on entity disclosure


ayuwono:
Based on https://jenkins.marklogic.com/view/DHF-9.0-Server-All/job/dhf-QS-develop-rh7-Commit/62/ (same fixes on this PR), it's green so check is passed. The failed check here is from dhf-core-pr.

PR #1435, opened 2018-10-12
RFC: HubWatchTask now extends ml-gradle's WatchTask

### LINKED ISSUE #1429: mlWatch doesn't load from src/main/ml-modules
This is on DHF 4.0.1. I believe it's just loading from "plugins". "mlWatch" is overwritten by the DHF plugin to use its own hub-specific task. 

I did the following on a project so load from src/main/ml-modules in the meantime (so I can quickly load marklogic-unit-test modules as I write them):

    project.tasks.replace("mlWatch", com.marklogic.gradle.task.client.WatchTask)

I hope to throw a PR together to address this.

rjrudin:
Don't merge this yet. This is my first attempt for making HubWatchTask do everything that ml-gradle's WatchTask does - to fix #1429 - plus load hub modules from the plugins directory. 

I made a tiny change to WatchTask in ml-gradle - https://github.com/marklogic-community/ml-gradle/commit/29165edd7570fd1a21a7f64557f034f050b1224a - so that there's a method for HubWatchTask to override where it can load hub modules. 

This is of course more fragile since it's using extension over composition. Another approach would be for the guts of WatchTask in ml-gradle to be refactored into a Java class in ml-gradle that e.g. provides a callback interface. HubWatchTask could then provide a callback for loading hub modules.

Longer term, I wonder if the "plugins" directory could go away and ./entities could just be under src/main/ml-modules/root (would get rid of the REST directories under each entity as well). That would remove a lot of custom code from DHF and allow for things like mlWatch to just work.

### DIFF
--- a/ml-data-hub-plugin/build.gradle
+++ b/ml-data-hub-plugin/build.gradle
@@ -42,6 +42,7 @@ repositories {
     jcenter()
     mavenCentral()
     maven { url 'http://developer.marklogic.com/maven2' }
+    mavenLocal()
 }
 
 dependencies {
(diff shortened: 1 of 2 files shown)


grechaw:
We've deprecated "(would get rid of the REST directories under each entity as well)" already, I thought they actually stopped working in 4.0.1.  Thanks for pushing us along with this, it's been a kind of a morass

PR #1447, opened 2018-10-19
MLUI-127: mlui integration - focus element directive

clockworked247:
Directive move and story complete

Added new shared/directives folder + added directive stories section


alexsmr:
One more thing. When open PR, try to have the naming with MLUI-122: prefix, so it is much easier to distinguish from other DHF PRs. Thanks.

clockworked247:
"One more thing. When open PR, try to have the naming with MLUI-122: prefix, so it is much easier to distinguish from other DHF PRs. Thanks."  Ok, I was following your previous comment on my last PR which said to name it to "mlui integration", I normally prefix all my PRs with the task ID, so I'll continue doing so.

alexsmr:
Hm. 'mlui-integration' should be used for integration PR against develop branch only. My previous comment is actually exactly the same: https://github.com/marklogic/marklogic-data-hub/pull/1389 use the story name for PR. It is a common sense: if the changes cover the story use the story name if the task then use the task name...

PR #1454, opened 2018-10-24
E2e/stable

ayuwono:
- Add more check on entity disclosure to make it more stable
- Use async await to make it more stable

Based on Jenkins CI result on qa branch, these changes have the tests green:
https://jenkins.marklogic.com/view/DHF-9.0-Server-All/job/NO_CI_dhf-qs-e2e-ui-rh7/


bsrikan:
@ayuwono Lets discuss tomorrow. await needs to be called on functions that are async. So the functions that need to be prepended with async are those in page objects.

PR #1458, opened 2018-10-25
DHFPROD-1246 flow operator/data-hub-user can deploy flows

grechaw:
This PR added three URI privileges to security configuration for DHF, and removes the any-uri execute privilege for data-hub-role.

A corresponding change needs to be made in DHS provisioning (note to self)


aebadirad:
This seems to patch the first step, allowing a data hub user to mess w/ data hub core files and flow modules.

PR #1469, opened 2018-10-29
Feature/globalsearch mlcp config

kghmanuel:
This is an mlcp option file to load the documentation as part of the globalsearch initiative. Not meant to be part of main branch nor released.

Targetted towards the docs branch.


load-to-globalsearch.mlcp
```diff
@@ -0,0 +1,36 @@
...
+-fastload
+false
+-host
+globalsearch.marklogic.com
+-port
+8011
+-modules_root
+/
+-transform_module
+"/MarkLogic/data-hub-framework/transforms/mlcp-flow-transform.xqy"
```
grechaw: This is the 3.0 URL, and will probably start failing at some point.  I'd recommend being on DHF 4+ and using /data-hub/4/transforms/mlcp-flow-transform.xqy

grechaw: You might also want to use port 8010.  I'd avoid using the FINAL appserver for anything related to DHF as the is the part that does not align with DHS

grechaw:
@kghmanuel do you want to push another commit here to fix the path issue?

grechaw:
@kghmanuel ping, would you like to change this PR or close it?

aebadirad:
Month since last contact, need this to be updated for 4.x and port 8010

PR #1495, opened 2018-11-02
MLUI-131: Job Output Component

clockworked247:


alexsmr:
@clockworked247, I think test failures are not related to this task. Only if adding the component to the app module would cause somehow module loading time to explode, causing the timeout on the login page... I see it started to be unstable for the last few builds with the same problem across the board. It could be hardware related too...

clockworked247:
Ok, thanks @alexsmr

PR #1512, opened 2018-11-06
MLUI-121: Scoping flows.component

pengzhouml:


quick-start/src/main/ui/app/flows/flows.component.ts
```diff
@@ -1,35 +1,22 @@
...
-
-import { Ng2DeviceService } from 'ng2-device-detector';
-
-import { distanceInWords } from 'date-fns';
-
-import * as _ from 'lodash';
+import { NewEntityComponent } from '../shared/components/new-entity/new-entity.component';
 
 @Component({
   selector: 'app-flows',
```
alexsmr: html file is excessive in smart component and template option should be used instead... also scss file must be in ui component location and should be deleted from here.

PR #1515, opened 2018-11-07
Fix for Support Ticket 20961 

srinathgit:
1. Loading value of 'finalSimpleSsl' 
2. fixing typos in property names
3. Fixing gradle plugin ssl/tls tests
4. Creating a task for the full cycle gradle tests


srinathgit:
Will create a new jenkins job for the full cycle tests that require installation / cleanup

grechaw:
Looks great

aebadirad:
We just need to remember in the next major version to take out hubconfigbuilder from this solution.

PR #1516, opened 2018-11-08
MLUI-124: harmonize-flow-options.component.ts

brucean52:


alexsmr:
@brucean52 please rebase on latest mlui-integration and wait until PR build is done.

alexsmr:
@brucean52 same here. It looks like it is not a clean rebase. Use merge rather than rebase for your changes in the forked branch. The rule of thumb: do rebase on branches without diffs, i.e. that can be fast forward to head with no diffs, but use merge, where you have diffs.

alexsmr:
@brucean52 same here.

PR #1524, opened 2018-11-09
Using jackson instead of gson in tests

srinathgit:
Change tests to use jackson api instead of gson


aebadirad:
Thank you very much :)

PR #1526, opened 2018-11-09
MLUI-150: Scoping traces.components and stories

pengzhouml:


quick-start/src/main/ui/app/traces/traces.component.html
```diff
@@ -1,54 +1,10 @@
...
-  </div>
-  <div *ngIf="loadingTraces" flex layout="column" layout-align="center center">
-    <h3>Loading Traces...</h3>
-    <div class="center-text">
-      <mdl-spinner single-color active></mdl-spinner>
-    </div>
-  </div>
-
-</div>
+<app-traces-ui
```
alexsmr: The idea is to remove UI artifacts from here to a shared place. Please remove this html file and use template string interpolation instead with this markup, as well as remove please original scss and reference to it here https://github.com/marklogic/marklogic-data-hub/blob/f3a44a429a165146fd003ce7cd70b09220b623dc/quick-start/src/main/ui/app/traces/traces.component.ts#L12

pengzhouml: As far as I understand, It actually server the sample purpose. I will follow the standard though.

alexsmr: Yes, we do not change logic, just structure, and some additional refactoring.

Issue #1259, opened 2018-08-08
Harmonize flow finishes with "CANCELED" status even though successful


danielholgate:
DHF 3.0.0

Under some conditions a Harmonize flow will consistantly return a status of "CANCELED" when it in fact successfully processed all documents and was not stopped/cancelled.

I have done a little debugging and the mis-reporting occurs in FlowRunnerImpl.java https://github.com/marklogic/marklogic-data-hub/blob/47764071f1e79e4d5fe9b36b4e1009dd344d8dd4/marklogic-data-hub/src/main/java/com/marklogic/hub/flow/impl/FlowRunnerImpl.java#L309

For most flows, uris.size() at this point correctly returns the original number of uris collected and the status reporting is correct. However, it seems over a certain number of uris (in my experience, 5000)  uris.size() will be the correct number before spawning the new thread here https://github.com/marklogic/marklogic-data-hub/blob/47764071f1e79e4d5fe9b36b4e1009dd344d8dd4/marklogic-data-hub/src/main/java/com/marklogic/hub/flow/impl/FlowRunnerImpl.java#L294

_but then is different by the time we evaluate the status and causes this problem_. Presumably some multi-threading thing

**To reproduce:**

1. Create an entity and ingest this simple csv test file. It has 5004 lines of data.
Make sure you have delimited text and "generate uris" option selected to avoid overwriting any URIs
[test_data.zip](https://github.com/marklogic/marklogic-data-hub/files/2270035/test_data.zip)

2. Create a harmonize flow. Do not edit/customise the flow at all, just let it move the documents from staging to final.
3. Run the flow (from Quickstart UI or gradle command line)
4. Job status incorrectly shows "CANCELED", as all 5004 documents can be found in the FINAL db
<img width="807" alt="screen shot 2018-08-08 at 1 09 33 pm" src="https://user-images.githubusercontent.com/13762269/43833831-5a3f97d8-9b0c-11e8-92b4-b7c6c2b64dfa.png">

I had some debugging code in during this scenario and uris.size() = 5004 before we start, but at the end when status is determined uris.size() = 4 for some reason and that causes the mis-reporting.

**Note: If we add [1 to 5000] to the collector and re-run we get a correct "FINISHED" status**
<img width="797" alt="screen shot 2018-08-08 at 1 12 33 pm" src="https://user-images.githubusercontent.com/13762269/43833974-ca8e6514-9b0c-11e8-8c1b-13a4b8cb6528.png">

[...]

PR #1551, opened 2018-11-16
Possible fix to the #1259 bug that keeps sporatically popping up

aebadirad:


aebadirad:
Comes from @miguelrgonzalez in the field and doesn't give a chance to cause a regression, so let's try it.

PR #1560, opened 2018-11-20
fix QuickStart type issue in getFlowType()

wooldridge:
See DHFPROD-1411 fix, need to type getFlowType arg as Flow.


ayuwono:
at least the build is successful

PR #1562, opened 2018-11-20
Syncing develop with master post 4.0.3 release

aebadirad:


examples/performance-sample/build.gradle
```diff
@@ -1,6 +1,6 @@
 plugins {
     id 'net.saliman.properties' version '1.4.6'
-    id 'com.marklogic.ml-data-hub' version '4.0.1'
+    id 'com.marklogic.ml-data-hub' version '4.0.2'
 }
```
srinathgit: Looks like version wasn't changed in this instance. Everything else looks good.

aebadirad: Good catch, fixed

PR #1567, opened 2018-11-20
create and init hub project

srinathgit:
create and init hub project


srinathgit:
there are a few failures, will create another PR

PR #1575, opened 2018-11-24
Using mlConfigPaths and only one custom command for database management

rjrudin:
This allows for mlDeployDatabases to work. It also removes some custom commands that aren't needed if we rely on ml-app-deployer's support for ordering databases based on their dependencies on triggers/schema databases. 

This depends on ml-app-deployer 3.10.1, which adds a couple enhancements to allow for DHF to remove most of the custom commands. 

This also removes some unused methods.


marklogic-data-hub/src/main/java/com/marklogic/hub/impl/DataHubImpl.java
```diff
@@ -657,9 +622,11 @@ private void runInDatabase(String query, String databaseName) {
...
 
         List<Command> dbCommands = new ArrayList<>();
-        dbCommands.add(new DeployHubOtherDatabasesCommand(hubConfig));
-        dbCommands.add(new DeployHubFinalTriggersDatabaseCommand(hubConfig));
-        dbCommands.add(new DeployHubFinalSchemasDatabaseCommand(hubConfig));
+        // TODO Hoping we don't need anything here, because the staging command list has everything we need
+        // Ideally we can get rid of the concept of "staging" commands and "final" commands
+//        dbCommands.add(new DeployHubOtherDatabasesCommand(hubConfig));
+//        dbCommands.add(new DeployHubFinalTriggersDatabaseCommand(hubConfig));
+//        dbCommands.add(new DeployHubFinalSchemasDatabaseCommand(hubConfig));
```
srinathgit: The database files in the path "ml-config/databases/" (final, final-schemas, final-triggers, modules) are not getting deployed. I think we may have to include this snippet to deploy those files

DeployOtherDatabasesCommand databaseCommand = new DeployOtherDatabasesCommand();
        dbCommands.add(new DeployHubStagingTriggersDatabaseCommand(hubConfig));	        databaseCommand.setDeployDatabaseCommandFactory(new HubDeployCommandDatabaseFactory(hubConfig));
        dbCommands.add(new DeployHubStagingSchemasDatabaseCommand(hubConfig));	        dbCommands.add(databaseCommand);

ml-data-hub-plugin/src/main/groovy/com/marklogic/gradle/DataHubPlugin.groovy
```diff
@@ -135,6 +139,18 @@ class DataHubPlugin implements Plugin<Project> {
...
+        appConfig.setNoRestServer(true)
+
+        // DHF defaults to a single config dir of the hub config dir (hub-internal-config)
+        // Add the user config dir as well
+        appConfig.getConfigDirs().add(new ConfigDir(hubConfig.getUserConfigDir().toFile()))
+
+        println "Will look for resource configuration files in the following directories:"
+        for (ConfigDir configDir : appConfig.getConfigDirs()) {
+            println "Configuration directory: " + configDir.getBaseDir()
+        }
```
srinathgit: Currently, there are separate staging and final schemas and triggers dbs. The expected behavior currently is that schemas under "hub-internal-config/databases/schemas" should be deployed to the data-hub-staging-SCHEMAS and the ones under "ml-config/databases/schemas" to data-hub-final-SCHEMAS. Similar behavior is expected for triggers as well. Will that hold up by adding "ml-config" dir to the configdir list

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/DataHubImpl.java
rjrudin: @srinathgit Have you tried a build of this? I did "gradle publishtomavenlocal" with version = 4.0.3.1 and then used it on a new DHF project; the final databases are all deployed correctly.

ml-data-hub-plugin/src/main/groovy/com/marklogic/gradle/DataHubPlugin.groovy
rjrudin: I'll test it locally, I don't think it should matter because I doubt that the code for loading schemas cares about mlConfigPaths (as schemas are not resources).

rjrudin: @srinathgit I'm working off a branch off develop off 4.0.3 - I added files to hub-internal-config/databases/schemas and ml-config/databases/schemas, and they were ignored. Only hub-internal-config/schemas was scanned and loaded from. 

I then tried using 4.0.3 and got the same behavior. So it doesn't look like those directories are supported. I also don't see any code in DHF that would provide this behavior. I only see this in HubConfigImpl:

```
ConfigDir configDir = new ConfigDir(getUserConfigDir().toFile());
config.setConfigDir(configDir);
config.setSchemasPath(getUserSchemasDir().toString());
```

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/DataHubImpl.java
srinathgit: Installation using java api seems to have failed in jenkins with the following message https://jenkins.marklogic.com/job/dhf-core-develop-rh7-PR/1759/consoleFull

11:32:36.275 [main] INFO  c.m.h.d.c.DeployUserServersCommand - Processing file: /space/jenkins/workspace/dhf-core-develop-rh7-PR/marklogic-data-hub/ye-olde-project/src/main/ml-config/servers/final-server.json
Exception in thread "main" 11:32:36.669 [main] ERROR c.m.r.util.MgmtResponseErrorHandler - Logging HTTP response body to assist with debugging: {"errorResponse":{"statusCode":"400", "status":"Bad Request", "messageCode":"MANAGE-INVALIDPAYLOAD", "message":"MANAGE-INVALIDPAYLOAD: (err:FOER0000) Payload has errors in structure, content-type or values. No such database"}}
11:32:36.670 [main] ERROR c.m.m.r.appservers.ServerManager - Error occurred while sending POST request to /manage/v2/servers; logging request body to assist with debugging: {"server-name":"data-hub-FINAL","server-type":"http","root":"/","group-name":"Default","port":8011,"modules-database":"data-hub-MODULES","content-database":"data-hub-FINAL","authentication":"digest","default-error-format":"json","error-handler":"/MarkLogic/rest-api/error-handler.xqy","url-rewriter":"/MarkLogic/rest-api/rewriter.xml","rewrite-resolves-globally":true}

Using your branch, I added that snippet in getFinalCommands() and installation/ uninstallation ran fine in my local

ml-data-hub-plugin/src/main/groovy/com/marklogic/gradle/DataHubPlugin.groovy
srinathgit: Sorry, I got the paths wrong. It is "hub-internal-config/schemas"  for loading to data-hub-staging-SCHEMAS  and "ml-schemas" for loading to data-hub-final-SCHEMAS

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/DataHubImpl.java
rjrudin: I believe this issue is because of the separation of getStagingCommands and getFinalCommands, with the latter being run first and then followed by the former. So with two separate lists, getStagingCommands doesn't need a set of database commands but getFinalCommands does (if both have the database commands, then the database files will all get processed twice).

On my other branch, I'm experimenting with just one set of commands and I'm not having any problems so far.

PR #1586, opened 2018-11-27
Dhfprod 1338 spring singleton

grechaw:
This PR, I think, just does some cleanup of interfaces and comments.   There is a blatant removal of static creation methods -- they don't make a lot of sense anymore and will just cause problems for anyone who uses that method.  We might have to consider a single entry point for Java though at this time


aebadirad:
Free sync w/ dev i guess.

PR #1587, opened 2018-11-28
Fix ssl issue

srinathgit:
Fix ssl issue

### DIFF
--- a/ml-data-hub-plugin/src/main/groovy/com/marklogic/gradle/DataHubPlugin.groovy
+++ b/ml-data-hub-plugin/src/main/groovy/com/marklogic/gradle/DataHubPlugin.groovy
@@ -21,10 +21,6 @@ import com.marklogic.appdeployer.command.Command
 import com.marklogic.appdeployer.impl.SimpleAppDeployer
 import com.marklogic.gradle.task.*
 import com.marklogic.hub.ApplicationConfig
-import com.marklogic.hub.DataHub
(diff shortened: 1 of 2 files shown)


ml-data-hub-plugin/src/main/groovy/com/marklogic/gradle/DataHubPlugin.groovy
```diff
@@ -45,13 +41,13 @@ import org.springframework.context.annotation.AnnotationConfigApplicationContext
...
-    private HubConfig hubConfig
+    private DataHubImpl dataHub
+    private ScaffoldingImpl scaffolding
+    private HubProjectImpl hubProject
+    private HubConfigImpl hubConfig
     private LoadHubModulesCommand loadHubModulesCommand
     private LoadUserStagingModulesCommand loadUserStagingModulesCommand
-    private MappingManager mappingManager
+    private MappingManagerImpl mappingManager
     private FlowManagerImpl flowManager
```
aebadirad: This really defeats the point of us having interfaces in the core. But not sure why we have them anyways given we need to create an object for everything anyways.

aebadirad:
So, technically the final simplessl setting is broken in 4.0.3 too, right, even though we just fixed the one for staging?

srinathgit:
This issue is same as the one that was fixed in 4.0.3 https://github.com/marklogic/marklogic-data-hub/pull/1515/files .

PR #1594, opened 2018-11-29
Better fix for passing Gradle properties into HubConfigImpl

rjrudin:
DataHubPlugin now passes in the already-processed set of Gradle properties. No other code was calling withPropertiesFromEnvironment, so there's no longer a need to load any properties from a file.

### DIFF
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/impl/HubConfigImpl.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/impl/HubConfigImpl.java
@@ -211,9 +211,6 @@ public class HubConfigImpl implements HubConfig
     private AppConfig stagingAppConfig;
     private AppConfig finalAppConfig;
 
-    private boolean usePropertiesFromEnvironment;
-    private String envString;
-
     private static final Logger logger = LoggerFactory.getLogger(HubConfigImpl.class);
 
     private ObjectMapper objmapper;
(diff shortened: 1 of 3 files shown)


srinathgit:
@rjrudin , @aebadirad 

I see two issues with this change:

1. Using java API, users can build hubconfig until 4.0.3 can create hubConfig with specified environment
HubConfig hubConfig = HubConfigBuilder.newHubConfigBuilder("/path/to/your/project")        .withPropertiesFromEnvironment("dhs").build();
using a gradle-dhs.properties. Wouldn't this remove that option of loading properties from different environment file when using  java API ? As far as gradle is concerned, I think the "net.saliman.properties" plugin takes care of loading specified environments .

2. Let's consider a project that has been initialized using hubInit command. If someone were to delete "gradle.properties" and "gradle-local.properties" and then run hubInit again, it throws the following exception. Even though in DataHubPlugin.groovy, hubConfig.initHubProject() once again creates "gradle.properties" and "gradle-local.properties" , "project" object  (in new ProjectPropertySource(project).getProperties() ) still has the old values (where "mlUser" is null). I see why we have to use new ProjectPropertySource(project).getProperties() to get the properties . But we hit this issue in this corner case.
```
FAILURE: Build failed with an exception.

* Where:
Build file '/tmp/testproj/build.gradle' line: 23

* What went wrong:
A problem occurred evaluating root project 'testproj'.
> Failed to apply plugin [id 'com.marklogic.ml-data-hub']
   > No user provided

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output.

BUILD FAILED
```

rjrudin:
@srinathgit @aebadirad For 1, given that HubConfigBuilder has gone away, how much backwards compatibility are we looking to preserve? I'm fine with the "withPropertiesFromEnvironment" being retained, but there were no references to it in the DHF codebase, and I'm wondering how likely it is that a Java API user is trying to load properties from multiple files, essentially duplicating what the Gradle properties plugin does. 

For 2 - I think that's a valid scenario because I've done that same thing before. I'll put together a PR for it.

PR #1595, opened 2018-11-29
1546 Using DeployOtherDatabasesCommand to deploy all databases

rjrudin:
The set of staging commands no longer has the database commands; since the final commands are executed first, they have the database commands, and DeployOtherDatabasesCommand is modified to use a custom impl of DeployDatabaseCommand that accounts for some DHF-specific stuff. 

Will soon have a single list of commands a single AppConfig. 

This PR includes the fix for gradle properties that's in the other PR I just submitted. It's also pointed at the Spring branch; if that's merged before this one, I'll resubmit this with it pointed to develop.


aebadirad:
Runs green, bringing it in

PR #1598, opened 2018-11-29
1547 Deploying all servers via a single command

rjrudin:


marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/HubAppDeployer.java
```diff
@@ -303,12 +303,6 @@ private void onStatusChange(int percentComplete, String message) {
         }
     }
 
-    private void onError() {
-        if (this.listener != null) {
-            this.listener.onError();
-        }
-    }
```
rjrudin: Unused private method, removed

PR #1613, opened 2018-11-30
Removing the modules from ml-modules in pii-test

srinathgit:
This PR
1 Removes all the modules from ml-modules and entity-config dir in pii-test resource directory
2. The hub modules are loaded from jar and not written to the project. Loading from ml-modules didn't work until now (in develop) because modulePaths is set incorrectly. Now that issue is fixed and hence all the these modules are loaded by LoadModulesCommand. It overwrites the existing hub modules and removes them from "hub-core-module" collection as it is not loaded by LoadHubModulesCommand.  A subsequent test (StreamCollectorTest) run dataHub.clearUserModules() and the core modules are considered non hub modules as they are not  part of "hub-core-module" collection. Hence the modules are wiped and tests fail with "com.marklogic.client.ResourceNotFoundException".


aebadirad:
Wow, great catch on the dead module/artifact clean up on the tests. Forgot we had worked around this myself, even. That's the danger of workarounds, I suppose.

PR #1625, opened 2018-12-03
Changing versions to "4.0.SNAPSHOT" and publishing the jars

srinathgit:
1. Changed versions to 4.0.SNAPSHOT (excluding example projects)
2. Changed xcc dependency version in scaffold to 9.0-7
3. Added info for publishing jars

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,1 +1,2 @@
-version=4.0.3
+version=4.0.SNAPSHOT
+publishUrl=file:../marklogic-data-hub/releases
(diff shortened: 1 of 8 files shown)


aebadirad:
We shouldn't have to, but it'd be nice to automatically assign version # based on version names. @itsshivaverma  has something that does most of this already I think?

srinathgit:
During release process, @itsshivaverma runs a script that updates the versions of gradle.properties, marklogic-data-hub/gradle.properties, ml-data-hub-plugin/gradle.properties,  quick-start/gradle.properties . The rest of the places including the example projects have to be done manually

PR #1626, opened 2018-12-03
fix regression failure due to locator mismatch

pengzhouml:
fix browse.ts and traces.ts


ayuwono:
Please merge this, it bring it down to 3 failures. The 3 failures are very intermittent

PR #1627, opened 2018-12-04
DHFPROD-1433 make entity editor scrollable

wooldridge:
Fixes entity editor extending beyond browser window when there are too many properties, making buttons not clickable:
- Make dialog body container scrollable after a max-height
- When new property is added, scroll to bottom


wooldridge:
@ayuwono, my latest commit should fix the build issue. Can you retest?

ayuwono:
@wooldridge, I think you need to rebase the branch with the latest one. After installation, it got stuck with the screen below. Clicking "Finished" will do nothing. We don't want this screen to appear

screen shot 2018-12-05 at 3 21 03 pm

wooldridge:
@ayuwono, will make a new PR after rebasing.

PR #1632, opened 2018-12-04
Fixing hubInit when gradle.properties doesn't exist

rjrudin:
This requires the user to run "hubInit" if the project hasn't been initialized yet. 

Also, initHubProject is no longer called if the project has not been initialized. If the project hasn't been initialized and the user isn't running hubInit, then the refreshProject call would likely fail because gradle.properties likely isn't present.


rjrudin:
@aebadirad Don't merge this yet, doing testing locally, something's amiss.

rjrudin:
@aebadirad Okay this should be good to go, tested it out locally

PR #1583, opened 2018-11-27
DHFPROD-1426 WIP index tooltips and help icons

wooldridge:
- Add new Bootstrap tooltips for index icons
- Replace existing MDL tooltips with Bootstrap tooltips
- Add help icons to mapping and entities views
- Destination docs page for help icons to come

PR #1636, opened 2018-12-04
change page object from fa-won to fa-krw

ayuwono:


aebadirad:
Can you sync this up with Mike's branch instead of develop so his PR passes? wooldridge:DHFPROD-1433-entity-editor-scroll

ayuwono:
This change is only for qa page object change, it's not against the entity editor scrolling. And it's more about that property tooltips PR that's been merged (https://github.com/marklogic/marklogic-data-hub/pull/1583). Current PR from Mike (wooldridge:DHFPROD-1433-entity-editor-scroll) is not passing because it can't build?

aebadirad:
Ah, my mistake. Mixed the two, apologies.

PR #1637, opened 2018-12-05
Change to support ml modules path by adding our absolute path, instea…

aebadirad:
…d of overwriting

### DIFF
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/LoadUserModulesCommand.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/LoadUserModulesCommand.java
@@ -16,7 +16,6 @@
 package com.marklogic.hub.deploy.commands;
 
 import com.marklogic.appdeployer.AppConfig;
-import com.marklogic.appdeployer.command.AbstractCommand;
 import com.marklogic.appdeployer.command.CommandContext;
 import com.marklogic.appdeployer.command.modules.AllButAssetsModulesFinder;
 import com.marklogic.appdeployer.command.modules.LoadModulesCommand;
@@ -51,6 +50,7 @@ import org.springframework.core.io.Resource;
 import org.springframework.scheduling.concurrent.ThreadPoolTaskExecutor;
 import org.springframework.stereotype.Component;
 
+import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.nio.file.*;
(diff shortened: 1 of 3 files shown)


quick-start/src/main/java/com/marklogic/quickstart/service/DataHubService.java
```diff
@@ -105,7 +105,7 @@ public void reinstallUserModules(HubConfig config, DeployUserModulesListener dep
         long startTime = PerformanceLogger.monitorTimeInsideMethod();
 
         try {
-            dataHub.clearUserModules();
+            //dataHub.clearUserModules();
```
rjrudin: @akshaysonvane  @srinathgit  This change from Alex seems to have fixed the problem where, when deploying from QS, nothing would be loaded from src/main/ml-modules.

akshaysonvane: QS tooltip for Redeploy says it will clear out the modules DB and reload them. Also removing this will break synchronization between the filesystem and server if the user removes a module from the filesystem.
The problem arises because we are relying on the custom user-modules-timestamp file for force loading the modules but the entries for this modules are in the module-timestamps file of ml-javaclient-util
A quick fix would be to delete both files while force loading 
or
Should we do away with the custom file and start relying on the module-timestamps file of ml-javaclient-util?

marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/LoadUserModulesCommand.java
```diff
@@ -98,8 +98,17 @@ public LoadUserModulesCommand() {
...
         PropertiesModuleManager pmm = new PropertiesModuleManager(timestampFile);
+
+        // Need to delete ml-javaclient-utils timestamp file as well as modules present in the standard gradle locations are now
+        // loaded by the modules loader in the parent class which adds these entries to the ml-javaclient-utils timestamp file
+        String filePath = hubConfig.getAppConfig().getModuleTimestampsPath();
+        File defaultTimestampFile = new File(filePath);
+
         if (forceLoad) {
             pmm.deletePropertiesFile();
+            if (defaultTimestampFile.exists()){
```
rjrudin: This seems like a good idea, I'll test it out locally.

quick-start/src/main/java/com/marklogic/quickstart/service/DataHubService.java
rjrudin: Let's keep both files in place for now - I took a brief look at getting rid of the hub-specific one, but I wasn't sure of the impacts. Something to consider post 4.1.0.

PR #1639, opened 2018-12-05
1592 hubUpdate now works when the project isn't fully initialized

rjrudin:
@aebadirad I'm not sure how to fully test this because of all the possible upgrade scenarios. The only concern I have with this is whether "refreshProject" can be called - that involves constructing a DatabaseClient to talk to the staging app server. If that's not possible because of what's in the user's gradle.properties file, then the hubUpdate will fail right away.

### DIFF
--- a/ml-data-hub-plugin/src/main/groovy/com/marklogic/gradle/DataHubPlugin.groovy
+++ b/ml-data-hub-plugin/src/main/groovy/com/marklogic/gradle/DataHubPlugin.groovy
@@ -17,7 +17,7 @@
 
 package com.marklogic.gradle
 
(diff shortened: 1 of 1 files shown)


srinathgit:
@rjrudin , It shouldn't be required but would it be fair to assume that gradle.properties has all the necessary info to call refreshProject()  at the time of upgrade ?

rjrudin:
@srinathgit  That is my thought - the reason that refreshProject can fail is because it tries to create a DatabaseClient for the staging database, and if properties don't specify how to create that connection, it'll blow up. But at least for an upgrade from 3.0, I believe we have those properties in gradle.properties so that that step shouldn't blow up.

PR #1644, opened 2018-12-06
Copying only necessary files during upgrade

srinathgit:
This PR

1. Copies job-database.json and staging-database.json from hub-internal-config/databases to src/main/hub-internal-config/databases
2. Copies module-database.json and final-database.json from hub-internal-config/databases to src/main/ml-config/databases
3. Doesn't copy any of the other db files (triggers, schemas, trace) from 3.0
4. Copies job-server.json and staging-server.json to src/main/hub-internal-config/servers and final-server.json to src/main/ml-config/servers
5. Copies the rest of hub-internal-config to src/main/hub-internal-config
6. 3.0 has "hub-internal-config" and "user-config" dirs. First the payloads in "hub-internal-config" is deployed and then the ones in "user-config" . But we don't
want to blindly copy all these files to ml-config as they can cause issues during deployment. So with regards to db and server files, I am not copying any files in user-config with the these names "job-database.json", "staging-database.json", "modules-database.json", "final-database.json", "trace-database.json", "triggers-database.json", "schemas-database.json", "job-server.json", "staging-server.json", "final-server.json", "trace-server.json" to ml-config. For changing configuration of hub related dbs and servers, users will have to manually update the respective hub-internal-config files .
7. The rest of user-config is copied over to ml-config


marklogic-data-hub/src/main/java/com/marklogic/hub/impl/HubProjectImpl.java
```diff
@@ -324,15 +326,27 @@ public void upgradeProject() throws IOException {
         Path oldEntityConfigDir = this.projectDir.resolve("entity-config.old");
         Path oldHubInternalConfigDir = this.projectDir.resolve("hub-internal-config.old");
         Path oldUserConfigDir = this.projectDir.resolve("user-config.old");
+        
+        //can be changed to traverse the old hub-internal-config and build the set
```
rjrudin: I think this is good, I think it's helpful to document all of the files that we expect from 3.x

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/HubProjectImpl.java
```diff
@@ -343,7 +357,97 @@ public void upgradeProject() throws IOException {
...
+                if(path.toLowerCase().equals("job-database.json")
+                || path.toLowerCase().equals("staging-database.json")) {
+                    try {
+                        FileUtils.copyFile(f.toFile(), getHubDatabaseDir().resolve(f.getFileName()).toFile());
+                    } catch (IOException e) {
+                        throw new RuntimeException(e);
+                    }
+                }
+              //copy modules,final db to ml-config
+                if(path.toLowerCase().equals("modules-database.json")
```
rjrudin: Is there a chance that these files already existing in ml-config? In which case, we should have a check saying - if the file exists, then don't replace it? I don't know what's best.

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/HubProjectImpl.java
```diff
@@ -343,7 +357,97 @@ public void upgradeProject() throws IOException {
...
+                    try {
+                        FileUtils.copyFile(f.toFile(), getHubServersDir().resolve(f.getFileName()).toFile());
+                    } catch (IOException e) {
+                        throw new RuntimeException(e);
+                    }
+                }
+              //copy final server to ml-config
+                if(path.toLowerCase().equals("final-server.json")) {
+                    try {
+                        FileUtils.copyFile(f.toFile(), getUserServersDir().resolve(f.getFileName()).toFile());
```
rjrudin: Same thing here, what if this file already exists?

I'll do some testing with a 3.0 build. I think the only reason the files would exist under src/main/ml-config already is if someone first ran hubInit using 4.1, then hubUpgrade. But it would be fine to overwrite the files, because the user wouldn't lose anything.

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/HubProjectImpl.java
```diff
@@ -343,7 +357,97 @@ public void upgradeProject() throws IOException {
...
+                }
+            });                        
+           // Files.copy(sourceDir, destDir, StandardCopyOption.REPLACE_EXISTING);
+            FileUtils.moveDirectory(sourceDir.toFile(), renamedSourceDir.toFile());
+        }       
+    }
+    
+    /*  "user-config" is the other config dir, it can contain "final-server.json", 
+     *  "staging-server.json" etc. In 3.0 ,they are deployed after hub-internal-config. But we don't
+     *  want to blindly copy all these files to ml-config as they can  cause issues during deployment.
```
rjrudin: I actually think it's okay to copy all of these over - with 4.1, it's fine if the user has e.g. staging-server.json or staging-database.json in ml-config.

rjrudin: Oh wait - is the issue that if the user has final-server.json in user-config and in hub-internal-config, then we'll first copy final-server.json over from hub-internal-config to ml-config, and then we'll overwrite it with the copy from user-config?

This seems like an issue with an upgrade from 3.0 to 4.0 as well - if e.g. final-database.json can be in both hub-internal-config and user-config in 3.0, but it can only be in one place in 4.0, then the two files should really be merged together, right?

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/HubProjectImpl.java
srinathgit: During upgrade process, we first call initProject(). So "modules-database.json" file will be present in ml-config but we are overwriting it with whatever user had in 3.0. The same is true for the other db and server files

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/HubProjectImpl.java
rjrudin: So worst case for the user, they've always got their original files in an ".old" directory right? So there's really no risk of losing anything - just some mild annoyance because some manual copying/editing may be required?

srinathgit: Yeah, they can find the file in ".old" dir but I will have to check with @sbayatpur-ml if we have to merge user-config files as well for 4.1

srinathgit:
I have updated the PR to

1. Replace SCHEMAS_DATABASE with mlStagingSchemasDbName and TRIGGERS_DATABASE with mlStagingTriggersDbName in jobs/staging-database.json
2. Replace the url rewriters in staging and job server json files
3. If "modules-database.json", "final-database.json" and "final-server.json" are present in user-config , merge them with the corresponding files in  ml-config (that were just written)  and write to src/main/ml-config dirs. The rest of files from user-config will be written to ml-config (except the obsolete files like trace-server, trace-database, schemas-database etc)

rjrudin:
Additions look good, I'm going to try to create unit tests for this

rjrudin:
I have unit tests in progress, will submit a PR with them into this branch.

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/HubProjectImpl.java
```diff
@@ -343,7 +354,148 @@ public void upgradeProject() throws IOException {
...
+                }
+                //copy jobs,staging server to hub-internal-config
+                else if(path.toLowerCase().equals("job-server.json")
+                || path.toLowerCase().equals("staging-server.json")) {
+                    ObjectMapper mapper = new ObjectMapper();
+                    JsonNode rootNode = null;
+                    File jsonFile = f.toFile();
+                    try {
+                        rootNode = mapper.readTree(jsonFile);
+                        ((ObjectNode) rootNode).put("url-rewriter", "/data-hub/4/tracing/tracing-rewriter.xml");
```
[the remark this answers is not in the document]
rjrudin: I think that's true - I'm making a fix on my branch that I'll soon submit as a PR.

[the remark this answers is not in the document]
srinathgit: @rjrudin , just pushed a commit fixing it

srinathgit:
@rjrudin , 
In the latest commit,
1. I am backing up hub-internal-config and ml-config directories if the dir structure is > = 4.0. I am writing this msg to stdout as it is required to run ./gradlew -i to get info messages in gradle. When upgrading flows, we print the msg to stdout , so just followed that
2. Added logging info messages during upgrade from 3.0
3. Update the gradle wrapper files (jar and properties) every time hubInit (project.initPotject() is called). 
4. Included your code for deleting obsolete files.

srinathgit:
@rjrudin , @akshaysonvane 

1. I am backing up hub-internal-config and ml-config directories if the dir structure is > = 4.0. I tried to include this code in HubProject.upgradeProject() but that code snippet required 'hubConfig' and 'versions' objects so , I am currently leaving it in DataHubImpl.upgradeHub()
2.  I am writing this msg to stdout as it is required to run ./gradlew -i to get info messages in gradle. When upgrading flows, we print the msg to stdout , so just followed that
```
The 4.0.3 src/main/hub-internal-config is now moved to src/main/hub-internal-config-4.0.3
The 4.0.3 src/main/ml-config is now moved to src/main/ml-config-4.0.3
Please copy the custom database, server configuration files from src/main/hub-internal-config-4.0.3 and src/main/ml-config-4.0.3 to their respective locations in  src/main/hub-internal-config and src/main/ml-config
```
3. Added logging info messages during upgrade from 3.0
4. Update the gradle wrapper files (jar and properties) every time hubInit (project.initProject() is called).
5. Merged your PR https://github.com/srinathgit/marklogic-data-hub/pull/2 in into this branch
6. Added a test HubProjectTest.upgrade300To403ToCurrentVersion() to verify that dirs get backed up in 4.0.x upgrade. 
7. Updated src/test/resources/upgrade-projects/dhf403from300/gradle.properties to include "mlDHFVersion","mlUsername/mlPassword" (and a few other properties).

rjrudin:
Great I'm testing this out locally now

rjrudin:
@srinathgit @akshaysonvane I think this is a separate issue - the jobs-database from 3.0.0 to 4.1.0 isn't getting the path range indexes. They do appear when going from 3.0.0 to 4.0.3 to 4.1.0. I'll investigate.

rjrudin:
When I do a 3.0 to 4.0.3 to upgrade using this branch, I get "hub-internal-config-2.0.0" as the backup directory. I think that's because myDHFVersion doesn't existing in gradle.properties - it's not added there via hubInit on 3.0 nor via hubUpdate.

PR #1646, opened 2018-12-06
do not initialize if project is already initialized

akshaysonvane:
Fixes https://project.marklogic.com/jira/browse/DHFPROD-1587


akshaysonvane:
Yes, you are right, there is a discrepancy between what initProject generates and what isInitialized checks for. Probably the way forward would be to close that gap.
But for now, it solves the issue https://project.marklogic.com/jira/browse/DHFPROD-1587
This workflow had skipped our testing.

PR #1647, opened 2018-12-06
1596 mlModulePaths and mlConfigPaths now work

rjrudin:
This modifies Alex's original PR and adds some tests. I believe the unit tests verify that the code is working, but I'm going to test it out on a local 4.1.0 project as well to ensure that mlModulePaths and mlConfigPaths both work.


akshaysonvane:
Let's get a Jenkins run on this.

PR #1658, opened 2018-12-11
HubConfig API changes

akshaysonvane:
Once Spring generates the hubConfig, the user can set properties using setters and call refreshProject which will take care of loading all properties and wiring clients based on these properties.


akshaysonvane:
Closing PR for now. Need to make a few more changes.

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/HubConfigImpl.java
```diff
@@ -78,28 +76,23 @@
...
-    @Autowired Versions versions;
+    @Autowired
+    FlowManagerImpl flowManager;
+    @Autowired
+    DataHubImpl dataHub;
+    @Autowired
+    Versions versions;
+
 
-    @Value("${mlHost}")
```
rjrudin: I'm nervous removing these - what problem does removing these annotations fix?

akshaysonvane:
Brought in @srinathgit changes from #1660

marklogic-data-hub/src/main/java/com/marklogic/hub/HubConfig.java
```diff
@@ -541,4 +539,12 @@
      * @return
      */
     HubConfig withPropertiesFromEnvironment(String environment);
+
+    /**
+     * Loads HubConfig object with values from gradle.properties (optionally overriden with
+     * gradle-(environment).properties). This method must be called every time after Spring creates
+     * HubConfig object and the project is initialized with createProject(String projectDirString)
+     * method
+     */
```
srinathgit: @akshaysonvane , should the doc include something to the effect of "make all changes wanted  to "hubConfig" using set methods and then call refreshProject" ?

akshaysonvane: Updated description 9d653dc

PR #1667, opened 2018-12-13
DHFPROD-1422 UX updates to mapping source

wooldridge:
- Add “URI:” label to source URI in Mapping view
- Add tooltip to URI label describing source URI
- Update tooltip for URI and only show it when truncated
- Update save button label to read “SAVE MAPPING”
- Expose help icon for Source heading, link to source info

### DIFF
--- a/quick-start/src/main/ui/app/mappings/map.component.html
+++ b/quick-start/src/main/ui/app/mappings/map.component.html
@@ -21,8 +21,8 @@
       </div>
 
       <div id="save-map">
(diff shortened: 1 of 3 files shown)


wooldridge:
Notes:
- Tooltip for URI value will only appear when URI length is greater than 45 characters. At that length the URI is truncated with a "..." prefix.
- Help icon next to "Source" links to: https://marklogic.github.io/marklogic-data-hub/harmonize/mapping/#changing-the-mapping-source-document
- Tooltips are the new Bootstrap-based ones now used on other pages.

PR #1681, opened 2018-12-17
Making ssl/cert-auth test work

srinathgit:
Trying to run the java api test with ssl/cert-auth


srinathgit:
The test passed with ssl settings from my branch as well
https://jenkins.marklogic.com/view/DHF-9.0-Server-All/job/dhf-core-develop-cert-ssl_9.0-7/22/

PR #1684, opened 2018-12-17
DHFPROD-1433 make entity editor scrollable 

wooldridge:
Fixes entity editor extending beyond browser window when there are too many properties, making buttons not clickable:
- Make dialog body container scrollable after a max-height
- When new property is added, scroll to bottom

Resubmit of after rebase: https://github.com/marklogic/marklogic-data-hub/pull/1627


wooldridge:
Closing and rebasing again.

PR #1690, opened 2018-12-18
DHFPROD-1645 Added fix for which databases should have their indexes …

rjrudin:
…updated

I wasn't aware of this requirement, but I think the test case captures the desired functionality. 

Note that if a user uses a filename other than the ones this is looking for, it won't get processed, even if it's referring to the staging or final or job database. But I doubt that is common.

### DIFF
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/impl/DataHubImpl.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/impl/DataHubImpl.java
@@ -508,16 +508,29 @@ public class DataHubImpl implements DataHub {
     public void updateIndexes() {
         HubAppDeployer deployer = new HubAppDeployer(getManageClient(), getAdminManager(), null, hubConfig.newStagingClient());
         deployer.setCommands(buildCommandMap().get("mlDatabaseCommands"));
-        DeployOtherDatabasesCommand c = null;
 
         AppConfig appConfig = hubConfig.getAppConfig();
         final boolean originalCreateForests = appConfig.isCreateForests();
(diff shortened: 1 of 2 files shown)


srinathgit:
@bsrikan  just mentioned that he is still seeing the same issue when he runs mlUpdateIndexes
```
Logging HTTP response body to assist with debugging: {"errorResponse":{"statusCode":"403", "status":"Forbidden", "messageCode":"", "message":"You do not have credentials to access /manage/v2/databases/data-hub-staging-SCHEMAS/properties ."}}
```

srinathgit:
@rjrudin ,

Since we are using HubAppDeployer in updateIndexes() and not SimpleAppDeployer, "prepareCommand(command, context);" which includes the pattern set is not getting called as deploy() method is overridden in HubAppDeployer.

srinathgit:
I changed locally to use 'SimpleAppDeployer' to update indexes and was able to run the command fine in DHS environment

rjrudin:
I applied a fix so that SimpleAppDeployer is used instead of HubAppDeployer. That seems fine for the updateIndexes method, as null was being passed in for the HubDeployStatusListener.

It's not good though that HubAppDeployer is extending SimpleAppDeployer and losing functionality in the process. I'll investigate that in a separate PR - I'll write up a JIRA ticket for it too.

aebadirad:
Yeah, I'm not liking we drop back to simpleappdeployer to just do indexes because we lost functionality on the hubappdeployer object.

While this 'works', let's make a separate ticket for it and address that concern.

bsrikan:
Fix is working now.

PR #1692, opened 2018-12-18
Initialize 'hubAdminUserName', 'hubAdminRoleName'  and fix build.gradle

srinathgit:
This PR

1. Initializes 'hubAdminUserName', 'hubAdminRoleName'  so that user defined values are set and not the ones  from dhf-default.props.
2. Doesn't get 'bootStrapHost' if not required (it caused issues in DHS env when running test),

### DIFF
--- a/marklogic-data-hub/build.gradle
+++ b/marklogic-data-hub/build.gradle
@@ -316,10 +316,11 @@ task testBootstrap(type: Test) {
 
 task setupSSL{
     doFirst{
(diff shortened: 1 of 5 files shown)


aebadirad:
Hold on this for post 4.1.0

srinathgit:
The PR fixes https://project.marklogic.com/jira/browse/DHFPROD-1646 , test issues Srikanth is facing in DHS environment

PR #1700, opened 2018-12-19
HubAppDeployer no longer loses functionality in SimpleAppDeployer

rjrudin:
HubAppDeployer now extends SimpleAppDeployer in a way that it doesn't lose any functionality.

This is not intended for the 4.1.0 release - hoping it can go into 4.1.1 or 4.2.x.


srinathgit:
The PR looks good. Should DataHubImpl.updateIndexes() be changed to use HubAppDeployer as part of this  PR ?

aebadirad:
It'd be good to do that, yes.

rjrudin:
AFAICT, there's no reason to use HubAppDeployer for updating indexes - unless the "telemetry" call in HubAppDeployer matters for this operation. But my hunch is that it doesn't, because that telemetry stuff seems specific to installs, not partial operations like updating indexes. Which makes me think it's better that SimpleAppDeployer is used instead of HubAppDeployer.

PR #1640, opened 2018-12-05
Added DHF4 project with test cases for verifying the deployment

rjrudin:
Most of the tests in VerifyDeploymentTest do not pass when using DHF 4.0.3.

I figure we can add to this too as we identify additional resources that a user is likely to deploy in a DHF app.

PR #1714, opened 2018-12-21
1580 Added DHF4 project with test cases for verifying the deployment …

rjrudin:
…(#1640)

Most of the tests in VerifyDeploymentTest do not pass when using DHF 4.0.3

@aebadirad Could this be pushed to master as well, since it's just an example project? The only files in this PR are under ./examples/dhf4-with-tests.


aebadirad:
That makes sense. Updating.

PR #1718, opened 2018-12-26
DHFPROD-1526 - Beautify trace errors on QuickStart UI

brucean52:


quick-start/src/main/ui/app/shared/components/ml-error/ml-error.component.scss
```diff
@@ -7,6 +7,8 @@
 .format-string {
   border: 1px solid rgba(0,0,0,0.10);
   padding: 10px;
+  max-height: 35vh;
```
alexsmr: Is there a reason for 35vh? This would scale with the height of the browser. For data-intensive applications, it is better to keep height fixed. Besides, is it realistic to have max for 35% of the viewport?

alexsmr: Can it be shortened to one line (like the first line with "..."), but with the link to open the modal for the long line. Does it make sense to do here? Or it would be overkill?

aebadirad: The truncation w/ click to display full might be nice for ease of use.

PR #1728, opened 2019-01-08
Fix for DHFPROD-1161

clockworked247:
Fixes refresh issue on map list in left-hand bar.  Fix also cleans up button code, making SAVE/RESET buttons in map-ui component HTML the "disabled" attribute dynamic, rather than two hard-coded versions of the same button.


clockworked247:
@ayuwono @wooldridge If everything looks good, can you merge this code?  I still don't have permissions.

ayuwono:
i don't have that power either =)

aebadirad:
Okay okay :) merging

clockworked247:
@aebadirad has the powah!

Issue #1652, opened 2018-12-07
contentContext creates "undefined" object wrapper around objects in an array


masyukun:
#### The issue

Short description of the problem:
When harmonizing, I want to assign an array of objects into the returned content object:
```javascript
  let email = xdmp.quote( !fn.empty(fn.head(source.xpath('//EmailAddress'))) 
    ? [].concat( fn.head(fn.head(source.xpath('//EmailAddress'))).toString().split("|").map(function(email){
         return { "emailAddress": email }
        })  )
      : null
  );

  // return the instance object
  return {
    '$attachments': attachments,
    '$type': 'Persona',
    '$version': '0.0.1',
    'id': id,
    'name': name,
    'email': email,
    'phone': phone
  }
```
From this code, I expect (and see returned when I run in Query console:

```json
{
	"envelope": {
		"headers": {},
		"triples": {},
		"instance": {
			"Persona": {
				"id": 234,
				"name": "Jim Driscoll",
				"email": [
					{
						"emailAddress": "jim.driscoll@marklogic.com"
					}
				],
				"phone": "555-123-4567"
			},
			"info": {
				"title": "Persona",
				"version": "0.0.1"
			}
		}
	}
}
```

Instead, I see:

```json
{
	"envelope": {
		"headers": {},
		"triples": {},
		"instance": {
			"Persona": {
				"id": 234,
				"name": "Jim Driscoll",
				"email": [
					{
						"undefined": {
							"emailAddress": "jim.driscoll@marklogic.com"
						}
					}
				],
				"phone": "555-123-4567"
			},
			"info": {
				"title": "Persona",
				"version": "0.0.1"
			}
		}
	}
}
```
What behavior are you expecting?
No wrapping {"undefined": XXX} object

#### Tech details

Which Operating System are you using? Any

Which version of MarkLogic are you using? 9.0-6

Which version of the Data Hub Framework are you using? 4.0.1

Is this a QuickStart UI bug? If so, which browser are you using? No

#### The devil is in the details...

If possible, rerun the command with -d for debugging output and attach the output:

  **for quickstart:**  
  `java -jar quickstart.war -d > myoutput.txt`

  **for gradle:**  
  `gradle mlYourCommand -d > myoutput.txt`

masyukun:
I know there is another issue related to this, but the outcome seems to be "this is how it's intended to work." https://github.com/marklogic/marklogic-data-hub/issues/1302

[...]

PR #1729, opened 2019-01-08
GH #1652 If $type is undefined, don't nest

ryanjdew:


aebadirad:
Is this consistent with  the xquery version (flow-lib.xqy) when handling nested json w/o a type?

ryanjdew:
@aebadirad it will be now.

PR #1730, opened 2019-01-08
add redeploy to stabilize

ayuwono:
minor stabilization


aebadirad:
5 seconds seems rather long to wait, but for now it'll do.

PR #1734, opened 2019-01-10
Add triggers for entity model TDE generation

ryanjdew:


marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/LoadEntityModelsCommand.java
```diff
@@ -0,0 +1,66 @@
...
+import com.marklogic.appdeployer.command.SortOrderConstants;
+import com.marklogic.client.DatabaseClient;
+import com.marklogic.client.datamovement.WriteEvent;
+import com.marklogic.client.document.DocumentWriteSet;
+import com.marklogic.client.document.JSONDocumentManager;
+import com.marklogic.hub.HubConfig;
+import com.marklogic.hub.deploy.util.EntityDeploymentUtil;
+
+@Component
+public class LoadEntityModelsCommand extends AbstractCommand {
```
rjrudin: I know this was merged already, but I think it's worth adding a comment to this class to explain why it exists and why it's separate from the commands for loading modules (I know there are good reasons for that, but those probably wouldn't be apparent to a typical reader of this code).

PR #1741, opened 2019-01-11
Fix for DHFPROD-590 and DHFPROD-1698

srinathgit:
This PR

1. Populates 'errors' and 'hasErrors' objects when there are changes to the code
2. Places the bug icon on the right line (unsure why they are off by one line currently)


ayuwono:
On initial UI testing, the bug icon appears again on the error line

PR #1742, opened 2019-01-14
Es alignment for 4.x

aebadirad:


aebadirad:
Holding on this. I've asked @srinathgit to create another command that just loads artifacts, as to isolate it away from modules, so we can have more control over grabbing ALL artifacts and deploying them independent (or in conjunction with) module loading.

PR #1744, opened 2019-01-15
Setting content db to final db

srinathgit:
Fix for https://project.marklogic.com/jira/browse/DHFPROD-1682

Setting content db to final db in the appConfig.


marklogic-data-hub/src/main/java/com/marklogic/hub/impl/HubConfigImpl.java
```diff
@@ -1721,6 +1721,7 @@ private void updateAppConfig(AppConfig config) {
         config.setTriggersDatabaseName(finalTriggersDbName);
         config.setSchemasDatabaseName(finalSchemasDbName);
         config.setModulesDatabaseName(modulesDbName);
+        config.setContentDatabaseName(finalDbName);
```
rjrudin: Yup, this makes sense. It'll make any ml-app-deployer command that uses getContentDatabaseName default to the final database. 

I recommend adding a unit test just to verify that that these four properties in particular default to the final databases and the modules database.

PR #1749, opened 2019-01-16
Running "LoadUserModulesCommand" after triggers are deployed

srinathgit:
This PR enables running "LoadUserModulesCommand" after triggers are deployed


rjrudin:
What's the purpose of this change (other than deleting code is always nice)? 

One potential issue is that since "hub user modules" (plugins specifically) aren't loaded when the other modules are loaded, any XQuery REST extension that depends on a library under "plugins" will fail to load since the library won't have been loaded yet. It's unlikely that someone would do this, but it's technically possible.

ryanjdew:
On first look, I think this may have a negative impact on user-defined triggers since any modules they provide for the triggers won't be available until after it attempts to load the triggers. Will look at it more closely.

srinathgit:
@rjrudin ,

I believe the original requirement was (@ryanjdew , please correct me if I am wrong)  to insert the entities after the triggers get deployed. Currently (in 4.1.0), entities are inserted as part of the "LoadUserModulesCommand". So,  the fix in this branch was to queue the entities during "LoadUserModulesCommand"  and then a new command 'LoadEntityModelsCommand' dequeues and writes the entities later after the triggers are deployed. What I attempted to do here, was to run the old 'LoadUserModulesCommand'(same as 4.1.0 where we insert entities) after "DeployTriggersCommand".

rjrudin:
I like the idea of a separate command because IMO, entity files aren't modules, they're content documents that are added to the content database. If the intent is for e.g. mlWatch to watch for changes to entity files, that change can be made to mlWatch. But otherwise, these are different from modules and a separate command makes sense.

srinathgit:
Closing this, will be creating a separate task for loading user artifacts

PR #1762, opened 2019-01-19
E2e/no toaster wait -- comment out waiting for toaster after updating index

ayuwono:
It seems that with the new npm update to resolve the vulns, the toaster is not recognizable when running e2e with jasmine. This PR will comment out all browser wait on toaster appearance. It will not affect the other tests. Commenting them out, as maybe on the future it could work again.


aebadirad:
It's weird, but it makes sense.

PR #1764, opened 2019-01-21
DHFPROD-1693 Use CMA with ml-gradle 3.11

ryanjdew:


marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/DeployHubAmpsCommand.java
```diff
@@ -63,53 +53,7 @@ public DeployHubAmpsCommand(HubConfig hubConfig) {
...
-
-            try (InputStream is = new ClassPathResource("hub-internal-config/configurations/amps.json").getInputStream()) {
-                String payload = IOUtils.toString(is, "utf-8");
-                payload = payload.replaceAll("data-hub-MODULES", modulesDatabaseName);
-                manageClient.postJsonAsSecurityUser("/manage/v3", payload);
-            } catch (IOException e) {
-                throw new DataHubConfigurationException(e);
-            }
-        }
+        super.execute(context);
```
srinathgit: I think removing the amps creation using the install-amps.xqy would slow the amps installation in 9.0-5

marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/DeployHubAmpsCommand.java
```diff
@@ -119,32 +63,13 @@ public void undo(CommandContext context) {
...
-                ampCall = ampCall.replaceAll("data-hub-MODULES", modulesDatabaseName);
-                call.xquery(ampCall);
-                call.eval();
-            } catch (IOException e) {
-                throw new DataHubConfigurationException(e);
-            }
+        if (serverVersion.matches("^[9]\\.0-([6789]|[0-9]{2,})(\\.\\d+)?")) {
+            // only on 9.0-5 are amps uninstalled at all.
+            logger.warn("Amps from uninstalled data hub framework to remain, but are disabled.");
+            super.execute(context);
```
srinathgit: We used to uninstall amps in 9.0-5 and not in any other version. Now it looks like we are uninstalling in all other versions except 9.0-5 ?

marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/DeployHubAmpsCommand.java
ryanjdew: The ticket states "make sure to scaffold out the 65ish amps and use them to deploy (but not remove) for 9.05 (and previous) and still use cma for 9.06 and later." I believe the reasoning is due to not wanting to maintain multiple sets of amps (the scaffolded out and the condensed set) and additional code for loading amps. I may be misinterpreting though.

srinathgit: @ryanjdew ,
The ticket does state to use the scaffolded amps for 9.0-5 installation and cma for 9.0-6 and above. I am guessing that the installation will be slow in 9.0-5 as a result. 
@aebadirad, is that ok ?

aebadirad: Yeah, 9.05 folks get the slow file by file request. It was to reduce the keeping of multiple amps. This one is okay to keep sanity going forward.

marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/DeployHubAmpsCommand.java
aebadirad: Yeah, let's make sure that we're consistent in the uninstallation of amps.

ryanjdew: Now checking for "-SNAPSHOT" to determine if amps are uninstalled. https://github.com/marklogic/marklogic-data-hub/blob/b437b77f29ac6b73c18e963cd4db40a5fa5d70fc/marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/DeployHubAmpsCommand.java#L67

aebadirad: This should split the middle between wanting one path to handle amps, while also not slowing our tests to a crawl. @srinathgit good for you?

PR #1768, opened 2019-01-21
Loading to staging schemas db from database-specific directory

srinathgit:
Loading to staging schemas db from database-specific directory

This PR

1. Scaffolds ml-config/databases/<staging-db-name>/schemas for the purpose of loading schemas and stops scaffolding the existing hub-internal-config/schemas dir
2. Makes the same change to the test
3. Fixes another test (where triggers count was incorrect)
4. Updates ml-gradle and ml-app-deployer to 3.11.0 and java client to 4.1.2

### DIFF
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/impl/HubProjectImpl.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/impl/HubProjectImpl.java
@@ -61,6 +61,7 @@ public class HubProjectImpl implements HubProject {
     private String projectDirString;
     private Path projectDir;
     private Path pluginsDir;
+    private Path stagingSchemasDir;
 
     protected final Logger logger = LoggerFactory.getLogger(this.getClass());
 
@@ -105,7 +106,11 @@ public class HubProjectImpl implements HubProject {
         return getHubConfigDir().resolve("security");
     }
 
-    @Override public Path getHubSchemasDir() { return getHubConfigDir().resolve("schemas"); }
+    @Override public Path getHubSchemasDir() { return this.stagingSchemasDir; }
+
+    private void setHubSchemasDir(Path stagingSchemasDir) {
+        this.stagingSchemasDir = stagingSchemasDir ;
+    }
 
     @Override public Path getHubTriggersDir() {
     	return getHubConfigDir().resolve("triggers"); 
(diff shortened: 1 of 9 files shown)


srinathgit:
The last commit https://github.com/marklogic/marklogic-data-hub/pull/1768/commits/96163ba7bb81ee5ed77b0b5ce5db19dbb0dcf505 is not related to the title of PR. It fixes 'LoadUserArtifactsCommand' and some of the test failures

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/HubProjectImpl.java
```diff
@@ -105,7 +106,11 @@ public void createProject(String projectDirString) {
         return getHubConfigDir().resolve("security");
     }
 
-    @Override public Path getHubSchemasDir() { return getHubConfigDir().resolve("schemas"); }
+    @Override public Path getHubSchemasDir() { return this.stagingSchemasDir; }
```
rjrudin: I would remove this, along with the private stagingSchemasDir variable. I don't think there's any reason for it - other than perhaps backwards compatibility, but I don't think that's necessary in this case because a feature is being removed - i.e. there's no need to keep track of this.

ml-data-hub-plugin/src/main/groovy/com/marklogic/gradle/DataHubPlugin.groovy
```diff
@@ -118,11 +118,13 @@ class DataHubPlugin implements Plugin<Project> {
...
 
         // This isn't likely to be used, but it's being kept for regression purposes for now
-        project.task("hubDeployUserModules", group: deployGroup, type: DeployUserModulesTask, description: "Installs user modules from the plugins and src/main/entity-config directories.")
+        project.task("hubDeployUserModules", group: deployGroup, type: DeployUserModulesTask,
+            description: "Installs user modules from the plugins and src/main/entity-config directories.")
+            .finalizedBy(["hubDeployUserArtifacts"])
 
         project.task("hubDeployUserArtifacts", group: deployGroup, type: DeployUserArtifactsTask,
             description: "Installs user artifacts such as entities and mappings.")
-            .mustRunAfter(["hubDeployUserModules"])
```
rjrudin: This looks like it should be in a separate PR as well.

rjrudin:
I also recommend removing updateSchemaCommandList from DataHubImpl and providing an upgrade step (manual or automated) that will move anything under hub-internal-config/schemas to ml-config/databases/(schemas db name)/schemas. Should get rid of all of the extra code now.

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/HubProjectImpl.java
srinathgit: @aebadirad , 
Do you want to retain the getHubSchemasDIr() for backward compatibility for 4.2 release?

aebadirad: Yes. Mark it deprecated though.

srinathgit:
Closing this , opened PR #1780

PR #1770, opened 2019-01-22
DHFPROD-1428: Improve the usability of text input elements

brucean52:


wooldridge:
As I wrote in the JIRA issue, I don't think this fix is going to work due to aesthetics. The mix of Bootstrap and MDL on the same page doesn't quite work, even if it does address the user confusion.

PR #1776, opened 2019-01-22
Feature/add dhf4 js only test examples

derms:
Added an example of how to write dhf unit tests using only javascript. Includes example of performing harmonization testing using javascript only


derms:
Now in pull request 1796

PR #1785, opened 2019-01-23
wait after checking entity box

ayuwono:
This PR adds a wait after checking the entity-only box. Sometimes on jenkins the spinner keeps spinning while waiting for the result. This will fix the one failure on jenkins


aebadirad:
Does this also affect the 4.x line?

PR #1788, opened 2019-01-23
DHFPROD-1710 - create space for the new dhf5 code (still rewriting to…

aebadirad:
(still rewriting to the 4.x code for right now)

Get familiar with this part of the code - this is what we're going to be working in


ryanjdew:
I assume all the XQuery code is currently copied over for the legacy flow. Should we instead have a wrapper reference the legacy code under the datahub/4 folder? 

If I'm misinterpreting or jumping the gun on work to be done (perhaps the method is copy over and prune as we go?), let me know.

aebadirad:
Yes, we'll be referencing the 4.x code. I just moved it all to 5 so we know the space we're playing in when begin the new work. Eventually the 5.x space should have no 4.x code in it at all. And no/minimal xqy

aebadirad:
Yes, yes it should be. Intellij obscured that strangely enough.

PR #1793, opened 2019-01-24
DHFPROD-1737 Add swagger mocks for flows

ryanjdew:


quick-start/api/swagger/mocks.json
```diff
@@ -20,6 +20,236 @@
...
+          }
+        }
+      },
+      "post": {
+        "tags": [
+          "flow"
+        ],
+        "summary": "Create flow",
+        "description": "",
+        "requestBody": {
```
brucean52: For POST requests, I believe the parameters array (line 99). See example below
        "parameters": [
          {
            "in": "body",
```suggestion
        "requestBody": {
```
            "name": "body",
            "description": "Pet object that needs to be added to the store",
            "required": true,
            "schema": {
              "$ref": "#/definitions/Pet"
            }
          }
        ],

ryanjdew: request body for post/put have been moved to parameters and they are appearing in the swagger ui properly

PR #1805, opened 2019-01-28
Feature/swagger mapping process

aebadirad:


quick-start/api/swagger/mocks.json
```diff
@@ -130,6 +130,207 @@
           }
         }
       }
+    },
+    "/mapping-process": {
```
ryanjdew: Should the URL be `/mapping-process`? I think the scheme should be `/process` and fileters for type should be params, like `/process?type=mapping`

PR #1812, opened 2019-01-29
Fixing the "Order" entity

srinathgit:
Fixing the "Order" entity


bsrikan:
I am running into a Null pointer exception when there is another entity. Say Order has Item and Customer as references. Item and Customer have none. TDE generate task for these 3 runs just fine.
When I add another entity say Person, run into the exception.

aebadirad:
Is there a matching person entity file so it knows about the person entity?

bsrikan:
Had it in filesystem not in the server. Deployed and generated TDE again and didnt run into the exception. Still would like to check all the persisted files in FINAL, final-SCHEMAS, staging and staging-SCHEMAS.

aebadirad:
let's log that as a ticket in jira to spell out the test enhancement for this

bsrikan:
Filed a bug because we should not get a null pointer exception.
DHFPROD-1773

PR #1821, opened 2019-01-30
Remove hub-internal-config/schemas as part of upgrade

srinathgit:
This is part of DHFPROD-1764 that got missed earlier. During upgrade from 4.x to 5.x , hub-internal-config/schemas is removed as the path no longer works


aebadirad:
great catch by @rahulvudutala on this and good work @srinathgit on buttoning up the loose end on this

PR #1828, opened 2019-01-30
DHFPROD-1662 - Stop overriding mlAppName if explicitly set - 4.x-develop

akshaysonvane:


aebadirad:
@rjrudin this is basically what you were considering, right?

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/HubConfigImpl.java
```diff
@@ -1710,8 +1710,10 @@ public DatabaseClient newModulesDbClient() {
      * @param config
      */
     private void updateAppConfig(AppConfig config) {
-        // This shouldn't be used for any resource names, but it does appear in logging, and DHF is a better choice than "my-app"
-        config.setName("DHF");
+        // If the user hasn't set the app name then override it to "DHF" instead of "my-app"
```
rjrudin: That's the ticket. It should be possible to write a unit test for this as well.

aebadirad: @akshaysonvane Sounds like we're good on it. Punch up a unit test for it and we'll move ahead

PR #1831, opened 2019-01-31
DHFPROD-1745: Primary key is not displayed on mapping entity table

brucean52:


quick-start/src/main/ui/app/shared/components/mappings/map-ui.component.ts
```diff
@@ -357,4 +357,13 @@ export class MapUiComponent implements OnChanges {
     return _.includes(this.chosenEntity.definition.pii, name);
   }
 
+    /**
+   * Is an entity property personally identifiable information?
```
wooldridge: Should reference PK not PII.

ayuwono:
I'm testing the UI, it seems that something goes wrong with the map if you have multiple entities. Will test around more and update the finding

ayuwono:
If I created two mappings from two entities. It seems that after creating the second mapping, I can't access the first mapping. It's weird that it seems that it's gone (I can't see it on the screen)

ayuwono:
Let me try it again

PR #1839, opened 2019-02-01
DHFPROD-1726 - "Update a Hub Project link" produces error

brucean52:
This is the same fix that was merged in 4.x-develop branch.


aebadirad:
Can we get a PR for this for 4.x-develop?

PR #1848, opened 2019-02-04
Dhfprod 1786

akshaysonvane:


srinathgit:
@akshaysonvane , I think changes must be made in tracing-rewriter.xml as well

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/rest-api/rewriter.xml
```diff
@@ -778,15 +778,15 @@
...
-                        <dispatch>/data-hub/5/rest-api/endpoints/unacceptable-type.xqy</dispatch>
+                        <dispatch>/data-hub/4/rest-api/endpoints/unacceptable-type.xqy</dispatch>
                     </match-header>
-                    <dispatch>/data-hub/5/rest-api/endpoints/apply-transform.xqy</dispatch>
+                    <dispatch>/data-hub/4/rest-api/endpoints/apply-transform.xqy</dispatch>
                 </match-content-type>
             </match-method>
-            <dispatch>/data-hub/5/rest-api/endpoints/apply-transform-bad-request.xqy</dispatch>
+            <dispatch>/data-hub/4/rest-api/endpoints/apply-transform-bad-request.xqy</dispatch>
         </match-path>
```
srinathgit: @aebadirad , @akshaysonvane ,
Why are we changing the data-hub/5/rest-api/rewriter.xml to point to 4's modules for default MarkLogic endpoints ? Shouldn't the change be made only for collector (from 5 to 4)  as it is the only legacy custom endpoint, the rest all can be dispatched to 5's modules ?

PR #1853, opened 2019-02-06
DHFPROD-1428 Improve the usability of text input elements

brucean52:


alexsmr:
Looks good. Thanks.

One note: as long as you're started using 7.x angular material/cdk modules, we better do update to 7.x for the whole application. 

I will create a separate story to upgrade to the next version of angular. This will also include removing MDL task, testing, and moving the rest to 7.x. I will present it to PM and we will groom it at some point. Just a heads up. If you have any small task related to that, you may start on your own, just add it to the sprint.

aebadirad:
Is this also for 4.x? Or is this just for develop?

brucean52:
I can create another PR for 4.x

PR #1854, opened 2019-02-06
e2e test fix to setup Express server when e2e testing + warnings off

clockworked247:
Fix to e2e test execution w/ support for running new Express server.


clockworked247:
@aebadirad Thanks for merging this.

PR #1856, opened 2019-02-07
DHFPROD-1781: storybook fixes

brucean52:


brucean52:
will uncommit package-lock

alexsmr:
@brucean52 Can you also made it against 4.x develop branch? My understanding that 4.x will be merged together with develop at some point. But we need it on 4.x

PR #1868, opened 2019-02-12
Do a case insensitive equality check for entity name when creating an…

srinathgit:
Do a case insensitive equality check for entity name when creating and running flows , if it returns more than one result, revert to strict check.


aebadirad:
Can we also get a version of this for 4.x?

PR #1899, opened 2019-02-21
DHFPROD-1861

wooldridge:
Start of Manage Flows view.


alexsmr:
Mike, maybe we should do that PR against "develop" branch, not feature branch? It will be much easier not to get lost between commits.

wooldridge:
@alexsmr, OK, I will make a new PR to develop.

PR #1900, opened 2019-02-21
DHFPROD-1851 pre/post hook and ensure MLCP transform compatibility

ryanjdew:


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/extensions/runFlow.sjs
```diff
@@ -19,27 +19,46 @@ const datahub = new DataHub();
...
+  else {
+    let options = params.options ? JSON.parse(params.options) : {};
+    let jobId = params.jobId || datahub.hubUtils.uuid();
+    let flow = datahub.flow.getFlow(flowName);
+    let query = null;
+    let uris = null;
+    if (params.uri) {
+      uris = datahub.hubUtils.normalizeToArray(params.uri);
+      query = cts.documentUriQuery(uris);
+    } else {
```
srinathgit: Is this meant to be "cts.documentQueryUris" ?

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/extensions/runFlow.sjs
```diff
@@ -19,27 +19,46 @@ const datahub = new DataHub();
...
+    if (input) {
+      content = input.toObject();
+    } else {
+      content = {};
+      datahub.hubUtils.queryLatest(function() {
+        for (let doc of cts.search(query, cts.indexOrder(cts.uriReference()))) {
+          content[xdmp.nodeUri(doc)] = doc;
+        }
+      }, flow.sourceDb || datahub.flow.globalContext.sourceDb);
+      uris = Object.keys(content);
```
srinathgit: @aebadirad , I thought the collector would run and provide the uris which will be passed to datahub.flow.runFlow() ? Has that changed ?

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/extensions/runFlow.sjs
aebadirad: Is that supposed to be cts.documentQuery?

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/extensions/runFlow.sjs
aebadirad: This still supports that, if I'm reading it correctly, but will also handle the case of running it on the server locally if no uri list is provided

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/jobs.sjs
```diff
@@ -51,15 +51,20 @@ class Jobs {
 
   getJobDocWithId(jobId) {
     let jobUri = "/jobs/" + jobId + ".json";
-    if (cts.contains(cts.doc(jobUri), cts.jsonPropertyValueQuery("jobId",jobId))) {
-      return this.getJobDocWithUri(jobUri);
-    }
+    return hubutils.queryLatest(function() {
```
srinathgit: can you please return fn.head of the query response ? Also in  getJobDocWithUri(jobUri) method

aebadirad:
Going to bring this in and make the changes, just so we can start integration

PR #1902, opened 2019-02-21
DHFPROD-1861: Set up Manage Flows

wooldridge:
Simple view, model, and service functionality for Manage Flows feature.

PR #1903, opened 2019-02-21
DHFPROD-1861: Set up Manage Flows

wooldridge:
Simple view, model, and service functionality for Manage Flows feature.


wooldridge:
@alexsmr, let's try this again. Replaces: https://github.com/marklogic/marklogic-data-hub/pull/1902

PR #1910, opened 2019-02-23
DHFPROD-1891 Create JSON/XML schema to validate against entity instances - 4.x

ryanjdew:


bsrikan:
Will take a look this now.

PR #1915, opened 2019-02-25
Deploy default artifacts

akshaysonvane:
Add LoadHubArtifactsCommand to deploy default steps and flows
Get resource as stream to work in both jar file and via IDE

TODO: Add tests for LoadHubArtifactsCommand


marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/LoadHubArtifactsCommand.java
```diff
@@ -110,75 +97,53 @@ public void execute(CommandContext context) {
...
     public void setHubConfig(HubConfig hubConfig) {
         this.hubConfig = hubConfig;
     }
 
+
+    private void addToWriteSet(InputStream inputStream, String docCollection, File file, DocumentWriteSet stagingDocumentWriteSet, DocumentWriteSet finalDocumentWriteSet, PropertiesModuleManager propertiesModuleManager) {
+        try {
+            StringHandle handle = new StringHandle(IOUtils.toString(inputStream));
+            inputStream.close();
+
```
hao1st: Could we put the close() to finally block?

PR #1923, opened 2019-02-26
DHFPROD-1870: some more improvements to the flow setting dialog

alexsmr:


wooldridge:
@alexsmr, looking good! Some comments:

- Arrange and label the buttons as specified in the wireframes here:

https://wiki.marklogic.com/display/ENGINEERING/Flow+UI+Design?preview=/102766514/102441533/Manage%20Flows%20v6%20(New%20Flow%20dialog%20open).png#FlowUIDesign-PageFlowDiagram

- The Description field is optional needs to be labeled that way.

- The dialog header/title area as it stands has a lot of padding and we'll want to settle on something more compact for all the dialog boxes. This worked for me in flow-settings-dialog.component.scss:

h1.mat-dialog-title {
  margin-top: -12px;
  margin-left: -4px;
  margin-bottom: 11px;
}

div.mat-dialog-content {
  padding: 10px 15px 20px 20px;
}

(We should figure out a way to set styles for those Material dialogs globally because we'll be using them a ton.)

- Batch Size has a default value of 100.

- Thread Count has a default value of 4. 

Default values and other information are here:

https://wiki.marklogic.com/display/ENGINEERING/Flow+UI+Design#FlowUIDesign-UIWireframes

PR #1924, opened 2019-02-26
Get resources as stream to make it accessible in a jar file

akshaysonvane:


marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/DeployDatabaseFieldCommand.java
```diff
@@ -42,16 +47,32 @@ public void execute(CommandContext context) {
     private List<DeployDatabaseCommand> buildDatabaseCommands() {
         List<DeployDatabaseCommand> dbCommands = new ArrayList<>();
 
-        String filePath = Objects.requireNonNull(getClass().getClassLoader().getResource("ml-database-field"))
-            .getFile();
-        File databaseFieldDir = new File(filePath);
+        InputStream inputStream = null;
+        try {
+            ResourcePatternResolver resolver = new PathMatchingResourcePatternResolver(getClass().getClassLoader());
```
hao1st: A subtle suggestion:
Could we use ResourcePatternResolver.class (compiler time) rather than runtime getClass()?

akshaysonvane: Sure @hao1st 
Changes incorporated 17819a8

PR #1933, opened 2019-02-26
Add Target Entity data to Manage Flows table

wooldridge:
Updates mock data for mapping steps to add targetEntity.

Also fixes handling of status data (i.e., “Never Run” if there is no latest job).

### DIFF
--- a/quick-start/src/main/ui/app/components/flows-new/models/flow.model.ts
+++ b/quick-start/src/main/ui/app/components/flows-new/models/flow.model.ts
@@ -27,7 +27,7 @@ export class Flow {
   constructor() {}
 
   get status(): string {
-    return this.latestJob['status'];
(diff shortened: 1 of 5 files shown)


quick-start/src/main/ui/app/components/flows-new/models/flow.model.ts
```diff
@@ -27,7 +27,7 @@ export class Flow {
   constructor() {}
 
   get status(): string {
-    return this.latestJob['status'];
+    return (this.latestJob['status']) ? this.latestJob['status'] : 'Never Run';
```
clockworked247: just to ensure you don't hit an issue with an empty latestJob, change this to:

`return (this.latestJob && this.latestJob['status']) ? this.latestJob['status'] : 'Never Run';`

wooldridge:
Thanks, @clockworked247. Added fix.

PR #1943, opened 2019-03-01
DHFPROD-1911 - Swagger updates Ingest, Mapping & Mastering step types

clockworked247:
Added the JSON in support of Ingest, Mapping & Mastering step types

Farmed Ingest from current develop settings and Mapping JSON from 4.x-develop, since mapping is broken on the develop branch.

Good first cut, but will likely need work next week when I return.   Should be enough for the UI team to get started with their work!

### DIFF
--- a/quick-start/api/swagger/mocks.json
+++ b/quick-start/api/swagger/mocks.json
@@ -1,5 +1,5 @@
 {
-  "openapi": "3.0.0",
+  "swagger": "2.0",
   "info": {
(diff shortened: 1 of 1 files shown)


alexsmr:
@aebadirad can you please merge that PR. Thanks.

wooldridge:
Sorry, just saw this. Nice work. @clockworked247 and I have different ideas with regard to where the matching/merging options that the API requires are constructed. We can discuss on Monday.

PR #1946, opened 2019-03-02
Bugfix/dhfprod 1930

aebadirad:


marklogic-data-hub/src/main/java/com/marklogic/hub/legacy/collector/DiskQueue.java
```diff
@@ -162,15 +163,17 @@ private static void closeQuietly(Closeable obj) {
 
     private void openFile() throws IOException {
         if (fileQueue == null) {
-            fileQueue = File.createTempFile(DiskQueue.class.getSimpleName() + "-backingstore-", null, tempDir);
+            fileQueue = File.createTempFile(com.marklogic.hub.collector.DiskQueue.class.getSimpleName() + "-backingstore-", null, tempDir);
             fileQueue.deleteOnExit();
```
srinathgit: Is there any reason the prefix of the temp file is changed to "com.marklogic.hub.collector.DiskQueue.class.getSimpleName()" rather the existing one "com.marklogic.hub.legacy.collector.DiskQueue.class.getSimpleName()"  for the legacy DiskQueue?

gradle.properties
```diff
@@ -1,2 +1,4 @@
 version=5.0-SNAPSHOT
 publishUrl=file:../marklogic-data-hub/releases
+systemProp.file.encoding=utf-8
+org.gradle.jvmargs='-Dfile.encoding=UTF-8'
```
srinathgit: Are these changes needed for handling the URI with diacritics ?

marklogic-data-hub/src/main/java/com/marklogic/hub/legacy/collector/DiskQueue.java
aebadirad: Whoops. Incorrect auto-classpath rendering. While it makes sense for now, in the near future we won't be duplicating like we are now, fixed.

gradle.properties
aebadirad: Not exactly. With this (and no other code changes), our gradle plugin will start handling them, because we're forcing the build to happen in UTF8 - though in general we should be always adhering to UTF-8 in our builds because that's what marklogic server uses by default.

PR #1957, opened 2019-03-05
DHFPROD-1924: Metadata and options recording (add option.headers)

hao1st:
https://project.marklogic.com/jira/browse/DHFPROD-1924

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/ingest/default/main.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/ingest/default/main.sjs
@@ -23,21 +23,17 @@ function main(id, content, options) {
     }
   }
 
-  let headers = createHeaders(id, content, options);
+  let headers = createHeaders(options);
+
   let envelope = datahub.flow.flowUtils.makeEnvelope(instance, headers, triples, outputFormat);
 
   return envelope;
 }
 
-function createHeaders(id, content, options) {
+function createHeaders(options) {
   let headers = {};
   for (let key in options.headers) {
-    headers[key] = options.headers[key];
-    if (headers[key] == datahub.flow.consts.CURRENT_DATE_TIME) {
-      headers[key] = fn.currentDateTime;
-    } else if (headers[key] == datahub.flow.consts.CURRENT_USER) {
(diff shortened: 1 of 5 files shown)


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/ingest/default/main.sjs
```diff
@@ -27,13 +23,25 @@ function main(id, content, options) {
...
 
   return envelope;
 }
 
+function createHeaders(id, content, options) {
+  let headers = {};
+  for (let key in options.headers) {
+    headers[key] = options.headers[key];
+    if (headers[key] == datahub.flow.consts.CURRENT_DATE_TIME) {
+      headers[key] = fn.currentDateTime;
```
aebadirad: should be fn.currentDateTime()

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/ingest/default/main.sjs
```diff
@@ -27,13 +23,25 @@ function main(id, content, options) {
...
-
+  let headers = createHeaders(id, content, options);
   let envelope = datahub.flow.flowUtils.makeEnvelope(instance, headers, triples, outputFormat);
 
   return envelope;
 }
 
+function createHeaders(id, content, options) {
+  let headers = {};
+  for (let key in options.headers) {
```
aebadirad: Is there a way we can functionize this to the hubUtils library for checking over the constants and returning the value so we just need to loop over the object and use the function to check and return the correct value?

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/ingest/default/main.sjs
hao1st: yes. i missed it when I copied from qconsole.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/ingest/default/main.sjs
hao1st: Sure. we can do that.

PR #1972, opened 2019-03-07
DHFPROD-1933: Create and verify different UI states

brucean52:


brucean52:
Note: Creating a mastering step will crash the app. I will address this in another PR. Creating ingest, mapping or custom step works fine.

wooldridge:
Thanks, @brucean52. Merging so we can move forward with the flow/step UI work.

wooldridge:
@alexsmr, can you review and approve? Looks like that is required for a merge since you requested changes.

PR #1980, opened 2019-03-09
DHFPROD-1931: fixing flows schema and refresh

alexsmr:


alexsmr:
The flows CRUD operations finally seem to start working end-to-end.

Note: mock server does not persist options changes. @clockworked247, please verify the problem.

PR #2002, opened 2019-03-13
DHFPROD-1969: Add Run flow and stop flow endpoint functionality in mock v2

brucean52:


quick-start/api/mocks/service/FlowsService.js
```diff
@@ -105,9 +105,32 @@ exports.updateFlow = function(flowId, body) {
...
+          clearInterval(interval);
+          currentFlow.isRunning = false;
+          currentFlow.latestJob.status = 'finished';
+        } else {
+          currentFlow.latestJob.runningPercent += 10;
+          if (currentFlow.latestJob.runningPercent > 100) {
+            currentFlow.latestJob.runningPercent = 100;
+            currentFlow.isRunning = false;
+            currentFlow.latestJob.status = 'finished';
+          }
```
hao1st: https://github.com/marklogic/marklogic-data-hub/blob/61488685842e765e25ce3f66ddcaccf08adb541a/quick-start/api/mocks/service/FlowsService.js#L124-L128

Do we really need to have those lines?

brucean52: Not sure if it's required, but it's a check to ensure the percentage doesn't go over 100.

alexsmr: It does not make sense to keep it here. But the check should be made at the beginning of both run/stop that the flow is not running or is running already. I do not see this logic. Otherwise, the client can by mistake execute the call twice and trigger two different interval updates.

Issue #1825, opened 2019-01-30
Parallel processing in Flow Runner is weak


josvanroosmalen:
Parallel processing in `FlowRunnerImpl` is weak. 

First, the current `awaitCompletion` contains a bug. Why do I pass a `unit` when `MILLISECONDS`is used? 

    @Override
    public void awaitCompletion(long timeout, TimeUnit unit) throws InterruptedException {
        if (runningThread != null) {
            runningThread.join(unit.convert(timeout, TimeUnit.MILLISECONDS));
        }
    }

Second, if I call `awaitCompletion` with a time-out, I do not get any information that it has been timed out or not. So this implementation need to change to something like:

    public void awaitCompletion(long timeout, TimeUnit unit) throws InterruptedException {
        if (runningThread != null) {
            runningThread.join(unit.convert(timeout, TimeUnit.MILLISECONDS));
            if ( runningThread.getState() != Thread.State.TERMINATED) {
                queryBatcher.stop();
                runningThread.interrupt();
                throw new InterruptedException("Time-out occurred.");
            }
        }
    }

Third, the `runningThread` cannot be managed from outside such as an `ExecutorService` eg obtained in an managed form from  an Application Server.  BTW: This is also the case from the MarkLogic Java API ThreadPool used by the `QueryBatcherImpl`.

aebadirad:
As of Data Hub 5, parallel processing is now blocked in flow runner and all tasks are serial in a blockingqueue. 

For parallel execution please either execute multiple flowrunners or, preferably, use an external orchestrator that is more in line with handling massive scale parallel orchestration such as NiFi, Kafka, Kinesis, etc.

PR #2003, opened 2019-03-13
Flow Runner improvements (fixing GitHub issue #1825 etc)

josvanroosmalen:
-	It optimized execution for maximum performance:
o	There is now an option to write only a job to the job database in case of a failure. This speeds up execution.  By default this is turned off (so always write to Jobs database). 

FlowRunner withOnlyWriteJobOnFailure(boolean onlyWriteJobOnFailure);

o	If the collector returns nothing, we bypass DMSDK/runningThread construction and immediately return a finished job. 
-	There is now support to run an Harmonization flow, without invoking the collector, e.g. in case you already know the harmonization ids via:
-	
-	Job run(Collection<String> uris);

-	Job has now a totalEvents count. That is because total cannot be guaranteed to be success + failed. 
-	awaitCompletion has now a bug fixed around timeUnits (it was passed but not used)
-	awaitCompletion now throws a TimeoutException in case of a Timeout, and it stops the underlying querybatcher properly. 
-	Some setters added to support hard-wiring instead of relying/requiring Spring Autowiring.

### DIFF
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/FlowManager.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/FlowManager.java
@@ -34,6 +34,12 @@ public interface FlowManager {
     String FLOW_FILE_EXTENSION = ".flow.json";
 
     /**
+     * Set the HubConfig
+     *
+     * @param hubConfig - the hubConfig to use
+     */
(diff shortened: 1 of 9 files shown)


aebadirad:
@josvanroosmalen we've moved commits from this PR into #2025 and merged them down into develop to better fit with the work in progress, so closing this PR.

PR #2011, opened 2019-03-15
DHFPROD-1828: Feature/qs to web

aebadirad:
Should clean up tests and build for new web app, produce proper named war file


bsrikan:
The passed check here is for dhf-core and not QS. QS job that had run before has failed. The jenkins build integration is not keeping both the runs.
https://jenkins.marklogic.com/job/dhf-QS-develop-rh7-Commit/812/console

PR #2022, opened 2019-03-18
DHFPROD-1989: Steps TODOs

brucean52:


brucean52:
Issue is related to cdk stepper not updating until ngAfterContentChecked.

PR #2038, opened 2019-03-19
DHFPROD-1957: FE Implementation: Custom step

brucean52:


wooldridge:
@brucean52, can you check if the form value in the Custom Module URI field has changed and keep the Save button disabled if there hasn't been a change? 

FWIW, for the Mapping config, I have the Save/Reset buttons in the step code and move the buttons up to the header via absolute positioning with CSS. I found it was easier to keep track of dirty state of the form that way and have it reflected in the buttons. YMMV.

https://github.com/marklogic/marklogic-data-hub/blob/develop/web/src/main/ui/app/components/flows-new/edit-flow/mapping/ui/mapping-ui.component.html#L8

https://github.com/marklogic/marklogic-data-hub/blob/develop/web/src/main/ui/app/components/flows-new/edit-flow/mapping/ui/mapping-ui.component.scss#L15

ayuwono:
It seems that there is a connection issue on PR merge. @SameeraPriyathamTadikonda, can we re-launch it?

PR #2042, opened 2019-03-20
DHFPROD-1957: build fixes and updated custom save

brucean52:
Disabling the save button does not work for the 'mocked' custom save step due to the customModuleUri = undefined.

Please create a new custom step to test the functionality.


wooldridge:
@brucean52 explained that when creating steps from scratch, which is what the user will be doing, the functionality works. And I confirmed that this is the case. Will approve.

alexsmr:
@brucean52 don't forget to create the PR against develop branch

PR #2050, opened 2019-03-20
Feature/dhfprod 1856

ryanjdew:


aebadirad:
Did this get stuck?

kkanthet:
Alex,

We have removed auto merge for feature branches  for now because its causing more work to rebase and merge .
Auto merge is only on develop for now . we will re visit this after this sprint or couple of sprints.,

Thank You,
Kiran

PR #2081, opened 2019-03-22
DHFPROD-1417: Feature/dhfprod 1417 new roles

aebadirad:
Running tests to check for new roles - also probably should find a better way than hard coding the different users and instead being able to swap between users on the hubconfig would be better


srinathgit:
cert-auth tests are going to fail. The common name in the certs  should match the 'username' in security db. So, the certs are invalid, new certs have to be generated or we should generate 2 users with old name ("data-hub-user and hub-admin-user) and assign the flowOperator and flowDeveloper roles to them

PR #2082, opened 2019-03-22
DHFPROD-1889: Feature/dhfprod 1885 1889

aebadirad:
Want to run the test against this branch


marklogic-data-hub/src/main/java/com/marklogic/hub/step/impl/QueryStepRunner.java
```diff
@@ -0,0 +1,457 @@
...
+                try {
+                    FlowResource flowResource;
+
+                    if (databaseClientMap.containsKey(batch.getClient())) {
+                        flowResource = databaseClientMap.get(batch.getClient());
+                    } else {
+                        flowResource = new FlowResource(batch.getClient(), destinationDatabase, flow);
+                        databaseClientMap.put(batch.getClient(), flowResource);
+                    }
+                    options.put("uris", batch.getItems());
```
ryanjdew: This is a shared Map Object so race conditions can occur when setting "uris". 

reqOptions
```suggestion
                    HashMap<> reqOptions = new HashMap<>(options);
                    reqOptions.put("uris", batch.getItems());
```

marklogic-data-hub/src/main/java/com/marklogic/hub/step/impl/QueryStepRunner.java
```diff
@@ -0,0 +1,457 @@
...
+
+                    RunStepResponse response = flowResource.run(job.getJobId(), step, options);
```
ryanjdew:
```suggestion
                    RunStepResponse response = flowResource.run(job.getJobId(), step, reqOptions);
```

PR #2084, opened 2019-03-22
DHFPROD-1872: Fix how tests are run

bsrikan:
Changes made so that tests can run in jenkins reliably. Once they do will create another PR ready to be merged for 1872 that has some tests and page objects for manage flow page.

Jenkins tests will now run in port 4200, since there is a blocker issue in running them on 8080.


bsrikan:
Verify following run where UI tests have run and passed:
https://jenkins.marklogic.com/view/DHF-9.0-Server-All/job/NO_CI_dhf-qs-develop-rh7/396/

PR #2085, opened 2019-03-22
DHFPROD-1888: Add endpoints for PUT, DELETE step

akshaysonvane:


web/src/main/java/com/marklogic/hub/web/service/FlowManagerService.java
```diff
@@ -114,24 +116,34 @@ public StepModel createStep(String flowName, String stepJson) {
         Step step = Step.create("dummy", Step.StepType.CUSTOM);
 
         // TODO: Transform stepModel to step, Save the step on disk, Add the stepJson to Flow
+        // NOTE: Only save step if step is of Custom type, for rest use the default steps.
```
hao1st: we had better throw new DataHubException(...) so that we can send error response payload.

PR #2096, opened 2019-03-25
DHFPROD-2039: Need to go back to Manage Flows page when deleting a flow from that flow page

brucean52:


web/src/main/ui/app/components/flows-new/edit-flow/ui/edit-flow-ui.component.ts
```diff
@@ -117,6 +117,7 @@ export class EditFlowUiComponent {
     dialogRef.afterClosed().subscribe(response => {
       if (response) {
         this.deleteFlow.emit(this.flow.id);
+        this.router.navigate(['flows']);
```
alexsmr: That shouldn't be done from ui implementation, please move it to logic component.

PR #2099, opened 2019-03-26
Make FlowRunnerImpl run mapping step

srinathgit:
Make FlowRunnerImpl run mapping step

### DIFF
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/flow/impl/FlowRunnerImpl.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/flow/impl/FlowRunnerImpl.java
@@ -16,6 +16,7 @@ import com.marklogic.hub.step.StepRunnerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 
+import java.text.SimpleDateFormat;
 import java.util.*;
 import java.util.concurrent.*;
 import java.util.concurrent.atomic.AtomicBoolean;
@@ -178,7 +179,7 @@ public class FlowRunnerImpl implements FlowRunner{
         @Override
         public void run() {
             RunFlowResponse resp = flowResp.get(runningJobId);
-            resp.setStartTime(Calendar.getInstance());
+            resp.setStartTime(new SimpleDateFormat("yyyy-mm-dd hh:mm:ss").format(Calendar.getInstance().getTime()));
             Queue<String> stepQueue = stepsMap.get(jobQueue.peek());
 
             AtomicInteger errorCount = new AtomicInteger();
@@ -238,8 +239,8 @@ public class FlowRunnerImpl implements FlowRunner{
                 jobUpdate.postJobs(jobId, JobStatus.FINISHED.toString(), runningStep.getName());
             }
             resp.setStepResponses(stepOutputs);
-            resp.setEndTime(Calendar.getInstance());
-            jobQueue.remove(jobQueue.peek());
+            resp.setEndTime(new SimpleDateFormat("yyyy-mm-dd hh:mm:ss").format(Calendar.getInstance().getTime()));
+            jobQueue.remove();
             if(!jobQueue.isEmpty()) {
                 initializeFlow((String) jobQueue.peek());
             }
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/flow/impl/FlowRunnerImpl.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/flow/impl/FlowRunnerImpl.java
@@ -36,6 +36,7 @@ public class FlowRunnerImpl implements FlowRunner{
 
     private AtomicBoolean isRunning = new AtomicBoolean(false);
     private AtomicBoolean isJobCancelled = new AtomicBoolean(false);
+    private AtomicBoolean isJobSuccess = new AtomicBoolean(true);
 
     private String runningJobId;
     private Step runningStep;
@@ -148,6 +149,7 @@ public class FlowRunnerImpl implements FlowRunner{
 
     private void initializeFlow(String jobId) {
         isRunning.set(true);
+        isJobSuccess.set(true);
         runningJobId = jobId;
         runningFlow = flowMap.get(runningJobId);
         if(jobUpdate == null) {
@@ -202,7 +204,6 @@ public class FlowRunnerImpl implements FlowRunner{
                     .withOptions(flow.getOverrideOptions())
                     .onItemFailed((jobId, itemId)-> {
                         errorCount.incrementAndGet();
-                        // TODO: Add to Flow Model
                         if(flow.isStopOnError()){
                             stopJob(jobId);
                         }
@@ -228,9 +229,12 @@ public class FlowRunnerImpl implements FlowRunner{
                 Job stepResp = stepRunner.run();
                 stepRunner.awaitCompletion();
                 stepOutputs.put(stepNum, stepResp);
+                if(! stepResp.isSuccess()) {
+                    isJobSuccess.set(false);
+                }
 
             }
-            if(errorCount.get() > 0){
+            if(! isJobSuccess.get()){
                 jobUpdate.postJobs(jobId, JobStatus.FINISHED_WITH_ERRORS.toString(), runningStep.getName());
             }
             else if(isJobCancelled.get()) {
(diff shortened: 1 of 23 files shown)


marklogic-data-hub/src/main/java/com/marklogic/hub/flow/impl/FlowRunnerImpl.java
```diff
@@ -222,8 +224,9 @@ public void run() {
                 }
 
                 Job stepResp = stepRunner.run();
-                stepOutputs.put(stepNum, stepResp);
                 stepRunner.awaitCompletion();
+                stepOutputs.put(stepNum, stepResp);
+
```
hao1st: https://github.com/marklogic/marklogic-data-hub/blob/423a73efb7ad178b25fd088458df1abcbf52196b/marklogic-data-hub/src/main/java/com/marklogic/hub/flow/impl/FlowRunnerImpl.java#L242-L245

I think we do not need to submit the thread again called in initializeFlow. We can do sth like:

Also I feel we do not need to peek from jobQueue. And remove it later. And you can use: jobQueue.remove() rather than jobQueue.remove(jobQueue.peek());

            while (!jobQueue.isEmpty()) {
                Queue<String> stepQueue = stepsMap.get(jobQueue.remove());
                while (!stepQueue.isEmpty()) {
                    //...
                }
            }

marklogic-data-hub/src/main/java/com/marklogic/hub/step/impl/WriteStepRunner.java
```diff
@@ -185,6 +185,7 @@ public void awaitCompletion(long timeout, TimeUnit unit) throws InterruptedExcep
     public Job run() {
         runningThread = null;
         Job job = createJob();
+        jobUpdate = new JobUpdate(hubConfig.newJobDbClient());
```
hao1st: I was wondering any benefit to create another thread for step runner (write/read) since we has spawned a thread in flowrunner and we execute step tasks concurrently...

marklogic-data-hub/src/main/java/com/marklogic/hub/flow/impl/FlowRunnerImpl.java
srinathgit: 1. I included jobQueue.remove() in the latest commit.
2. FlowRunnerTask is meant to run a single flow, that's why I create another Runnable and submit to the executor service in case if the queue is not empty

marklogic-data-hub/src/main/java/com/marklogic/hub/step/impl/WriteStepRunner.java
srinathgit: I am not sure if I understand this correctly.  StepRunners are just adoption of the previous 4.x FlowRunnerImpl

marklogic-data-hub/src/main/java/com/marklogic/hub/flow/impl/FlowRunnerImpl.java
hao1st: We can discuss tomorrow.  My point is that we can let the available jobs in jobQueue run one by one (outer loop) in the thread until the thread exits, and we do not need to submit the runnable again in the run method.

hao1st: Also why do we need to threadPool.shutdownNow() for each flowrunner thread? we can use a shutdownHook thread to manage the pool when jvm exits.

srinathgit:
@hao1st ,  intellij automatically converts to wild card (*) when more than certain number of classes from the package are imported.  I didn't want to clutter the code with a lot of lines of import statements

web/src/main/java/com/marklogic/hub/web/service/FlowManagerService.java
```diff
@@ -221,14 +224,26 @@ public void deleteStep(String flowName, String stepId) {
...
         }
-        return resp.getJobId();
+        return this.getFlow(flowName);
+    }
+
+    public Flow stop(String flowName) {
+        if (flowRunner.getRunningFlow().equals(flowName)) {
+            flowRunner.stopJob(flowRunner.getRunningJobId());
+        }
+        else{
```
akshaysonvane: We also need to handle flows that are not running but are still part of the job queue.

web/src/main/java/com/marklogic/hub/web/web/FlowController.java
```diff
@@ -133,7 +133,14 @@
...
-    public String runFlow(@PathVariable String flowName, @RequestBody String[] steps) {
-        return flowManagerService.runFlow(flowName, steps);
+    public ResponseEntity<?> runFlow(@PathVariable String flowName, @RequestBody List<String> steps) {
+        Flow flow = flowManagerService.runFlow(flowName, steps);
+        return new ResponseEntity<Flow>(flow, HttpStatus.OK);
+    }
+
+    @RequestMapping(value = "/flows/{flowName}/stop", method = RequestMethod.POST)
+    @ResponseBody
+    public String stopFlow(@PathVariable String flowName) {
```
akshaysonvane: The return type should be `ResponseEntity<?>`

marklogic-data-hub/src/main/java/com/marklogic/hub/step/impl/StepImpl.java
```diff
@@ -26,6 +26,8 @@
 import java.util.Map;
 
 public class StepImpl implements Step {
+    public final static int DEFAULT_BATCH_SIZE = 100;
+    public final static int DEFAULT_THREAD_COUNT = 4;
```
akshaysonvane: We don't need defaults as we fallback on the properties set in the Flow.

web/src/main/java/com/marklogic/hub/web/web/FlowController.java
```diff
@@ -133,7 +133,14 @@
...
+        return flowManagerService.stop(flowName);
```
akshaysonvane: Should be something like this
`return new ResponseEntity<Flow>(flow, HttpStatus.OK);`

web/src/main/java/com/marklogic/hub/web/service/FlowManagerService.java
srinathgit: I spoke with @alexsmr,  he mentioned that from UI a flow cannot be started again until previous run completes, so I don't think we have to handle that case.

web/src/main/java/com/marklogic/hub/web/web/FlowController.java
srinathgit: Fixed in the latest commit

marklogic-data-hub/src/main/java/com/marklogic/hub/step/impl/StepImpl.java
srinathgit: Fixed in the latest commit

web/src/main/java/com/marklogic/hub/web/web/FlowController.java
srinathgit: Fixed in the latest commit

web/src/main/java/com/marklogic/hub/web/service/FlowManagerService.java
akshaysonvane: Okay got it.

srinathgit:
@aebadirad ,
Can you merge this PR if you are ok with these changes ?

web/src/main/java/com/marklogic/hub/web/service/FlowManagerService.java
```diff
@@ -287,7 +303,7 @@ private String getStepKeyInStepMap(Flow flow, String stepId) {
             String[] key = new String[1];
 
             flowManager.getSteps(flow).forEach((k, v) -> {
-                if (name.equals(v.getName()) && type.equals(v.getType().toString())) {
+                if (name.equals(v.getName()) && type.toLowerCase().equals(v.getType().toString().toLowerCase())) {
```
akshaysonvane: Can we use this? Makes code more readable.
```suggestion
                if (name.equals(v.getName()) && type.equalsIgnoreCase(v.getType().toString())) {
```

aebadirad: Yeah, it's a minor nitpick, but it makes it easier

srinathgit:
I will make the change and open clean PR, some of commits are appearing again here

aebadirad:
> I will make the change and open clean PR, some of commits are appearing again here

thanks :)

PR #2108, opened 2019-03-27
DHFPROD-1678: Feature/DHFPROD-1678 Add custom merge options, namespaces

wooldridge:
Add custom merge options for merging configuration.
Add namespace settings for custom match options, custom merge options, and custom threshold actions.
Uppercase button names in mastering. (Confirm dialogs still need work.)
Use truncate pipe for long values

### DIFF
--- a/web/src/main/ui/app/components/flows-new/edit-flow/mastering/matching/match-options.model.ts
+++ b/web/src/main/ui/app/components/flows-new/edit-flow/mastering/matching/match-options.model.ts
@@ -77,6 +77,7 @@ export class MatchOption {
   public zip9match5: number;
   public customUri: string;
   public customFunction: string;
+  public customNs: string;
   public editing: boolean = false;
 
   constructor (mOpt: any = {}) {
@@ -93,6 +94,7 @@ export class MatchOption {
     if (mOpt.zip9match5) this.zip9match5 = mOpt.zip9match5;
     if (mOpt.customUri) this.customUri = mOpt.customUri;
     if (mOpt.customFunction) this.customFunction = mOpt.customFunction;
+    if (mOpt.customNs) this.customNs = mOpt.customNs;
     // Adjust property names if reduce type
     if (mOpt.matchType === 'reduce' && mOpt.propertiesReduce) {
       this.propertyName = mOpt.propertiesReduce.map(prop => {
@@ -139,6 +141,7 @@ export class MatchOption {
       });
       if (alg.at) result.customUri = alg.at;
       if (alg.function) result.customFunction = alg.function;
+      if (alg.namespace) result.customNs = alg.namespace;
     }
     return result;
   }
--- a/web/src/main/ui/app/components/flows-new/edit-flow/mastering/matching/match-thresholds.model.ts
+++ b/web/src/main/ui/app/components/flows-new/edit-flow/mastering/matching/match-thresholds.model.ts
@@ -39,6 +39,7 @@ export class MatchThresholds {
     if (thr.action === 'custom') {
       if (thr.customUri) mThr.customUri = thr.customUri;
       if (thr.customFunction) mThr.customFunction = thr.customFunction;
+      if (thr.customNs) mThr.customNs = thr.customNs;
     }
     this.thresholds.splice(index, 1, mThr);
   }
@@ -67,6 +68,7 @@ export class MatchThreshold {
   public action: string;
   public customUri: string;
   public customFunction: string;
+  public customNs: string;
 
   constructor(mThr: any) {
     if (mThr.label) this.label = mThr.label;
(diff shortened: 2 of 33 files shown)


ayuwono:
@wooldridge, on editing custom match threshold, it seems that I can't edit the URI, function, and namespace. These fields don't appear on edit settings

wooldridge:
@ayuwono the threshold issues should be fixed with the latest commit.

wooldridge:
Feature updates to merge strategy management:

- When you update values for a strategy, the merge-options values that are defined using that strategy are updated as well
- You cannot edit merge-option values that are defined via a strategy directly (edit-in-place is turned off for those values). You must edit the strategy values (see previous bullet).
- When you delete a strategy, merge options defined using that strategy are deleted as well. There is a warning in the delete confirmation dialog about this.

@ayuwono can you test these updates?

ayuwono:
> Feature updates to merge strategy management:
> 
> * When you update values for a strategy, the merge-options values that are defined using that strategy are updated as well
> * You cannot edit merge-option values that are defined via a strategy directly (edit-in-place is turned off for those values). You must edit the strategy values (see previous bullet).
> * When you delete a strategy, merge options defined using that strategy are deleted as well. There is a warning in the delete confirmation dialog about this.
> 
> @ayuwono can you test these updates?

sure, testing it

ayuwono:
The last commit looks good @wooldridge, just one thing (scenario on bullet no 2):
- When editing max values and max source directly from a merge strategy, the values on merge option that's using that strategy are not updated. You need to edit setting on that merge option, and click on save to see the values updated. 
The other values are fine: source name, source weight, and length weight. The updates are reflected directly when doing edit-in-place on merge strategy

ayuwono:
So, does it mean we can have multiple default strategies?

Screen Shot 2019-03-28 at 2 23 27 PM

wooldridge:
@ayuwono, I've added one more setting to mastering, the Timestamp Path setting under the Merging tab. It stands alone as a text box and Save button unlike all the other settings that are in tables. I think the UI for the Mastering step (for Matching and Merging) should be complete with all the planned features now.

I also (tried to) fix the value persistence issue. Edited values should now properly persist if you reload the page or go to Manage Flows and back. 

Note that for the edit-in-place of strategy values, you will still see slightly different behavior between Source Weights and Length Weight, and Max Values and Max Sources. The values for the former change in real time across tables while the values for the latter require a submit (via an Enter keypress or click outside) for the corresponding information to change in the other table. I haven't figured out how to rectify this yet. I think it is a minor issue (values should still be properly updated), but it is worth filing a bug for. (I'm hoping you're seeing the same behavior I'm seeing; let's discuss if you do not.)

wooldridge:
@ayuwono Re: your comment about the duplicate default strategies... I don't think we should allow this since I can't think of any reason a user would want such a configuration and, if it was allowed, it's not clear to me how the system would parse it. (How do you deal with conflicting values?) Can we add this issue as a bug and tackle it when we work on improving validation?

ayuwono:
> @ayuwono Re: your comment about the duplicate default strategies... I don't think we should allow this since I can't think of any reason a user would want such a configuration and, if it was allowed, it's not clear to me how the system would parse it. (How do you deal with conflicting values?) Can we add this issue as a bug and tackle it when we work on improving validation?

thanks @wooldridge , will test this feature and file the necessary issues

ayuwono:
@wooldridge , for the persisted values on edit-in-place, it seems that it's still not working on all values 
if you don't hit return. For example: 
1) Change the value on one of the Match Options weight (don't hit return)
2) Go back to Manage Flows page
3) Go to the mastering step where you changed the value, you'll see the value is not changed

If you hit return after changing the values, the new values are retained after you go back to manage flows page

Are you seeing the same thing? Maybe you see them changes because you hit return every time? If not, I can webex you to show the behavior

wooldridge:
@ayuwono the intent of the design was to require the user to hit return to save a changed value. (Clicking away from the field being edited also triggers a save.)

ayuwono:
> @ayuwono the intent of the design was to require the user to hit return to save a changed value. (Clicking away from the field being edited also triggers a save.)

@wooldridge hitting the return is working to save the value, clicking away from the field is not saving the value (value is not modified when revisit the step later)

wooldridge:
@ayuwono the issue with changed edit-in-place values not persisting after a click-away should be fixed.

ayuwono:
Test on the latest commit looks good, edit-in-place is saving values and persisting them everywhere

wooldridge:
@ayuwono I've added edit-in-place for Distance Threshold and Zip values. Can you test?

ayuwono:
> @ayuwono I've added edit-in-place for Distance Threshold and Zip values. Can you test?

testing it

ayuwono:
Everything looks good

wooldridge:
Merged with latest develop with no ill effects, as far as I can tell. @alexsmr @brucean52 can you review so I can merge?

PR #2109, opened 2019-03-27
DHFPROD-2046: Provenance merge tracking

clockworked247:
@ryan321 

Features:

- Added ability to track merge of properties on a merge step through the creation of document property records, merge records and a new document record based on those merge records.
- Added a queue with commit() and discard() features that allow queued provenance records to be committed in a single call or discarded on error. A flag for turning on this feature must be added to the module.

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/prov.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/prov.sjs
@@ -29,13 +29,14 @@ class Provenance {
     this.config = {};
     this.config.granularityLevel    = config && config.granularityLevel || 'coarse';
     this.config.JOBDATABASE         = defaultConfig.JOBDATABASE || 'data-hub-JOBS';
-    this.config.autoCommit          = config && config.autoCommit || true;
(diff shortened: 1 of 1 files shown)


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/prov.sjs
```diff
@@ -227,79 +230,79 @@ class Provenance {
...
+   *      - influencedBy (mastering)
    *      - influencedBy (component [aka. custom code])
    *      - associatedWith (flow id)
    *    attributes:
    *      - location (doc URI)
    */
-  _createMasteringStepRecord(jobId, flowId, stepType, docURI, info) {
-    let provId = this.hubutils.uuid();
-    let provTypes = ['ps:Flow','ps:EntityProperty','dhf:Mastering'];
+  _createMasteringStepRecord(jobId, flowId, docURI, info) {
```
ryanjdew:
```suggestion
  _createMasterStepRecord(jobId, flowId, docURI, info) {
```
We are using the step type "master" instead of "mastering", so I believe this needs to change for call on line 357 to work properly.

PR #2110, opened 2019-03-27
DHFPROD-1885: feature/dhfprod 1885 1889

hao1st:


web/src/main/java/com/marklogic/hub/web/model/FlowStepModel.java
```diff
@@ -0,0 +1,182 @@
...
+        }
+        final String[] count = {"1"};
+        steps.forEach(s -> {
+            JSONObject stepJson = new JSONObject((JsonNode) s);
+            String id = stepJson.getString("id");
+            String stepName = stepJson.getString("name");
+            Step step;
+            if (keyById.containsKey(id)) { //update
+                step = existingSteps.get(keyById.get(id));
+            } else {
```
akshaysonvane: Not required as Request is not going to update Step in here.
Any update to the Steps will go via the Steps API
Only reordering of Steps will be done via Flows endpoint.

web/src/main/java/com/marklogic/hub/web/model/FlowStepModel.java
```diff
@@ -0,0 +1,182 @@
...
+        List<StepSummary> stepModels = new ArrayList<>();
+        fsm.setStepModels(stepModels);
+
+        steps.forEach((name, step) -> {
+            StepSummary sm = new StepSummary();
+            sm.id = step.getId();
+            sm.name = step.getName();
+            sm.type = step.getType() == null ? "" : step.getType().toString();
+            if (step.getOptions() != null && step.getOptions().get("targetEntity") != null) {
+                sm.targetEntity = (String) step.getOptions().get("targetEntity").toString();
```
akshaysonvane: Cast to TextNode and use `asText()` on it

marklogic-data-hub/src/main/java/com/marklogic/hub/flow/impl/FlowImpl.java
```diff
@@ -97,18 +97,18 @@ public String getName() {
...
-
     @Override
     public String getId() {
         return id;
     }
 
     @Override
     public void setId(String id) {
-        this.id = id;
+        this.id = id == null ? UUID.randomUUID().toString() : id;
```
akshaysonvane: Don't need ID field as flowId == flowName
That is how we search for the Flow in the filesystem.

marklogic-data-hub/src/main/java/com/marklogic/hub/step/impl/StepImpl.java
```diff
@@ -177,9 +174,20 @@ public void incrementVersion() {
         setVersion(getVersion() + 1);
     }
 
+    @Override
+    public void setId(String id) {
```
akshaysonvane: Don't need ID field as stepId == stepName-stepType
That is how we search for the Step in the filesystem.

PR #2111, opened 2019-03-28
DHFPROD-2019: Update UI to work with new JSON payload changes

brucean52:


brucean52:
OK, that seems to fix it.

brucean52:
Should I also change config to options? I think it would make the terminology more consistent across the entire stack.

alexsmr:
@brucean52 yes, that will be great. It is more work though than ingestion.

brucean52:
I think I got all the changes done. Please check.

PR #2113, opened 2019-03-28
Refactor code

akshaysonvane:
Handle some more exceptions
Add detailed error messages


web/src/main/java/com/marklogic/hub/web/model/StepModel.java
```diff
@@ -23,6 +23,8 @@
 import com.marklogic.hub.util.json.JSONObject;
 import com.marklogic.hub.web.model.entity_services.JsonPojo;
 
+import java.util.Iterator;
+
 public class StepModel extends JsonPojo {
```
hao1st: do we have to extends JsonPojo?

akshaysonvane: Nope, we absolutely don't require that thanks to your library.

PR #2114, opened 2019-03-28
DHFPROD-1951: ingest step develop merge

alexsmr:


ayuwono:
@alexsmr , looks good. Some observations:
1) I thought we should not be able to set target entity on ingest step? I'm able to set target entity on the ingest step dialog
2) Input file path works fine when creating a new ingest step, I'm able to browse/navigate the folders. But it seems that it's not working on the Mock API flows/steps (I'm not able to navigate to subfolders). Is this expected to not working with Mock API? Try to change the filepath on the first flow "Order Flow 01"

alexsmr:
@ayuwono, I needed to have a collection name based on the Target Entity. So I kept it. We may need to remove it later. Depends on integration decisions. With mocking API it provides the path that cannot be matched with my local path. That is why the problem. That should be started with the current location ".", and then save the selection out of it. We use an old file browser, so it works the same way. Probably we should improve error handling for this component later.

ayuwono:
> @ayuwono, I needed to have a collection name based on the Target Entity. So I kept it. We may need to remove it later. Depends on integration decisions. With mocking API it provides the path that cannot be matched with my local path. That is why the problem. That should be started with the current location ".", and then save the selection out of it. We use an old file browser, so it works the same way. Probably we should improve error handling for this component later.

Ok sounds good, then this feature should be good in overall. We will log the issues later separately

ayuwono:
> I see that I can submit the comma-delimited list of output permissions with spaces around the commas. That might cause failure if that value is turned into a command line param. Or it may be OK if the back end sanitizes things. @ayuwono, any thoughts on this?

The old UI didn't do any validation on this field. I don't think the backend sanitizes this either, it will just throw an error when you run the flow. I think we can just throw any string on this field. So we can put regex restriction on the UI, sanitize the backend, or put the instruction on the hover over help. Sounds like an enhancement for the future sprint?

alexsmr:
@ayuwono @wooldridge If we will have time to add validation here we will do. I plan to remove a SAVE button to save as the change happens and will add some sanity check to it as a part of the feature.

PR #2115, opened 2019-03-28
DHFPROD-2062: Add Binary as a valid format type.

akshaysonvane:


bsrikan:
@akshaysonvane Can we add this fix to the mapping main.sjs as well

PR #2121, opened 2019-03-30
Feature/dhfprod 1885 1889

srinathgit:
This PR

1. Handles exceptions that happen during step run and writes the step output events to job doc
2. Uses new options object when starting each step and when adding uris to options in the step
3. Writes the status of steps that completed with errors as "completed with error step <stepnum>"
4. Return FlowStepModel to UI rather than Flow
5. Minor null checks

### DIFF
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/flow/impl/FlowRunnerImpl.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/flow/impl/FlowRunnerImpl.java
@@ -100,11 +100,11 @@ public class FlowRunnerImpl implements FlowRunner{
             stepNums = new ArrayList<String>(flow.getSteps().keySet());
         }
 
-        if(destDB != null){
+        if(StringUtils.isNotEmpty(destDB)){
             flow.setOverrideDestDB(destDB);
         }
 
-        if(sourceDB != null){
+        if(StringUtils.isNotEmpty(sourceDB)){
             flow.setOverrideSourceDB(sourceDB);
         }
 
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/flow/impl/FlowRunnerImpl.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/flow/impl/FlowRunnerImpl.java
@@ -13,6 +13,7 @@ import com.marklogic.hub.job.JobUpdate;
 import com.marklogic.hub.step.Step;
 import com.marklogic.hub.step.StepRunner;
 import com.marklogic.hub.step.StepRunnerFactory;
+import org.apache.commons.lang3.StringUtils;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 
@@ -20,7 +21,7 @@ import java.text.SimpleDateFormat;
 import java.util.*;
 import java.util.concurrent.*;
 import java.util.concurrent.atomic.AtomicBoolean;
-import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicLong;
 import java.util.stream.Collectors;
 
 @Component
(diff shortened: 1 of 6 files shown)


web/src/main/java/com/marklogic/hub/web/service/FlowManagerService.java
```diff
@@ -255,10 +255,10 @@ public Flow runFlow(String flowName, List<String> steps) {
             steps.forEach((step) -> restrictedSteps.add(this.getStepKeyInStepMap(flow, step)));
             resp = flowRunner.runFlow(flowName, restrictedSteps);
         }
-        return flowManager.getFlow(flowName);
+        return FlowStepModel.transformFromFlow(flowManager.getFlow(flowName));
```
hao1st: you can call getFlow(flowName) directly without using flowManager

web/src/main/java/com/marklogic/hub/web/service/FlowManagerService.java
```diff
@@ -267,7 +267,7 @@ public Flow stop(String flowName) {
         while(itr.hasNext()){
             flowRunner.stopJob(itr.next());
         }
-        return flowManager.getFlow(flowName);
+        return FlowStepModel.transformFromFlow(flowManager.getFlow(flowName));
```
hao1st: you can call getFlow(flowName) directly without using flowManager

marklogic-data-hub/src/main/java/com/marklogic/hub/flow/impl/FlowRunnerImpl.java
```diff
@@ -103,15 +104,15 @@ public RunFlowResponse runFlow(String flowName, List<String> stepNums, String jo
             flow.setOverrideDestDB(destDB);
         }
 
-        if(sourceDB == null){
```
hao1st: Please use if (StringUtils.isNotEmpty(sourceDB)) {..} the same for other checks.
https://github.com/marklogic/marklogic-data-hub/blob/9fdff17adb1ec88e052ce5de153691fdc70bf01c/marklogic-data-hub/src/main/java/com/marklogic/hub/flow/impl/FlowRunnerImpl.java#L119

        if (options != null && !options.isEmpty()) {
            flow.setOverrideOptions(options);
        }

marklogic-data-hub/src/main/java/com/marklogic/hub/step/impl/QueryStepRunner.java
```diff
@@ -197,10 +197,16 @@ public Job run() {
...
+            job.withSuccess(false);
             StringWriter errors = new StringWriter();
             e.printStackTrace(new PrintWriter(errors));
             job.withJobOutput(errors.toString());
-            jobUpdate.postJobs(jobId, JobStatus.FAILED.toString(), step);
+            try {
+                jobUpdate.postJobs(jobId, JobStatus.FAILED.toString(), step);
+            }
+            catch (Exception ex) {
+                throw (ex);
```
hao1st: throw ex

marklogic-data-hub/src/main/java/com/marklogic/hub/flow/impl/FlowRunnerImpl.java
srinathgit: 1. options is a map and can be empty, so no need for the check
2. will use StringUtils.isNotEmpty for sourceDB

hao1st: if the map is empty, do you still want to ovveride the Options in Flow?

srinathgit: It is not going to affect anything as they will be passed to stepRunner,  which can add its options( if present)  and will be sent to the server where they are combined.

PR #2123, opened 2019-03-31
DHFPROD-2070: Exception thrown when running query step (collector)

hao1st:


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/endpoints/collector.sjs
```diff
@@ -54,16 +54,16 @@ if(method === 'GET') {
     resp = fn.error(null, "RESTAPI-SRVEXERR", Sequence.from([404, "Not Found", "The requested flow was not found"]));
   } else {
     let stepDoc = flowDoc.steps[step];
-    let baseStep = datahub.flow.step.getStepByNameAndType(stepDoc.name, stepDoc.type);
+    let baseStep = datahub.flow.step.getStepByNameAndType("default-ingest", "INGEST");
```
srinathgit: Why is the flow name and type hardcoded ? Also for INGEST flows, collector is not run

srinathgit:
@hao1st 
There are 6 commits for making change in one file . Can this be done in a single commit ? Does this PR address any existing bug ?

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/endpoints/collector.sjs
hao1st: @srinathgit i tried to understand and fix the exception. I guess the baseStep is our default. The stepDoc is the customized step. if I pass flow name (flow1 for example) and type (INGEST), datahub.flow.step.getStepByNameAndType return null.  I am not quite sure how to fix it correctly. But I works after the change. I might miss something here.

hao1st: @srinathgit Query step runner works for you?

PR #2124, opened 2019-03-31
DHFPROD-1889: fix switch bug and set step options

hao1st:


marklogic-data-hub/src/main/java/com/marklogic/hub/step/StepRunnerFactory.java
```diff
@@ -33,7 +35,8 @@ public StepRunner getStepRunner(Flow flow, String stepNum) {
             .withBatchSize(step.getBatchSize())
             .withThreadCount(step.getThreadCount())
             .withSourceClient(hubConfig.newStagingClient(step.getSourceDatabase()))
-            .withDestinationDatabase(step.getDestinationDatabase());
+            .withDestinationDatabase(step.getDestinationDatabase())
+            .withOptions(step.getOptions());
```
srinathgit: We don't have to set  step options, they are already available in the server. The only 'options' that has to be set is the one overridden from the command line which is ready set in FlowRunnerImpl

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/endpoints/collector.sjs
```diff
@@ -54,16 +54,16 @@ if(method === 'GET') {
     resp = fn.error(null, "RESTAPI-SRVEXERR", Sequence.from([404, "Not Found", "The requested flow was not found"]));
   } else {
     let stepDoc = flowDoc.steps[step];
-    let baseStep = datahub.flow.step.getStepByNameAndType(stepDoc.name, stepDoc.type);
+    let baseStep = datahub.flow.step.getStepByNameAndType("default-ingest", "INGEST");
```
srinathgit: I don't think 'baseStep' is the "default-ingest" step, so we shouldn't hardcode it. In fact, for ingest step, we don't run collector. collector runs only for mapping step

marklogic-data-hub/src/main/java/com/marklogic/hub/step/StepRunnerFactory.java
hao1st: @srinathgit I am confused about why we do not need step options? The configuration is deployed to ML server. We do not need them in DMSDK? 
How about switch problem?

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/endpoints/collector.sjs
hao1st: @srinathgit 2019-03-30 13:59:45.343 Notice:+in /data-hub/5/endpoints/collector.sjs [javascript]
2019-03-30 13:59:45.808 Info: Status 500: JS-JAVASCRIPT: let combinedOptions = Object.assign({}, baseStep.options, stepDoc.options, flowDoc.options, options); -- Error running JavaScript request: TypeError: Cannot read property 'options' of undefined

https://project.marklogic.com/jira/browse/DHFPROD-2070

marklogic-data-hub/src/main/java/com/marklogic/hub/step/StepRunnerFactory.java
srinathgit: 1. We only have to set the 'options' overriden from command line  in FlowRunner which is already done in https://github.com/marklogic/marklogic-data-hub/blob/feature/DHFPROD-1885-1889/marklogic-data-hub/src/main/java/com/marklogic/hub/flow/impl/FlowRunnerImpl.java#L212 . The other options are available in server and it will build a combined option using all of them.
2. We require the 'break' statement, since we are only running mapping step with flowrunner so far, it goes to the default case which returns a 'QueryStepRunner'

hao1st: @srinathgit thanks for quick response. Because of no break,  it went to QueryStepRunner. So I got https://project.marklogic.com/jira/browse/DHFPROD-2070 problem (my flow step is "INGEST",  not "mapping" type...)

srinathgit: Currently, FlowRunner cannot run "INGEST" step , "WriteStepRunner" still requires work which will be done in this sprint.

PR #2090, opened 2019-03-23
DHFPROD-1266: run mastering step

ryanjdew:

PR #2126, opened 2019-04-01
DHFPROD-2041: Adjust flow scaffolding to put identifier in option

ryanjdew:
The other changes surrounding this ticket were already made in the Smart Mastering feature branch. https://github.com/marklogic/marklogic-data-hub/pull/2090/files#diff-5e944b09cf7c5a31ae3bc89610e7b97cR32


bsrikan:
We should probably address this in the default mapping step artifact as well.

PR #2127, opened 2019-04-01
DHFPROD-1885: add utility methods for converting map to json string

hao1st:
@srinathgit you can use the following code snippet for converting a map to a json string
        //option is a Map<String, Object> 
        JSONObject jsonObject = new JSONObject();
        String optionString = jsonObject.convertMapToJsonString(options);


aebadirad:
Can we get the java docs filled out?

hao1st:
@aebadirad sure

PR #2144, opened 2019-04-03
DHFPROD-1417: first cut of 4.x renaming roles/users and adding the th…

aebadirad:
…ird flow developer, running tests


marklogic-data-hub/src/main/resources/hub-internal-config/security/users/flow-developer-user.json
```diff
@@ -0,0 +1,6 @@
+{
+  "user-name": "%%mlFlowDeveloperUserName%%",
+  "description": "A user that is used for reading and writing from MarkLogic Data Hub databases and for maintaining DHF modules",
+  "password": "%%mlFlowDeveloperUserPassword%%",
+  "role": ["%%mlFlowDeveloperRole%%","%%mlFlowOperatorRole%%"]
```
bsrikan: another typo here. Should be mlFlowDeveloperPassword.

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/HubConfigImpl.java
```diff
@@ -1667,29 +1697,32 @@ public DatabaseClient newModulesDbClient() {
         customTokens.put("%%mlFinalSchemasDbName%%", finalSchemasDbName == null ? environment.getProperty("mlFinalSchemasDbName") : finalSchemasDbName);
         customTokens.put("%%mlFinalSchemasForestsPerHost%%", finalSchemasForestsPerHost == null ? environment.getProperty("mlFinalSchemasForestsPerHost") : finalSchemasForestsPerHost.toString());
 
-        customTokens.put("%%mlHubUserRole%%", hubRoleName == null ? environment.getProperty("mlHubUserRole") : hubRoleName);
-        customTokens.put("%%mlHubUserName%%", hubUserName == null ? environment.getProperty("mlHubUserName") : hubUserName);
+        customTokens.put("%%mlFlowOperatorRole%%", flowDeveloperRoleName == null ? environment.getProperty("mlFlowOperatorRole") : flowDeveloperRoleName);
```
bsrikan: Noting down the typos for reference so I remember to change them. Variable name should align to that of flowOperator

bsrikan:
Fixes issue DHFPROD-2089

bsrikan:
Expecting close to 41 tests to fail after PII tests. Requested @srinathgit to help take a look.
Except these failures + DHFPROD-2105, 1417 user story should be more or less complete.
Need to test upgrade once identified issues noted above are fixed.

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/DataHubImpl.java
```diff
@@ -859,7 +859,10 @@ public boolean upgradeHub(List<String> updatedFlows) throws CantUpgradeException
                 text = Pattern.compile("^(\\s*)id\\s+['\"]com.marklogic.ml-data-hub['\"]\\s+version.+$", Pattern.MULTILINE).matcher(text).replaceAll("$1id 'com.marklogic.ml-data-hub' version '" + version + "'");
                 text = Pattern.compile("^(\\s*)compile.+marklogic-data-hub.+$", Pattern.MULTILINE).matcher(text).replaceAll("$1compile 'com.marklogic:marklogic-data-hub:" + version + "'");
                 FileUtils.writeStringToFile(buildGradle, text);
-                hubConfig.getHubSecurityDir().resolve("roles").resolve("flow-operator.json").toFile().delete();
+                hubConfig.getHubSecurityDir().resolve("roles").resolve("data-hub-role.json").toFile().delete();
+                hubConfig.getHubSecurityDir().resolve("roles").resolve("hub-admin-role.json").toFile().delete();
+                hubConfig.getHubSecurityDir().resolve("users").resolve("hub-admin-user.json").toFile().delete();
+                hubConfig.getHubSecurityDir().resolve("users").resolve("data-hub-user.json").toFile().delete();
             }
```
bsrikan: @aebadirad I dont think we should delete them. We are altering the .old file that way and user will lose what they had pre upgrade. Collecting the obsolete files and not moving them over seems to be the right approach. What do you think? Co hubInit is going to create the new roles and users.

aebadirad: Fair enough. Then let's add them under hub-internal-config obsolete files (all 4)

bsrikan: Looks like we only had data-hub-user and data-hub-role in 3.0
hub-admin-role and hub-admin-user probably came in 4.0+. Will test by marking them obsolete for 4.x to 4.3 upgrade.

bsrikan:
@aebadirad and @srinathgit please review

srinathgit:
I see that most of the tests are running as data-hub-admin user , so they are passing. After installation , none of the tests should run as "data-hub-admin" user. Ideally, we will have to 
1. Install with mlUser/Password set to "data-hub-admin-user"
2. Create/Deploy modules flows with mlUser/Password "flow-developer" and enable debug/trace and run flows with  mlUser/Password "flow-operator"
3. Negative tests for the above cases

bsrikan:
@aebadirad can you pls help create the certs for flowDeveloper... We had flowOperator and hubAdmin certs.. but we need flowDeveloper now maybe?

aebadirad:
Certs are user based. There's no flow developer user.

bsrikan:
Tests should pass now.

PR #2151, opened 2019-04-03
DHFPROD-2034: Debug improvement #1: Add host information to Jobs document that helps with debugging

barkhaML:
Added fields in createBatch func.

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/jobs.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/jobs.sjs
@@ -142,6 +142,11 @@ class Jobs {
         batchStatus: "started",
         timeStarted:  fn.currentDateTime(),
         timeEnded: "N/A",
+        hostName: xdmp.hostName(),
+        reqTimeStamp: xdmp.requestTimestamp(),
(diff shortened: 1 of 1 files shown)


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/jobs.sjs
```diff
@@ -142,6 +142,11 @@ class Jobs {
         batchStatus: "started",
         timeStarted:  fn.currentDateTime(),
         timeEnded: "N/A",
+        hostName: xdmp.hostName(),
+        reqTimeStamp: xdmp:request-timestamp(),
```
ryanjdew: This needs to be the JavaScript signature. "xdmp.requestTimestamp()".

srinathgit:
@barkhaML ,
There are 7 commits for changes in one file. Can this be done in a single commit as all the changes pertain to the same bug ?

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/jobs.sjs
barkhaML: updated syntax for dmp.requestTimestamp()

PR #2153, opened 2019-04-04
DHFPROD-2067: Jobs view mock data

brucean52:
Changed the name for a few parameters
added flowId
jobStatus -> status
committed -> successfulEvents
errors -> failedEvents
timeStarted -> startTime
timeEnded -> endTime
stepStatus -> status


web/api/mocks/service/JobsService.js
```diff
@@ -7,12 +7,47 @@ let Error = require('./ErrorService');
...
  *
  * jobId String Id of Job to be fetched
- * returns flow
+ * returns job
  **/
 exports.getJob = function(jobId) {
   let resp;
   if (jobId) {
-    resp = Storage.get('jobs', jobId);
+    resp = Storage.get('job-detail', jobId);
```
clockworked247: This should remain "jobs" and the use of job-detail should be removed across all code.

web/api/mocks/service/JobsService.js
```diff
@@ -7,12 +7,47 @@ let Error = require('./ErrorService');
...
+ * Delete a Job
+ * ....
+ *
+ * jobId String Id of Job to be fetched
+ * returns job that is deleted
+ **/
+exports.deleteJob = function(jobId) {
+  let resp;
+  if (jobId) {
+    Storage.delete('job-all', jobId);
```
clockworked247: Just use "jobs" the details of a job should be within each individual object.

web/api/mocks/service/StorageService.js
```diff
@@ -13,7 +13,9 @@ let collections = {
   'flows': require('../data/flows.json'),
   'steps': require('../data/steps.json'),
   'collections': require('../data/collections.json'),
-  'jobs': []  // generated on the fly when /flow/{flow id}/run triggered
+  'jobs': [],  // generated on the fly when /flow/{flow id}/run triggered
+  'job-detail': require('../data/job.json'),
```
clockworked247: Seeding "jobs" with the Array in jobs.json is what should happen here and all additions/deletions/gets/updates should happen against the "job" collection.

web/api/swagger/mocks.yaml
```diff
@@ -723,24 +809,87 @@ definitions:
         type: "string"
         format: "date-time"
         description: "Timestamp job started"
-      output:
+      steps:
         type: "array"
-        description: "Array with Job output info (not required for UI atm - skip implementation)"
+        description: "Array of Step output info"
         items:
-          type: "string"
```
clockworked247: Good catch here, thanks.

web/api/swagger/mocks.yaml
```diff
@@ -723,24 +809,87 @@ definitions:
...
-      stepId:
```
clockworked247: From: https://wiki.marklogic.com/display/ENGINEERING/DHF+Web+Service+REST+API+Data+Model?showComments=true#DHFWebServiceRESTAPIDataModel-Ex.Step:Mastering
Job should contain:

    "stepId": "my-ingest-step-1-ingest",
    "stepName": "My Ingest Step 1",
    "stepRunningPercent": null,

The addition of "lastAttemptedStep" and "lastCompleted" step don't match this structure.  By default, the "stepName" & "stepId" reflect the step being currently executed or, in the case of a failed job, the last step that executed.  So all we need are "stepName" & "stepId".

web/api/swagger/mocks.yaml
```diff
@@ -723,24 +809,87 @@ definitions:
...
+        description: "Name of the flow being processed"
+      lastAttemptedStep:
+        type: "integer"
+        format: "int64"
+        description: "last step that was attempted to run"
+      lastCompletedStep:
+        type: "integer"
+        format: "int64"
+        description: "last step that was completed"
+      jobStatus:
```
clockworked247: should be named "status", we're already inside a job, so no need to prefix it.

web/api/swagger/mocks.yaml
```diff
@@ -723,24 +809,87 @@ definitions:
...
+      successfulEvents: 0
+      flow: "Flow 01"
+      steps: []
+      lastAttemptedStep: 0
+      lastCompletedStep: 0
+      startTime: "2000-01-23T04:56:07.000+00:00"
+      id: "id"
+      endTime: "2000-01-23T04:56:07.000+00:00"
+      jobStatus: "canceled"
+  jobs:
```
clockworked247: You can remove this definition.  "job" is the correct definition and is all that is required.

web/api/mocks/service/StorageService.js
```diff
@@ -76,9 +76,19 @@ exports.getCollection = function(cName) {
 exports.save = function(cName, id, obj) {
   return new Promise(function(resolve, reject) {
     if (collections[cName]) {
-      let index = _.findIndex(collections[cName], ['id', id]);
-      let item = _.find(collections[cName], ['id', id]);
-      let data = Object.assign({}, item || {}, obj);
+      let index = null;
```
clockworked247: I see why this change went in, but as part of our clean-up next sprint, we should change this to be "id", not "jobId".

PR #2160, opened 2019-04-04
DHFPROD-2069: create api/jobs with static data

hao1st:


alexsmr:
Reopen in different PR

PR #2161, opened 2019-04-04
DHFPROD-2042: reduce cts queries, attempt to identify envelope if pre…

aebadirad:
…sent


bsrikan:
Rahul tested and looks like json document is not getting harmonized correctly.

PR #2166, opened 2019-04-05
DHFPROD-2009: WriteStepRunner

srinathgit:
1. Implement 'noBatchWrite' option to stop writing batch docs
2. Adding a new server transform for running dmsdk ingest
3. Fix bug in 'case' statement in StepRunnerFactory, handling cases when first step run fails without job doc created
4. Update java client version
5. Making WriteStepRunner handle options and use them for running

### DIFF
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/flow/impl/FlowRunnerImpl.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/flow/impl/FlowRunnerImpl.java
@@ -1,6 +1,5 @@
 package com.marklogic.hub.flow.impl;
 
-import com.marklogic.hub.DatabaseKind;
 import com.marklogic.hub.FlowManager;
 import com.marklogic.hub.flow.Flow;
 import com.marklogic.hub.flow.FlowRunner;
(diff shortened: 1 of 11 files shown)


srinathgit:
Initial check-in so that @hao1st  can start working on his jobs endpoint

marklogic-data-hub/src/main/java/com/marklogic/hub/flow/impl/FlowRunnerImpl.java
```diff
@@ -281,14 +286,15 @@ public void run() {
...
-                threadPool.shutdownNow();
+            finally {
+                resp.setStepResponses(stepOutputs);
+                resp.setEndTime(DATE_TIME_FORMAT.format(new Date()));
+                jobQueue.remove();
+                if (!jobQueue.isEmpty()) {
+                    initializeFlow((String) jobQueue.peek());
+                } else {
+                    threadPool.shutdownNow();
+                }
```
hao1st: We should add  isRunning.set(false);

hao1st: we will fix later.

PR #2175, opened 2019-04-09
DHFPROD-2106: Enable search via raw cts json query

akshaysonvane:


web/src/main/java/com/marklogic/hub/web/service/SearchService.java
```diff
@@ -70,6 +71,15 @@ public StringHandle search(SearchQuery searchQuery) {
 
         ArrayList<StructuredQueryDefinition> queries = new ArrayList<>();
 
+        if (searchQuery.query != null && !searchQuery.query.isNull()) {
```
hao1st: if the condition is true, we will return. Then How to get the line?
https://github.com/marklogic/marklogic-data-hub/blob/8b3813ffeaad6359f866feee4f600a2582b96067/web/src/main/java/com/marklogic/hub/web/service/SearchService.java#L114

akshaysonvane:
@wooldridge as per your earlier suggestion, I have created a new endpoint and model.
Benefits:
We don't need to retrofit the current query parameter with new capabilities.
We can build on the new endpoint if we require more capabilities from it.

aebadirad:
@hao1st Whoops, sorry, accidentally requested a review from you again instead of @wooldridge 

Please reapprove :|

akshaysonvane:
I've changed the model to:
`{
    "count": 1,
    "database": "STAGING",
    "ctsQuery": {
        "query": {
            "wordQuery": {
                "text": [
                    "sunglasses"
                ]
            }
        }
    },
    "start": 1
}`
I've renamed the "query" param in the model to "ctsQuery" so as to avoid confusion with actual json query.

wooldridge:
@akshaysonvane, will the new endpoint worked with free text and structured queries as well as cts queries? If so, we should make the endpoint more generic than "ctsSearch".

wooldridge:
@akshaysonvane, thanks for the explanation, looks good when I run the following:
```
POST http://localhost:4200/api/search/ctsSearch
{ "database": "STAGING", "ctsQuery": { "query": { "wordQuery": { "text": [ "sunglasses" ] } } }, "start": 1, "count": 1 }
```

PR #2193, opened 2019-04-12
DHFPROD-2158: Validate mapping (and other step) data

wooldridge:
Adds validation when creating or editing a step. This was required as part of the mapping step update because missing data (target entity, source information) leads to a broken config view for mapping:

- Updates to new-step-dialog to ensure complete data is submitted.
- Custom validator checks for target entity and source data.
- Save button in new-step-dialog is disabled until data is valid
- Mark required fields in new-step-dialog with asterisk (i.e. the radio buttons label).
- For a new step, present a Step Type menu in the dialog and display the rest of the inputs after the type has been selected.

Follows guidelines described in the spreadsheet:

https://project.marklogic.com/jira/browse/DHFPROD-2148

To test, create or edit a step and test that empty required fields cannot be submitted.

### DIFF
--- a/web/src/main/ui/app/components/flows-new/edit-flow/mastering/matching/match-options.model.ts
+++ b/web/src/main/ui/app/components/flows-new/edit-flow/mastering/matching/match-options.model.ts
@@ -101,6 +101,26 @@ export class MatchOption {
         return prop.name;
       })
     }
+    // Set algorithmRef based on matchType
+    if (!mOpt.algorithmRef) {
+      switch(mOpt.matchType) {
+        case "synonym":
+          this.algorithmRef = "thesaurus";
+          break;
+        case "double metaphone":
+          this.algorithmRef = "double-metaphone";
+          break;
+        case "zip":
+          this.algorithmRef = "zip-match";
+          break;
+        case "reduce":
+          this.algorithmRef = "standard-reduction";
+          break;
+        case "custom":
+          this.algorithmRef = mOpt.customFunction;
+          break;
+      }
+    }
   }
 
   /**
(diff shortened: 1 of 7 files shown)


ayuwono:
@wooldridge , I found more issue, it's rather higher priority
When you add a Match Option with type other than Exact, it's saved as Exact type after you revisit the step
1) On Mastering Step, add a Match Option with type other than Exact (Synonym, Double Metaphone, Zip, etc). After setting it, it seems that it's saved properly
2) Go back to Manage Flows page
3) Go to the Mastering Step where you add Match Option with type other than Exact on step 1

Every match option is saved as Exact type

wooldridge:
@ayuwono, just tried to reproduce your steps above (create 4-step flow, add reduce option, etc.) and I was successful, didn't run into your error. I'm running with runUImock, BTW. I will look at the specific Match Option issue now.

wooldridge:
@ayuwono, OK, I'm seeing the error you describe (can't select step) after I refresh the page. Looking into it.

ayuwono:
> @ayuwono, just tried to reproduce your steps above (create 4-step flow, add reduce option, etc.) and I was successful, didn't run into your error. I'm running with runUImock, BTW. I will look at the specific Match Option issue now.

@wooldridge , after creating the reduce option, did you go back to manage flows page and then go to the flow?

ayuwono:
> > @ayuwono, just tried to reproduce your steps above (create 4-step flow, add reduce option, etc.) and I was successful, didn't run into your error. I'm running with runUImock, BTW. I will look at the specific Match Option issue now.
> 
> @wooldridge , after creating the reduce option, did you go back to manage flows page and then go to the flow?

alright, saw your comment above

wooldridge:
@ayuwono, that problem with the match options should be fixed now.

ayuwono:
looks good @wooldridge

PR #2200, opened 2019-04-12
DHFPROD-1816: Mapping configuration

wooldridge:
Finish mapping config UI work for 5.0, includes:

- Auto-save of mapping changes
- Update mapping config view when mapping settings change
- Support defining source for mapping via cts query
- Validate mapping step data upon creation (and any type pf step data for that matter)
- algorithmRef bug fix
- Fix radio button color

### DIFF
--- a/web/src/main/ui/app/components/flows-new/edit-flow/mapping/ui/mapping-ui.component.html
+++ b/web/src/main/ui/app/components/flows-new/edit-flow/mapping/ui/mapping-ui.component.html
@@ -8,12 +8,12 @@
 
     <div id="map-heading">
 
-      <!-- Save/Reset happens via header buttons
+      <!-- Save/Reset happens via header buttons -->
       <div id="save-map">
         <button mat-raised-button color="primary" [disabled]="!mapChanged()" (click)="mapChanged() && onSaveMap()">Save</button>
         <!-- <button mat-raised-button color="primary" [disabled]="!mapChanged()" (click)="mapChanged() && onResetMap()">Reset</button> -->
       </div>
-      -->
+
     </div>
 
     <div id="source">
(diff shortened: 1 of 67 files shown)


web/src/main/ui/app/components/flows-new/edit-flow/mapping/ui/mapping-ui.component.html
```diff
@@ -34,7 +28,7 @@
       </div>
 
       <!-- Dropdown row for each entity property -->
-      <div *ngFor="let entityProp of chosenEntity?.definition.properties; let i = index;" class="src-prop-container">
```
clockworked247: Let's leave these selectors the same.   Wiki and QA already using the old one.  Less churn.

wooldridge: Can you tell me exactly what to change back? I'm not sure what I did. :-)

ayuwono: @wooldridge , it's the class name, the old one is src-prop-container. @clockworked247 , I think it's okay to have the new name source-prop-container so it's consistent and more readable? Wiki and QA can change the naming

wooldridge: @clockworked247 sounds like @ayuwono doesn't need this changed?

clockworked247: Ok, just update then wiki for me then.

wooldridge:
@alexsmr, THANK YOU.

PR #2204, opened 2019-04-13
DHFPROD-2144: Remove tooltips and tooltip icons

wooldridge:
Most of the current tooltips apply to MLCP command-line settings or are empty so we’re going to remove them for the initial release.


wooldridge:
@alexsmr, thanks. The tooltip icon remains whether you include the tooltipText (and tooltipPlacement) attribute or not. I left the component (and the config information for the Ingest view). I will put back some commented out HTML so the usage is clear.

wooldridge:
@alexsmr, OK, did that.

PR #2205, opened 2019-04-13
DHFPROD-2140: Fix Mastering tab selection

clockworked247:
Fixes issue with tab selection when the container is not rendered yet due to Angular Material bug https://github.com/angular/material2/issues/7006


wooldridge:
@clockworked247 nice fix.

PR #2224, opened 2019-04-16
DHFPROD-2092: Able to create multiple default merge strategies on mastering step

barkhaML:
Added a check for default merge strategy, a popup to replace the existing one and not allowing to create duplicate default.


wooldridge:
Nice work, @barkhaML. @ayuwono, would you mind reviewing?

wooldridge:
Merging work with DHFPROD-2096 here: https://github.com/marklogic/marklogic-data-hub/pull/2242

PR #2225, opened 2019-04-16
DHFPROD-1980: Fix web service entities endpoint

ryanjdew:


aebadirad:
Before we bring this in @ryanjdew we need to stop the 4.x plugins/ entities from deploying to the server, only from /entities when they get copied over.

Otherwise we're running into conflicts.

wooldridge:
@aebadirad @ryanjdew can we make those updates and push ASAP since QuickStart views are very broken with the entities issue?

aebadirad:
Sure, we can push this in and open a new PR.

PR #2231, opened 2019-04-16
DHFPROD-1980: Fix issue of duplicate entities

ryanjdew:


marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/LoadUserArtifactsCommand.java
```diff
@@ -146,28 +142,6 @@ public String toURI(Resource r) throws IOException {
         FlowDefModulesFinder flowDefModulesFinder = new FlowDefModulesFinder();
         try {
             //first let's do the entities paths
```
srinathgit: @ryanjdew ,
Should deploying 4.x entities and mappings be removed  , I think this will cause all 4.x flows would fail

ryanjdew: Yes. hubUpdate will move them to the proper place.

PR #2234, opened 2019-04-16
DHFPROD-1417: Updating flow tests to have correct permissions

bsrikan:
Fixing the FlowManager and FlowRunner test as they were failing in DHS with flow-developer-role does not exist. On ingestion the permissions being set was that of flow-developer-role instead of flow-operator-role. Fixed that.


aebadirad:
Ah, great catch.

bsrikan:
@SameeraPriyathamTadikonda I forgot to append bug number and so the pipeline seems to have failed while trying to add comment in JIRA. Hope we can go ahead and rebase+merge manually?

PR #2235, opened 2019-04-16
DHFPROD-2081: step model creation and new step dialog updates

brucean52:
Updated syntax and new step dialog on the UI. I will do further UI updates with @clockworked247's updated mock api


web/src/main/ui/app/components/flows-new/models/step.model.ts
```diff
@@ -1,16 +1,51 @@
-import { Options } from './step-options.model';
+import { IngestionOptions } from './ingestions-options.model';
+import { MappingOptions } from './mapping-options.model';
+import { MasteringOptions } from './mastering-options.model';
+import { CustomOptions } from './custom-options.model';
+
 
 export class Step {
```
alexsmr: Please create a private empty constructor so no one could create Step without factory methods. This may cause a number of errors across the app (which is a good thing). Please test the build from scratch.

web/src/main/ui/app/components/flows-new/edit-flow/ui/new-step-dialog-ui.component.ts
```diff
@@ -22,12 +19,13 @@ export class NewStepDialogUiComponent implements OnInit {
...
   @Input() flow: Flow;
+  @Input() projectDirectory: string;
   @Output() getCollections = new EventEmitter();
   @Output() cancelClicked = new EventEmitter();
   @Output() saveClicked = new EventEmitter();
 
-  public newStep: Step = new Step;
-  readonly stepOptions = ['ingest', 'mapping', 'mastering', 'custom'];
+  public newStep: Step;
+  readonly stepOptions = ['INGESTION', 'MAPPING', 'MASTERING', 'CUSTOM'];
```
alexsmr: Because there is a number of other files that use step types as a string, it becomes dangerous to do refactorings like these all the time. I suggest also along the Step model type add an enum like this:
``` typescript
export enum StepType {
  INGESTION = "INGESTION",
  MAPPING = "MAPPING",
  MASTERING = "MASTERING",
  CUSTOM = "CUSTOM"
};

export class Step {
  public id: string;
  public type: StepType;
}
```
and update all string references used in different expressions in templates as well as in TS files.

aebadirad: Agreed. Backend does this for the exact reason, we also will be adding new types for sure. So this is an excellent suggestion by @alexsmr

web/src/main/ui/app/components/flows-new/edit-flow/ui/new-step-dialog-ui.component.html
```diff
@@ -3,31 +3,31 @@ <h1 mat-dialog-title>{{title}}</h1>
...
     <label>Source Type: </label>
     <mat-radio-group id="step-source-type" [(ngModel)]="selectedSource" [ngModelOptions]="{standalone: true}">
       <mat-radio-button id="step-source-type-collection-radio" value="collection">Collection</mat-radio-button>
       <mat-radio-button id="step-source-type-query-radio" value="query">Query</mat-radio-button>
     </mat-radio-group>
   </div>
-  <mat-form-field *ngIf="selectedSource === 'collection' && newStepForm.value.type != 'ingest'">
-    <mat-select id="step-source-collection" class="type-select" placeholder="Source Collection">
+  <mat-form-field *ngIf="selectedSource === 'collection' && newStepForm.value.stepDefinitionType !== 'INGESTION'">
+    <mat-select id="step-source-collection" class="type-select" placeholder="Source Collection" formControlName="sourceQuery">
```
clockworked247: This will still be stored as sourceCollection in the UI model, but we will need to add a property sourceQuery to that model that is a string based on the collection selection made here.  See more here: https://project.marklogic.com/jira/browse/DHFPROD-2177

web/src/main/ui/app/components/flows-new/edit-flow/ui/new-step-dialog-ui.component.html
```diff
@@ -45,7 +45,7 @@ <h1 mat-dialog-title>{{title}}</h1>
           </mat-panel-title>
         </mat-expansion-panel-header>
 
-        <mat-form-field *ngIf="newStepForm.value.type != 'ingestion'">
+        <mat-form-field *ngIf="newStepForm.value.stepDefinitionType !== 'INGESTION'">
```
alexsmr: The whole idea of using StepType enum is to remove 'INGESTION' here (and everywhere in the code) and use StepType.INGESTION instead.

brucean52: I missed this one. Thanks!

alexsmr: Oh sorry, somehow I was looking into your old commit. I see you actually substituted the string usages.)))

web/src/main/ui/app/components/flows-new/edit-flow/ui/new-step-dialog-ui.component.ts
```diff
@@ -22,12 +19,14 @@ export class NewStepDialogUiComponent implements OnInit {
...
   @Input() flow: Flow;
+  @Input() projectDirectory: string;
   @Output() getCollections = new EventEmitter();
   @Output() cancelClicked = new EventEmitter();
   @Output() saveClicked = new EventEmitter();
 
-  public newStep: Step = new Step;
-  readonly stepOptions = ['ingest', 'mapping', 'mastering', 'custom'];
+  public newStep: Step;
+  public stepType: typeof StepType = StepType;
```
alexsmr: You don't need to instantiate the enum to use it. StepType.<whatever> should work.

brucean52: If I don't, it doesn't work on the HTML templating side

alexsmr: Hm. Yeah. My bad. The template is translated and run within the execution of its component. Technically it does not see a global context. Thanks.

PR #2246, opened 2019-04-17
DHFPROD-2096: Fix unit tests

ryanjdew:


aebadirad:
Bringing this in prematurely, because it resolves a major amount of test failures

marklogic-data-hub/src/test/java/com/marklogic/hub_integration/EndToEndFlowTests.java
```diff
@@ -719,10 +719,11 @@ private void createFlows(String prefix, CreateFlowListener listener) {
 
     private void createFlow(String prefix, CodeFormat codeFormat, DataFormat dataFormat, FlowType flowType, boolean useEs, CreateFlowListener listener) {
```
hao1st: I guess we can change the method name from createFlow -> createLegacyFlow, and we do not need to change local variable name with prefix "legacy" or mix name (somewhat confused?) ?

PR #2260, opened 2019-04-17
rename step type to stepDefinitionType

hao1st:


aebadirad:
@hao1st @akshaysonvane are we taking into account stepDefinitionName here too?

akshaysonvane:
> @hao1st @akshaysonvane are we taking into account stepDefinitionName here too?

We've only substituted 'type' with 'stepDefinitionType'. 
'name' is still the one that the user provides from the UI for that step.

PR #2270, opened 2019-04-18
DHFPROD-2183: Porting changes for new roles/users - DO NOT APPROVE

bsrikan:
Ran the tests locally and see same set of failures for core and web. Gradle runs clean.


marklogic-data-hub/src/main/resources/hub-internal-config/security/roles/data-hub-admin-role.json
```diff
@@ -0,0 +1,109 @@
+{
+  "role-name": "%%mlFlowDeveloperRole%%",
```
hao1st: "role-name": "%%mlDataHubAdminRole%%" ?

bsrikan: Thanks Hao! Fixed in latest commit.

bsrikan:
@ryanjdew I have created a PR against #2096 as well. Lets see if we can fix tests with these changes there itself.

PR #2272, opened 2019-04-18
DHFPROD-2198: Handle no-jobs case for UI display

wooldridge:
Edit Flow view when no jobs:
- Display gray “View Jobs” text (no link)
- Don’t display timestamp or timestamp prefix
- Display “Never run” as status

Manage Flows view:
- Display “Never run” as status

Generally:
- Display status with spaces (convert underscores) and capitalization of first word, e.g.: “Finished with errors”
- Display friendly timestamps with spaces and capitalization of first word, e.g.: “An hour ago”
- Handle linking correctly by checking for null properties in latestJob object

### DIFF
--- a/web/src/main/ui/app/components/flows-new/edit-flow/ui/stepper.component.html
+++ b/web/src/main/ui/app/components/flows-new/edit-flow/ui/stepper.component.html
@@ -21,14 +21,17 @@
           color="primary"
           (click)="this.stopClicked()">STOP</button>
         <div id="latest-job-info">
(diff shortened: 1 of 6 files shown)


wooldridge:
Updated Edit Flow page when no jobs have run:

Screen Shot 2019-04-18 at 12 10 37 AM

wooldridge:
@ayuwono, thanks! This should fix some of the other linking errors that were popping up in the new merged UI, similar to what @clockworked247 fixed yesterday.

PR #2274, opened 2019-04-18
Fix empty collection issue

akshaysonvane:
Add keys to fileLocations object

### DIFF
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/step/impl/IngestionStepDefinitionImpl.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/step/impl/IngestionStepDefinitionImpl.java
@@ -21,6 +21,8 @@ import com.marklogic.hub.step.AbstractStepDefinition;
 import com.marklogic.hub.step.StepDefinition;
 import com.marklogic.hub.util.json.JSONObject;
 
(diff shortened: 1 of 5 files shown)


marklogic-data-hub/src/main/java/com/marklogic/hub/step/impl/IngestionStepDefinitionImpl.java
```diff
@@ -30,11 +32,21 @@
     public IngestionStepDefinitionImpl(String name) {
         setName(name);
         setType(StepDefinitionType.INGESTION);
+
+        JSONObject jsonObject = new JSONObject();
+        jsonObject.put("inputFilePath", "");
+        jsonObject.put("outputURIReplacement", "");
+        jsonObject.put("inputFileType", "");
+        fileLocations = jsonObject.jsonNode();
         setFileLocations(fileLocations);
```
hao1st: https://github.com/marklogic/marklogic-data-hub/blob/805acf5f716ec7e46079abfadb150d2bd5348d69/marklogic-data-hub/src/main/java/com/marklogic/hub/step/impl/IngestionStepDefinitionImpl.java#L40-L41

this.fileLocations = jsonObject.jsonNode();

PR #2286, opened 2019-04-19
DHFPROD-2222: Job endpoint returns null for id & Modify project watchers

hao1st:


web/src/main/java/com/marklogic/hub/web/web/CurrentProjectController.java
```diff
@@ -300,8 +300,5 @@ private void clearAuthenticationAttributes(HttpServletRequest request) {
     @Override
     public void onLogoutSuccess(HttpServletRequest request, HttpServletResponse response, Authentication authentication) throws IOException, ServletException {
         request.getSession().invalidate();
-        String pluginDir = hubConfig.getHubPluginsDir().toString();
-        watcherService.removeListener(this);
-        watcherService.unwatch(pluginDir);
```
srinathgit: Shouldn't we unwatch flows, stedef, entities dir snd also removelistener ?

hao1st: @srinathgit add them.

web/src/main/java/com/marklogic/hub/web/web/CurrentProjectController.java
```diff
@@ -230,15 +230,15 @@ public String preInstallCheck() {
...
         String flowsDir = hubConfig.getFlowsDir().toString();
         String stepDefinitionsDir = hubConfig.getStepDefinitionsDir().toString();
         String entitiesDir = hubConfig.getHubEntitiesDir().toString();
+        String mappingsDir = hubConfig.getHubMappingsDir().toString();
         if (!watcherService.hasListener(this)) {
-            watcherService.watch(pluginDir);
             watcherService.watch(flowsDir);
             watcherService.watch(stepDefinitionsDir);
             watcherService.watch(entitiesDir);
+            watcherService.watch(mappingsDir);
```
akshaysonvane: Now that we are watching the entire mapping directory, should we remove the watch service from the mappingManagerService for the individual mapping folder name?

hao1st: We can leave them there now. Let us test after beta release.

akshaysonvane:
@aebadirad I was under the impression that we need watcher service on the plugins dir as the 4.x flows still reside there. Has that been changed?

aebadirad:
We just deploy & run them, but we offer no way to interact with the 4.x /plugins currently. So the watcher isn't needed once an upgrade has been completed. I'd rather not keep an active watch that's 'hidden' from the user, giving them unexpected behavior.

web/src/main/java/com/marklogic/hub/web/web/CurrentProjectController.java
aebadirad: Yes we should also unwatch them to clear it.

PR #2289, opened 2019-04-19
DHFPROD-2221:fix for few more bugs, refactoring, adding flowrunner test

srinathgit:
In this PR
1. Renamed Job -> RunStepResponse, RunStepResponse (it was a POJO holding response from ml:runFlow)  -> ResponseHolder
2. In case uncaught exceptions are thrown, continue with next step execution (existing code stopped current flow execution and proceeded to next flow execution)
3. Added tests for FlowRunner
4. Remove sourceDB and targetDB options from command line as they are part of options
5. Fix options and stepconfig (apply them during step run as they have the highest precedence)
6. FIx DHFPROD-2214 (incorrect doc count, In the dmsdk callback , one additional 'null' item is always returned, unsure why)
7. Fallback to default dbs in case staging or final is not set in options in impl/flow.sjs
8. Add stepDefinitionName,type and step name to response and remove getRunningStepKey() method from StepRunner
9. Fix for DHFPROD-2221
10. Making options precedence consistent with server for WriteStepRunner

PR #2292, opened 2019-04-19
DHFPROD-2223: moving to absolute path in ingest step

alexsmr:


ayuwono:
It works on both 4200 and 8080

timur-isangulov:
It doesn't work for me on Windows using the latest fix.

aebadirad:
@ayuwono @alexsmr  verify this on windows please

timur-isangulov:
On the latest develop ingest shows absolute path on Windows. However, when I run it's still returns error "java.lang.RuntimeException: java.lang.RuntimeException: java.util.IllegalFormatConversionException: c != java.util.UUID"

aebadirad:
I think we need to pull in #2289 to correct this, as that's in the java layer. Add the comment there :)

PR #2295, opened 2019-04-19
DHFPROD-2216: Connect Jobs Endpoints to UI

brucean52:
I don't think pagination and filtering are setup yet for job details view.


clockworked247:
"I don't think pagination and filtering are setup yet for job details view."
Please file a Jira issue with Hao to implement that.

clockworked247:
@ayuwono thanks for helping keep things moving by filing separate bugs.

PR #2298, opened 2019-04-19
DHFPROD-2194: fix new step dialog validation

clockworked247:
fix new step dialog validation + small clean-up for incorrect properties from before 2096 prop name changes.

### DIFF
--- a/web/src/main/ui/app/components/flows-new/edit-flow/ui/step.component.ts
+++ b/web/src/main/ui/app/components/flows-new/edit-flow/ui/step.component.ts
@@ -35,7 +35,7 @@ export class StepComponent implements OnChanges {
   ngOnChanges(changes: any) {
     if (changes && 
       changes.selectedStepId && 
-      this.step.type === 'mastering' && 
+      this.step.stepDefinitionType === this.stepType.MASTERING && 
       this.step.id === changes.selectedStepId.currentValue) {
       setTimeout(() => {
         this.masteringTabGroup.realignInkBar();  
--- a/web/src/main/ui/app/components/flows-new/validators/new-step-dialog.validator.ts
+++ b/web/src/main/ui/app/components/flows-new/validators/new-step-dialog.validator.ts
@@ -1,10 +1,12 @@
 import { FormGroup } from '@angular/forms';
+import { StepType } from '../models/step.model';
 import * as _ from "lodash";
 
 export function NewStepDialogValidator(group: FormGroup) {
+  let stepType: typeof StepType = StepType;
   let errors = {};
-  if (group.value.type === 'mapping' ||
-      group.value.type === 'mastering') {
+  if (group.value.stepDefinitionType === stepType.MAPPING ||
+      group.value.stepDefinitionType === stepType.MASTERING) {
     if (!group.value.targetEntity) {
       errors['noTargetEntity'] = true;
     }
--- a/web/src/main/ui/app/components/flows-new/edit-flow/ui/new-step-dialog-ui.component.html
+++ b/web/src/main/ui/app/components/flows-new/edit-flow/ui/new-step-dialog-ui.component.html
@@ -17,23 +17,23 @@
     <mat-hint>Optional</mat-hint>
   </mat-form-field>
   <div *ngIf="newStepForm.value.stepDefinitionType !== stepType.INGESTION">
-    <label>Source Type: </label>
+    <label>Source Type: <span *ngIf="newStepForm.value.stepDefinitionType !== stepType.INGESTION" class="source-req">*</span></label>
     <mat-radio-group id="step-source-type" [(ngModel)]="selectedSource" [ngModelOptions]="{standalone: true}">
       <mat-radio-button id="step-source-type-collection-radio" value="collection">Collection</mat-radio-button>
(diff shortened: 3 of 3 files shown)


web/src/main/ui/app/components/flows-new/edit-flow/ui/step.component.ts
```diff
@@ -35,7 +35,7 @@ export class StepComponent implements OnChanges {
   ngOnChanges(changes: any) {
     if (changes && 
       changes.selectedStepId && 
-      this.step.type === 'mastering' && 
+      this.step.stepDefinitionType === 'MASTERING' &&
```
alexsmr: Please use StepType instead of a string

web/src/main/ui/app/components/flows-new/validators/new-step-dialog.validator.ts
```diff
@@ -3,8 +3,8 @@ import * as _ from "lodash";
 
 export function NewStepDialogValidator(group: FormGroup) {
   let errors = {};
-  if (group.value.type === 'mapping' ||
-      group.value.type === 'mastering') {
+  if (group.value.stepDefinitionType === 'MAPPING' ||
```
alexsmr: Please use StepType enum instead of a string

web/src/main/ui/app/components/flows-new/edit-flow/ui/new-step-dialog-ui.component.ts
```diff
@@ -49,7 +49,7 @@ export class NewStepDialogUiComponent implements OnInit {
       description: [this.step ? this.step.description : ''],
       sourceQuery: [this.step ? this.step.options.sourceQuery : ''],
       sourceCollection: [this.step ? this.step.options.sourceCollection : ''],
-      targetEntity: [this.step ? this.step.options.targetEntity : ''],
+      targetEntity: [this.step ? this.step.options.targetEntity : '', Validators.required],
```
alexsmr: This field is not shown on the Ingestion step. You need to add required Validator only if it is not ingestion...

clockworked247: Thanks Sasha.  Caught the problem.

web/src/main/ui/app/components/flows-new/edit-flow/ui/step.component.ts
clockworked247: Good feedback, thanks.

web/src/main/ui/app/components/flows-new/validators/new-step-dialog.validator.ts
clockworked247: Good feedback, thanks.

web/src/main/ui/app/components/flows-new/edit-flow/ui/new-step-dialog-ui.component.ts
clockworked247: This case is already handled by NewStepDialogValidator

`
  if (group.value.stepDefinitionType === this.stepType.MAPPING ||
      group.value.stepDefinitionType === this.stepType.MASTERING) {
    if (!group.value.targetEntity) {
      errors['noTargetEntity'] = true;
    }
`

clockworked247: So I'm removing this standard validator

clockworked247:
Thanks @ayuwono 
See it working for all steps:
https://drive.google.com/file/d/1UUhBLviw6NkHHdaKts9fieTXSw4MG9YY/view

clockworked247:
@ayuwono @wooldridge if the step name is the same as another step you have, it won't let you create it.  Are you using a unique name?

clockworked247:
@ayuwono  @wooldridge If I create a Step by choosing a collection, it works.  But, if I choose "query" first, then switch back to collection, the Save button stays grey'd out.  So it appears there's a validation issue.  I just want to confirm that's what you were seeing.

clockworked247:
@wooldridge @ayuwono Fixed form submission issue.  It seems Angular Material has a bug where you can't make conditionally rendered fields required via the "required" HTML attribute (which adds the asterisks).  So, for now, just to get this in, I added an asterisk to the placeholder text.  The one or the other (query or collection) fields are still required via the custom form validation.

If you want any more refinement, let's file a bug for after beta...

See video:
https://drive.google.com/file/d/1GJIFW6ARyh6ewh5XpK0H1NlIwsXKZDxc/view

clockworked247:
@ayuwono Thanks for the approval.

"aside that custom step issue where you have to specify collection and target entity, everything looks good"

please file a separate bug.  This issue tracks validation fixes.

clockworked247:
Code has been rebased against latest.  Please merge @wooldridge

clockworked247:
requiring two approved reviews once again...

wooldridge:
@clockworked247, thanks.

Remaining issues:

1. Can submit a mapping or mastering step with an empty source query field. (To do this, define a source collection first, then switch to source query via radio button.)

2. Source info and target entity are required for custom. 

I have a possible fix for the second issue, will create a separate PR.

wooldridge:
@clockworked247 need second approval, @alexsmr @aebadirad @ayuwono ?

PR #2299, opened 2019-04-19
DHFPROD-2229: Remove unnecessary modules deploy in hubUpdate

ryanjdew:


aebadirad:
Let's make sure we're deploying everything properly after this step in upgrade. 4.3 flows should appear from under /plugins and the rest of 5.0 should be there too.

bsrikan:
@aebadirad We seem to have missed including flow-developer-role in the module permissions like we did in 4.3. Can you pls let me know if we should do the same for 5.0.

aebadirad:
@bsrikan ah, yes, we should add that in. Open a new bug and PR for it please?

PR #2301, opened 2019-04-19
DHFPROD-2237: added support for txt and binary output formats for ingestion

alexsmr:


srinathgit:
@alexsmr ,
I just chatted with @aebadirad , I think we will have to change "txt" to "text" for  inputFileType and outFormat

PR #2302, opened 2019-04-19
DHFPROD-2151: Buttons non functional in the flows page

brucean52:


web/src/main/ui/app/components/flows-new/manage-flows/manage-flows.component.ts
```diff
@@ -69,6 +71,29 @@ export class ManageFlowsComponent implements OnInit {
     });
   }
 
+  runFlow(runObject): void {
+    this.manageFlowsService.runFlow(runObject).subscribe(resp => {
+      // TODO add response check
+      // console.log('run flow resp', resp);
+      this.running = timer(0, 500)
```
alexsmr: This is similar functionality that exists in the edit-flow component. I suggest refactor it to the same place. It can be the Run button itself. 

We do not have much time now for it though. So please add a TODO comment to do that later. Open also a Jira ticket and reference it in TODO comment in both places.

brucean52: Yeah I agree, I figured we will optimize it once we test it out more.

PR #2306, opened 2019-04-19
DHFPROD-2218: can't create step if you don't point on the last step

hao1st:


hao1st:
@ayuwono @aebadirad Thanks for quickly finding the problem. I just fixed it.

PR #2308, opened 2019-04-19
Bugfix/dhfprod 2206

akshaysonvane:
TODO: Change the sourceQuery in all the artifacts


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/extensions/collections.sjs
```diff
@@ -20,9 +20,28 @@ const datahub = new DataHub();
...
-    resp = xdmp.eval("cts.collections(null, ['map'])", null, {database: xdmp.database(database)});
+    if (sourceQuery != null) {
+      let uris = xdmp.eval("cts.uris(null, ['limit=" + limit + "'], " + sourceQuery + ")", null, {database: xdmp.database(database)});
+      uris = datahub.hubUtils.normalizeToArray(uris);
+
+      let i;
+      resp = [];
+      for (i in uris) {
+        let uri = uris[i];
+        let doc = xdmp.eval("cts.doc('" + uri + "')", null, {database: xdmp.database(database)});
```
akshaysonvane: Can we move the for loop in the eval?
I need to pass the "resp" object to the eval and then get access to it so as to return it back.

aebadirad: cts.documentQuery(uris) and strip the loop entirely.


use a map function in js to return response you want.

PR #2314, opened 2019-04-20
DHFPROD-2087: add final/final as target/source db for mastering step …

aebadirad:
…in default flow, added defaults to built in step definitions

### DIFF
--- a/marklogic-data-hub/src/main/resources/hub-internal-artifacts/flows/default-ingestion.flow.json
+++ b/marklogic-data-hub/src/main/resources/hub-internal-artifacts/flows/default-ingestion.flow.json
@@ -1,17 +1,17 @@
 {
-  "name" : "default-ingestion",
-  "description" : "This is the default flow for ingesting into data hub",
+  "name": "default-ingestion",
+  "description": "This is the default flow for ingesting into data hub",
   "options": {
   },
-  "steps" : {
-	  "1": {	 
-	   "stepDefinitionType": "ingestion",
-	   "stepDefinitionName": "default-ingestion",
-	   "retryLimit": 0,
-	   "options": {
-        "outputFormat" : "json",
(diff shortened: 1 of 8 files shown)


marklogic-data-hub/src/main/resources/hub-internal-artifacts/flows/default-map-and-master.flow.json
```diff
@@ -1,27 +1,30 @@
...
-	  }
-	}
+  "steps": {
+    "1": {
+      "stepDefinitionType": "mapping",
+      "stepDefinitionName": "default-mapping",
+      "retryLimit": 0,
+      "options": {
+        "outputFormat": "json",
+        "collections": "default-mapping"
```
rahulvudutala: Can we have the collections as an array to make it consistent across all the flow and step def artefacts

marklogic-data-hub/src/main/resources/hub-internal-artifacts/flows/default-ingestion.flow.json
```diff
@@ -1,17 +1,17 @@
...
-	  }
-	}
+  "steps": {
+    "1": {
+      "stepDefinitionType": "ingestion",
+      "stepDefinitionName": "default-ingestion",
+      "retryLimit": 0,
+      "options": {
+        "outputFormat": "json",
+        "collections": "default-ingestion"
```
rahulvudutala: Can we have the collections as an array to make it consistent across all the flow and step def artefacts

marklogic-data-hub/src/main/resources/hub-internal-artifacts/flows/default-map-and-master.flow.json
aebadirad: Sure can.

PR #2317, opened 2019-04-21
DHFPROD-2263: Status and status time display fixes

wooldridge:
- Stop displaying the “Ended” prefix when flows are running and there is no end time available.
- Add defensive checks for status and time values


wooldridge:
@ayuwono @aebadirad, thanks for reviewing and merging!

PR #2318, opened 2019-04-21
DHFPROD-2265: Adding flow-developer-role to modulePermissions

bsrikan:
@aebadirad staging, final and job ports were enclosed in quotes. Not sure if that was intentional. I have removed them. Please review and let me know if I should switch it back.


aebadirad:
Yeah, we can remove them I believe. Not sure why there is a reason for the quotes - we can't make app servers or databases with spaces in the name regardless.

This shouldn't cause a backwards incompatibility with 4.x, either.

bsrikan:
> I agree with removing the quotes, unless @aebadirad knows a specific reason for the quotes. If we do remove the quotes, we need to remove the quote in the getCustomTokens functions as well. https://github.com/marklogic/marklogic-data-hub/blob/develop/marklogic-data-hub/src/main/java/com/marklogic/hub/impl/HubConfigImpl.java#L1699 https://github.com/marklogic/marklogic-data-hub/blob/develop/marklogic-data-hub/src/main/java/com/marklogic/hub/impl/HubConfigImpl.java#L1705 https://github.com/marklogic/marklogic-data-hub/blob/develop/marklogic-data-hub/src/main/java/com/marklogic/hub/impl/HubConfigImpl.java#L1712

Thanks @ryanjdew. That explains the test diff as well. Fixed it. Will commit after tests run fine locally.

PR #2320, opened 2019-04-22
DHFPROD-2264: Documents weren't being mastered through UI

ryanjdew:


ryanjdew:
@wooldridge I think the last time you checked it was due to not targeting the right collection. I added another commit that on the creation of a step if collections aren't specified it will default to the name of the step for mapping and mastering like ingestion is already doing. You'll need to delete and recreate your mapping step after pulling down the latest.

wooldridge:
With help from @ayuwono, I just figured out the problem. I was setting the source collection for my mastering to the name of the previous mapping step. But it looks like (in my current branch, anyway) data that is being mapped is only having the tag "default-mapping" applied and not a tag corresponding to the step name. When I master to a default-mapping source instead, I'm seeing mastering work. Great!

Sounds like maybe you addressed this issue per your last comment?

ryanjdew:
@wooldridge Yes. The last commit addresses the issue. Now documents should end up in a collection matching the step name.

ayuwono:
I don't see the documents being merged? I got new mastering and mdm collections, but the docs are all the same @ryanjdew

PR #2322, opened 2019-04-22
DHFPROD-2270: Adjust Job Details summary since no target entity

wooldridge:
Target entity data is missing from jobs payloads. Adjust Job Details UI to address this, see: https://project.marklogic.com/jira/browse/DHFPROD-2270


aebadirad:
Any reason why this is failing the QS tests @ayuwono ?

ayuwono:
@aebadirad , I think the same old installation issue that sometimes failing, relaunch it again

PR #2331, opened 2019-04-22
DHFPROD-2228: reset LatestJob info for UI to deal with a running flow…

hao1st:
… job which takes time to start

https://project.marklogic.com/jira/browse/DHFPROD-2228


clockworked247:
@brucean52 is there a jira task tracking that work so I can follow on after you get those small fixes in?

PR #2332, opened 2019-04-23
DHFPROD-2288: BUG: Update job status check on UI

brucean52:


web/src/main/ui/app/components/flows-new/models/flow.model.ts
```diff
@@ -27,7 +27,11 @@ export class Flow {
   }
 
   get status(): string {
-    return (this.latestJob && this.latestJob['status']) ? this.latestJob['status'] : 'Never Run';
+    if (this.latestJob === null) {
+      return '';
+    } else {
+      return (this.latestJob && this.latestJob['status']) ? this.latestJob['status'] : 'Never Run';
```
wooldridge: Status text should be "Never run" (capitalize first word, lowercase those that follow).

PR #2345, opened 2019-04-23
DHFPROD-2122: Display the correct available source databases in the UI

brucean52:
removed the jobs database as an option


clockworked247:
@wooldridge would you merge plz?  thx

PR #2359, opened 2019-04-24
DHFPROD-1971: Flow creation with no flow name should not be allowed

alexsmr:


alexsmr:
There is more work to be done for validation and presentation however. Added https://project.marklogic.com/jira/browse/DHFPROD-2314 for post beta goals.

bsrikan:
FYI&A This is not working as expected. I still can create a flow with no name.

PR #2369, opened 2019-04-25
DHFPROD-2320: Status doesn't change from "Run" to "Stop" when flow is run in Flows page

brucean52:


timur-isangulov:
Bruce, I looked at few cases:
1. Run the flow from the inside the flow. Result: RUN changes to STOP while job is running. 
2. Run the flow from the inside the flow and refresh the page. Result: button is RUN, the job still runs. 
3. Run the flow from the Manage flows view. Result: button is RUN and the job still runs. 
4. Run the flow from the inside the flow and immediately open Manage flows view. Result: button is RUN and the job still runs.

brucean52:
> Bruce, I looked at few cases:
> 
> Run the flow from the inside the flow. Result: RUN changes to STOP while job is running.
> Run the flow from the inside the flow and refresh the page. Result: button is RUN, the job still runs.
> Run the flow from the Manage flows view. Result: button is RUN and the job still runs.
> Run the flow from the inside the flow and immediately open Manage flows view. Result: button is RUN and the job still runs.
> Can you please fix cases 2,3,4.
> Thanks.

Live updates for jobs running isn't supported when navigating between views. I'm adding this functionality for DHFPROD-2241

PR #2371, opened 2019-04-25
DHFPROD-2368: e2e tests flow page objects fix for 5.0

timur-isangulov:

### DIFF
--- a/web/e2e/page-objects/flows/editFlow.ts
+++ b/web/e2e/page-objects/flows/editFlow.ts
@@ -43,7 +43,7 @@ export class EditFlow extends AppPage {
   }
 
   get runFlowStatus() {
-    return element(by.css("a.ng-star-inserted"));
+    return element(by.id("latest-job-status"));
   }
 
   get jobStartedTimestamp() {
--- a/web/e2e/page-objects/flows/editFlow.ts
+++ b/web/e2e/page-objects/flows/editFlow.ts
@@ -43,7 +43,7 @@ export class EditFlow extends AppPage {
   }
 
   get runFlowStatus() {
-    return element(by.id("latest-job-status"));
+    return element(by.css("#latest-job-info #latest-job-status"));
   }
 
   get jobStartedTimestamp() {
--- a/web/e2e/page-objects/flows/manageFlows.ts
+++ b/web/e2e/page-objects/flows/manageFlows.ts
@@ -216,14 +216,14 @@ export class ManageFlows extends AppPage {
    */
 
   async clickPaginationDropDown() {
-    return await element(by.id("mat-select-0")).click();
+    return await element(by.id("mat-select-arrow")).click();
   }
 
-  itemsPerPage(value: number) {
-    return element(by.css(`mat-option[ng-reflect-value="${value}"]`));
+  itemsPerPage(value: string) {
+    return element(by.cssContainingText("mat-option .mat-option-text", value));
   }
 
-  async selectItemsPerPage(value: number) {
+  async selectItemsPerPage(value: string) {
     let pageNum = this.itemsPerPage(value);
     return await pageNum.click();
   }
@@ -233,7 +233,7 @@ export class ManageFlows extends AppPage {
    * @param option = ["Previous page"/"Next page"]
    */
   async clickPageNavigation(option: string) {
-    return await element(by.css(`button[ng-reflect-message="${option}"]`)).click();
+    return await element(by.css(`button[aria-label="${option}"]`)).click();
   }
 
   get pageRangeText() {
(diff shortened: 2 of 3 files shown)


web/e2e/page-objects/flows/manageFlows.ts
```diff
@@ -216,14 +216,14 @@ export class ManageFlows extends AppPage {
...
 
   async clickPaginationDropDown() {
-    return await element(by.id("mat-select-0")).click();
+    return await element(by.id("mat-select-arrow")).click();
   }
 
-  itemsPerPage(value: number) {
-    return element(by.css(`mat-option[ng-reflect-value="${value}"]`));
+  itemsPerPage(value: string) {
+    return element(by.cssContainingText("mat-option .mat-option-text", value));
```
ayuwono: isn't the same thing? it goes to the same selector, it doesn't need to be changed

web/e2e/page-objects/flows/manageFlows.ts
```diff
@@ -233,7 +233,7 @@ export class ManageFlows extends AppPage {
    * @param option = ["Previous page"/"Next page"]
    */
   async clickPageNavigation(option: string) {
-    return await element(by.css(`button[ng-reflect-message="${option}"]`)).click();
+    return await element(by.css(`button[aria-label="${option}"]`)).click();
```
ayuwono: isn't the same thing? it goes to the same selector, it doesn't need to be changed

web/e2e/page-objects/flows/manageFlows.ts
```diff
@@ -216,14 +216,14 @@ export class ManageFlows extends AppPage {
...
   }
 
-  itemsPerPage(value: number) {
-    return element(by.css(`mat-option[ng-reflect-value="${value}"]`));
+  itemsPerPage(value: string) {
+    return element(by.cssContainingText("mat-option .mat-option-text", value));
   }
 
-  async selectItemsPerPage(value: number) {
+  async selectItemsPerPage(value: string) {
```
ayuwono: if we don't change the itemsPerPage, this doesn't need to be change either

web/e2e/page-objects/flows/manageFlows.ts
timur-isangulov: Is ng-reflect-value attribute of web element? I didn't find it. Was it there before?

web/e2e/page-objects/flows/manageFlows.ts
timur-isangulov: I didn't see the attribute ng-reflect-message.

web/e2e/page-objects/flows/editFlow.ts
```diff
@@ -43,7 +43,7 @@ export class EditFlow extends AppPage {
   }
 
   get runFlowStatus() {
-    return element(by.css("a.ng-star-inserted"));
+    return element(by.id("latest-job-status"));
   }
```
bsrikan: Just the id is matching in 2 places. Locators should always match to exactly 1 we intend to test. Please change this to use by.css(.flow-container #latest-job-status) or by.css(#latest-job-info #latest-job-status) whichever works

web/e2e/page-objects/flows/manageFlows.ts
```diff
@@ -216,14 +216,14 @@ export class ManageFlows extends AppPage {
    */
 
   async clickPaginationDropDown() {
-    return await element(by.id("mat-select-0")).click();
+    return await element(by.id("mat-select-arrow")).click();
   }
```
bsrikan: mat-select-row is not an id, its a class. Please use by.css
Like earlier comment mat-select-row by itself returns 4 matches. Pls use by.css(".mat-select .mat-select-arrow")

web/e2e/page-objects/flows/manageFlows.ts
bsrikan: Yes it was there when I first added it back in March. Lot of changes since then. Its not there anymore. Changing the type to string has impact in other places where itemsPerPage is used. For instance selectItemsPerPage and the corresponding tests. If string works then pls change all the affected places.

web/e2e/page-objects/flows/manageFlows.ts
bsrikan: Was there before. Not anymore. Could have used label or the class. Lets chk with UI team once for best practice. If class name is something that will stay longer for the element(Apparently label name has changed or been removed) then lets use class by.css{`button.mat-paginator-navigation-${option}`} where option is previous or next

web/e2e/page-objects/flows/manageFlows.ts
ayuwono: Hmm...you can't see it? I'm using latest develop:
Screen Shot 2019-04-26 at 11 15 00 AM

ayuwono:
Do not merge yet, as we will have another PR to minimize confilcts

ayuwono:
We can close this as PR-2392 includes these changes

PR #2374, opened 2019-04-26
DHFPROD-2314: added validations and error messages to the flow dialog

alexsmr:


wooldridge:
I'm seeing the same as @ayuwono. What is the call on spaces and other special characters? Seems like consistency with step names would be a good thing.

wooldridge:
@aebadirad is OK with keeping the same restrictions on flow names as we do step names. Let's do that and keep consistent error messaging on those as well. If anyone sees a problem with that, please raise it.

alexsmr:
Moved back to Step name restriction pattern

PR #2375, opened 2019-04-26
DHFPROD-2330: Avoid unnecessary saving when managing steps in a flow

hao1st:
https://project.marklogic.com/jira/browse/DHFPROD-2330

It also fixed https://project.marklogic.com/jira/browse/DHFPROD-2350 on the server side (save with a new version only if the mapping configuration gets changed).

### DIFF
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/flow/Flow.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/flow/Flow.java
@@ -166,4 +166,12 @@ public interface Flow {
      * @return this mapping
      */
     Flow deserialize(JsonNode json);
+
+    /**
+     * Get step Info by step id
+     *
+     * @param stepId
+     * @return
+     */
+    Step getStepById(String stepId);
 }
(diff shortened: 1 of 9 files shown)


akshaysonvane:
Now that @wooldridge PR resolves the PUT request problem from UI, do we need this anymore? 
If so should we also override the `hashCode` method to comply with the contract for hashCode allowing it to work with hash-based collections?

hao1st:
> Now that @wooldridge PR resolves the PUT request problem from UI, do we need this anymore?
> If so should we also override the `hashCode` method to comply with the contract for hashCode allowing it to work with hash-based collections?

@akshaysonvane Good thinking! I thought about it. However,  the Step and MappingModel objects include mutable members and they should not be used as keys in hash-based collections. Most cases we should override the two methods. In our case, maybe I could use another method name "isEqual(...)" to avoid the confusion.

aebadirad:
Let's add another method that doesn't override hashcode, and does an "isEqual" or "isDuplicate".

I still think we should do this check on the backend.

PR #2376, opened 2019-04-26
DHFPROD-2285: Make committed docs links functional

ryanjdew:


aebadirad:
Need to fix that smart mastering test again somehow and also handle cases of multiple jobs for a doc.

wooldridge:
After further thought, I see issues with all the committed links currently.

If I have a flow that commits data to STAGING and FINAL, committed links from the job will be incorrect since we can only search on STAGING or FINAL (as described in my comment above).

If I have a flow where multiple steps commit to STAGING or FINAL, committed links from a specific job step will be incorrect since it will show all the docs committed by all the steps for the database. I see the beginnings of a fix for this (add constraint by step) here:

https://github.com/marklogic/marklogic-data-hub/pull/2376/files#diff-cd770ad6645ae020efd5172ef27b4d29R480

ryanjdew:
@ayuwono Can you inspect your network traffic? I suspect that the new STAGING/FINAL options weren't deployed. In the response from the /api/search REST call, do you get something to the effect of  "Warning unknown constraint [createdByJob]"?

@wooldridge I'll add an additional filter for the step. As for a federated search across databases, that gets messy real quick, as you'll need to read out a lot more documents for filtering/sorting and compare the contents across the databases. There might be a way to do it with a superdatabase, but that is a drastic change that I don't think would be worth the benefit of this use case. I do try to determine whether the last step was run in staging or final to determine the job-level link and attempt to direct individual step links to the proper target database as well.

wooldridge:
@ryanjdew, I'd forgotten about superdatabases.

ryanjdew:
@wooldridge Yeah, probably best to forget them for most instances. Taking a little closer look, this requirement alone pretty much makes it off the table.

"Because super-databases and their sub-databases are effectively a single database, you cannot have documents with the same URI in super-databases and their sub-databases. It is a best practice to use directories to ensure that your document URIs are unique."

ayuwono:
@ryanjdew , I don't see the job constraint on network traffic, but I see this on the console:
ERROR TypeError: Cannot read property 'status' of undefined
    at JobDetailsUiComponent.push../src/main/ui/app/components/jobs-new/ui/job-details-ui.component.ts.JobDetailsUiComponent.ngOnInit (job-details-ui.component.ts:32)
    at checkAndUpdateDirectiveInline (core.js:22000)
    at checkAndUpdateNodeInline (core.js:23264)
    at checkAndUpdateNode (core.js:23226)
    at debugCheckAndUpdateNode (core.js:23860)
    at debugCheckDirectivesFn (core.js:23820)
    at Object.eval [as updateDirectives] (JobDetailsComponent.html:2)
    at Object.debugUpdateDirectives [as updateDirectives] (core.js:23812)
    at checkAndUpdateView (core.js:23208)
    at callViewAction (core.js:23449)

On your latest commit, I still see documents not constrained by the collection (created-by-job), I still see other documents from the other collections displayed:

Screen Shot 2019-04-30 at 1 38 51 PM

ryanjdew:
@ayuwono Pushed another commit to resolve the console errors you were seeing, but this filter isn't a collections filter. It is a filter on the documents that have the job ID in their createdByJob metadata.

ryanjdew:
@ayuwono @wooldridge The latest commits preserve the metadata for mastered docs and add the target database to the committed links on the manage jobs page.

ryanjdew:
@ayuwono  I'm not getting the error, but I also see the metadata field in my database settings.
Screen Shot 2019-05-02 at 9 56 10 AM

aebadirad:
Yes, make sure you do a full mlDeploy due to the new path indexes.

ayuwono:
Let me try with a clean env

wooldridge:
Testing now...

ryanjdew:
@wooldridge That is what I'd expect. the 1000 count is the sum of both mapping and mastering runs, which get written to the same URI. The extra count beyond 500 is the extra mastering docs that are created for merges, etc. So the count of documents "committed" is the count of docs passed into a step and succeeded, not necessarily the count of documents created by the step.

PR #2381, opened 2019-04-29
DHFPROD-2369: Mapping json and xml docs to work with sourceContext as "/" and "/root"

rahulvudutala:


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mapping/default/lib.sjs
```diff
@@ -15,12 +15,33 @@ function getMappingWithVersion(mappingName, version) {
...
+    } else {
+      sourceContext += "/" + element;
+    }
+  });
+  return sourceContext;
+}
+
 function processInstance(model, mapping, content) {
- return  extractInstanceFromModel(model, model.info.title, mapping, content);
+ return extractInstanceFromModel(model, model.info.title, mapping, content);
```
aebadirad: what if there's no info or info.title present? Probably want to do an existence check before we throw an exception here.

rahulvudutala: We can't deploy an entity without the info object in {entity-name}.entity.json file. It will throw the following exception.
Local message: failed to apply resource at documents: Internal Server Error. Server Message: TDE-INDEX: Error applying template entity-types-info-json.tdex to document /entities/Order2.entity.json: TDE-EVALFAILED: Eval for Object='$title' returns an empty sequence . See the MarkLogic server error log for further detail.

Should we still have this check.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mapping/default/lib.sjs
```diff
@@ -31,10 +52,11 @@ function extractInstanceFromModel(model, modelName, mapping, content) {
     instance['$version'] = '0.0.1';
   }
 
-  if (content.nodeName === 'envelope' || (content.nodeKind === 'document' && content.root && content.root.envelope)) {
+  if (content.nodeName === 'envelope' || (content.nodeKind === 'document')) {
```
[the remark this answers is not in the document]
rahulvudutala: If dhf transforms are used, there should be envelope right. Do we also support for docs that are ingested as in using mlcp without dhf transforms for mapping?

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mapping/default/lib.sjs
rahulvudutala: @aebadirad Thanks for letting know that TDE's can be turned off. I will add the check

PR #2385, opened 2019-04-29
DHFPROD-2281: required field indicators removed from new custom step

clockworked247:
https://drive.google.com/file/d/1fVJk8VmzspnF36ftY-F8_xEP-oomCWnB/view

### DIFF
--- a/web/src/main/ui/app/components/flows-new/edit-flow/ui/new-step-dialog-ui.component.ts
+++ b/web/src/main/ui/app/components/flows-new/edit-flow/ui/new-step-dialog-ui.component.ts
@@ -28,9 +28,17 @@ export class NewStepDialogUiComponent implements OnInit {
   public stepType: typeof StepType = StepType;
   readonly stepOptions = Object.keys(this.stepType);
 
-  selectedSource = '';
+  selectedSource: string = '';
   newStepForm: FormGroup;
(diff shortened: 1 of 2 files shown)


clockworked247:
Rebased to latest develop.  Please approve and @wooldridge if you'd merge we can do https://github.com/marklogic/marklogic-data-hub/pull/2386

web/src/main/ui/app/components/flows-new/edit-flow/ui/new-step-dialog-ui.component.ts
```diff
@@ -28,9 +28,17 @@ export class NewStepDialogUiComponent implements OnInit {
   public stepType: typeof StepType = StepType;
```
brucean52: Since we are not using the ENUM in the HTML anymore, we don't need instantiate stepType. I think this should be part of a separate code clean up task.

alexsmr: +1

clockworked247: Good point, thanks for bringing attention to it Bruce.

PR #2396, opened 2019-05-01
DHFPROD-2334: Exclude internal collections

akshaysonvane:


akshaysonvane:
> @akshaysonvane , you missed the http://marklogic.com/entity-services. I can see this collection here
> 
> Screen Shot 2019-05-01 at 2 19 43 PM

Should work now.

PR #2401, opened 2019-05-02
DHFPROD-2396: correct attachments on raw json to json, massage custom…

aebadirad:
… module scaffolding, streamline some flow utils stuff, and handle nullnodes in hubutils

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/flow-utils.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/flow-utils.sjs
@@ -112,6 +112,10 @@ class FlowUtils {
       if(content instanceof Element &&  content.nodeName.toLowerCase() === 'root' && content.namespaceURI.toLowerCase() === ""){
         instance = Sequence.from(content.xpath('/root/node()'));
       } else {
+        if(content['$attachments']) {
+          attachments = content['$attachments'];
+          delete content['$attachments'];
+        }
         instance = content;
       }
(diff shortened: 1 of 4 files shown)


aebadirad:
@ayuwono @bsrikan just want to make sure this doesn't cause a regression in what you test on either

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/flow-utils.sjs
```diff
@@ -223,26 +227,25 @@ class FlowUtils {
...
-      }
-      else if (dataFormat === this.consts.XML) {
-        return json.transformFromJson(resp, json.config("custom"));
-      }
-      else {
-        return resp;
-      }
-    } else if(isXml && resp) {
-      if ((resp instanceof ArrayNode || resp instanceof Array) && dataFormat === this.consts.XML) {
+      let kind = resp ? xdmp.nodeKind(resp) : null;
```
rjrudin: I know this fix is needed, as I'm running into a null error from it locally

marklogic-data-hub/src/main/resources/scaffolding/custom-module/main.sjs
```diff
@@ -48,13 +48,16 @@ function main(content, options) {
   }
 
   //get our instance, default shape of envelope is envelope/instance, else it'll return an empty object/array
-  let instance = datahub.flow.flowUtils.getInstance(doc);
+  let instance = datahub.flow.flowUtils.getInstance(doc) || {};
```
rjrudin: And these seem like useful fixes too so that null isn't passed in

PR #2404, opened 2019-05-02
DHFPROD-2171: Resolve errors loading online-store in QS

ryanjdew:


examples/online-store/flows/LoadOrders.flow.json
```diff
@@ -1,11 +1,22 @@
...
-    "1": {
-      "stepDefinitionName": "IngestOrders",
-      "stepDefinitionType": "CUSTOM"
+    "1" : {
+      "name" : "IngestOrders",
+      "description" : "null",
+      "options" : { },
+      "customHook" : null,
+      "retryLimit" : 0,
+      "batchSize" : 0,
```
akshaysonvane: Should be null or matching the flow
```suggestion
      "batchSize" : null,
```

examples/online-store/flows/LoadOrders.flow.json
```diff
@@ -1,11 +1,22 @@
...
+      "threadCount" : 0,
```
akshaysonvane: Should be null or matching the flow
```suggestion
      "threadCount" : null,
```

marklogic-data-hub/src/main/java/com/marklogic/hub/step/impl/Step.java
```diff
@@ -133,15 +133,19 @@ public static Step deserialize(JsonNode json) {
         Step step = new Step();
 
         JSONObject jsonObject = new JSONObject(json);
-        step.setName(jsonObject.getString("name"));
+        step.setStepDefinitionName(jsonObject.getString("stepDefinitionName"));
+        step.setStepDefinitionType(StepDefinition.StepDefinitionType.getStepDefinitionType(jsonObject.getString("stepDefinitionType")));
+        String stepName = jsonObject.getString("name");
+        if (stepName == null) {
+            stepName = step.getStepDefinitionName();
```
akshaysonvane: Is step name optional?

aebadirad: Yes, but that's not encouraged.

PKey for the step is the #.

ryanjdew: Good question. The backend doesn't require it and I don't believe the middle-tier does either, but Quick Start seems to.

akshaysonvane: We seem to track both step name and step-def name in prov.
Should we make it required as going forward when we have a step-def bank the step name will come in handy to track the relation between both?

aebadirad: Yeah. Going forward we should.

aebadirad:
Rebase needed

PR #2406, opened 2019-05-02
DHFPROD-2339: Flow name should not be editable in QS and server should throw error

alexsmr:


web/src/main/ui/app/components/flows-new/edit-flow/ui/new-step-dialog-ui.component.html
```diff
@@ -7,7 +7,7 @@ <h1 mat-dialog-title>{{title}}</h1>
       <mat-option class="type-option" *ngFor="let option of stepOptions" [value]="option">{{capitalFirstLetter(option)}}</mat-option>
     </mat-select>
   </mat-form-field>
-  <div *ngIf="newStepForm.value.stepDefinitionType || editingStep">
+  <div *ngIf="newStepForm.value.stepDefinitionType || isUpdate">
```
brucean52: If we are going to use isUpdate boolean in new step dialog, we need to add the parameter to the interface in new-step-dialog.component.ts and then update the data objects when opening the new step dialog in edit-flow-ui.component.ts and step.component.ts

alexsmr: Agree. I add it to the interface, but the values are already updated.

PR #2408, opened 2019-05-02
DHFPROD-2346: 5.0 QuickStart is resolving project relative path in in…

hao1st:
…gest flow

Also fixed inputFilePath does not save correctly...

### DIFF
--- a/web/src/main/java/com/marklogic/hub/web/web/UtilController.java
+++ b/web/src/main/java/com/marklogic/hub/web/web/UtilController.java
@@ -16,16 +16,9 @@
  */
 package com.marklogic.hub.web.web;
 
+import com.marklogic.hub.impl.HubConfigImpl;
 import com.marklogic.hub.util.FileUtil;
 import com.marklogic.hub.web.model.SearchPathModel;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-import org.springframework.stereotype.Controller;
-import org.springframework.web.bind.annotation.RequestMapping;
-import org.springframework.web.bind.annotation.RequestMethod;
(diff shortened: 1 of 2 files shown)


web/src/main/java/com/marklogic/hub/web/web/UtilController.java
```diff
@@ -16,46 +16,56 @@
...
+
 	@RequestMapping(value = "/searchPath", method = RequestMethod.GET)
 	@ResponseBody
 	public Map<String, Object> searchPath(@RequestParam String path) {
-		logger.debug("Search Path:" + path);
-		List<SearchPathModel> folders = new ArrayList<>();
-		List<SearchPathModel> files = new ArrayList<SearchPathModel>();
-		Path currentPath = Paths.get(".").toAbsolutePath().normalize();
-		Path relativePath;
+        String basePath = ".";
```
srinathgit: We always had this behavior. Why are we changing this now ? During  bootstrapping, when the project is not initialized and hence when project dir not set , this may return null

aebadirad: Check if hubprojectdir() is set before doing this, or else use that behavior.

srinathgit: The behavior with this change seems to be 
1. When QS is started and hubConfig is not populated, the dir listing is relative to location from which war file was started (this can be seen in the dir picker in the login page)
2. Once logged in and hubProject/ hubConfig are populated, the dir listing is relative to current proj dir (this can be seen in dir picker in ingestion step page)
3. Once user logs out (old hubConfig/hubProject are still populated), when user gets back to login page and picks dir, it is relative to the last proj dir

PR #2410, opened 2019-05-03
DHFPROD-2331: Set all step/flow start/end times to server

srinathgit:
This PR,

1. Remove job doc creation/updating from collector 
2. Moves all step/flow start/end times to server- changes to extensions/jobs.sjs to handle creation/updating of stepresponse in the job doc at the start of step (when status is changed to "running step #" ) and at the end of step (when status is changed to "completed (or other statuses) step #" ) 
3. Corresponding changes to FlowRunner, Write/QueryStepRunners
4. Adds lastAttemptedStep, lastCompletedStep, user info to RunFLowResponse
5. Refactor code (JobUpdate renamed to JobDocManager and move common methods in Write/QueryStepRunner to StepRunnerUtil)

### DIFF
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/flow/impl/FlowRunnerImpl.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/flow/impl/FlowRunnerImpl.java
@@ -335,24 +335,6 @@ public class FlowRunnerImpl implements FlowRunner{
                 logger.error(e.getMessage());
             }
             finally {
-                if (!isJobSuccess.get()) {
-                    try {
-                        flowStatusListeners.forEach((FlowStatusListener listener) -> {
-                            listener.onStatusChanged(jobId, runningStep, jobStatus, currPercentComplete[0], currSuccessfulEvents[0], currFailedEvents[0], JobStatus.FAILED.toString());
-                        });
-                    } catch (Exception ex) {
-                        logger.error(ex.getMessage());
-                    }
-                } else {
-                    try {
-                        flowStatusListeners.forEach((FlowStatusListener listener) -> {
-                            listener.onStatusChanged(jobId, runningStep, jobStatus, currPercentComplete[0], currSuccessfulEvents[0], currFailedEvents[0], JobStatus.FINISHED.toString());
-                        });
-                    } catch (Exception ex) {
-                        logger.error(ex.getMessage());
-                    }
-                }
-
                 JsonNode jobNode = null;
                 try {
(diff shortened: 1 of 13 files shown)


srinathgit:
@hao1st ,
I ran a basic test on QS and it seems to render values fine. Can you please confirm if any changes are required on the web layer ?

marklogic-data-hub/src/main/java/com/marklogic/hub/flow/impl/FlowRunnerImpl.java
```diff
@@ -353,25 +353,26 @@ else if (!isJobSuccess.get()) {
...
+                        resp.setStartTime(jobDoc.getStartTime());
+                        resp.setEndTime(jobDoc.getEndTime());
+                        resp.setUser(jobDoc.getUser());
+                        resp.setLastAttemptedStep(jobDoc.getLastAttemptedStep());
+                        resp.setLastCompletedStep(jobDoc.getLastCompletedStep());
+                    }
+                    catch (Exception e) {
+                        logger.error(e.getMessage());
+                    }
                 }
```
hao1st: Nice job with updating start/end time logic on the e-node side! 👍

If you switch the order of the following codes by putting the first snippet before the second snippet, it should be fine.

https://github.com/marklogic/marklogic-data-hub/blob/c7cbd0297c17d5d37ff82abf3157d5959e125cfd/marklogic-data-hub/src/main/java/com/marklogic/hub/flow/impl/FlowRunnerImpl.java#L356-L376

https://github.com/marklogic/marklogic-data-hub/blob/c7cbd0297c17d5d37ff82abf3157d5959e125cfd/marklogic-data-hub/src/main/java/com/marklogic/hub/flow/impl/FlowRunnerImpl.java#L338-L354

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/endpoints/collector.sjs
```diff
@@ -41,13 +41,6 @@ if(method === 'GET') {
   const database = requestParams.database;
   let options = requestParams.options ? JSON.parse(requestParams.options) : {};
 
-  let jobDoc = datahub.jobs.getJobDocWithId(jobId);
```
hao1st: we do not need to have the lines:
https://github.com/marklogic/marklogic-data-hub/blob/29f69f0379c4a7167e69baa5178073f22d46bb02/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/endpoints/collector.sjs#L40-L41

PR #2411, opened 2019-05-03
DHFPROD-2377: reload mapping after flow run

wooldridge:
Running a flow can ingest sample documents required for mapping.
Account for this by reloading any mapping steps when a job ends.

1. edit-flow.component sets flowEnded flag with job ID at end of job
2. flowEnded passed to edit-flow-ui.component, which recognizes change in value
3. flowEnded passed to step.component, which triggers mapping load on change if mapping step

Screencast: https://drive.google.com/file/d/1to9CTpCugHLNLEPVKBJ5Z5xKh0WZqcqP/view


aebadirad:
Fixed the merge conflict.

PR #2412, opened 2019-05-03
DHFPROD-2400: Prefix metadata with datahub

ryanjdew:


aebadirad:
We're picking up a few test failures @ryanjdew

ryanjdew:
@rjrudin I believe we're moving to just Data Hub, instead of Data Hub Framework in our docs/published material. Personally, I don't have a preference for what the prefix is, but I agree we should be consistent.

aebadirad:
I think datahub is good for now. I agree we have a datahub vs. hub delta, but for now let's go with datahub (since we use that server side all over) and see if we can close the deltas elsewhere in future versions.

PR #2414, opened 2019-05-03
DHFPROD-2391: Handle whitespace in mapping for json source docs

akshaysonvane:


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mapping/default/lib.sjs
```diff
@@ -33,6 +33,19 @@ function getSourceContext(sourceContext) {
   return sourceContext;
 }
 
+function getPath(sourceContext, connector, propertyName) {
+  let path;
+  if (xdmp.castableAs("http://www.w3.org/2001/XMLSchema", "NCName", propertyName)) {
+    path = `${sourceContext}${connector}${propertyName}`;
+  } else {
+    if (connector.includes("*:")) {
+      connector = "";
```
ryanjdew: Should this replace just the "*:" with blank instead of the whole connector as blank? 

A first glance, it looks like it could lose "/", which may be significant.

akshaysonvane: Yes, that right, @ryanjdew. It should only replace "*:"

PR #2418, opened 2019-05-04
DHFPROD-2419: Handle 0 committed values in Jobs views

wooldridge:
Details in: https://project.marklogic.com/jira/browse/DHFPROD-2419

You can test by running a flow job that fails (e.g., put in bad URI replacement text for an ingestion). This fix addresses the missing 0 values in the committed column.

Plus minor CSS alignment tweak.


aebadirad:
These test failures are getting rather annoying :/

PR #2419, opened 2019-05-05
DHFPROD-2414: QS somehow persists previously used data-hub name

hao1st:
https://project.marklogic.com/jira/browse/DHFPROD-2414

### DIFF
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/impl/HubConfigImpl.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/impl/HubConfigImpl.java
@@ -2016,4 +2016,75 @@ public class HubConfigImpl implements HubConfig
         manageClient = null;
     }
 
+    public void resetHubConfigs() {
+        stagingDbName = null;
+        stagingHttpName = null;
+        stagingForestsPerHost = null;
+        stagingPort = null;
+        stagingAuthMethod = null;
+        stagingScheme = null;
+        stagingSimpleSsl = null;
+
+        stagingSslContext = null;
+        stagingSslHostnameVerifier = null;
+        stagingCertFile = null;
+        stagingCertPassword = null;
+        stagingExternalName = null;
+        stagingTrustManager = null;
+
+        finalDbName = null;
+        finalHttpName = null;
+        finalForestsPerHost = null;
+        finalPort = null;
+        finalAuthMethod = null;
+        finalScheme = null;
+
+        finalSimpleSsl = null;
+        finalSslContext = null;
+        finalSslHostnameVerifier = null;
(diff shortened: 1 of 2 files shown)


aebadirad:
@ayuwono we're getting some test failures....is this related to the flushing of the hugconfig, but I don't see how the test failures are related.

So approving, but just wanted to let you know

ayuwono:
> @ayuwono we're getting some test failures....is this related to the flushing of the hugconfig, but I don't see how the test failures are related.
> 
> So approving, but just wanted to let you know

The test failures are because of DHFPROD-2421 (NPE) and The About button has no unique id (using "login-button" as id -- DHFPROD-2420)

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/HubConfigImpl.java
```diff
@@ -2016,4 +2016,75 @@ public void resetAppConfigs() {
         manageClient = null;
     }
 
+    public void resetHubConfigs() {
+        stagingDbName = null;
```
rjrudin: I think reflection would be useful here to lessen the maintenance burden - i.e. any time a new field is added to HubConfigImpl, we have to hope that someone remembers to null it out in this method if necessary. Could reflect on all the fields and null out anything starting with "staging" or "final" or "job" or "modules", and then also null out the handful of other fields not starting with one of those prefixes.

srinathgit: @hao1st , I agree with @rjrudin .  HubTestBase.resetProperties() nulls out fields using reflection. You can modify the logic.

```
    public void resetProperties() {
        Field[] fields = HubConfigImpl.class.getDeclaredFields();
        Set<String> s =  Stream.of("hubProject", "environment", "flowManager", 
                "dataHub", "versions", "logger", "objmapper", "projectProperties").collect(Collectors.toSet());

        for(Field f : fields){
            if(! s.contains(f.getName())) {
                ReflectionUtils.makeAccessible(f);
                ReflectionUtils.setField(f, adminHubConfig, null);                
            }
            
        }
    }
```

aebadirad: @rjrudin Could we instead shift hubconfigimpl for the next minor release to be an extension of appconfig and resolve it in that effort?

Not saying we shouldn't use reflection atm, but just that the longer-term solution consist of a better way to handle configs? We'll esp have to start considering multi-users logged in for operational use (maybe not modifying the project, but just different users running flows/getting data, and right now that's tied at the hip with the hubconfig.)

PR #2421, opened 2019-05-05
DHFPROD-2421: move creation/release of client into the service method…

aebadirad:
… to avoid null pointer

### DIFF
--- a/web/src/main/java/com/marklogic/hub/web/service/FlowJobService.java
+++ b/web/src/main/java/com/marklogic/hub/web/service/FlowJobService.java
@@ -44,6 +44,9 @@ public class FlowJobService extends ResourceManager {
 
     public FlowJobService() {
         super();
+        if(hubConfig != null) {
+            this.setupClient();
+        }
     }
 
     private void setupClient() {
@@ -51,7 +54,9 @@ public class FlowJobService extends ResourceManager {
     }
 
     public FlowJobs getJobs(String flowName) {
-        this.setupClient();
+        if(this.client == null) {
+            this.setupClient();
+        }
         try {
             return cachedJobsByFlowName.get(flowName, () -> {
                 client.init(ML_JOBS_NAME, this);
@@ -149,10 +154,14 @@ public class FlowJobService extends ResourceManager {
         }
         return cachedJobsByFlowName.getIfPresent(flowName);
     }
-
+    //TODO: fix this whole client mess and properly report exceptions
     private void release() {
         if(this.client != null) {
-            this.client.release();
+            try {
+                this.client.release();
+            } catch (Exception e) {
+                this.client = null;
+            }
         }
     }
 }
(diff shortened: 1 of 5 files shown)


ayuwono:
@aebadirad , if you see QS jenkins failures, it's because of the old installation issue:
java.lang.RuntimeException: Error occurred while loading REST modules: Error occurred while loading modules; host: localhost; port: 8010; cause: java.net.ConnectException: Failed to connect to localhost/0:0:0:0:0:0:0:1:8010
	at com.marklogic.client.ext.modulesloader.impl.DefaultModulesLoader.rethrowRestModulesFailureIfOneExists(DefaultModulesLoader.java:172)
	at com.marklogic.client.ext.modulesloader.impl.DefaultModulesLoader.loadModules(DefaultModulesLoader.java:156)
	at com.marklogic.hub.deploy.commands.LoadHubModulesCommand.execute(LoadHubModulesCommand.java:104)
	at com.marklogic.appdeployer.impl.AbstractAppDeployer.executeCommand(AbstractAppDeployer.java:111)
	at com.marklogic.hub.deploy.HubAppDeployer.executeCommand(HubAppDeployer.java:78)
	at com.marklogic.appdeployer.impl.AbstractAppDeployer.deploy(AbstractAppDeployer.java:70)
	at com.marklogic.hub.deploy.HubAppDeployer.deploy(HubAppDeployer.java:58)
	at com.marklogic.hub.impl.DataHubImpl.install(DataHubImpl.java:531)

web/src/main/java/com/marklogic/hub/web/service/FlowJobService.java
```diff
@@ -40,17 +40,18 @@
 
     //use a cache with ttl 30 minutes or longer to reduce frequency for fetching data from DB
```
rjrudin: The comment is out of sync with the code now - I recommend removing the ttl part so don't have to keep it in sync, just say "use a chance to reduce...".

aebadirad: Well, I think I'd like to remove it entirely, as MarkLogic should be a source of truth

aebadirad:
That makes no sense why refreshing the browser would affect the endpoint. Is something occurring on the client that's causing this too?

web/src/main/java/com/marklogic/hub/web/service/FlowJobService.java
hao1st: well, I felt would it be nice to have a cache (5 min is good, 30 min may be too optimistic) in web-tier to avoid unnecessary query against job db (even if ml database has cache mechanism). Unless a user feel the data is not consistent because other apps or he manually changed the data in the database.

hao1st: we might have bug to release client for error handling...

PR #2423, opened 2019-05-06
DHFPROD-2126: Added custom hook to the online-store project

rjrudin:
Also fixed a bug in hub-utils where "user" was being passed instead of "userId".


examples/online-store/build.gradle
```diff
@@ -5,7 +5,7 @@ buildscript {
         mavenLocal()
     }
     dependencies {
-        classpath("com.marklogic:ml-data-hub:${mlDHFVersion}")
+        classpath("com.marklogic:ml-data-hub:5.0-SNAPSHOT")
```
rjrudin: I wasn't sure what to put here, as in order for the custom hook to work, 5.0-SNAPSHOT needs to be published locally so that the fix to hub-utils.sjs is available.

examples/online-store/build.gradle
```diff
@@ -30,6 +26,21 @@ configurations {
 }
 
 dependencies {
-    mlcp "com.marklogic:mlcp:9.0.8.2"
+    mlcp "com.marklogic:mlcp:9.0.9"
     mlcp files("lib")
 }
+
+task loadOrders(type: com.marklogic.gradle.task.MlcpTask) {
```
rjrudin: Converted the mlcp script into a Gradle task.

examples/online-store/settings.gradle
```diff
@@ -1,12 +0,0 @@
-
```
rjrudin: I removed this because I didn't see why it was needed, and mlGradleVersion isn't defined anywhere.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/hub-utils.sjs
```diff
@@ -89,7 +89,7 @@ class HubUtils {
     xdmp.invoke(moduleUri, parameters, {
       ignoreAmps: true,
       database: database ? xdmp.database(database): xdmp.database(),
-      user: xdmp.user(user)
+      userId: xdmp.user(user)
```
rjrudin: This fix is needed for a custom hook to work.

rjrudin:
The Gradle task "loadOrders" runs fine. If you run in a second time, the previously inserted Order documents are all archived. So the custom hook works fine there.

But the flow doesn't yet work in QuickStart - presumably because it's missing paths to the files to ingest? I wasn't sure how best to address this, or even if it should be addressed.

rjrudin:
Also note that for testing this locally, mlDHFVersion needs to be 5.0-SNAPSHOT, and you'll need to publish that locally so that you get the fix for hub-utils.sjs.

aebadirad:
> 
> 
> The Gradle task "loadOrders" runs fine. If you run in a second time, the previously inserted Order documents are all archived. So the custom hook works fine there.
> 
> But the flow doesn't yet work in QuickStart - presumably because it's missing paths to the files to ingest? I wasn't sure how best to address this, or even if it should be addressed.

What's the error you're running into?

examples/online-store/step-definitions/custom/IngestOrders/IngestOrders.step.json
```diff
@@ -1,9 +1,25 @@
 {
+  "language" : "zxx",
   "name" : "IngestOrders",
+  "description" : null,
   "type" : "CUSTOM",
```
ryanjdew: Changing the type to "ingestion" here and in the flow might resolve the issue with LoadOrders not working through the QuickStart UI.

aebadirad: Yeah, this should cause a pull to happen when ran. If its of type ingest, it'll 'pull' when 'ran' from DHF.

rjrudin: @ryanjdew Will I also need to define the files to ingest? Currently those are defined in the MlcpTask in build.gradle. I'm not clear on how to define a flow and its files to load that will work via both QS and MlcpTask.

ryanjdew: @rjrudin The MLCP task will ignore any file input information in the step/step definition. To the step in the flow, you can information that looks like this, so it will run in QuickStart (fingers crossed) https://github.com/marklogic/marklogic-data-hub/blob/develop/marklogic-data-hub/src/test/resources/master-test/flows/myNewFlow.flow.json#L11

rjrudin:
Closing this for now, going to rework this to see if I can get a custom hook working from QS first.

PR #2424, opened 2019-05-06
DHFPROD-2430: Fix to runBefore set false, also consider step before s…

aebadirad:
…tepdefinition in custom hook setting

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/flow.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/flow.sjs
@@ -265,7 +265,10 @@ class Flow {
     }
 
       let hookOperation = function() {};
-      let hook = processor.customHook;
+      let hook = step.customHook;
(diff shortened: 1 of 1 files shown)


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/flow.sjs
```diff
@@ -265,15 +265,18 @@ class Flow {
     }
 
       let hookOperation = function() {};
-      let hook = processor.customHook;
+      let hook = step.customHook;
+      if(!hook || !hook.module){
+        hook = processor.customHook;
+      }
       if (hook && hook.module) {
         let parameters = Object.assign({uris}, processor.customHook.parameters);
```
ryanjdew:
```suggestion
        let parameters = Object.assign({uris}, hook.parameters);
```

aebadirad: Ah, great catch. Yes.

PR #2427, opened 2019-05-06
DHFPROD-2432: fixes for the build

aebadirad:


web/src/main/ui/app/components/header/ui/header-ui.component.html
```diff
@@ -17,8 +17,8 @@
       <a id="browser-tab" class="mdl-navigation__link" [ngClass]="{'active': isActive('/browse')}" [routerLink]="['browse']"><i class="mdi mdi-magnify"></i> Browse Data</a>
       <a id="jobs-tab" class="mdl-navigation__link" [ngClass]="{'active': isActive('/jobs')}" [routerLink]="['jobs']"><i class="fa fa-tasks"></i> <span *ngIf="runningJobs !== 0" [mdl-badge]="runningJobs">Jobs</span><span *ngIf="runningJobs === 0">Jobs</span></a>
       <a id="settings-tab" class="mdl-navigation__link" [ngClass]="{'active': isActive('/settings')}" [routerLink]="['settings']"><i class="fa fa-gear"></i> Settings</a>
-      <mdl-button id="header-menu" #menuButton="mdlButton" (click)="headerMenu.toggle($event, menuButton)" mdl-ripple title="{{settings.mlUsername}}">
-        <i class="mdi mdi-account mdl-color--white"></i> <span class="user-name">{{settings.mlUsername | truncate: 11 }}</span>
+      <mdl-button id="header-menu" #menuButton="mdlButton" (click)="headerMenu.toggle($event, menuButton)" mdl-ripple title="{{settings.username}}">
+        <i class="mdi mdi-account mdl-color--white"></i> <span class="user-name">{{settings.username | truncate: 11 }}</span>
```
alexsmr: the endpoint is actually provided mlUsername:

image

So, either the HubSettings 'username' property must be changed or additional `this.settings.username = json.mlSettings.mlUsername` must be used in the service.

bsrikan: Can @alexsmr @brucean52 @wooldridge contribute to the branch DHFPROD-2432 to implement the change requested. Its a blocker now to test any bugs in UI since buildUI is broken.

web/src/main/ui/app/models/hub-settings.model.ts
```diff
@@ -3,6 +3,7 @@ export class HubSettings {
   name: string = 'data-hub';
 
   projectDir: string = null;
+  dhfversion: string = null;
```
ayuwono: camel case for consistency with the other props: dhfVersion

wooldridge:
Added fix for username reference. The fix @ayuwono is requesting is a backend change (dhfversion -> dhfVersion) that will result in the need for a frontend update.

bsrikan:
@alexsmr I am guessing you tested on 4200.. Did you try building the UI

alexsmr:
I ran a local 'ng build'. It was OK. But you're right: 'ng build --prod' if failing there.

web/src/main/ui/app/components/header/ui/header-ui.component.html
bsrikan: @alexsmr are you sure that the mlUsername in the snapshot you provided is not from gradle.properties. I was able to build OK, with just using username instead of mlUsername. I could not find any usage of mlUsername for web anywhere.

aebadirad:
thx @alexsmr

PR #2428, opened 2019-05-06
DHFPROD-2431: checking to see if sample doc has the envelope, if not,…

aebadirad:
… go from root


aebadirad:
Oh, yes, it'll require a redeploy. It's a core e-node code change

PR #2431, opened 2019-05-06
DHFPROD-2435: Accurate write tx and dateTime

ryanjdew:
write functions now return information about the transaction and it is stored in the job batch document.

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/jobs.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/jobs.sjs
@@ -154,10 +154,10 @@ class Jobs {
         timeStarted:  fn.currentDateTime(),
         timeEnded: "N/A",
         hostName: xdmp.hostName(),
-        reqTimeStamp: xdmp.requestTimestamp(),
+        reqTimeStamp: xdmp.timestampToWallclock(xdmp.requestTimestamp()),
         reqTrnxID: xdmp.transaction(),
-        writeTimeStamp: fn.currentDateTime(),
-        writeTrnxID: xdmp.transaction(),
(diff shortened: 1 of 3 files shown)


aebadirad:
@ryanjdew huh, we picked up a test failure

ryanjdew:
@aebadirad Isn't that failure related to https://github.com/marklogic/marklogic-data-hub/pull/2427?

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/jobs.sjs
```diff
@@ -211,6 +211,8 @@ class Jobs {
       }
       docObj.batch.error = `${error.name || error.code}: ${error.message}`;
     }
+    docObj.batch.writeTrnxID = writeTransactionInfo.transaction;
+    docObj.batch.writeTimeStamp = writeTransactionInfo.dateTime;
```
aebadirad: is this wallclock time too?

ryanjdew: no

aebadirad: Could you make it a wallclock timestamp?

We should be consistent with the timestamps in the docs

ryanjdew: sorry, I misread. It is wallclock. It uses fn.currentDateTime, since xdmp.requestTimestamp returns an empty sequence in update transactions.

aebadirad: ah right.

PR #2433, opened 2019-05-06
DHFPROD-2422: Fix NPE for step-def type

akshaysonvane:


aebadirad:
Should we be doing this, or should we be throwing an error back if there's no type being set?

akshaysonvane:
The error "Invalid step type" is currently being thrown. This just allows the stepResponse object to not be null and thus show the o/p with all 0's for counts.

aebadirad:
Fair enough.

PR #2434, opened 2019-05-07
DHFPROD-2126: Added project for demonstrating a custom hook

rjrudin:
Also fixed a bug in hub-utils.sjs and added to the amount of context passed to a custom hook.


rjrudin:
I figured it's easier to add a new project than trying to upgrade online-store, which still has some DHF 2 stuff in it.

examples/dhf5-custom-hook/build.gradle
```diff
@@ -0,0 +1,47 @@
...
+}
+plugins {
+    id 'net.saliman.properties' version '1.4.6'
+}
+
+apply plugin: 'com.marklogic.ml-data-hub'
+
+repositories {
+    jcenter()
+    maven { url "http://developer.marklogic.com/maven2/" }
```
aebadirad: we should probably remove the internal, svn only accessible repo from public consumption

aebadirad:
@rahulvudutala what's with the failing 30 tests?

examples/dhf5-custom-hook/build.gradle
rjrudin: That one is public - though I don't think it's needed, I'll check it out. 

Also going to do some performance testing of this, so don't merge this quite yet. Want to see if there's any noticeable impact with passing a bunch of context to the custom hook.

PR #2435, opened 2019-05-07
DHFPROD-2443: Unpath mastering timestamp from doc root

ryanjdew:


ryanjdew:
@wooldridge Ran through it again this morning and I'm getting the 2019 date and tracing is showing the date being sorted on. Can you verify that the changes were deployed to the modules on your machine?

PR #2440, opened 2019-05-07
DHFPROD-2370/DHFPROD-2307 - Update per Ganesh's correction

eltesoro-ml:


bsrikan:
We mention in the docs to use 8004 to bootstrap DHF (running hubInstallModules) 
gradle.properties later in the docs shows port 8011. So should we have a gradle.properties with port 8004 for bootstrapping and make it explicit in the docs?

PR #2442, opened 2019-05-07
DHFPROD-2423: Modified entity e2e tests and page object.

timur-isangulov:
Please don't merge yet, I'd like to run a few times on Jenkins.


timur-isangulov:
Can you please approve it. I tested it locally and on the Jenkins server. Thanks.

PR #2444, opened 2019-05-07
DHFPROD-2427: Added namespace and namespacePrefix to entity model

rjrudin:
I tested this out manually - just needed the namespace/namespacePrefix properties in the various model classes.


rjrudin:
@wooldridge I added the properties to entity.model.ts, though FWIW, they didn't appear necessary in the testing I did, which was:

- Load up an entity definition with namespace and namespacePrefix defined
- Edit the entity definition it in the entities editor
- Verify that the file on disk (and in ML) still has namespace and namespacePrefix

PR #2445, opened 2019-05-07
STORYBOOK-fix: updating storybook versions and fixing the build

alexsmr:


alexsmr:
@ayuwono looks like there is a compilation error in e2e tests:

image

PR #2446, opened 2019-05-07
DHFPROD-2395: Firefox browser issue: Entities not showing up in the GUI

brucean52:


aebadirad:
Test may need fixing (Error: Timeout - Async callback was not invoked within timeout specified by jasmine.DEFAULT_TIMEOUT_INTERVAL. si the response) but this works, so bringing this in. @ayuwono if you could see why this is the cause for the test, that'd be neat :)

PR #2447, opened 2019-05-07
DHFPROD-2313: sort by alphabetical order for folder and files

aebadirad:


aebadirad:
Error: Timeout - Async callback was not invoked within timeout specified by jasmine.DEFAULT_TIMEOUT_INTERVAL.

on the failure, so I'm marching it on, we can fix the isolated test that has nothing to do w/ this

PR #2452, opened 2019-05-08
DHFPROD-2453: Don't let null through as collection

ryanjdew:


ryanjdew:
@bsrikan Missed that the e-node doesn't have control over ingest collections. Updated the middle-tier to filter out nulls.

bsrikan:
On it now.

clockworked247:
@ryanjdew Can we remove the UI changes to edit-flow.component.ts?  The correct change lives in my PR: 
https://github.com/marklogic/marklogic-data-hub/pull/2457

PR #2455, opened 2019-05-08
DHFPROD-2407: Adding missing data to default flow template

rjrudin:
Also ensuring that config.tuning exists before checking for maxScan. 

Note that this does not yet generate a mapping file. I believe the current expectation, based on the warning message generated by hubCreateFlow, is that the user must either provide a mapping file or removing the mapping step.

Also - @akshaysonvane is putting together a separate PR for allowing for steps with a name starting with "default-" to work properly. I skirted around that issue by modifying the default flow template so that no steps have a name starting with "default-".

### DIFF
--- a/ml-data-hub-plugin/src/main/groovy/com/marklogic/gradle/task/CreateFlowTask.groovy
+++ b/ml-data-hub-plugin/src/main/groovy/com/marklogic/gradle/task/CreateFlowTask.groovy
@@ -32,7 +32,7 @@ class CreateFlowTask extends HubTask {
                 scaffolding.createDefaultFlow(flowName.toString())
 
                 println "IMPORTANT: Your new flow configuration file \"flows/" + flowName + ".flow.json\" contains step templates with " +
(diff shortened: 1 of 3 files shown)


ml-data-hub-plugin/src/main/groovy/com/marklogic/gradle/task/CreateFlowTask.groovy
```diff
@@ -32,7 +32,7 @@ class CreateFlowTask extends HubTask {
                 scaffolding.createDefaultFlow(flowName.toString())
 
                 println "IMPORTANT: Your new flow configuration file \"flows/" + flowName + ".flow.json\" contains step templates with " +
-                    "example values. The flow will not run as is. " +
+                    "example values, such as 'inputFilePath' and 'entity-name'. The flow will not run as is. " +
```
rjrudin: I added this to give the user some clue as to what needs to be changed.

PR #2457, opened 2019-05-08
DHFPROD-2453: Null step collection fix

clockworked247:


clockworked247:
Payload no longer passes a null collection:
https://drive.google.com/file/d/1NleAPSBa3mB2u41Ul0fFyyOxxoqwIZWF/view

aebadirad:
kicking this through so @bsrikan can do a holistic QA on develop

clockworked247:
Thanks

PR #2460, opened 2019-05-08
DHFPROD-2399: fix to browse pagination when going back/forth

clockworked247:
Pagination when going from the document details to browser page list fixed.  Page # now resets to first page or results, which is the same page as being shown in the list.


clockworked247:
Fix in action:  https://drive.google.com/file/d/10lkAf0K6qVkzGApLt-RpeiLb616WuNt1/view

aebadirad:
I love these demo videos

PR #2461, opened 2019-05-08
DHFPROD-2407: Handle a flow with a non-existent target entity

wooldridge:
Gradle-generated flows may have a non-existent target entity set for mapping steps.
Check for this in the mapping step component and display appropriate messaging in UI.


wooldridge:
Attaching a test flow: [gradle-flow2.flow.json.zip](https://github.com/marklogic/marklogic-data-hub/files/3159398/gradle-flow2.flow.json.zip)

Test scenario:

1. Put above flow in project/flows directory.
2. Load Manage Flows view in UI, drill down to gradle-flow2, select mapping step.
3. Mapping step has a non-existent target entity set so the following messaging is displayed:
```
Before configuring this mapping you must set a target entity. You can create new entities in the Entities view.
```
4. To get the mapping step working, you can associate an entity with the mapping step and also get source documents loaded (as required previously).

PR #2468, opened 2019-05-09
DHFPROD-2461: Control when we should reset the latest job info and wh…

hao1st:
…en to refresh job from DB

### DIFF
--- a/web/src/main/java/com/marklogic/hub/web/service/FlowJobService.java
+++ b/web/src/main/java/com/marklogic/hub/web/service/FlowJobService.java
@@ -89,7 +89,7 @@ public class FlowJobService extends ResourceManager {
 
         ServiceResultIterator resultItr = this.getServices().get(params);
         if (resultItr == null || !resultItr.hasNext()) {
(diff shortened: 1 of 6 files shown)


web/src/main/java/com/marklogic/hub/web/service/FlowJobService.java
```diff
@@ -56,127 +55,132 @@ public FlowJobService() {
...
+        }
+        client.init(ML_JOBS_NAME, this);
+        RequestParameters params = new RequestParameters();
+        if (StringUtils.isNotEmpty(flowName)) {
+            params.add("flow-name", flowName);
+        }
+
+        ServiceResultIterator resultItr = this.getServices().get(params);
+        if (resultItr == null || !resultItr.hasNext()) {
+            throw new RuntimeException("Unable to get job document");
```
rjrudin: I recommend tossing the flow name into this message (would it really ever be null, given the method name?) - e.g. "No jobs found for flow with name: " + flowName

PR #2470, opened 2019-05-09
DHFPROD-2461: better control multiple flow jobs running at the same time

hao1st:


marklogic-data-hub/src/main/java/com/marklogic/hub/step/impl/WriteStepRunner.java
```diff
@@ -405,16 +405,14 @@ private RunStepResponse runIngester(RunStepResponse runStepResponse, Collection<
...
+                } catch (Exception e) {
+                    throw e;
+                }
                 return StepRunnerUtil.getResponse(jobDoc, step);
-            }
-            catch (Exception ex)
-            {
+            } catch (Exception ex) {
+            } finally {
                 return runStepResponse;
```
srinathgit: Why is return runStepResponse moved to finally block here ?

hao1st: i saw the stepResponses missing some value like flowName for RunFlowResponse object for the case. It is just data completion problem when i debugged.

aebadirad:
Merging to consolidate

PR #2476, opened 2019-05-09

rjrudin:
Added marklogic-unit-test as part of this. Will call a meeting - probably post-MLW - to go over this with the team and how we can start using it.

### DIFF
--- a/marklogic-data-hub/build.gradle
+++ b/marklogic-data-hub/build.gradle
@@ -62,7 +62,12 @@ dependencies {
       exclude group: 'org.apache.avro', module: 'avro-tools'
       exclude group: 'org.apache.commons', module: 'commons-csv'
     }
+
+    // Provides support for invoking marklogic-unit-test tests via JUnit
+    testCompile "com.marklogic:marklogic-junit:0.11.0"
+    mlRestApi "com.marklogic:marklogic-unit-test-modules:0.12.0"
 }
+
 import com.marklogic.mgmt.ManageClient
 import com.marklogic.mgmt.resource.hosts.HostManager
 
@@ -262,14 +267,10 @@ ext {
         customTokens.put("%%mlModulesDbName%%", mlModulesDbName)
         customTokens.put("%%mlStagingTriggersDbName%%", mlStagingTriggersDbName)
         customTokens.put("%%mlStagingSchemasDbName%%", mlStagingSchemasDbName)
-
-        modulePaths = ["marklogic-data-hub/src/main/resources/ml-modules"]
-
     }
 }
 
 def propFile = file("gradle.properties")
-String fileContents = propFile.text
 boolean sslRun = Boolean.parseBoolean(System.properties['ssl'])
 boolean certAuth = Boolean.parseBoolean(System.properties['certauth'])
 def props = new Properties()
@@ -363,6 +364,10 @@ task setupSSL{
 
 setupSSL.dependsOn(testClasses)
 bootstrap.dependsOn(setupSSL)
+
+// The bootstrap process must load modules so that the marklogic-unit-test modules are loaded
+bootstrap.finalizedBy mlLoadModules
+
 test.dependsOn(bootstrap)
 testIntegration.dependsOn(bootstrap)
 testAcceptance.dependsOn(bootstrap)
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/impl/DbConfigsManager.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/impl/DbConfigsManager.java
@@ -1,0 +1,42 @@
+package com.marklogic.hub.impl;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import com.marklogic.client.DatabaseClient;
+import com.marklogic.client.extensions.ResourceManager;
+import com.marklogic.client.extensions.ResourceServices;
+import com.marklogic.client.io.JacksonHandle;
+import com.marklogic.client.util.RequestParameters;
+
+import java.io.IOException;
+import java.util.List;
+
+class DbConfigsManager extends ResourceManager {
+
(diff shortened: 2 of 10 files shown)


marklogic-data-hub/build.gradle
```diff
@@ -262,14 +262,10 @@ ext {
         customTokens.put("%%mlModulesDbName%%", mlModulesDbName)
         customTokens.put("%%mlStagingTriggersDbName%%", mlStagingTriggersDbName)
         customTokens.put("%%mlStagingSchemasDbName%%", mlStagingSchemasDbName)
-
-        modulePaths = ["marklogic-data-hub/src/main/resources/ml-modules"]
```
rjrudin: Moved this into gradle.properties

marklogic-data-hub/build.gradle
```diff
@@ -262,14 +262,10 @@ ext {
...
-
     }
 }
 
 def propFile = file("gradle.properties")
-String fileContents = propFile.text
```
rjrudin: Unused var

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/DbConfigsManager.java
```diff
@@ -0,0 +1,42 @@
...
+import com.marklogic.client.DatabaseClient;
+import com.marklogic.client.extensions.ResourceManager;
+import com.marklogic.client.extensions.ResourceServices;
+import com.marklogic.client.io.JacksonHandle;
+import com.marklogic.client.util.RequestParameters;
+
+import java.io.IOException;
+import java.util.List;
+
+class DbConfigsManager extends ResourceManager {
```
rjrudin: Extracted this from EntityManagerImpl so it can be reused easily in a test

marklogic-data-hub/src/test/java/com/marklogic/hub/impl/GenerateIndexesTest.java
```diff
@@ -0,0 +1,142 @@
...
+    private static final String CODEPOINT = "http://marklogic.com/collation/codepoint";
+    private static final String ROOT = "http://marklogic.com/collation/";
+
+    private List<JsonNode> entities = new ArrayList();
+    private ObjectNode indexes;
+
+    @Test
+    public void sharedPropertyWithNullNamespace() {
+        final String namespace = null;
+        givenAnEntityWithTitleProperty("Book", namespace, CODEPOINT);
```
rjrudin: I often like using given/when/then methods in a JUnit test to mirror what ideally will be in the JIRA ticket.

rjrudin:
I'm making a fix so that the test modules are loaded via "gradle bootstrap".

rjrudin:
"gradle bootstrap" will now deploy the test modules, the lack of which was causing the Jenkins job to fail.

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/EntityManagerImpl.java
```diff
@@ -150,12 +150,11 @@ public boolean saveDbIndexes() {
             List<JsonNode> entities = getAllEntities();
 
             if (entities.size() > 0) {
-                DbIndexGenerator generator = new DbIndexGenerator(hubConfig.newReverseFlowClient());
-                String indexes = generator.getIndexes(entities);
+                DbConfigsManager generator = new DbConfigsManager(hubConfig.newReverseFlowClient());
+                ObjectNode indexNode = generator.generateIndexes(entities);
```
hao1st: i think that we should call release() in a finally block for the database client object.

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/DbConfigsManager.java
```diff
@@ -0,0 +1,42 @@
...
+    private static final String NAME = "ml:dbConfigs";
+
+    private RequestParameters params = new RequestParameters();
+
+    public DbConfigsManager(DatabaseClient client) {
+        super();
+        client.init(NAME, this);
+    }
+
+    public ObjectNode generateIndexes(List<JsonNode> entities) {
```
hao1st: could it be better if we add throws exception in the method declaration...

rjrudin: Yeah I left that in place, not knowing why it was there. Currently, if an error occurs, no exception is thrown, but rather an empty object node is returned. That doesn't break the client. I didn't want to change this to throw an error since it's out of scope of this JIRA ticket, and I didn't want to sort through the ramifications.

marklogic-data-hub/src/test/ml-modules/root/test/suites/com.marklogic.hub.impl/generate-indexes.sjs
```diff
@@ -0,0 +1,124 @@
+const test = require("/test/test-helper.xqy");
```
jamesagardner: With marklogic-unit-test version 1.0 you don't need to have a `com.marklogic.hub.impl` directory, you can actually use directories to organize your test suites.

Not sure if you want to reproduce the entire `com/marklogic/hub/impl directory` structure in `test/suites`, it may be fine to start with `hub/impl` to see how that goes.

marklogic-data-hub/build.gradle
```diff
@@ -62,7 +62,12 @@ dependencies {
       exclude group: 'org.apache.avro', module: 'avro-tools'
       exclude group: 'org.apache.commons', module: 'commons-csv'
     }
+
+    // Provides support for invoking marklogic-unit-test tests via JUnit
+    testCompile "com.marklogic:marklogic-junit:0.11.0"
+    mlRestApi "com.marklogic:marklogic-unit-test-modules:0.12.0"
```
jamesagardner: You should give version 1.0.beta a spin.

rjrudin:
Will resubmit this with marklogic-unit-test 1.0.beta in action. Also still having some issues with getting all the tests running - DataHubInstallTest re-installs the app but doesn't load the test modules.

PR #2480, opened 2019-05-09
DHFPROD-2501: If 'uris'is null and job is stopped return the response immediately

srinathgit:
If 'uris'is null and job is stopped return the response immediately. We can run this on win/linux cluster and ensure tests are running in cluster as well before merging


srinathgit:
Thanks @hao1st  for catching it

PR #2491, opened 2019-05-13
DHFPROD-2500: Fix XML attributes in mapping

ryanjdew:


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mapping/default/lib.sjs
```diff
@@ -35,7 +35,9 @@ function getSourceContext(sourceContext) {
 
 function getPath(sourceContext, connector, propertyName) {
   let path;
-  if (xdmp.castableAs("http://www.w3.org/2001/XMLSchema", "NCName", propertyName)) {
+  // validExtractPath will recognize complex XPath, like attributes and filtered steps
+  // if not a validExtractPath, likely a JSON property with XPath incompatible name
+  if (cts.validExtractPath(propertyName)) {
```
rjrudin: I think some marklogic-unit-test tests would be easy to write here and would serve as better documentation of what this is addressing.

I'm fixing an issue over on my 2496 branch where the unit-test modules weren't being loaded via "gradle bootstrap" - will have that fixed shortly.

PR #2499, opened 2019-05-13
DHFPROD-2499: hubCreateMapping task doesn't populate target entity pr…

ryanjdew:
…operly

### DIFF
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/EntityManager.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/EntityManager.java
@@ -60,6 +60,10 @@ public interface EntityManager {
 
     boolean deployStagingQueryOptions();
 
+    HubEntity getEntity(String entityName);
+
+    HubEntity getEntity(String entityName, String version);
+
     List<HubEntity> getEntities();
 
     HubEntity saveEntity(HubEntity entity, Boolean rename) throws IOException;
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/EntityManager.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/EntityManager.java
@@ -60,9 +60,9 @@ public interface EntityManager {
 
     boolean deployStagingQueryOptions();
 
-    HubEntity getEntity(String entityName);
+    HubEntity getEntityFromProject(String entityName);
 
-    HubEntity getEntity(String entityName, String version);
+    HubEntity getEntityFromProject(String entityName, String version);
 
     List<HubEntity> getEntities();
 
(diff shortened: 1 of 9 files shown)


marklogic-data-hub/src/main/java/com/marklogic/hub/impl/EntityManagerImpl.java
```diff
@@ -296,6 +297,22 @@ private HubModuleManager getPropsMgr() {
...
 
+    public HubEntity getEntity(String entityName) {
+        return getEntity(entityName, null);
+    }
+
+    public HubEntity getEntity(String entityName, String version) {
+        HubEntity entity = null;
+        for (HubEntity e: getEntities()) {
+            InfoType info = e.getInfo();
+            if (info.getTitle().equals(entityName) && (version == null || info.getVersion().equals(version))) {
```
hao1st: info.getTitle() can be null?

marklogic-data-hub/src/main/java/com/marklogic/hub/EntityManager.java
```diff
@@ -60,6 +60,10 @@
 
     boolean deployStagingQueryOptions();
 
+    HubEntity getEntity(String entityName);
+
+    HubEntity getEntity(String entityName, String version);
```
rjrudin: One thing that's not obvious to me from a number of DHF methods is whether the entity is being retrieved from the project files or from the ML database. That's not specific to these methods, but I'm wondering if we could standardize on verbs that imply whether the data is retrieved from project files or a database. Or just make the method names more verbose - i.e. "getEntityFromProject" and "getEntityFromDatabase".

rjrudin: I just noticed this calls "getEntities", which reads from the project. I think "getEntityFromProject" would be a good choice, with the notion that the "project" is a data store separate from the ML database.

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/EntityManagerImpl.java
ryanjdew: Title is a required field for entities, so it should never be null.

marklogic-data-hub/src/main/java/com/marklogic/hub/EntityManager.java
ryanjdew: @rjrudin I believe that all of our managers read items from the project. We do make calls to the server to generate some of our items, like TDEs. I do think that could be a good practice to make it clear. Since these are new functions it would be a good place to start. I'll leave existing function signatures as is.

marklogic-data-hub/src/main/java/com/marklogic/hub/mapping/MappingImpl.java
```diff
@@ -49,6 +53,25 @@ public MappingImpl(String name) {
         this.targetEntityType = "http://example.org/modelName-version/entityType";
     }
 
+    public MappingImpl(String name, HubEntity entity) {
```
rjrudin: I think a unit test would be useful here and fairly easy to write as well, as there's what looks like some important logic. 

The test could also verify that the new "getEntityFromProject" method works as expected.

Issue #2493, opened 2019-05-13
DHF 5.0.0: Jobs throws 500


prestonmcgowan:
#### The issue
While running a flow, if I select the Jobs tab or if I navigate from the Flows tab by clicking the "running" link, the Jobs network call throws a 500 http error.

What behavior are you expecting?
I did not expect a 500.

#### Tech details

Which Operating System are you using?
MacOS 10.14.4 (18E226)

Which version of MarkLogic are you using?
9.0-9

Which version of the Data Hub Framework are you using?
5.0.0

Is this a QuickStart UI bug? If so, which browser are you using?
I don't think so

Adding HAR output for the network calls and a GIF of the issue

[BUG - jobs not loading_8053_ErrorLog.txt](https://github.com/marklogic/marklogic-data-hub/files/3174192/BUG.-.jobs.not.loading_8053_ErrorLog.txt)

[BUG - jobs not loading.har.txt](https://github.com/marklogic/marklogic-data-hub/files/3174193/BUG.-.jobs.not.loading.har.txt)

![BUG - jobs not loading](https://user-images.githubusercontent.com/1315891/57643445-c0f7a280-7587-11e9-9667-3372b6df7d7a.gif)

ayuwono:
Tracking on JIRA: https://project.marklogic.com/jira/browse/DHFPROD-2523

ayuwono:
Fixed on 5.0.1

PR #2500, opened 2019-05-13
DHFPROD-2523: Clicking Jobs link while running the flows produces 500…

hao1st:
… internal error

https://github.com/marklogic/marklogic-data-hub/issues/2493
https://project.marklogic.com/jira/browse/DHFPROD-2523

### DIFF
--- a/web/src/main/java/com/marklogic/hub/web/service/NewJobService.java
+++ b/web/src/main/java/com/marklogic/hub/web/service/NewJobService.java
@@ -228,16 +228,16 @@ public class NewJobService extends ResourceManager {
             Optional.ofNullable(step.getOptions()).ifPresent(o -> js.targetDatabase = ((TextNode) o.getOrDefault("targetDatabase", new TextNode(""))).asText());
 
             JsonNode res = stepRes.get(key);
-            js.totalEvents = res.get("totalEvents") != null ? res.get("totalEvents").asLong() : 0;
-            js.successfulEvents = res.get("successfulEvents") != null ? res.get("successfulEvents").asLong() : 0;
-            js.failedEvents = res.get("failedEvents") != null ? res.get("failedEvents").asLong() : 0;
+            js.totalEvents = getLong(res,"totalEvents");
+            js.successfulEvents = getLong(res, "successfulEvents");
+            js.failedEvents = getLong(res,"failedEvents");
 
             counters[0] += js.successfulEvents;
             counters[1] += js.failedEvents;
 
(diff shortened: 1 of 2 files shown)


web/src/main/java/com/marklogic/hub/web/service/NewJobService.java
```diff
@@ -228,16 +228,16 @@ private void createJobs(List<JobModel> jobModels, JSONObject jobJson) {
             Optional.ofNullable(step.getOptions()).ifPresent(o -> js.targetDatabase = ((TextNode) o.getOrDefault("targetDatabase", new TextNode(""))).asText());
 
             JsonNode res = stepRes.get(key);
-            js.totalEvents = res.get("totalEvents").asLong();
-            js.successfulEvents = res.get("successfulEvents").asLong();
-            js.failedEvents = res.get("failedEvents").asLong();
+            js.totalEvents = res.get("totalEvents") != null ? res.get("totalEvents").asLong() : 0;
```
rjrudin: I recommend using:

    res.has("totalEvents")

The check for a long value is repeated enough times that a separate method to handle it would be useful - e.g. private long getLong(JsonNode node, String propertyName).

web/src/main/java/com/marklogic/hub/web/web/NewJobController.java
```diff
@@ -20,18 +21,28 @@
...
     @ResponseBody
     public ResponseEntity<?> getJobs(@RequestParam(value = "flowName", required = false) String flowName) throws IOException {
-        jobService.setupClient();
-        List<JobModel> jobModels = jobService.getJobs(flowName);
-        jobService.release();
+        List<JobModel> jobModels;
+        try {
+            jobService.setupClient();
+            jobModels = jobService.getJobs(flowName);
+            jobService.release();
```
rjrudin: I know @ehennum has mentioned before that there's some overhead with construction of a DatabaseClient, and since it's a thread-safe object, it's ideal for performance if it can be instantiated once and used many times. 

Since NewJobService is a singleton object managed by Spring, is there any risk to having it instantiate a new Jobs DatabaseClient in a PostConstruct-annotated method on startup? The only risk I can see is if the config in HubConfigImpl changed... perhaps that's possible when a new project is selected? Though ideally, that would refresh the Spring container with the new config.

web/src/main/java/com/marklogic/hub/web/service/NewJobService.java
hao1st: OK, agree.  @rjrudin👍

PR #2511, opened 2019-05-14
DHFPROD-2530: Remove 'CMASettings' as 5.x is supported on versions > 9.0-7

srinathgit:
Remove 'CMASettings' as 5.x is supported on versions > 9.0-7


rjrudin:
Looks good - always nice to delete code! Just FYI, in ml-app-deployer 3.15.0, I'm going to default to using CMA for all supported resources.

PR #2516, opened 2019-05-16

rjrudin:
Also added marklogic-unit-test. Had to add support for loading test modules during the test suite because other tests were deleting them since they're not in the "hub-core-module" collection.

Ran all the tests locally - was getting the same failures as Jenkins (due to modules being modified by some tests), but now they're all passing.


rjrudin:
> maybe we could change src/test/ml-modules => src/test/resources/ml-modulues

"src/test/ml-modules" is the standard location for modules in a marklogic-unit-test project. I believe the reasoning behind "src/main/resources/ml-modules" is so that ml-modules is by default included in the QuickStart WAR file, as everything under src/main/resources is on the classpath.

rjrudin:
Ack - I forgot to upgrade marklogic-unit-test to 1.0.beta. Will resubmit this shortly.

PR #2519, opened 2019-05-16

rjrudin:
Also added marklogic-unit-test. Had to add support for loading test modules during the test suite because other tests were deleting them since they're not in the "hub-core-module" collection.

Upgraded to 1.0.beta of marklogic-unit-test.


rjrudin:
Argh, still having issues in Jenkins, RunMarkLogicUnitTestsTest is not finding tests to run. Debugging.

PR #2523, opened 2019-05-20
DHFPROD-2500: Add Unit test and avoid error in building attribute path

ryanjdew:


bsrikan:
Kicked off the unit test again. Last one seemed to be a random IO exception failure trying to delete .tmp directory
https://jenkins.marklogic.com/blue/organizations/jenkins/Datahub_CI/detail/PR-2523/1/pipeline/

PR #2524, opened 2019-05-20
DHFPROD-2526: A flow keeps running status when the Data-hub app serve…

hao1st:
…r is down during the running period

https://project.marklogic.com/jira/browse/DHFPROD-2526

### DIFF
--- a/web/src/main/java/com/marklogic/hub/web/service/FlowJobService.java
+++ b/web/src/main/java/com/marklogic/hub/web/service/FlowJobService.java
@@ -7,27 +7,31 @@ import com.marklogic.client.DatabaseClient;
 import com.marklogic.client.extensions.ResourceManager;
 import com.marklogic.client.extensions.ResourceServices.ServiceResult;
 import com.marklogic.client.extensions.ResourceServices.ServiceResultIterator;
+import com.marklogic.client.io.Format;
 import com.marklogic.client.io.JacksonHandle;
+import com.marklogic.client.io.StringHandle;
 import com.marklogic.client.util.RequestParameters;
 import com.marklogic.hub.FlowManager;
 import com.marklogic.hub.flow.Flow;
 import com.marklogic.hub.impl.HubConfigImpl;
+import com.marklogic.hub.job.JobStatus;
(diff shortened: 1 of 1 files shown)


web/src/main/java/com/marklogic/hub/web/service/FlowJobService.java
```diff
@@ -166,11 +179,33 @@ private FlowJobs retrieveJobsByFlowName(String flowName) {
...
+    private void updateJobStaleStates(List<String> staleStateJobIds) {
+        RequestParameters params = new RequestParameters();
+        for (String jobId : staleStateJobIds) {
+            params.add("jobid", jobId);
+            params.add("status", JobStatus.FAILED.toString());
+
+            try {
+                this.getServices().post(params, new StringHandle("{}").withFormat(Format.JSON));
+            } catch (Exception e) {
+                logger.error(e.getMessage());
```
aebadirad: Can we log the job-id explicitly that was associated with the error in addition to the exception?

web/src/main/java/com/marklogic/hub/web/service/FlowJobService.java
```diff
@@ -149,6 +153,15 @@ private FlowJobs retrieveJobsByFlowName(String flowName) {
         }
         latestJob.status = jobJson.getString("jobStatus");
 
+        List<String> staleStateJobIds = new ArrayList<>();
+        if (cachedJobsByFlowName.size() == 0) {
+            if (StringUtils.isNotEmpty(latestJob.status) && latestJob.status.startsWith(RUNNING_PREFIX) && StringUtils.isNotEmpty(latestJob.id)) {
+                //invalid state on the job database that may be caused by unexpected server shutdown or crashed
+                staleStateJobIds.add(latestJob.id);
+                latestJob.status = JobStatus.FAILED.toString();
```
aebadirad: Are we also considering the possibility that another orchestrator maybe running the job that isn't this client?

hao1st: Yeah. I only considered QS to recover. How can I detect it and work around ?

PR #2525, opened 2019-05-21
DHFPROD-2546: Can now install/upgrade DHF via an executable jar

rjrudin:
See the comments in Installer for how to build and run it. 

Also removed the "main" method from ApplicationConfig as I don't believe there's any use case for it.

### DIFF
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/cli/Installer.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/cli/Installer.java
@@ -1,0 +1,158 @@
+package com.marklogic.hub.cli;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ArrayNode;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import com.marklogic.appdeployer.ConfigDir;
+import com.marklogic.hub.ApplicationConfig;
+import com.marklogic.hub.DataHub;
+import com.marklogic.hub.impl.HubConfigImpl;
+import com.marklogic.mgmt.resource.security.UserManager;
+import com.marklogic.rest.util.ResourcesFragment;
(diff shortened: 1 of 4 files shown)


marklogic-data-hub/src/main/java/com/marklogic/hub/cli/Installer.java
```diff
@@ -0,0 +1,158 @@
...
+
+    /**
+     * If a user already exists, we don't want to overwrite its password.
+     */
+    protected void removePasswordsFromUserFiles(File projectDir, HubConfigImpl hubConfig) {
+        ResourcesFragment existingUsers = new UserManager(hubConfig.getManageClient()).getAsXml();
+        File dir = new File(projectDir, "src/main/hub-internal-config");
+        if (dir.exists()) {
+            final Map<String, String> tokens = hubConfig.getAppConfig().getCustomTokens();
+
```
hao1st:
```suggestion
if (tokens == null) {
   logger.warn(...)
   return;
}
```

marklogic-data-hub/src/main/java/com/marklogic/hub/cli/Installer.java
```diff
@@ -0,0 +1,158 @@
...
+            File usersDir = new ConfigDir(dir).getUsersDir();
+            if (usersDir != null && usersDir.exists()) {
```
hao1st: usersDir can be null ?

rjrudin: Good point - I'll remove that

marklogic-data-hub/src/main/java/com/marklogic/hub/cli/Installer.java
rjrudin: The token map wouldn't be null, as it's eagerly instantiated by AppConfig. AppConfig could be null, which would mean that the hubConfig.refresh call failed, and nothing's going to work, in which case an exception will naturally occur.

PR #2540, opened 2019-05-23
DHFPROD-2548: Throw error if entity is already present

akshaysonvane:


ayuwono:
@akshaysonvane , can you synch-up with UI dev so QS can have the same behavior? Either they can fix it here or we make a separate issue for QS

wooldridge:
Thanks, @ayuwono @akshaysonvane. I created a subtask for adding an appropriate validation fix in the QuickStart UI:

https://project.marklogic.com/jira/browse/DHFPROD-2573

akshaysonvane:
Let's get this in as we now have a separate task for the UI work.

PR #2547, opened 2019-05-23
DHFPROD-2324: Set source context correctly for XML

ryanjdew:


rjrudin:
Nice to see marklogic-unit-test! I'll try this out locally just to make sure the tests pass, particularly from within JUnit (I want to make sure that's reliable so folks trust it).

rjrudin:
Nice, tests passed, and both test suites were reported by RunMarkLogicUnitTestsTest.

PR #2552, opened 2019-05-23
Develop

browda:


build.gradle
```diff
@@ -13,8 +13,8 @@ allprojects {
 subprojects {
     apply plugin: 'java'
 
-    sourceCompatibility = "1.8"
-    targetCompatibility = "1.8"
+    sourceCompatibility = "1.11"
```
rjrudin: @browda @aebadirad I'm wondering if bumping this up constitutes a breaking change (as DHF code will no longer run on Java 8/9/10), which by the rules of semver would require the next release to be 6.x. Regardless of the status of Java 8, my hunch is a significant number of DHF customers are still on Java 8 and thus would be surprised that e.g. 5.0.1 or 5.1.x of DHF would require them to upgrade. I'm for the upgrade, I'm just wondering if it's safe to do in a bugfix or minor release.

browda:
Since this is an incorrectly-named pull request and the build fails in jenkins, there needs to be more work.  Also, @rjrudin makes an excellent point with respect to numbering and general support issues.

PR #2557, opened 2019-05-24
DHFPROD-2271: Some lags on loading flows and jobs if you have a lot of them

brucean52:
Added indeterminate progress bar.


brucean52:
@ayuwono @wooldridge Thanks for the feedback. I'll fix the bug empty flow and job page and try the spinner

brucean52:
@wooldridge One thing about the spinner component is there is some css for .spinner-container (border-bottom, background) that make it specific for within our tables. I don't think it adds value to make the values customizable. You can add the material spinner element by itself for other use cases.

I'll work on maker it smaller like the browse data view and add it to the job details view

PR #2565, opened 2019-05-28
DHFPROD-2555: Add collection for entity with mapping step

ryanjdew:
Also, adjusts the mastering step to set entity collection on the proper documents.

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/com.marklogic.smart-mastering/impl/process.xqy
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/com.marklogic.smart-mastering/impl/process.xqy
@@ -180,6 +180,7 @@ declare function proc-impl:process-match-and-merge-with-options(
         merge-impl:options-from-json($merge-options)
       else
         $merge-options
+  let $target-entity := $matching-options/matcher:target-entity ! fn:string(.)
   let $actions := fn:distinct-values(($matching-options/matcher:actions/matcher:action/@name ! fn:string(.), $const:MERGE-ACTION, $const:NOTIFY-ACTION))
   let $thresholds := $matching-options/matcher:thresholds/matcher:threshold[(@action|matcher:action) = $actions]
   let $threshold-labels := $thresholds/(@label|matcher:label)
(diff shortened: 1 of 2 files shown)


marklogic-data-hub/src/main/resources/ml-modules/root/com.marklogic.smart-mastering/impl/process.xqy
```diff
@@ -251,19 +252,22 @@ declare function proc-impl:process-match-and-merge-with-options(
...
-                    return
-                      map:entry(
-                        $uri,
-                        $write-object
-                          => map:get("context")
-                          => map:get("collections")
-                      )
-                  )),
-                  $on-merge-options
+                (
```
rjrudin: I looked at the function that this code is in - it's about 250 lines, and of course a very important one. I think when we change it, it'd be nice to extract smaller functions to make it easier to understand, and then we can write unit tests for those smaller functions. Perhaps this could be a "build-collections-for-merged-uri" function? Or "build-context-for-merged-uri", which would include the permissions too? Or "build-merge-result" for the entire map? (I like the verb "build" for anything that builds something in-memory but doesn't have save it (I like "create" when something is saved)).

marklogic-data-hub/src/main/resources/ml-modules/root/com.marklogic.smart-mastering/impl/process.xqy
```diff
@@ -335,9 +339,12 @@ declare function proc-impl:process-match-and-merge-with-options(
           => map:get("context")
         let $current-collections := $write-context
           => map:get("collections")
-        let $new-collections := coll-impl:on-no-match(
-          map:entry($uri, $current-collections),
-          $on-no-match
+        let $new-collections := (
```
rjrudin: And then "build-collections-for-merged-uri"? Could put this entire block into that function:

```
let $write-context := $write-object
          => map:get("context")
        let $current-collections := $write-context
          => map:get("collections")
        let $new-collections := coll-impl:on-no-match(
          map:entry($uri, $current-collections),
          $on-no-match
        )
        let $_update := $write-context
          => map:put("collections", $new-collections)
```

marklogic-data-hub/src/main/resources/ml-modules/root/com.marklogic.smart-mastering/impl/process.xqy
aebadirad: Can we handle that in a technical task ticket separate from this bug fix?

marklogic-data-hub/src/main/resources/ml-modules/root/com.marklogic.smart-mastering/impl/process.xqy
aebadirad: Let's make a new task to refactor this and unit test this.

PR #2566, opened 2019-05-28
DHFPROD-2529: Deploying protected paths when updating indexes

rjrudin:


examples/dhf5-custom-hook/build.gradle
```diff
@@ -1,19 +1,8 @@
-buildscript {
```
rjrudin: I used this project for ad hoc testing and updated it to reference 5.0.0, now that that's available.

marklogic-data-hub/src/test/java/com/marklogic/hub/impl/DeployProtectedPathsWhenUpdatingIndexesTest.java
```diff
@@ -0,0 +1,68 @@
...
+import java.io.IOException;
+
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+@ExtendWith(SpringExtension.class)
+@ContextConfiguration(classes = ApplicationConfig.class)
+public class DeployProtectedPathsWhenUpdatingIndexesTest extends HubTestBase {
+
+    private final static String TEST_PATH_EXPRESSION = "/envelope//instance/SomeEntity/someProperty";
```
hao1st: specially reason with double // here?

rjrudin: That's how DHF builds a path when it generates a new protected path.

PR #2567, opened 2019-05-28
DHFPROD-2567: To run a custom step, user can't leave the source collection empty

brucean52:
Also truncated step name (DHFPROD-2514)

### DIFF
--- a/web/src/main/ui/app/components/flows-new/validators/new-step-dialog.validator.ts
+++ b/web/src/main/ui/app/components/flows-new/validators/new-step-dialog.validator.ts
@@ -11,8 +11,6 @@ export function NewStepDialogValidator(group: FormGroup) {
       if (!group.value.targetEntity) {
         errors['noTargetEntity'] = true;
       }
-      /* falls through */
-    case StepType.CUSTOM:
       if (
         (group.value.selectedSource === 'collection' && !group.value.sourceCollection) ||
         (group.value.selectedSource === 'query' && !group.value.sourceQuery) ||
--- a/web/src/main/ui/app/components/flows-new/validators/new-step-dialog.validator.ts
+++ b/web/src/main/ui/app/components/flows-new/validators/new-step-dialog.validator.ts
@@ -19,14 +19,6 @@ export function NewStepDialogValidator(group: FormGroup) {
         errors['noSource'] = true;
       }
       break;
-    case StepType.CUSTOM:
-      if (
-        (group.value.selectedSource === 'collection' && !group.value.sourceCollection) ||
-        (group.value.selectedSource === 'query' && !group.value.sourceQuery)
-      ) {
-        errors['noSource'] = true;
-      }
-      break;
     default:
       break;
   }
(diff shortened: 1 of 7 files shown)


web/src/main/ui/app/components/flows-new/validators/new-step-dialog.validator.ts
```diff
@@ -3,19 +3,24 @@ import { StepType } from '../models/step.model';
...
+  switch (group.value.stepDefinitionType) {
+    case StepType.MAPPING:
+    case StepType.MASTERING:
+      if (!group.value.targetEntity) {
+        errors['noTargetEntity'] = true;
+      }
+      if (
+        (group.value.selectedSource === 'collection' && !group.value.sourceCollection) ||
+        (group.value.selectedSource === 'query' && !group.value.sourceQuery) ||
+        (group.value.selectedSource === '')
```
clockworked247: As Aries mentioned, choosing a selected source isn't required.   But, if they do choose a source type, they must have a sourceCollection value with collection type and sourceQuery with query type.

clockworked247: so the statement

> (group.value.selectedSource === '')

should be removed

brucean52:
@clockworked247 @wooldridge I removed the source validation requirement for custom step and added logic to allow empty source collection/query. We are relying on the new step validators to not allow mapping and mastering to have empty source collection/query

brucean52:
@ayuwono Good catch. I added the condition for default source query if no radio selection is made

ayuwono:
> @brucean52 if I do the following:
> 
> 1. Create Custom step.
> 2. Select "Query" as the Source Type.
> 3. DO NOT specify a source query and save the step.
> 
> I do not see a (default) sourceQuery value in the Custom step artifact. Should there be?

@wooldridge , I'm seeing this after running your step:
{
  "name": "FlowD",
  "description": "",
  "batchSize": 100,
  "threadCount": 4,
  "stopOnError": false,
  "options": {},
  "version": 0,
  "steps": {
    "1": {
      "name": "CustomD",
      "description": "",
      "options": {
        "sourceQuery": "cts.collectionQuery([])",
        "targetEntity": "",
        "collections": [
          "CustomD"
        ],
        "sourceDatabase": "data-hub-STAGING",
        "sourceCollection": "",
        "targetDatabase": "data-hub-FINAL"
      },
      "customHook": null,
      "retryLimit": null,
      "batchSize": null,
      "threadCount": null,
      "stepDefinitionName": "CustomD",
      "stepDefinitionType": "CUSTOM",
      "modulePath": "/custom-modules/custom/CustomD/main.sjs"
    }
  }
}

wooldridge:
@ayuwono thanks for checking this. I rescind my previous comment as now I see the default value filled in now and the step runs fine. I suspect it was a hiccup in my editor environment.

brucean52:
@ayuwono @wooldridge Thanks for testing this. I updated the truncate in the top stepper container and added truncating to the bottom step container. We can revisit the tooltip option in another JIRA ticket.

ayuwono:
On the bottom step container

At first I see this when my browser window is not extended:
Screen Shot 2019-05-30 at 1 44 16 PM

Then I extend my browser window and the truncate looks fine, but the vertical dots and chevron menu icon are outside the pane:
Screen Shot 2019-05-30 at 1 44 29 PM

ayuwono:
Very close @brucean52 =) , I like having the "..." and the end of the truncate if you can, instead of abruptly cut

Screen Shot 2019-05-30 at 2 44 55 PM

brucean52:
@ayuwono Thanks for all the testing. It should be good now :)

PR #2572, opened 2019-05-29
DHFPROD-2588: Fix for resolving custom transform module paths

rjrudin:
I'm running this locally to make sure that my modification to modulePaths doesn't break anything. 

Pinging @derms  on this too.

### DIFF
--- a/marklogic-data-hub/build.gradle
+++ b/marklogic-data-hub/build.gradle
@@ -281,7 +281,7 @@ ext {
         customTokens.put("%%mlStagingTriggersDbName%%", mlStagingTriggersDbName)
         customTokens.put("%%mlStagingSchemasDbName%%", mlStagingSchemasDbName)
 
-        modulePaths = ["marklogic-data-hub/src/main/resources/ml-modules"]
+        modulePaths = ["marklogic-data-hub/src/main/resources/ml-modules", "marklogic-data-hub/src/test/ml-modules"]
     }
 }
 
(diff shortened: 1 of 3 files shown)


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/rest-api/lib/extensions-util.xqy
```diff
@@ -484,7 +484,7 @@ declare function extut:get-extension-function(
         )
 };
 
-declare private function extut:get-extension-function(
+declare function extut:get-extension-function(
```
rjrudin: I had to remove the "private" so I could write tests against this.

marklogic-data-hub/build.gradle
```diff
@@ -281,7 +281,7 @@ ext {
         customTokens.put("%%mlStagingTriggersDbName%%", mlStagingTriggersDbName)
         customTokens.put("%%mlStagingSchemasDbName%%", mlStagingSchemasDbName)
 
-        modulePaths = ["marklogic-data-hub/src/main/resources/ml-modules"]
+        modulePaths = ["marklogic-data-hub/src/main/resources/ml-modules", "marklogic-data-hub/src/test/ml-modules"]
```
jamesagardner: Won't this cause the src/test/ml-modules to be loaded everywhere?  Should this not live in an environment specific properties file to prevent test code from being loaded into non-test environments?

rjrudin: This build.gradle file doesn't impact the use of DHF - it just impacts how the DHF source code is deployed so that it can be tested. It's equivalent to the build.gradle file for ml-app-deployer and ml-gradle - it has no impact on users of either project.

PR #2573, opened 2019-05-29
DHFPROD-2343: Remove the entity definition from database on deletion from UI

akshaysonvane:
Things to consider:
1. Security concerns for adding 
`void deleteDocument(String uri, DatabaseKind databaseKind);`
2. With this PR, we are moving away from the trend of clearing all modules and then reinstalling all of them. Instead, we are specifically removing a document if the user removes the entity from UI. This causes an issue if a user removes the entity from the filesystem and then expects it to be removed from the server on redeploy using UI.

### DIFF
--- a/web/src/main/java/com/marklogic/hub/web/service/EntityManagerService.java
+++ b/web/src/main/java/com/marklogic/hub/web/service/EntityManagerService.java
@@ -143,11 +143,12 @@ public class EntityManagerService {
     }
 
     public void deleteEntity(String entity) throws IOException {
-        File entitiesFile = hubConfig.getHubEntitiesDir().resolve(entity + ENTITY_FILE_EXTENSION).toFile();
+        String entityFileName = entity + ENTITY_FILE_EXTENSION;
+        File entitiesFile = hubConfig.getHubEntitiesDir().resolve(entityFileName).toFile();
         if (entitiesFile.exists()) {
             em.deleteEntity(entity);
-            dataHubService.deleteDocument("/entities/" + entity + ENTITY_FILE_EXTENSION, DatabaseKind.STAGING);
-            dataHubService.deleteDocument("/entities/" + entity + ENTITY_FILE_EXTENSION, DatabaseKind.FINAL);
+            dataHubService.deleteDocument("/entities/" + entityFileName, DatabaseKind.STAGING);
+            dataHubService.deleteDocument("/entities/" + entityFileName, DatabaseKind.FINAL);
         }
     }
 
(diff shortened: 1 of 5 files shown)


web/src/main/java/com/marklogic/hub/web/service/EntityManagerService.java
```diff
@@ -142,9 +143,12 @@ public EntityModel saveEntity(EntityModel entity) throws IOException {
     }
 
     public void deleteEntity(String entity) throws IOException {
-        File entitiesFile = hubConfig.getHubEntitiesDir().resolve(entity + ENTITY_FILE_EXTENSION).toFile();
+        String entityFileName = entity + ENTITY_FILE_EXTENSION;
```
rjrudin: I think this is the right method to expose, and I wouldn't yet expose a general-purpose "deleteDocument" method until there's at least one more use case for it. 

The concept of deleting a document isn't really specific to a data hub, it's just a general ML feature. The concept of deleting an entity is specific to a data hub, and so this method makes a lot of sense and seems very handy. 

FWIW, I have no issue with this code deleting individual documents, I think that's the way to go.

web/src/main/java/com/marklogic/hub/web/service/EntityManagerService.java
```diff
@@ -142,9 +143,12 @@ public EntityModel saveEntity(EntityModel entity) throws IOException {
...
+        File entitiesFile = hubConfig.getHubEntitiesDir().resolve(entityFileName).toFile();
         if (entitiesFile.exists()) {
             em.deleteEntity(entity);
+            dataHubService.deleteDocument("/entities/" + entityFileName, DatabaseKind.STAGING);
+            dataHubService.deleteDocument("/entities/" + entityFileName, DatabaseKind.FINAL);
```
rjrudin: I don't know what our rules of thumb are for logging yet, but I think there's value in info-level logging of each URI being deleted (that would be an argument for a general-purpose "delete document" method, but I'd still hold off on that until I have another use case for it) - e.g. "Deleting URI (uri) from (database kind) database".

web/src/main/java/com/marklogic/hub/web/service/EntityManagerService.java
akshaysonvane: I made a general purpose "deleteDocument" to delete entity, flow, and step-def artifacts.
The current PR only deals with entity artifacts as per the scope of the bug but I think we should also delete the flow artifacts when they are deleted from the UI. (If agreed, I can take this up as a new task)
The step-defs can be shared among flows and thus we might only want to delete them if they are not backing any steps in any flows. (We can take this on when we introduce the concept of step-defs in the UI)

web/src/main/java/com/marklogic/hub/web/service/EntityManagerService.java
akshaysonvane: Yes, the info level logging makes sense. I will add that.

PR #2574, opened 2019-05-30
DHFPROD-2570: Extend JSON array support

ryanjdew:


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/flow-utils.sjs
```diff
@@ -371,8 +379,16 @@ class FlowUtils {
...
+                if (item instanceof Node) {
                   nb.addNode(item);
+                } else if (item instanceof Number) {
+                  nb.addNumber(item);
+                } else if (item instanceof Boolean) {
+                  nb.addBoolean(item);
+                } else if (item === null) {
+                  nb.addNull();
+                } else {
+                  nb.addText(fn.string(item));
```
rjrudin: This is the same as the modified block above - I think it's a good candidate for a separate function with a set of unit tests, where a NodeBuilder is passed in and then the item that was added is verified.

ryanjdew: Refactored to make it more DRY.

PR #2576, opened 2019-05-30
DHFPROD-2517: Support temporal document inserts/updates

ryanjdew:


gvaidees:
One thing with the code that *could* be an issue. If multiple temporal collections are passed by mistake by the user, it looks like the first one in the list will be used for the temporal insert. Should we throw an error instead?

PR #2579, opened 2019-05-31
DHFPROD-1878: Long facet text

clockworked247:
The following is the proposed change to the UI to not only fix the described bug with the text bumping down the chicklets but also the general size and "..." truncation of longer names.

What changed?

- The facets column became one column wider, while the results became one smaller. There is no negative impact on the results, as far as I can tell.
- As the browser shrinks or on smaller screen sizes, there was previously code that made the text bump down in size, making it unreadable. I've removed that and stayed at a fixed, readable size.
- We're now using CSS to handle the ellipsis, rathar than truncating by # of characters. This makes our truncation be more accurate.
- The chiclet size is much less wide with less padding, leaving more room for the collection name.

Old Behavior:
https://drive.google.com/file/d/1wFhbM1C1xMkW1wJkeDd35mI6NByhQjbO/view

New Behavior:
https://drive.google.com/file/d/1XvGI6TkRNWlZ9_h2DCZBo1Na-P400TwB/view

### DIFF
--- a/web/src/main/ui/app/components/facets/facets.component.html
+++ b/web/src/main/ui/app/components/facets/facets.component.html
@@ -5,7 +5,7 @@
         mdl-colored="primary" mdl-button mdl-button-type="raised"
         *ngFor="let value of facet.values"
         (click)="toggle(facet.__key, value)">
-        <span class="chiclet-name" title="{{ value }}">{{ facet.__key }}: {{ value | truncate : 20 }}</span>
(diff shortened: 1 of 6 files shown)


clockworked247:
@wooldridge Understood, thanks for the feedback.

clockworked247:
@wooldridge @brucean52 rebased off latest develop and ready to merge.  See:

https://drive.google.com/file/d/1HHpQuv7AhJI6IbRKmdwUAz8WKUevzDiY/view

PR #2580, opened 2019-05-31
DHFPROD-2591: Delete flow artifact from both DB and FS

akshaysonvane:


ayuwono:
@akshaysonvane , if the flow contains some mappings, will all the mapping versions be removed too? Right now all of these files are stored under mappings on filesystem and on staging+final db

akshaysonvane:
> @akshaysonvane , if the flow contains some mappings, will all the mapping versions be removed too? Right now all of these files are stored under mappings on filesystem and on staging+final db

Nope, mapping artifacts will not be deleted from either FS or DB. Reason being mappings can be shared among flows.

PR #2582, opened 2019-05-31
DHFPROD-2514: css to handle long flow names and steps

brucean52:
Handles long flow and step names on Manage Flows, Edit Flow, Jobs, Job Details, and the Run Dialog


brucean52:
@ayuwono I fixed this down to 1000px width since this is a desktop application.

PR #2583, opened 2019-06-01
DHFPROD-2452: Processing text docs that have xml/json content and a test for it

srinathgit:
Processing text docs that have xml/json content and a test for it

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/ingestion/default/main.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/ingestion/default/main.sjs
@@ -16,20 +16,38 @@ function main(content, options) {
   let instance = content.value.root || content.value;
   if (instance.nodeType === Node.BINARY_NODE || outputFormat === datahub.flow.consts.BINARY || outputFormat === datahub.flow.consts.TEXT || outputFormat === Node.TEXT_NODE) {
     return content;
-  } else {
-    let triples = [];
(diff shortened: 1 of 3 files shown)


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/ingestion/default/main.sjs
```diff
@@ -16,20 +16,38 @@ function main(content, options) {
   let instance = content.value.root || content.value;
   if (instance.nodeType === Node.BINARY_NODE || outputFormat === datahub.flow.consts.BINARY || outputFormat === datahub.flow.consts.TEXT || outputFormat === Node.TEXT_NODE) {
     return content;
```
aebadirad: the last check "outputFormat === node.text_node" needs to be eliminated

srinathgit: If the "outputFormat" is text, we still don't process it and just return the content right ? I am trying to handle only the scenarios where input format is text and outputFormat is xml/json . If it is a well formed xml/json, I am processing it as such else I am throwing an error

aebadirad: Right, but the very last one isn't a possible check. Node.text_node is a js object value, not an output format.

The original check should have been instance.nodeType == node.text_node and been removed here.

PR #2587, opened 2019-06-03
DHFPROD-2506: adding null check and use the proper key to search for …

aebadirad:
…a batch document


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/flow-utils.sjs
```diff
@@ -125,7 +125,7 @@ class FlowUtils {
     }
 
     if (dataFormat === this.consts.JSON) {
-      if(instance.root) {
+      if(instance && instance.root) {
```
rjrudin: FWIW, I think this is a reasonable commit to omit tests for, as the fixes are both obvious and (more importantly) extremely unlikely to be undone.

PR #2590, opened 2019-06-03
DHFPROD-2573: UI fix: Check for existing entity names when creating new entity

brucean52:


brucean52:
> Minor request, can we change the messaging to the following:
> 
> "This entity name already exists."
> 
> This is more consistent with what we have for the step name check.

@wooldridge I went with 'title' over 'name' since we call it Title for the input

PR #2591, opened 2019-06-03
E2E UI regression facets fix

ayuwono:
This PR fixed facets regression issue as the result of dev fix


ayuwono:
@SameeraPriyathamTadikonda & @timur-isangulov , please review and get this merge to develop. This is to fix a regression issue in which dev modified UI element. If it's not merged, all PR commits will fail

PR #2592, opened 2019-06-03
DHFPROD-2589: The status of the cancelled flow doesn't update. 

brucean52:


brucean52:
@timur-isangulov I doubled checked this and it's working for me. Could you test again?

timur-isangulov:
> @timur-isangulov I doubled checked this and it's working for me. Could you test again?

I just tested again. I still see the same behavior as on video above, after clicking on the stop button of the running job the status doesn't change to canceled.

ayuwono:
> > @timur-isangulov I doubled checked this and it's working for me. Could you test again?
> 
> I just tested again. I still see the same behavior as on video above, after clicking on the stop button of the running job the status doesn't change to canceled.

I think what @timur-isangulov sees could be a race condition, as he is using a single csv document that's very quick to ingest. Try to run ingestion that takes longer. But if the race condition is there for ingesting one document, it could be a separate issue

timur-isangulov:
> > > @timur-isangulov I doubled checked this and it's working for me. Could you test again?
> > 
> > 
> > I just tested again. I still see the same behavior as on video above, after clicking on the stop button of the running job the status doesn't change to canceled.
> 
> I think what @timur-isangulov sees could be a race condition, as he is using a single csv document that's very quick to ingest. Try to run ingestion that takes longer. But if the race condition is there for ingesting one document, it could be a separate issue

I actually tried to ingest a single orders.csv file. I see the same issue with a larger dataset as well. 
[orders.zip](https://github.com/marklogic/marklogic-data-hub/files/3253671/orders.zip)
[OR30.zip](https://github.com/marklogic/marklogic-data-hub/files/3253677/OR30.zip)

PR #2594, opened 2019-06-03
DHFPROD-2506: Change 'step' to 'stepNumber' in test batch docs

srinathgit:
This should fix the failing test


aebadirad:
Ah, good catch.

PR #2596, opened 2019-06-03
DHFPROD-2607: Fix uber-model generation on ML nightly

bsrikan:
Entity services has a fix for bugtrack 52006, which requires DHF to have xdmp:xslt-invoke privilege to our roles.

### DIFF
--- a/marklogic-data-hub/src/main/resources/hub-internal-config/security/roles/data-hub-admin-role.json
+++ b/marklogic-data-hub/src/main/resources/hub-internal-config/security/roles/data-hub-admin-role.json
@@ -36,6 +36,11 @@
       "kind": "execute"
     },
     {
+      "privilege-name": "xdmp:xslt-invoke",
+      "action": "http://marklogic.com/xdmp/privileges/xslt-invoke",
+      "kind": "execute"
+    },
+    {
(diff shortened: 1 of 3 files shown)


marklogic-data-hub/src/main/resources/hub-internal-config/security/roles/data-hub-admin-role.json
```diff
@@ -35,6 +35,11 @@
       "action": "http://marklogic.com/xdmp/privileges/xslt-eval",
       "kind": "execute"
     },
+    {
```
rjrudin: One thing I've found helpful on field engagements is to include a README.md file in a config directory to explain why e.g. roles have certain privileges. The README file is needed because comments are awkward to include in JSON files (though IIRC, ml-app-deployer will strip them out for you). 

Otherwise, when looking at this comment, it's not possible to know why this privilege is needed unless you have access to the internal JIRA instance.

aebadirad: Yeah, let's never put comments in JSON files (they're actually not valid json), and instead place readme files in places. Or better yet, an explanation at the root of the project outlining basic security principles and roles.

PR #2597, opened 2019-06-03
DHFPROD-2340: Add outputFormat option to all steps

brucean52:


ayuwono:
@wooldridge, this menu is under Advanced settings. Maybe we can just let the target output type be in the ingestion step page?

ayuwono:
@wooldridge, we can talk about it. I think it's okay to remove it from the page as long as the users know what kind of output they are having

brucean52:
> Get rid of the "Target File Type" menu in the Ingestion configuration since this setting is now covered in the New Step dialog.
> 
> Change the label from "Output Format" to "Target Format" since we are using Source/Target terminology across the app, let's be consistent. I also like "Format" rather than "File Type" here since arguably it's not a file when it's in the database.

@wooldridge  I agree with @ayuwono with keeping the 'Target File Type' in the Ingestion component since it is an important option for ingestion. I'd rather remove the 'Output Format' from the advanced settings for new Ingestion step. 

I like Format over File Type, but I feel like Output Format is a better definition than Target Format. My reasoning is we are doing Input/Output for ingestion step. For databases, we are doing source/target.

wooldridge:
@ayuwono and @brucean52 I agree that there is value to having the output format setting more visible for the ingestion step. So remove it from the New Step Advanced Settings for ingestion. But use "Target Format" for the label.

brucean52:
@wooldridge I made the changes for filetype -> format and removed target format for advanced settings for new ingestion steps. These changes only effect the UI labels, the parameters for the step objects are 'inputFileType' and 'outputFormat.' Changing these parameters would also require server changes, so I left them as is.

wooldridge:
@brucean52 one more thing, can we move the Target Format menu in the Advanced Settings section under the Target Database menu, so the Target-related settings are side by side. So:

Source Database
Target Database
Target Format

ayuwono:
Not sure if it's easily reproducible. I'm having a hard time to reproduce it

PR #2601, opened 2019-06-04
DHFPROD-2610: Upgrading to ml-gradle/ml-app-deployer 3.15.0

rjrudin:


marklogic-data-hub/src/main/java/com/marklogic/hub/impl/DataHubImpl.java
```diff
@@ -530,15 +527,10 @@ public void install(HubDeployStatusListener listener) {
                 throw new DataHubConfigurationException(e);
             }
         }
-        AppConfig appConfig = hubConfig.getAppConfig();
```
rjrudin: 3.15.0 defaults to using CMA for these resources, so no longer need to set these.

marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/DeployDatabaseFieldCommand.java
```diff
@@ -45,7 +45,7 @@
     private final static Namespace MANAGE_NS = Namespace.getNamespace("http://marklogic.com/manage");
 
     public DeployDatabaseFieldCommand() {
-        setExecuteSortOrder(SortOrderConstants.DEPLOY_CONTENT_DATABASES + 1);
+        setExecuteSortOrder(SortOrderConstants.DEPLOY_OTHER_DATABASES + 1);
```
rjrudin: Using a non-deprecated constant

PR #2609, opened 2019-06-04
DHFPROD-2027: fixing display icons on multiple lines

jasonmy419:
DHFPROD-2027 fixed by displaying icons on multiple lines, simple css changes.


web/src/main/ui/app/components/flows-new/manage-flows/ui/step-icons-ui.component.scss
```diff
@@ -1,14 +1,18 @@
 .step-icons {
-  position: absolute;
+  //position: absolute;
```
brucean52: You can remove the position, top, and left css parameters since they are not used.

jasonmy419: Ok, I am gonna do it tmr.

PR #2611, opened 2019-06-05
DHFPROD-2566: Add insurance example project

wooldridge:
Add the insurance example project demonstrated at MLW 2019 to the examples subfolder.
Create a README for the insurance example project.
Update the README for the examples directory.


rjrudin:
Hey Mike - I think it's better to check in those sample customer files as a single zip, which can be unzipped when the application is deployed. If you agree, let me know, I can add the Gradle task to take care of that.

wooldridge:
@rjrudin since this is UI-oriented sample project rather than a gradle-oriented sample project, I'd like to avoid requiring the user do anything extra like unzipping files to work with it. An ease-of-use benefit, especially for a non-technical user. I reduced the number of sample files in the Advantage dataset folder from 500 to 100 anticipating this issue. I know this issue was raised for DHFPROD-2606 but I believe those files numbered a lot more?

rjrudin:
@wooldridge If the user is just starting up QuickStart and has no reason to run any Gradle tasks, then I agree about not introducing a Gradle task just to unzip the file.

wooldridge:
@rjrudin great, I'm glad we can avoid the unzipping in this case.

PR #2613, opened 2019-06-05
DHFPROD-2601: fix the propagation of headers/triples within the mappi…

aebadirad:
…ng step


rjrudin:
@aebadirad I'll resubmit this, I'm close to having a unit test working for this. Need to move a little code around.

PR #2619, opened 2019-06-06
DHFPROD-2627: UX fix: Add appropriate UI feedback after starting a flow run

brucean52:


brucean52:
> The disabled run button is good
> 
> But I still see some delay after clicking to run flow on new flow and new step. @brucean52 I will approve if we are not handling the delay on this PR
> 
> https://drive.google.com/file/d/1RrQW6a75X-QhMzNh_TF7TabPT2dOJfDL/view
> 
> On existing flow and steps that have been run before, there is no delay, the run button is disabled right away after clicking on it

@ayuwono it should be good now.

brucean52:
@wooldridge I added the spinner to the edit flow view. Spinner disappears and gets replaced with status text once a step is running.

brucean52:
@wooldridge for mange flow view, I added the spinner to where status is.

PR #2620, opened 2019-06-06
E2E UI - stabilize flow run with await

ayuwono:
Add await on flow run to stabilize the tests


aebadirad:
Rebase off develop and let's see if we can get a clean run on this

PR #2621, opened 2019-06-06
DHFPROD-2608: Handle entity file rewrite in a case insensitive file-system

akshaysonvane:


ayuwono:
Logged and assigned to @brucean52 so QS UI will handle this too
https://project.marklogic.com/jira/browse/DHFPROD-2636

PR #2624, opened 2019-06-07
DHFPROD-2358: e2e UI dashboard tests

timur-isangulov:
This pr contains 5 tests for the QS dashboard page.


timur-isangulov:
I also removed old entity tests that are not used anymore.

web/e2e/page-objects/dashboard/dashboard.ts
```diff
@@ -43,6 +43,28 @@ export class DashboardPage extends AppPage {
     await browser.sleep(2000);
   }
 
+  async clearStagingDatabase() {
+    await browser.wait(EC.visibilityOf(this.clearStagingButton()));
+    await this.clearStagingButton().click();
+    await browser.sleep(500);
```
ayuwono: if you need to sleep here, might as well to put it to 1000 rather than 500. 500 is 0.5 sec

timur-isangulov: I noticed sometimes protractor works more stable if I add a short sleep timeout like after clicking on a button. Generally, we need to remove all timeout or make them as short as possible. In this particular case, 0.5 sec. timeout works well.

web/e2e/specs/dashboard/dashboard.ts
```diff
@@ -0,0 +1,65 @@
...
+
+    it('should remove jobs documents', async function () {
+      await appPage.clickDashboardTab();
+      await dashboardPage.clearJobDatabase();
+      await expect(dashboardPage.jobCount().getText()).toBe('0');
+    });
+
+    it('should remove all documents', async function () {
+      await appPage.clickDashboardTab();
+      await dashboardPage.clearAllDatabases();
```
ayuwono: We can more the clear all database test case after redeploy and then verify the docs. It should be no harm since we will do uninstall on the next step

timur-isangulov: Done.

PR #2639, opened 2019-06-10
DHFPROD-2383: added customized tooltips for matching tab

jasonmy419:
Also removed conflicting default tooltips and corrected a variable name typo


wooldridge:
@jasonmy419 please make this PR to the following feature branch:

https://github.com/marklogic/marklogic-data-hub/tree/feature/DHFPROD-2383-matching-tooltips

PR #2643, opened 2019-06-10
DHFPROD-2635: Users should be able to set delimiter (other than comma)…

xnikhil08:
Added just a text field, labeled as "Source Format Delimiter", in the ingestion step. Using this field the user will be able to provide an appropriate delimiter character, if the value for the field "source format " is CSV.

### DIFF
--- a/web/src/main/ui/app/components/flows-new/edit-flow/ingest/ui/ingest-ui.component.ts
+++ b/web/src/main/ui/app/components/flows-new/edit-flow/ingest/ui/ingest-ui.component.ts
@@ -1,4 +1,5 @@
 import {Component, EventEmitter, Input, OnInit, Output} from "@angular/core";
+import { FlowsTooltips } from '../../../tooltips/flows.tooltips';
 
 const settings = {
   inputFilePath: {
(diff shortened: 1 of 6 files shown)


web/src/main/ui/app/components/flows-new/edit-flow/ingest/ui/ingest-ui.component.ts
```diff
@@ -32,6 +32,11 @@ const settings = {
       }
     ]
   },
+  SourceFormatDelimiter: {
+    label: 'Source Format Delimiter',
+    description: 'The delimiter character used in the delimited file. Accepted values are comma(,), tab(" "), Default: (,)',
```
ayuwono: Change the description:
'The delimiter character used in the delimited file. Accepted values are comma(,) and tab(" "). Default: comma (,)'

ayuwono:
1) If we only have two delimiter options, should we make it as a menu? How the user will know to put a tab? (by typing " " ?)

2) After putting the delimiter, there is no change on the flow json file on the filesystem:
      "fileLocations" : {
        "inputFilePath" : "/Users/ayuwono/builds/macosx-64/b9_0/dhf-qa/input/bedrock",
        "inputFileType" : "csv",
        "outputURIReplacement" : "/Users/ayuwono/builds/macosx-64/b9_0/dhf-qa/input/bedrock,'/bedrock'"
      }

3) If we still opt on text field, we should put a validation where user puts incorrect delimiter (other than comma and tab)

wooldridge:
This issue was addressed in another PR so will close this one: https://github.com/marklogic/marklogic-data-hub/pull/2657

PR #2644, opened 2019-06-11
DHFPROD-2508: Add filename to header and metadata properties when ingesting csv docs

srinathgit:
Add filename to header and metadata properties when ingesting csv docs and a test

### DIFF
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/step/impl/WriteStepRunner.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/step/impl/WriteStepRunner.java
@@ -539,6 +539,15 @@ public class WriteStepRunner implements StepRunner {
             while(itr.hasNext()) {
                 try {
                     File file = new File((String) itr.next());
+                    if("csv".equalsIgnoreCase(inputFileType)) {
+                        //Add "file" to options so that it gets added to header in enode code
+                        options.put("file", file.getAbsolutePath());
+                        optionString = jsonToString(options);
+                        serverTransform.remove("options");
+                        serverTransform.addParameter("options", optionString);
+                        //Add additional document metadata
+                        metadataValues.add("datahubCreatedUsingFile" , file.getAbsolutePath());
+                    }
                     addToBatcher(file, fileFormat);
                 }
                 catch (Exception e) {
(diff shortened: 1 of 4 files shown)


marklogic-data-hub/src/main/java/com/marklogic/hub/step/impl/WriteStepRunner.java
```diff
@@ -539,6 +539,15 @@ private RunStepResponse runIngester(RunStepResponse runStepResponse, Collection<
             while(itr.hasNext()) {
                 try {
                     File file = new File((String) itr.next());
+                    if("csv".equalsIgnoreCase(inputFileType)) {
+                        //Add "fileName" to options so that it gets added to header in enode code
+                        options.put("fileName", file.getAbsolutePath());
```
rjrudin: I would call this "file" or "filePath" or "pathname" instead of "fileName", as fileName/filename typically imply just the name of the file and not the full path. 

This would be consistent with the only other place where "filename" is used in DHF - in the tasks for importing/exporting jobs. In those tasks, "filename" is just for the name of a file, not including the path.

marklogic-data-hub/src/main/java/com/marklogic/hub/step/impl/WriteStepRunner.java
```diff
@@ -539,6 +539,15 @@ private RunStepResponse runIngester(RunStepResponse runStepResponse, Collection<
...
+                        optionString = jsonToString(options);
+                        serverTransform.remove("options");
+                        serverTransform.addParameter("options", optionString);
+                        //Add additional document metadata
+                        metadataValues.add("datahubCreatedUsingFile" , file.getAbsolutePath());
```
rjrudin: Since only "File" is used here, then I think that suggests using just "file" above for consistency.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/flow-utils.sjs
```diff
@@ -505,6 +505,9 @@ class FlowUtils {
     for (let key in options.headers) {
       headers[key] = this.evalSubstituteVal(options.headers[key]);
     }
+    if(options.fileName) {
+      headers["createdUsingFile"] = options.fileName;
```
aebadirad: if you're doing file, file here too instead of filename

PR #2645, opened 2019-06-11
DHFPROD-2599: Deleting a step deletes artifacts associated with it

rjrudin:
For a mapping step, the related mapping artifact is deleted.

For a custom step, the related step definition is deleted - but not the custom module, as other modules may depend on or reference that module.

Note this is for 5.0.2, not 5.0.1.


rjrudin:
Will reopen this, adding support for deleting documents too.

PR #2646, opened 2019-06-11
DHFPROD-2357: UI automated tests for QS Browse Data page. 

timur-isangulov:


web/e2e/page-objects/browse/browse.ts
```diff
@@ -73,13 +119,84 @@ export class BrowsePage extends AppPage {
...
+      }
+      await browser.sleep(100);
+    }
+  }
+
+  /**
+   *
+   * @param page: first, previous, next, last
+   */
+  async paginateClick(page: string) {
```
ayuwono: for naming consistency with the rest, rename the function to: clickPaginate()

timur-isangulov: Fixed.

web/e2e/specs/browse/browse-data.ts
```diff
@@ -38,10 +38,10 @@ export default function (qaProjectDir) {
...
     it('should verify navigation to the next, last, previous, first pages', async function () {
-      await browsePage.paginateClick('next');
-      await browsePage.paginateClick('previous');
-      await browsePage.paginateClick('last');
-      await browsePage.paginateClick('first');
+      await browsePage.clickPaginate('next');
+      await browsePage.clickPaginate('previous');
+      await browsePage.clickPaginate('last');
+      await browsePage.clickPaginate('first');
     });
```
bsrikan: There is no validation that after clickPaginate we really land at the intended page. So if this test is really validating the pagination button isnt dead then its fine. Otherwise IMO we need some validation.

timur-isangulov: Added validation for the number of docs > 0.  Thanks.

PR #2647, opened 2019-06-11
DHFPROD-2644: tests for extractInstance method in default-mapping step lib.sjs

rahulvudutala:


rjrudin:
Looks good, I'll try it out locally to make sure the tests pass

PR #2650, opened 2019-06-11
DHFPROD-2620: Fixed by adding real-time error checking

jasonmy419:
DHFPROD-2653:  Fixed by calling selectedTypeChanged method in ngOnInit


web/src/main/ui/app/components/flows-new/edit-flow/mastering/matching/ui/add-match-threshold-dialog.component.ts
```diff
@@ -16,6 +17,7 @@ export interface DialogData {
 })
 export class AddMatchThresholdDialogComponent {
 
+  instantErrorMatcher: InstantErrorStateMatcher;
```
brucean52: you can define and instantiate the 'instantErrorMatcher' all here and not do it in the OnInit method.
instantErrorMatcher: InstantErrorStateMatcher = new InstantErrorStateMatcher();

PR #2651, opened 2019-06-12
DHFPROD-2383: Matching tooltip edits

wooldridge:
- Move Reduce Properties tooltip up to section label.
- Make title attributes empty (no space) strings.
- Nudge the tooltip containers to the right and widen container for match weight tooltip (keep to one line).
- Add tooltip for Action in Match Thresholds dialog.
- Edit tooltip content, mostly for length.

### DIFF
--- a/web/src/main/ui/app/components/flows-new/edit-flow/mastering/matching/ui/add-match-option-dialog.component.html
+++ b/web/src/main/ui/app/components/flows-new/edit-flow/mastering/matching/ui/add-match-option-dialog.component.html
@@ -51,7 +51,7 @@
     [matTooltip]="selectedType === 'reduce' ? tooltips.weightReduce : tooltips.weightOption"
     matTooltipPosition="left"
     matTooltipShowDelay="500"
-    matTooltipHideDelay="500000"
+    matTooltipHideDelay="500"
     matTooltipClass="weight-tooltip"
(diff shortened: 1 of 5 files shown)


web/src/main/ui/app/components/flows-new/edit-flow/mastering/matching/ui/add-match-option-dialog.component.html
```diff
@@ -51,10 +51,11 @@ <h1 mat-dialog-title>{{getDialogTitle()}}</h1>
     [matTooltip]="selectedType === 'reduce' ? tooltips.weightReduce : tooltips.weightOption"
     matTooltipPosition="left"
     matTooltipShowDelay="500"
-    matTooltipHideDelay="500"
+    matTooltipHideDelay="500000"
```
briantang: Should the delay here be 500 ms for consistency? Other than that, everything looks good for me!

wooldridge: Thanks for catching that, fixed in my latest commit.

jasonmy419:
So basically UI changes, remove unnecessary tooltips and add a new one, tooltips text changes, and CSS   
 changes, fix the conflict of default browser tooltips and custom tooltips look great!

PR #2652, opened 2019-06-12
DHFPROD-2599: Deleting related artifacts when a step is deleted

rjrudin:
Does not yet include deleting custom modules, as it's difficult to reliably delete a module without knowing that something else depends on it (such as another step definition or another module).


marklogic-data-hub/src/main/java/com/marklogic/hub/flow/Flow.java
```diff
@@ -107,10 +107,10 @@
 
     /**
      * Gets the step specified by string key
-     * @param stepNum the string key of the step (usually 1, 2, 3 etc)
+     * @param stepKey the string key of the step (usually 1, 2, 3 etc)
      * @return the step object that is stored in the flow
      */
-    Step getStep(String stepNum);
+    Step getStep(String stepKey);
```
akshaysonvane: +1 for standardizing to `stepKey` everywhere

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/FlowManagerImpl.java
```diff
@@ -145,17 +156,45 @@ public void deleteFlow(String flowName) {
...
+        int stepOrder = Integer.parseInt(key);
+
+        Step removedStep = null;
+
+        if (stepOrder == stepMap.size()) {
+            removedStep = stepMap.remove(key);
+        } else {
+            Map<String, Step> newStepMap = new LinkedHashMap<>();
+            final int[] newStepOrder = {1};
+            final int[] stepIndex = {1};
```
akshaysonvane: We can get rid of this hack as we are not using a lambda expression anymore.

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/FlowManagerImpl.java
```diff
@@ -184,18 +223,102 @@ public boolean isFlowExisted(String flowName) {
...
+     *
+     * @param flow
+     * @param removedStep
+     */
+    protected void deleteRelatedStepArtifacts(Flow flow, Step removedStep) {
+        if (removedStep == null) {
+            return;
+        }
+
+        if (removedStep.isMappingStep() && !stepIsReferencedByAFlow(removedStep.getName(), StepDefinition.StepDefinitionType.MAPPING)) {
```
akshaysonvane: Multiple steps can have the same mapping name. 
(Although that's not possible from the UI yet as mapping name is autogenerated to "flow-name-mapping-name")
Better to check for `mappingIsReferencedByAFlow` to avoid UI logic spilling in the core.

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/FlowManagerImpl.java
```diff
@@ -184,18 +223,102 @@ public boolean isFlowExisted(String flowName) {
...
+
+    protected void deleteStepDefinitionArtifacts(Step removedStep) {
+        logger.info("Deleting custom step as it's no longer referenced by any flows: " + removedStep.getName() +
+            ". The module associated with this step will not be deleted in case other modules refer to it.");
+        StepDefinition stepDef = StepDefinition.create(removedStep.getName(), StepDefinition.StepDefinitionType.CUSTOM);
+        stepDefinitionManager.deleteStepDefinition(stepDef);
+        deleteDocumentsInDirectory(format("/step-definitions/%s/%s/", stepDef.getType().toString(), stepDef.getName()));
+    }
+
+    protected void deleteDocumentsInDirectory(String directory) {
```
akshaysonvane: We can probably reuse `dataHubService.deleteDocument(String uri, DatabaseKind databaseKind)`

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/FlowManagerImpl.java
```diff
@@ -36,16 +41,23 @@
...
 import java.nio.file.Paths;
 import java.util.ArrayList;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
 
 @Component
-public class FlowManagerImpl implements FlowManager {
+public class FlowManagerImpl extends LoggingObject implements FlowManager {
```
hao1st: My two cents here. We have an interface String format(String s, Object... args) in the LoggingObject along with  a logger declaration. It is just one reason to extends the class? It may cause the inconsistency for different Managers' implementation. Would be better if we could extend a class with more useful contracts or implementations?

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/FlowManagerImpl.java
```diff
@@ -184,18 +223,102 @@ public boolean isFlowExisted(String flowName) {
...
+    }
+
+    protected void deleteDocumentsInDirectory(String directory) {
+        final String query = format("declareUpdate(); " +
+                "cts.uris(null, null, cts.directoryQuery('%s', 'infinity'))." +
+                "toArray().forEach(uri => xdmp.documentDelete(uri))",
+            directory
+        );
+
+        if (logger.isInfoEnabled()) {
```
hao1st: do we really need if (logger.isInfoEnabled()) here?

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/FlowManagerImpl.java
hao1st: Each mapping file has different version. Would it be better only clean those files that are not referenced by all the steps across all the flows?

hao1st: I agree. Currently, we do not support a shared Mapping across all the steps/flows.

akshaysonvane: > Not sure what you mean - stepIsReferencedByAFlow is checking the step definition type as well.

It's not possible to share a mapping with multiple flows from UI but that's possible from gradle or core java. So if we have the code to delete mapping in `FlowManager` then we need to check for mappings in all flows. (and use something like `mappingIsReferencedByAFlow` to check that)

The assumption
`final String mappingName = flow.getName() + "-" + removedStep.getName();`
is only applicable in case of QS UI. 
Considering that we should move this logic to the `FlowManagerService` from `FlowManager`

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/FlowManagerImpl.java
akshaysonvane: What do you mean by multiple documents? Wouldn't there only be a single document for mapping/step-def for a step that is being deleted?

The URI's for step-defs and mappings obey a set standard. So we know the URI in advance and can delete them using one or more calls to delete the URIs.

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/FlowManagerImpl.java
rjrudin: I updated this to look at options/mapping/name.

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/FlowManagerImpl.java
rjrudin: A mapping directory can have multiple documents in it, one for each version. I think a step definition can only have a single document, but it seems safer for the future to just delete the directory since we know everything in the directory should be deleted.

akshaysonvane: Makes sense.

PR #2653, opened 2019-06-12
DHFPROD-2656: Added unit tests for building invoke options

rjrudin:
Added "testUnit" task as well so it's easy to run marklogic-unit-test tests. 

This also fixes an issue where TestConfig was being mistakenly picked up by the web tests.

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/hub-utils.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/hub-utils.sjs
@@ -113,19 +113,26 @@ class HubUtils {
   }
 
   invoke(moduleUri, parameters, user = null, database) {
+    let options = buildInvokeOptions(user, database);
+    xdmp.invoke(moduleUri, parameters, options)
+  }
+
+  buildInvokeOptions(user = null, database) {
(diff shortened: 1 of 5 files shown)


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/hub-utils.sjs
```diff
@@ -113,19 +113,26 @@ class HubUtils {
...
 
-    if(user && user !== xdmp.getCurrentUser()) {
+    if (user && user !== xdmp.getCurrentUser()) {
       options.userId = xdmp.user(user);
     }
-    
-    if(database) {
-      options.database = database ? xdmp.database(database): xdmp.database();
+
+    if (database) {
```
rjrudin: I removed the "database ?" check here too.

hao1st: guess if no database is specified, the default is current database without requiring to set options.database ?

aebadirad: correct, its assumed if nothing is provided, its the current db

PR #2654, opened 2019-06-12
DHFPROD-1982: Add fine granularity for prov

ryanjdew:


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/extensions/runFlow.sjs
```diff
@@ -14,7 +14,7 @@
  limitations under the License.
  */
 'use strict';
-const DataHub = require("/data-hub/5/datahub.sjs");
+const DataHubSingelton = require("/data-hub/5/datahub-singleton.sjs");
```
rjrudin: Little typo - "Single"

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/flow.sjs
```diff
@@ -219,14 +221,58 @@ class Flow {
...
       let info = {
-        derivedFrom: content.previousUri || content.uri,
+        derivedFrom: previousUris,
         influencedBy: stepName,
         status: (stepDefTypeLowerCase === 'ingestion') ? 'created' : 'updated',
         metadata: {}
       };
-      let provResult = this.datahub.prov.createStepRecord(jobId, flowName, stepName, stepRef.stepDefinitionName, stepDefTypeLowerCase, content.uri, info);
+      let provResult;
+      if (prov.granularityLevel() === prov.FINE_LEVEL && content.provenance) {
```
rjrudin: This is a meaty chunk of code that seems like it would benefit from being in an e.g. "buildFineProvenanceData" function (or "create" or "save" if it's persisting the data as well and not just building an in-memory data structure).

rjrudin:
Going to try this out locally - will try dh-5-example before these changes and then after it to see what the difference is.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/flow.sjs
rjrudin: @ryanjdew Since the function is inserting a bunch of documents, I would use "create" or "save" here instead of "build".

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/flows/suite-setup.xqy
```diff
@@ -0,0 +1,44 @@
+xquery version "1.0-ml";
```
rjrudin: This setup seems like it'd be very useful for any test that wants to easily setup and run a flow with mapping and mastering. No need to move it yet, but I'm guessing it would be useful to have this outside of the suites directory and in an e.g. "/test/com.marklogic.hub/" directory full of DHF-specific helpers.

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/flows/mapping-prov-test.sjs
```diff
@@ -0,0 +1,30 @@
...
+  ], { provenanceGranularityLevel: 'fine'}, 1);
+  let provCount = fn.head(
+    xdmp.invokeFunction(function() {
+      return cts.estimate(cts.collectionQuery([
+        'http://marklogic.com/provenance-services/record'
+      ]));
+    }, { database: xdmp.database('data-hub-JOBS')})
+  );
+  return [
+    test.assertTrue(provCount > 1, `Provenance document count should be greater than 1 (was: ${provCount})`)
```
rjrudin: I think it'd be useful to add assertions on one of the property-change provenance documents, just to verify and document that the expected content is in one of them. 

I noticed that a few of them have an encoded path as a type - e.g. "_2F__2A__3A_envelope_2F__2A__3A_instance_2F__2A__3A_postal". Is that expected?

ryanjdew: It needs to be encoded because the PROV library stores the value as an xs:QName. It may be encoded in more places that it really needs to. I'll look at isolating the encoded value to only the location that is storing the as an xs:QName.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/flow.sjs
ryanjdew: It doesn't technically get saved until the prov.commit() call later on.

aebadirad:
We're getting a failure on a prov test for mapping: https://jenkins.marklogic.com/blue/organizations/jenkins/Datahub_CI/detail/PR-2654/37/tests

Maybe we need to update the test or account for the shift here.

ryanjdew:
The PROV test is passing locally. I was hoping that changing the test to use config.JOBSDATABASE, instead of the literal would have fixed it for Jenkins, but I haven't figured out why we're getting different results in Jenkins yet. The assertion is failing, so an error isn't being thrown prior that I can tell.

ryanjdew:
@aebadirad @rjrudin So the users we provide don't have the capability to read PROV documents, so my tests can't test if they're created. If I grant read capability, my test will pass but a bunch of other tests will fail when trying to clear out the jobs database. 

Whether our provided users should be granted read capability to the PROV documents perhaps questionable, but I'm pretty sure we don't want to grant the internal role that has update/delete privileges on the PROV docs...

ryanjdew:
Adjusted the unit test to only test assertions if the test is run by a user with proper permissions to read PROV data, which unfortunately means running manually as admin/ps-user for the time being in order for it to be meaningful. 

Also adjusted it to continue creating the mastering version of PROV documents for now, in addition to the fine level PROV if specified by configuration. With that change, I did add the ability to exclude individual documents from PROV so that we avoid creating PROV documents of mastering's PROV documents.

aebadirad:
I agree Rob. We should add significantly more detail

PR #2606, opened 2019-06-04
DHFPROD-2616: Disable editing entity title

brucean52:
Consistent with Flows and Steps. This also prevents issues with renaming to duplicate Entity name and breaking Target Entity links in Steps.

PR #2655, opened 2019-06-12
DHFPROD-2636: UI fix preventing creation of duplicate entities with case sensitivity, include DHFPROD 2305 

briantang:
Now shows "Entity title already exists" when user tries entering a duplicate name with different case sensitivity.


ayuwono:
This was the issue: https://project.marklogic.com/jira/browse/DHFPROD-2616, but it was fixed and closed a week ago

Here is the PR that @brucean52 had with the fix: https://github.com/marklogic/marklogic-data-hub/pull/2606

ayuwono:
Talked with @briantang. It's weird that the issue is disappeared after you create a flow

PR #2657, opened 2019-06-12
DHFPROD-2592: tab delimited support for ingest csv

paxtonhare:
DHFPROD-2592
also fixes DHFPROD-2635


ayuwono:
> On QuickStart:
> 
> * There is no label
> * I can only put one character on the input field. The separator is \t
> 
> Screen Shot 2019-06-12 at 2 51 29 PM

But it seems that the separator is keystroke "tab"

aebadirad:
This should resolve the issues found previous and make it more user friendly at the same time

wooldridge:
Made a few more updates to the separator setting with my latest commits. Also performed manual tests with all four separator types, ingestions all worked: [sep-types.zip](https://github.com/marklogic/marklogic-data-hub/files/3283671/sep-types.zip)

marklogic-data-hub/src/main/java/com/marklogic/hub/step/impl/WriteStepRunner.java
```diff
@@ -280,6 +283,13 @@ public RunStepResponse run() {
         inputFilePath = (String)fileLocation.get("inputFilePath");
         inputFileType = (String)fileLocation.get("inputFileType");
         outputURIReplacement = (String)fileLocation.get("outputURIReplacement");
+        if(inputFileType.equalsIgnoreCase("csv") && fileLocation.get(separator) != null) {
```
srinathgit: The "if" condition has to be 
if(inputFileType.equalsIgnoreCase("csv") && fileLocation.get("separator") != null)

hao1st: https://github.com/marklogic/marklogic-data-hub/blob/574d4b7f2ac4c9dd0bb6f9fb2ccb6bda3d10c0b8/marklogic-data-hub/src/main/java/com/marklogic/hub/step/impl/WriteStepRunner.java#L286-L291

 line 291 separator cannot be null in the if block. We can check like:
csvColumnSeparator = separator.isNotEmpty ? separator.charAt(0) :  ',';

hao1st:
```suggestion
we do not need a member variable "csvColumnSeparator", instead
private char csvColumnSeparator = ','; => private static final CSV_DEFAULT_COLUMN_SEPARATOR = ','; 
       separator = Character.toString(',');
        if (inputFileType.equalsIgnoreCase("csv") && fileLocation.get("separator") != null) {
           String tmp =((String) fileLocation.get("separator")).trim(); 
           if (tmp.equals("\\t") {
               separator = "\t";
           }
        }
        
     https://github.com/marklogic/marklogic-data-hub/blob/574d4b7f2ac4c9dd0bb6f9fb2ccb6bda3d10c0b8/marklogic-data-hub/src/main/java/com/marklogic/hub/step/impl/WriteStepRunner.java#L627
     
     changed line 627 => .withColumnSeparator(separator.charAt(0));
```

PR #2662, opened 2019-06-13
DHFPROD-2631: Adding a step is broken after pointing manually to the last step

brucean52:


aebadirad:
Is the splash radius for this small enough to consider bringing in?

ayuwono:
@aebadirad , this is an edge case, we can bring this in for 5.0.2

wooldridge:
@aebadirad I agree with @ayuwono, fine for 5.0.2.

PR #2668, opened 2019-06-13
DHFPROD-2668: Don't try to carry forward temporalDocURI metadata

ryanjdew:


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/hub-utils.sjs
```diff
@@ -66,6 +66,10 @@ class HubUtils {
       let metadata = context.metadata;
       let temporalCollection = collections.find((col) => temporalCollections[col]);
       if (temporalCollection) {
+        // temporalDocURI is managed by the temporal package and must not be carried forward.
+        if (metadata) {
+          delete metadata.temporalDocURI;
+        }
```
rjrudin: Just curious, are there any downsides to pulling most of the code in this eval'ed string into its own function in hub-utils.sjs and then having the eval'ed string just invoke that function? That would make it easier to unit test the logic in this code. 

If done, an easy way to test this code would be to make a function that contains all the logic for building the JSON object containing permissions, collections, and metadata. A unit test could then easily invoke that (without doing a document insert) and verify that temporalDocURI is removed if a temporalCollection is found.

PR #2673, opened 2019-06-14
DHFPROD-2449: Headers and triples no longer contain empty XML elements

rjrudin:
Also moved buildInvokeHookOptions into a directory based on its module name


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/flow-utils.sjs
```diff
@@ -531,15 +531,21 @@ class FlowUtils {
 
   getHeaders(doc) {
     let headers = fn.head(doc.xpath('/*:envelope/*:headers'));
-    if(fn.count(headers) === 0) {
+    if (fn.count(headers) === 0) {
+      headers = null;
+    }
+    else if (fn.count(fn.head(doc.xpath('/*:envelope/*:headers/*'))) == 0) {
```
rjrudin: I'm guessing there's a better implementation technique for determining if headers/triples have any child elements / contents.

PR #2678, opened 2019-06-14
DHFPROD-2661: Add header info in default flow scaffold

ryanjdew:


aebadirad:
Not sure why the pipeline got stuck on this.

PR #2679, opened 2019-06-14
DHFPROD-2632:Revert "Changing validation tests to accommodate for changes in 10.x server behavior"

srinathgit:
Revert "Changing validation tests to accommodate for changes in 10.x server behavior"
This reverts commit f75402fd
The issue has been fixed in 10.x server


aebadirad:
We tripped an e2e failure

PR #2681, opened 2019-06-14
DHFPROD-2587: Reduce latency of getFlows API call on the backend server

hao1st:
https://project.marklogic.com/jira/browse/DHFPROD-2587

1. Introduce Jaeger for performance tracing;
2. Use Asyn callable/thread pool to improve the performance of getFlows api call;
3. Reduce getFlow call with disk access;
4. Sorte by flow name as default for getFlows response
5. Refractor code

### DIFF
--- a/web/src/main/java/com/marklogic/hub/web/service/AsyncFlowService.java
+++ b/web/src/main/java/com/marklogic/hub/web/service/AsyncFlowService.java
@@ -27,6 +27,7 @@ import io.opentracing.Scope;
 import io.opentracing.Span;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorCompletionService;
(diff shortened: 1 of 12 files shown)


web/src/main/java/com/marklogic/hub/web/service/AsyncFlowService.java
```diff
@@ -0,0 +1,220 @@
...
+
+    ExecutorService executor;
+
+    @PostConstruct
+    public void init(){
+        //we could turn on/off jaeger to trace performance of any call stacks by
+        //injecting customized span code along with tags, e.g. try (Scope scope = JaegerConfig.activate(span))
+        //default tracing is off, in order to turn on, just add a VM option -DJaegerServiceName=data-hub
+        //also see: https://www.jaegertracing.io/docs/1.12/getting-started/
+        String jaegerServiceName = System.getProperty("JaegerServiceName");
```
rjrudin: Instead of pulling a system prop, I recommend using a Spring Value annotation. That allows for the property to be set in a number of ways, not just via system props.

hao1st: Now Jaeger is only internally used for us to trace the bottleneck of performance and then try to improve it. Once you turn on it with the system property flag, and measure what codes you want to trace by simply injecting code snippet, adding some tags. It will print out detail log information. The most important thing is you can analyze the performance using a clean graphical web interface once you run the Jaeger server locally or remotely (docker image).

hao1st: @rjrudin good suggestion! done.

hao1st:
https://www.jaegertracing.io/docs/1.12/getting-started/#all-in-one

$ docker run -d --name jaeger \
  -e COLLECTOR_ZIPKIN_HTTP_PORT=9411 \
  -p 5775:5775/udp \
  -p 6831:6831/udp \
  -p 6832:6832/udp \
  -p 5778:5778 \
  -p 16686:16686 \
  -p 14268:14268 \
  -p 9411:9411 \
  jaegertracing/all-in-one:1.12

then you can go to http://localhost:16686/search to find the tracing history. They are stored in the cache (cassandra) as default. Once you stop the docker image, the history is gone.

aebadirad:
You can also persist it if you keep the disk around for the docker image.

rjrudin:
I'm trying this out locally.

PR #2682, opened 2019-06-15
DHFPROD-2148: Validations for Mastering step

jasonmy419:
1. added instant error matcher
2. added "required" if necessary


wooldridge:
@jasonmy419 can you resolve the above conflicts? And address the distance-threshold validation issue if it still exists?

ayuwono:
@jasonmy419 , can you also add the validation message on create/edit flow dialog's batch size and max count? 

Steps:
1) Click to create a new flow or Edit existing flow
2) Put a string on batch size or max count
3) The label become red and the create button is disabled. But there is no validation message. Should add the validation message like the others. These fields have to be number

Screen Shot 2019-06-18 at 1 32 54 PM

jasonmy419:
Hi Aries,

I did not see threshold option in the dialog, does the max count mean thread count in the dialog?

wooldridge:
@jasonmy419, here's how to access the Distance Threshold setting, which should be validated as a number:

https://drive.google.com/file/d/1wMUH6utvyQWOjHnrJd3eafZbJ6MwygCk/view

PR #2695, opened 2019-06-18
DHFPROD-2560: Can now install DHF into DHS via an executable jar

rjrudin:
Includes a command for verifying the DHF installation

Instructions for building and using the installer are at https://wiki.marklogic.com/display/ENGINEERING/DHF+Installer+Usage


ryanjdew:
Looks good. One discrepancy with the documentation that I'm seeing is that if I don't pass parameters I get the error "Caused by: com.beust.jcommander.ParameterException: The following options are required: [--password], [--username], [--path]" instead of a list all available options.

PR #2713, opened 2019-06-19
DHFPROD-2135: FE Implementation Display MLCP command in ingest step

xnikhil08:
1. Added the MLCP Command in the ingest step.
2. It can be pasted directly onto the terminal, after replacing the password parameter with actual values. Currently set as an encrypted value "*****".
MLCPCommand
@wooldridge - Please review it.

### DIFF
--- a/marklogic-data-hub/build.gradle
+++ b/marklogic-data-hub/build.gradle
@@ -217,6 +217,7 @@ publishing {
             from components.java
             artifact sourcesJar
             artifact javadocJar
+	    artifact bootJar
 
             pom.withXml {
                 asNode().appendNode('description', pomCustomizations.textdescription.text())
@@ -228,9 +229,6 @@ publishing {
                 asNode().append(pomCustomizations.licenses)
             }
         }
-	bootJava(MavenPublication) {
-			artifact bootJar
-		}
     }
     repositories {
         maven {


web/src/main/ui/app/components/flows-new/edit-flow/edit-flow.component.ts
```diff
@@ -180,7 +174,6 @@ export class EditFlowComponent implements OnInit, OnDestroy {
   }
   updateStep(step) {
     this.manageFlowsService.updateStep(this.flow.id, step.id, step).subscribe(resp => {
-      this.editFlowUi.stepUpdated(step);
```
brucean52: It looks like some functionality that's on develop has been removed with this PR. Please look into this

web/src/main/ui/app/components/flows-new/edit-flow/edit-flow.component.ts
```diff
@@ -55,8 +54,6 @@ export class EditFlowComponent implements OnInit, OnDestroy {
   projectDirectory: string;
   flowEnded: string = '';
   stepType: typeof StepType = StepType;
-  runFlowClicked: boolean = false;
-  disableSelect: boolean = false;
```
brucean52: It looks like some functionality that's on develop has been removed with this PR. Please look into this

srinathgit:
The step number should also be included, else, it will always run the first step

aebadirad:
May want to rebase with develop.

web/src/main/ui/app/components/flows-new/edit-flow/edit-flow.component.ts
```diff
@@ -141,15 +138,14 @@ export class EditFlowComponent implements OnInit, OnDestroy {
     });
   }
   runFlow(runObject): void {
-    this.runFlowClicked = true;
     this.manageFlowsService.runFlow(runObject).subscribe(resp => {
-      // TODO add error handling for error response, set runFlowClicked = false;
+      // TODO add response check
+      console.log('run flow resp', resp);
```
rjrudin: Not sure of the team standards yet, but I usually vote on removing log statements like this

PR #2721, opened 2019-06-21
DHFPROD-2164: Change label names for Zip type Match option

jasonmy419:
Already changed name

### DIFF
--- a/web/src/main/ui/app/components/flows-new/edit-flow/mastering/matching/ui/add-match-option-dialog.component.html
+++ b/web/src/main/ui/app/components/flows-new/edit-flow/mastering/matching/ui/add-match-option-dialog.component.html
@@ -111,7 +111,7 @@
       matTooltipPosition="left"
       matTooltipShowDelay="500"
       matTooltipHideDelay="500" layout-fill>
-      <mat-label>5-Matches-9 Boost</mat-label>
+      <mat-label>Five-vs-Nine Match Weight</mat-label>
       <input id="match-option-zip-zip5match9" matInput formControlName="zip5match9" required title="" [errorStateMatcher]="instantErrorMatcher"/>
       <mat-error id="match-option-zip-zip5match9-error" *ngIf="form.get('zip5match9').invalid">A number is required</mat-error>
     </mat-form-field>
@@ -120,7 +120,7 @@
       matTooltipPosition="left"
       matTooltipShowDelay="500"
       matTooltipHideDelay="500" layout-fill>
-      <mat-label>9-Matches-5 Weight</mat-label>
+      <mat-label>Partial Match Weight</mat-label>
       <input id="match-option-zip-zip9match5" matInput formControlName="zip9match5" required title="" [errorStateMatcher]="instantErrorMatcher"/>
       <mat-error id="match-option-zip-zip9match5-error" *ngIf="form.get('zip9match5').invalid">A number is required</mat-error>
     </mat-form-field>
--- a/web/src/main/ui/app/components/flows-new/edit-flow/mastering/matching/ui/add-match-option-dialog.component.html
+++ b/web/src/main/ui/app/components/flows-new/edit-flow/mastering/matching/ui/add-match-option-dialog.component.html
@@ -111,7 +111,7 @@
       matTooltipPosition="left"
       matTooltipShowDelay="500"
       matTooltipHideDelay="500" layout-fill>
-      <mat-label>Five-vs-Nine Match Weight</mat-label>
+      <mat-label>5-vs-9 Match Weight</mat-label>
       <input id="match-option-zip-zip5match9" matInput formControlName="zip5match9" required title="" [errorStateMatcher]="instantErrorMatcher"/>
       <mat-error id="match-option-zip-zip5match9-error" *ngIf="form.get('zip5match9').invalid">A number is required</mat-error>
     </mat-form-field>
@@ -120,7 +120,7 @@
       matTooltipPosition="left"
       matTooltipShowDelay="500"
       matTooltipHideDelay="500" layout-fill>
-      <mat-label>Partial Match Weight</mat-label>
+      <mat-label>9-vs-5 Match Weight</mat-label>
       <input id="match-option-zip-zip9match5" matInput formControlName="zip9match5" required title="" [errorStateMatcher]="instantErrorMatcher"/>
       <mat-error id="match-option-zip-zip9match5-error" *ngIf="form.get('zip9match5').invalid">A number is required</mat-error>
     </mat-form-field>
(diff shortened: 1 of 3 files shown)


aebadirad:
Can we resolve this? It's been open for 3 weeks :O

wooldridge:
This change contradicts what is described here for the Zip algorithm:

https://marklogic-community.github.io/smart-mastering-core/docs/match-algorithms/#zip

@ryanjdew we need clarity on how the Zip algorithm works so we can document it properly, see:

https://project.marklogic.com/jira/browse/DHFPROD-2164

ryanjdew:
@wooldridge It works as described in the documentation at https://marklogic-community.github.io/smart-mastering-core/docs/match-algorithms/#zip with one slight error in the docs:

Effect:

If the original document has a 5-digit zip:
- A potential match with the same 5-digit zip will get 5 points (from the add).
- A potential match with a 9-digit zip that starts with the same five digits will get ~(5+3=)8~ 3 points. The add will not match a 9-digit zip code.

If the original document has a 9-digit zip:
- A potential match with the same 9-digit zip will get 5 points.
- A potential match with a 5-digit zip that matches the first five digits of the original document will get 2 points.

ayuwono:
> @wooldridge It works as described in the documentation at https://marklogic-community.github.io/smart-mastering-core/docs/match-algorithms/#zip with one slight error in the docs:
> 
> Effect:
> 
> If the original document has a 5-digit zip:
> 
> * A potential match with the same 5-digit zip will get 5 points (from the add).
> * A potential match with a 9-digit zip that starts with the same five digits will get ~(5+3=)8~ 3 points. The add will not match a 9-digit zip code.
> 
> If the original document has a 9-digit zip:
> 
> * A potential match with the same 9-digit zip will get 5 points.
> * A potential match with a 5-digit zip that matches the first five digits of the original document will get 2 points.

@ryanjdew , so this point will be added on top of the weights that we input on the UI?

ryanjdew:
@ayuwono 
Given the scenario of setting add with a weight of 10, origin 5-digit zip code weight of 7, and an origin 9-digit zip code with a weight of 3 there will be 3 possible weights.

1. Exact matches (5-digit = 5-digit or 9-digit = 9-digit) get 10 added to their score.
2. 5-digit zip matching 9-digit via a wildcard query will get 7 added to your score
3. 9-digit zip matching 5-digit via by truncating the 9-digit for query will get 3 added to your score

wooldridge:
Thanks so much for the clarification, @ryanjdew. 

Let's say I have a dataset of two documents (A and B) that only have a zip property. A Zip Match Option is defined as @ryanjdew describes above. 

Document A has a 5-digit zip and document B has a 9-digit zip, and the first 5 digits of each match.

I have a Match Threshold set to 10 for merging.

Comparing A against B (5-digit against 9-digit) yields a score of 7. 

Comparing B against A (9-digit against 5-digit) yields a score of 3. 

So will the total score generated for A and B be 

1. 10 so there's a merge? 
2. 3 for one comparison and 7 for another (separate) comparison so no merge?

If it's number 1, then what's the point of specifying different option values since they should always add up?

If it's number 2, what is the reason for separating the two option definitions? I don't understand why directionality matters here, maybe I'm missing something.

ryanjdew:
@wooldridge It is number 2. 

Honestly, I wasn't a part of the Labs team when the Zip algorithm was added so I don't know the whole use case. The expanding the query is certainly useful if you have mixed 5-digit/9-digit zip codes, but I see what you're getting at. I honestly can't think of a scenario where you wouldn't want exact, 5-digit to 9-digit, and 9-digit to 5-digit to all have the same weight.

ayuwono:
> Thanks so much for the clarification, @ryanjdew.
> 
> Let's say I have a dataset of two documents (A and B) that only have a zip property. A Zip Match Option is defined as @ryanjdew describes above.
> 
> Document A has a 5-digit zip and document B has a 9-digit zip, and the first 5 digits of each match.
> 
> I have a Match Threshold set to 10 for merging.
> 
> Comparing A against B (5-digit against 9-digit) yields a score of 7.
> 
> Comparing B against A (9-digit against 5-digit) yields a score of 3.
> 
> So will the total score generated for A and B be
> 
> 1. 10 so there's a merge?
> 2. 3 for one comparison and 7 for another (separate) comparison so no merge?
> 
> If it's number 1, then what's the point of specifying different option values since they should always add up?
> 
> If it's number 2, what is the reason for separating the two option definitions? I don't understand why directionality matters here, maybe I'm missing something.

agree with @wooldridge, both are kind of like partial match weight, maybe we can just use one field?

But, thanks for the explanation @ryanjdew

aebadirad:
So, what's the output of this? We need to resolve it.

wooldridge:
Given our discussion with @ryanjdew , we should update the Zip labels in Matching to the following:

- 5-vs-9 Match Weight
- 9-vs-5 Match Weight

Let me know if anyone thinks otherwise.

@jasonmy419 , these appear in the Match Options dialog for the Zip type and in the Match Options table under Other for Zip types. Looks like you need to add updates here:
```
match-options-ui.component.html
```
Why are there separate configuration settings for when comparing a 9-digit zip against a 5-digit zip and for when comparing a 5-digit zip against a 9-digit zip? That is still unclear. But that is how the feature works and how it needs to be documented.

@eltesoro-ml, could you fix the Smart Mastering documentation here so that it is accurate? 

https://marklogic-community.github.io/smart-mastering-core/docs/match-algorithms/#zip

At the bottom of the Zip section, the first bullet's second bullet should read:

A potential match with a 9-digit zip that starts with the same five digits will get 3 points.

wooldridge:
One more thing, update the tooltip, I did that here: https://github.com/jasonmy419/marklogic-data-hub/pull/2

eltesoro-ml:
@wooldridge: This solution doesn't address the commutativity issue that we talked about a while back. There is no way the user can tell whether the "original document" is the 5-digit zip or the 9-digit zip. So the same two zip codes could potentially return different weights depending on which is used as the "original document", which (I would guess) depends on the sort order, random conditions, the phase of the moon, and/or the alignment of the planets.

This solution would be fine if the user could tell when the 5-digit is the original and when the 9-digit is the original. But they cannot. Neither can we. So it's essentially a toss of a coin what weight they get.

I like what @ayuwono hinted at ("what is the reason for separating the two option definitions?"). Maybe we just need one UI field and, under the covers, we plug in the same weight value whether the original is the 5-digit or the 9-digit. That way, it becomes commutative through Data Hub, without requiring a change in the Smart Mastering Core code.

Yes, I can document it the way it behaves now, but I'm adding a very big warning that they should make the weights the same, unless they enjoy games of chance embedded in their business software. (Does that sound dire enough?)

(Sorry, I'm late responding. I'm still catching up on old emails.)

wooldridge:
I created a Jira ticket for reviewing and improving the Zip algorithm: https://project.marklogic.com/jira/browse/DHFPROD-2871

PR #2732, opened 2019-06-25
DHFPROD-2386: Add uri preview feature to ingestion step page

srinathgit:
Add uri preview feature to ingestion step page. I am creating a draft PR for review now and will rebase with Nikhil's changes (mlcp command) once they are merged

### DIFF
--- a/web/src/main/ui/app/components/flows-new/edit-flow/ingest/ui/ingest-ui.component.ts
+++ b/web/src/main/ui/app/components/flows-new/edit-flow/ingest/ui/ingest-ui.component.ts
@@ -90,6 +90,10 @@ const settings = {
   outputURIReplacement: {
     label: 'Target URI Replacement',
     description: 'Specify a prefix to prepend to the default URI. Used to construct output document URIs. For details, see Controlling Database URIs During Ingestion.'
+  },
+  outputURIPreview :{
+    label: 'Target URI Preview (Read-Only)',
+    description: 'Previews the URI of the ingested documents'
   }
 };
 interface MlcpOptions {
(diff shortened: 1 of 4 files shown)


web/src/main/ui/app/components/flows-new/edit-flow/ingest/ui/ingest-ui.component.ts
```diff
@@ -128,6 +134,75 @@ export class IngestUiComponent implements OnInit{
   }
 
   onChange() {
+    this.buildURIPreview();
     this.saveStep.emit(this.step);
   }
+  buildURIPreview(): void {
```
rjrudin: @srinathgit @wooldridge @aebadirad On the two largest projects I've been on, we never found the investment in automated testing of the UI to be worth the potential benefits. But I'm wondering how easily we could setup automated testing for logic like this, where the output can be verified without running a web browser or looking at any HTML. This is non-trivial "presentation logic" which in theory, we could write simple JS tests for that would run very quickly. 

It might be that only 10 to 20% of the total UI code would fit into this bucket, but if testing it were easy, I think it'd be worth the investment.

wooldridge: @rjrudin agreed, the new MLCP command on this page also fits into this bucket. @ayuwono, are we doing this?

srinathgit:
@wooldridge ,
I will make changes to error messages and add "Read-Only" to "Modules Database". With regards to trailing slash, I believe for csv , the target URI Replacement has to be /PATH/TO,'test/' as the regex replacement is done on the parent dir in case of csv files (whereas the regex replacement is done on abs path of the files in case of xml, json, binary or text

web/src/main/ui/app/components/flows-new/edit-flow/ingest/ui/ingest-ui.component.ts
ayuwono: @wooldridge, for the URI preview, I usually just verify the text as we input a predetermined value. For the MLCP, I think we will verify some subset of strings from the existing scenario that we have

bsrikan: To add to what Aries mentioned, both the mlcp command verification and uri preview verification can be part of unit test. I dont think they are candidates in an e2e test.

PR #2742, opened 2019-07-03
DHFPROD-2716:Create a xqy custom module, add "format" input switch to gradle command

srinathgit:
Create a xqy custom module, add "format" input switch to gradle command. The xqy module is not meant to be xquery implementation of datahub lib but only serves as an example for users to write their xquery modules


bsrikan:
Custom xqy module generation is tested. XQY module itself will be validated as part of a separate task.

PR #2743, opened 2019-07-03
DHFPROD-2534: validation of entities, default baseURI in UI

aebadirad:


marklogic-data-hub/src/test/resources/e2e-test/e2eentity.entity.json
```diff
@@ -7,7 +7,6 @@
   "definitions" : {
     "e2eentity" : {
       "required" : [ ],
-      "primaryKey": "id",
```
bsrikan: I am trying to check how this primaryKey can make the entity invalid. Will report back.

bsrikan:
I think somehow we have introduced an expectation that baseUri is a required property. Entity Services does not mandate it. If missing, entity services uses a default that is of the form "http://example.org/"

aebadirad:
Correct. We are going to begin enforcing it for data hub.

marklogic-data-hub/src/test/resources/e2e-test/e2eentity.entity.json
bsrikan: There seems to be an issue with json tests. xml tests are passing ok. I dont think there is anything wrong with the entity having primaryKey.
The error for all the failed MappingE2E tests is:
`Info: Status 500: TDE-INDEX: xdmp:eval("xdmp.documentInsert(\"/8390440750948294347.json\",trace,xdmp.def...", {trace:{trace:{jobId:"10e61e72-45ac-4545-8651-5d4f74bd5880", format:"json", traceId:"8390440750948294347", ...}}}, {database:"15450808607671818307", commit:"auto", update:"true", ...}) -- Error applying template /tde/e2eentity-0.0.1.tdex to document /8390440750948294347.json: TDE-EVALFAILED: Eval for $subject-iri='sem:iri(concat("http://www.marklogic.com/e2eentity-0.0.1/e2eentity/", fn:encode-for-uri(./id)))' returns TDE-BADVALEXPRESSION: Invalid val expression: XDMP-ARGTYPE: (err:XPTY0004) arg1 is not of type xs:string?`

aebadirad:
The error is most likely from the scope of the TDE, as it's looking at the trace documents generated, not just the final entities themselves.

The removal of the PK was because a) Pk wasn't being used for that entity by the tests and b) the pk made it so the trace documents matched.

We could (and should) look at modifying the TDE to ignore anything not in the proper entity collection.

bsrikan:
Thanks for the details Alex. I think you are right about PK not being used in the test. Will test again once the TDE fix is in place.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/4/triggers/entity-model-validate-trigger.xqy
```diff
@@ -0,0 +1,86 @@
...
+(: Create map version so we can update $ref to external links if we have to :)
+let $entity-def-map as map:map := $entity-def
+let $entity-base-uri := $entity-def/info/baseUri
+let $entity-title := $entity-def/info/title
+let $entity-version := $entity-def/info/version
+let $validate-base-uri := try {
+    xdmp:json-validate-node($entity-def/info, xdmp:unquote('{type: "object", properties: {"baseUri":{type:"string", format:"uri"}}, required: ["baseUri"]}'), "strict")
+  } catch ($ex) {
+    if(($ex/error:code eq "XDMP-JSVALIDATEINVFORMAT" or $ex/error:code eq "XDMP-JSVALIDATEMISSING") and fn:contains($ex/error:format-string/text(), "baseUri")) then (
+      fn:error(fn:QName("error","XDMP-JSVALIDATEINVFORMAT"),fn:concat("A valid baseUri is required (ex: http://www.example.org/) for entity: ", $entity-title, "."), fn:concat("Error Message: ", $ex/error:format-string/text()))
```
rjrudin: I tested this out on a new project with a single entity model file with an invalid baseUri. When I install the app using QS, the deployment fails because of the error thrown by the entity validation trigger, though it happens late enough in the process that QS still reports back to the user that the installation succeeds. 

The entity shows up in the UI though - QS still loads the file. But the entity document isn't in staging or final (nor are any flows or steps). 

If I edit the entity, there isn't any indication that there's an error with the Base URI. If I save it, I get the error message on Base URI. 

So while this prevents a user from creating an entity with a bad Base URI, I agree with @bsrikan's comment about this being a breaking change, and it seems like it'll be difficult for a user to figure out that the breaking change occurs if they're using QS.

rjrudin: Also, I'd make this message as similar as possible to the one provided in QS for consistency purposes.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/4/triggers/entity-model-validate-trigger.xqy
```diff
@@ -0,0 +1,86 @@
...
+        $ref-entity
+      (: Otherwise, update to be external reference URI to ensure we still have a valid entity model :)
+      else
+        $prop => map:put("$ref", $entity-uri)
+  )
+let $info-map as map:map := $entity-def-map => map:get("info")
+(: build uber model with original info :)
+let $uber-model := hent:uber-model((xdmp:to-json($entity-def-map)/object-node(), $supporting-entity-defs)) => map:with("info", $info-map)
+return (
+  es:model-validate($uber-model)
```
rjrudin: I think this is all new, right? If so, it'd be worth making unit tests to cover some of the scenarios in this code. We don't need to test that es:model-validate works, but there's other stuff going on in here.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/4/triggers/entity-model-validate-trigger.xqy
aebadirad: The UI just doesn't handle errors.

It's not a 'breaking change' anymore than anytime a deploy fails silently in QS.

I'd imagine the UI to report failed save attempts to be its own ticket, as this doesn't modify any of that behavior.

At a certain point we have to decide on a) allowing users to insert bad entities like we do now, which causes failed tdes and models which breaks then during flow running or b) force compliance.

It's not a breaking change anymore than it used to be. We get an issue a day about invalid models silently not being deployed.

rjrudin: That's fair. I think it's worth making a note of in our release notes though - i.e. that if you have entity models without a base URI, you'll need to add one. What's the best way to ensure something is added to the release notes? Pinging @sbayatpur and @eltesoro-ml

marklogic-data-hub/src/test/resources/e2e-test/e2eentity.entity.json
aebadirad: This is due to invalid scoping of the tde's generated by ES. Ugh.

ayuwono:
> Jenkins Build failure
> https://jenkins.marklogic.com/job/5.x/job/dhf-QS-develop-rh7-Commit/2589/testReport
> 183 tests run, 18 skipped, 14 failed.

Tests are failing because of invalid base URI tests. @timur-isangulov , can you change the base URI to a valid one on these tests? It needs to be like http://example.org/something/. I think you can make a commit against this branch. @aebadirad , can you confirm that the base URI needs to be trailed by a slash ('/')

Screen Shot 2019-07-12 at 1 23 25 PM

timur-isangulov:
I added formated Base URIs to the tests.

rjrudin:
@aebadirad Is this ready for review now or still in-progress?

aebadirad:
It would be if I could get a this test to stop failing whereas it passes locally.

aebadirad:
There's a delta between nightly and released at the moment. I'll address this shortly.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/4/triggers/entity-model-validate-trigger.xqy
```diff
@@ -0,0 +1,86 @@
...
+(: Create map version so we can update $ref to external links if we have to :)
+let $entity-def-map as map:map := $entity-def
+let $entity-base-uri := $entity-def/info/baseUri
+let $entity-title := $entity-def/info/title
+let $entity-version := $entity-def/info/version
+let $validate-base-uri := try {
+    xdmp:json-validate-node($entity-def/info, xdmp:unquote('{type: "object", properties: {"baseUri":{type:"string", format:"uri"}}, required: ["baseUri"]}'), "strict")
+  } catch ($ex) {
+    if(($ex/error:code eq "XDMP-JSVALIDATEINVFORMAT" or $ex/error:code eq "XDMP-JSVALIDATEMISSING") and fn:contains($ex/error:format-string/text(), "baseUri")) then (
+      fn:error(fn:QName("error","XDMP-JSVALIDATEINVFORMAT"),fn:concat("A valid Base URI is required (e.g. http://example.org/) for entity: ", $entity-title, "."), fn:concat("Error Message: ", $ex/error:format-string/text()))
```
rjrudin: Just confirming - I tested this out by deploying an app with an existing entity definition that doesn't conform to the new requirements, and the deployment failed (as expected) with this message:

```
* What went wrong:
Execution failed for task ':mlDeployApp'.
> Local message: failed to apply resource at documents: Internal Server Error. Server Message: A valid Base URI is required (e.g. http://example.org/) for entity: Customer. (XDMP-JSVALIDATEINVFORMAT): Error Message: XDMP-JSVALIDATEINVFORMAT: Invalid node value: Node value is not valid per format uri, found invalid at baseUri:fn:doc("/entities/Customer.entity.json")/info/text("baseUri") using schema "" . See the MarkLogic server error log for further detail.
```

I think a hard failure is good there, this will hopefully help people catch this problem quickly for existing entity definitions.

web/src/main/ui/app/components/entity-modeler/entity-editor/entity-editor.component.html
```diff
@@ -19,7 +19,8 @@
       <mdl-textfield type="text" label="Description" [(ngModel)]="entity.info.description" floating-label></mdl-textfield>
     </div>
     <div>
-      <mdl-textfield type="text" label="Base URI" [(ngModel)]="entity.info.baseUri" floating-label></mdl-textfield>
+      <mdl-textfield type="text" label="Base URI" required [(ngModel)]="entity.info.baseUri" floating-label></mdl-textfield>
+      <span *ngIf="!validBaseUri" class="alert-text">A valid Base URI is required, e.g. http://example.org/</span>
```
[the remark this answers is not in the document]
bsrikan: I tested this part with baseURI and the title field. The alert text behavior seems to be consistent ie the error message goes away only after hitting Save button. Able to click on Save after correcting the baseURI. To me Save button is enabled at all times.

bsrikan:
@aebadirad can you please rebase with develop. Merge is blocked because of that.

aebadirad:
See #2814

PR #2745, opened 2019-07-05
Latest mastering changes with performance tracing and enhancements

ryanjdew:
Items addressed:

- Adds SM-PERFORMANCE trace to better track where time is spent
- Minor performance enhancements in reducing the number of searches
- Moves merge options code into a separate library module at /com.marklogic.smart-mastering/survivorship/merging/options.xqy to reduce /com.marklogic.smart-mastering/survivorship/merging/base.xqy size
- Adds a helper function for building queries to reduce the duplicate code to determine whether json-property, element-value, or range queries should be used in the match algorithms
- Normalizes the output from a merge JavaScript function for the user, so the user can output array instead of sequence and JSON values will be converted to nodes for the user.

These changes current exist in the develop branch of smart-mastering-core (https://github.com/marklogic-community/smart-mastering-core/tree/develop) with all unit test passing.

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/com.marklogic.smart-mastering/algorithms/double-metaphone.xqy
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/com.marklogic.smart-mastering/algorithms/double-metaphone.xqy
@@ -8,6 +8,9 @@ import module namespace spell = "http://marklogic.com/xdmp/spell"
 import module namespace const = "http://marklogic.com/smart-mastering/constants"
   at "/com.marklogic.smart-mastering/constants.xqy";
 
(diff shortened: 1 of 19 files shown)


ryanjdew:
@rjrudin I was testing with the attached QConsole Workspace and with the smart-mastering-core marklogic-unit-test app server.
[SM - Optimization.xml.zip](https://github.com/marklogic/marklogic-data-hub/files/3369893/SM.-.Optimization.xml.zip)

rjrudin:
I wasn't able to get the workspace working - the scripts depend on a /test/suites/large-match-and-merge/lib/lib.xqy module that isn't present - but the existing unit tests and MasterTest pass.

PR #2754, opened 2019-07-08
DHFPROD-2301: Fix the percent complete value for csv ingestion

srinathgit:
Fix the percent complete value for csv ingestion

### DIFF
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/step/impl/WriteStepRunner.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/step/impl/WriteStepRunner.java
@@ -52,7 +52,6 @@ import java.util.*;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicBoolean;
-import java.util.concurrent.atomic.AtomicLong;
 import java.util.stream.Stream;
 
 public class WriteStepRunner implements StepRunner {
@@ -64,6 +63,8 @@ public class WriteStepRunner implements StepRunner {
     private DatabaseClient stagingClient;
     private String destinationDatabase;
     private int previousPercentComplete;
+    protected long csvFilesProcessed;
+    private String currentCsvFile;
     private Map<String, Object> options;
     private boolean stopOnFailure = false;
     private String jobId;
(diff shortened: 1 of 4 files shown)


hao1st:
https://github.com/marklogic/marklogic-data-hub/blob/5d9f58fa6e4aef9acc1817a226382ab31a682c0b/marklogic-data-hub/src/main/java/com/marklogic/hub/step/impl/WriteStepRunner.java#L770-L789

It should be feasible to combine the two functions into one...

marklogic-data-hub/src/main/java/com/marklogic/hub/step/impl/WriteStepRunner.java
```diff
@@ -760,6 +776,17 @@ private void runStatusListener(long totalRunBatches, double batchCount, AtomicLo
             });
         }
     }
+    //percentComplete for csv files is (csvFilesProcessed/ urisCount) * 100.0
+    //The number of csv files would probably be less than that of regular files, so the step status listeners are updated more frequently
+    private void runCsvStatusListener(double urisCount, AtomicLong successfulEvents, AtomicLong failedEvents) {
```
rjrudin: I think it's worth making this protected so that some unit tests can be written against it. You can call onStatusChanged on WriteStepRunner and pass in a test implementation of StepStatusListener. That implementation can save the args that are passed to it by this code, and your test can then verify that it received the expected args.

marklogic-data-hub/src/main/java/com/marklogic/hub/step/impl/WriteStepRunner.java
```diff
@@ -476,7 +478,12 @@ private RunStepResponse runIngester(RunStepResponse runStepResponse, Collection<
                 successfulEvents.addAndGet(batch.getItems().length-1);
                 successfulBatches.addAndGet(1);
                 logger.debug(String.format("Current SuccessfulEvents: %d - FailedEvents: %d", successfulEvents.get(), failedEvents.get()));
-                runStatusListener(successfulBatches.get()+failedBatches.get(), batchCount, successfulEvents, failedEvents);
+                if("csv".equalsIgnoreCase(inputFileType)) {
+                   runCsvStatusListener(uris.size(), successfulEvents, failedEvents);
+                }
+                else {
+                    runStatusListener(successfulBatches.get()+failedBatches.get(), batchCount, successfulEvents, failedEvents);
```
rjrudin: A refactor to consider is that the successfulEvents, failedEvents, successfulBatches, and failedBatches variables are often passed to other methods. That's a good indication that they should be grouped together into a new object class - e.g. "StepStatus". 

That would also help avoid the duplication of this code block. You could instead have a method like this:

protected void runStatusListener(String inputFileType, Collection<String> uris, StepStatus stepStatus)

(I don't think batchCount needs to be passed in, as that can be quickly calculated whenever it's needed). 

That method would check to see if inputFileType == csv. You could then call that method here and in the block below so that the if/else isn't duplicated.

marklogic-data-hub/src/main/java/com/marklogic/hub/step/impl/WriteStepRunner.java
```diff
@@ -576,27 +574,27 @@ private RunStepResponse runIngester(RunStepResponse runStepResponse, Collection<
...
-                listener.onStatusChange(runStepResponse.getJobId(), 100, stepStatus, successfulEvents.get(), failedEvents.get(), "Ingestion completed");
+                listener.onStatusChange(runStepResponse.getJobId(), 100, stepStatus, stepMetrics.getSuccessfulEventsCount(), stepMetrics.getFailedEventsCount(), "Ingestion completed");
             });
 
             stepFinishedListeners.forEach((StepFinishedListener::onStepFinished));
 
             dataMovementManager.stopJob(writeBatcher);
 
-            runStepResponse.setCounts(successfulEvents.get() + failedEvents.get(),successfulEvents.get(), failedEvents.get(), successfulBatches.get(), failedBatches.get());
+            runStepResponse.setCounts(stepMetrics.getSuccessfulEventsCount() + stepMetrics.getFailedEventsCount(),stepMetrics.getSuccessfulEventsCount(), stepMetrics.getFailedEventsCount(), stepMetrics.getSuccessfulBatchesCount(), stepMetrics.getFailedBatchesCount());
```
rjrudin: No need to change now, but I think it'd be preferable to pass the StepMetrics object into "setCounts", and the RunStepResponse can figure out how to do the math.

PR #2764, opened 2019-07-09
DHFPROD-2674: Fixes for deploying user files to DHS

rjrudin:
8000 isn't available, so defaulting appServicesPort to finalPort.

Also, was running into some issues where timestamps files were preventing modules from being loaded. When running dhsDeploy, we rarely will have a desire to not load all the modules, so the module timestamps path is set to null.

### DIFF
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/impl/DataHubImpl.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/impl/DataHubImpl.java
@@ -577,17 +577,23 @@ public class DataHubImpl implements DataHub {
     }
 
     public void dhsInstall(HubDeployStatusListener listener) {
-        prepareAppConfigForInstallingIntoDhs(hubConfig.getAppConfig());
+        prepareAppConfigForInstallingIntoDhs(hubConfig);
 
         HubAppDeployer dhsDeployer = new HubAppDeployer(getManageClient(), getAdminManager(), listener, hubConfig.newStagingClient());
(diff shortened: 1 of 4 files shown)


aebadirad:
Can we swap this from 8011 to 'staging' and use whatever staging is? I'd like to future proof this.

bsrikan:
I have verified using mlAPpServicePort=8010 as well. dhsDeploy task deploys user modules as expected

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/DataHubImpl.java
```diff
@@ -577,17 +577,23 @@ protected void disableSomeCmaUsage(AppConfig appConfig) {
...
+    protected void prepareAppConfigForInstallingIntoDhs(HubConfig hubConfig) {
+        AppConfig appConfig = hubConfig.getAppConfig();
+
+        appConfig.setModuleTimestampsPath(null);
         appConfig.setCreateForests(false);
         appConfig.setResourceFilenamesIncludePattern(buildPatternForDatabasesToUpdateIndexesFor());
         disableSomeCmaUsage(appConfig);
+
+        // 8000 is not available in DHS
+        appConfig.setAppServicesPort(hubConfig.getPort(DatabaseKind.FINAL));
```
aebadirad: STAGING please, FINAL may not be always a thing.

rjrudin: Done!

PR #2767, opened 2019-07-09
DHFPROD-2780: The parameter "output_collections" of MLCP command in the ingest step, needs to be updated.

xnikhil08:
Updated the -output_collections parameter to include step name and removed flow name & input collection names.


xnikhil08:
@ayuwono - Updated the output collections under this ticket already. Please review and approve this.

ayuwono:
@xnikhil08 , I think you need to put a colon ":" on between DHFPROD id and the title, so this PR can trigger the CI jenkins pipeline

PR #2769, opened 2019-07-10
DHFPROD-2502: The sourceContext in a new mapping is now always "/"

rjrudin:
I updated all of the test files to have "/" as well.


rjrudin:
> This all looks good, but I don't get why there's a failing test. We cycled logs on it, so we should probably figure out the single failure before merging.

I ran EntityManagerServiceTest locally and it passed. It passed in the previous runs for this PR as well. The error is due to not being to delete ye-olde-test-project/.tmp - that strikes me as an intermittent environment-based issue.

PR #2774, opened 2019-07-12
DHFPROD-2725: custom step in Insurance example

bsrikan:
New flow is added to the existing Insurance example and can be run by itself to demonstrate the custom step. It doesnt depend on other flows in the project.

### DIFF
--- a/examples/insurance/README.md
+++ b/examples/insurance/README.md
@@ -5,6 +5,11 @@
 - Sunrise Insurance has bought two other insurance companies, Advantage Insurance and Bedrock Insurance.
 - Each company has a database of customers; in each database, customers are represented differently.
 - Sunrise Insurance wants to integrate the customers from Advantage and Bedrock with its own.
+- Sunrise Insurance wants to enrich Advantage Insurance data via a custom step called enrichAdvantage that
+    - Takes the U.S. five-digit zip code from a customer document
+    - Looks up the corresponding latitude and longitude coordinates
+      - This is achieved by calling a library(zipcodeData.sjs) invoked within the custom step 
+    - Writes those coordinates as new document properties
 
 ## Predefined Flows
 
@@ -13,6 +18,10 @@ The project has flows predefined for integrating the customer data.
 - **AdvantageFlow**: Has steps for ingesting and mapping Advantage customer data.
 - **BedrockFlow**: Has steps for ingesting and mapping Bedrock customer data.
 - **CustomerMastering**: Has a mastering step for matching and merging duplicate customers across the Advantage and Bedrock datasets.
+
+There is also a flow to enrich data
+
+- **customEnrichment**: Has steps for ingestion and a custom step to enrich customer data
 
 You can finish configuring the flows and run the steps to complete the integration.
 
@@ -29,6 +38,14 @@ You can finish configuring the flows and run the steps to complete the integrati
 9. Run the `BedrockMap` step in the `BedrockFlow` flow. This harmonizes the 100 Bedrock customer documents into the final database. You can view the documents in the Browse Data view.
 10. Run the `CustomerMaster` flow to master the Advantage and Bedrock customer data. This merges documents for two matching customers in the final database. You can view the results in the Browse Data view.
 
+
+## How to enrich Customer Data using custom step
+
+1. Start the Data Hub and select the project folder: examples/insurance
+2. View the `customEnrichment` flow. Configure the `ingestAdvantage` step by setting the Source Directory Path to the `datasets/advantage` directory (exact path will depend on your filesystem).
+3. The scaffolded custom main.sjs has been edited(line 90-94) to insert code that manipulates the instance to insert geo-spatial information corresponding to the postal code
+4. The uri has also been manipulated(line 103) and is appended with "/enriched" in the FINAL database  
+5. You can now click on **Run** button to run the flow and verify enriched data in FINAL database
 
 ## Example Customer Data
 
@@ -47,6 +64,26 @@ You can finish configuring the flows and run the steps to complete the integrati
   "Phone": "(870) 409-2724",
   "PIN": 6454,
   "Updated": "2015-05-17T08:24:16"
+}
+```
+
+### Snippet of enriched data
+
+```
+"instance": {
+  "ObjectID": {
+    "$oid": "5cd0da4d4162c033d57dc2f6"
+  },
+  "CustomerID": "bc0dd434-232b-4ff9-bef9-c0cd2fcb72be",
+  "FirstName": "Camille",
+  "LastName": "Case",
+  "Email": "camillecase@comvex.com",
+  "Postal": "45348-4317",
+  "Phone": "(935) 438-3459",
+  "PIN": 4019,
+  "Updated": "2018-02-12T03:33:41",
+  "latitude": "40.316833",
+  "longitude": "-84.633911"
 }
 ```
 
--- a/examples/insurance/README.md
+++ b/examples/insurance/README.md
@@ -27,7 +27,7 @@ You can finish configuring the flows and run the steps to complete the integrati
 
 ## How to Integrate the Customer Data
 
-1. Start the Data Hub and select the project folder: examples/insurance
+1. Using Data Hub QuickStart with a clean MarkLogic server, select and install the project folder: examples/insurance
 2. View the `AdvantageFlow` flow. Configure the `AdvantageIngest` step by setting the Source Directory Path to the `datasets/advantage` directory (exact path will depend on your filesystem).
 3. Ingest the Advantage dataset by running the `AdvantageIngest` step. This ingests 100 Advantage customer documents into the staging database. You can view the documents in the Browse Data view.
 4. With the Advantage customer data ingested, configure the `AdvantageMap` step in the `AdvantageFlow` flow.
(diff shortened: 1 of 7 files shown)


aebadirad:
Really nice :)

examples/insurance/src/main/ml-modules/root/custom-modules/custom/enrichAdvantage/main.sjs
```diff
@@ -0,0 +1,114 @@
...
+      headers: datahub.flow.flowUtils.getHeaders(doc) || {},
+      triples: datahub.flow.flowUtils.getTriples(doc) || [],
+      instance: datahub.flow.flowUtils.getInstance(doc)
+    }
+  };
+   
+   
+  //insert code to manipulate the instance, triples, headers, uri, context metadata, etc.
+     
+  if (instance["Postal"]) {
```
rjrudin: This is the only interesting part of the module. I think we need some changes to custom steps so that a developer doesn't need much code besides this.

examples/insurance/src/main/ml-modules/root/custom-modules/custom/enrichAdvantage/main.sjs
```diff
@@ -0,0 +1,114 @@
...
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+*/
+
+const DataHub = require("/data-hub/5/datahub.sjs");
+const datahub = new DataHub();
+// Require module for mapping latitude and longitude from zip codes
+const zipcodeData = require("/custom-modules/utils/zipcodeData.sjs");
```
rjrudin: This file is fairly large for adding to version control - I recommend slimming it down to a handful of zip codes that will guarantee some matches against our ingested data.

examples/insurance/flows/customEnrichment.flow.json
```diff
@@ -0,0 +1,69 @@
...
+            "name" : "customEnrichment"
+          } ],
+          "createdOn" : "currentDateTime",
+          "createdBy" : "currentUser"
+        },
+        "sourceQuery" : "cts.collectionQuery([])",
+        "collections" : [ "ingestAdvantage" ],
+        "permissions" : "rest-reader,read,rest-writer,update",
+        "outputFormat" : "json",
+        "targetDatabase" : "insurance-STAGING"
```
rjrudin: I cloned this locally, and in QuickStart, I had to initialize the project - that's because gradle.properties isn't checked in. That results in the default database names being used. When I run this flow, I get an error because "insurance-STAGING" doesn't exist. So I think this example project should assume that "data-hub-STAGING" and the other default names will be used as well.

examples/insurance/src/main/ml-modules/root/custom-modules/custom/enrichAdvantage/main.sjs
```diff
@@ -0,0 +1,114 @@
...
+  let id = content.uri;
+   
+  //here we can grab and manipulate the context metadata attached to the document
+  let context = content.context;
+   
+  //let's set our output format, so we know what we're exporting
+  let outputFormat = options.outputFormat ? options.outputFormat.toLowerCase() : datahub.flow.consts.DEFAULT_FORMAT;
+   
+  //here we check to make sure we're not trying to push out a binary or text document, just xml or json
+  if (outputFormat !== datahub.flow.consts.JSON && outputFormat !== datahub.flow.consts.XML) {
```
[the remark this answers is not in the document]
rjrudin: I simplified the module to what's below, and it still worked correctly. Maybe it's too much, but I think generally, the less code, the better. 

```
let doc = content.value;
  if (doc && (doc instanceof Document || doc instanceof XMLDocument)) {
    doc = fn.head(doc.root);
  }

  let instance = datahub.flow.flowUtils.getInstance(doc).toObject() || {};
  instance['$attachments'] = doc;

  if (instance["Postal"]) {
    let zipcode = instance["Postal"].substring(0, 5);
    instance["latitude"] = zipcodeData.getLatitude(zipcode);
    instance["longitude"] = zipcodeData.getLongitude(zipcode);
  }

  let headers = datahub.flow.flowUtils.getHeaders(doc) || {};
  let triples = datahub.flow.flowUtils.getTriples(doc) || [];
  let outputFormat = options.outputFormat ? options.outputFormat.toLowerCase() : datahub.flow.consts.DEFAULT_FORMAT;

  content.value = datahub.flow.flowUtils.makeEnvelope(instance, headers, triples, outputFormat);

  return content;
```

examples/insurance/flows/customEnrichment.flow.json
```diff
@@ -0,0 +1,69 @@
...
+        "inputFilePath" : "/Users/sbalasub/Desktop/ML/DHF/QA/insurance/datasets/advantage",
+        "inputFileType" : "json",
+        "outputURIReplacement" : ".*datasets,''",
+        "separator" : ","
+      },
+      "mappingStep" : false,
+      "customStep" : false,
+      "mappingName" : null
+    },
+    "2" : {
```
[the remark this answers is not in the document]
rjrudin: I was able to simplify the step definition to the following:

```
{
  "language": "zxx",
  "name": "enrichAdvantage",
  "description": "custom step to enrich Advantage customer data",
  "type": "CUSTOM",
  "version": 1,
  "modulePath": "/custom-modules/custom/enrichAdvantage/main.sjs"
}
```

And then the flow definition has this:

```
"2" : {
      "name" : "enrichAdvantage",
      "description" : "custom step to enrich Advantage customer data",
      "options" : {
        "additionalCollections" : [ ],
        "sourceQuery" : "cts.collectionQuery([\"ingestAdvantage\"])",
        "targetEntity" : "Customer",
        "collections" : [ "enrichAdvantage", "Customer" ],
        "sourceDatabase" : "data-hub-STAGING",
        "sourceCollection" : "ingestAdvantage",
        "outputFormat" : "json",
        "targetDatabase" : "data-hub-FINAL"
      },
      "customHook" : null,
      "retryLimit" : 0,
      "batchSize" : 100,
      "threadCount" : 4,
      "stepDefinitionName" : "enrichAdvantage",
      "stepDefinitionType" : "CUSTOM",
      "mappingStep" : false,
      "customStep" : true,
      "mappingName" : null
    }
```

I don't know if that's the best separation though. To me, the step definition should know as little as possible, other than what the custom module is. The flow then determines how to incorporate the step.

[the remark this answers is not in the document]
bsrikan: I used QS to create these, so its the default config we create. Seems like we can make some improvement to the artifacts that QS generates.

examples/insurance/flows/customEnrichment.flow.json
bsrikan: Agree. I will make that change so it uses default app name.

examples/insurance/src/main/ml-modules/root/custom-modules/custom/enrichAdvantage/main.sjs
bsrikan: Yes, there are quite a few postal code not covered and we get null. I will reduce it to around 100 or so to match ingested data.

examples/insurance/src/main/ml-modules/root/custom-modules/custom/enrichAdvantage/main.sjs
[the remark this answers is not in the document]
bsrikan: I didnt want to cut out anything from the generated module. Maybe I should to make it lean and only have code that is relevant to the example.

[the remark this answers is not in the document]
rjrudin: I think it's better for sake of example to lean it out. The generated module will have all the boilerplate in it, so a user can keep that if they want. But for an example, I think simplicity is key. Docs can be exhaustive.

examples/insurance/flows/customEnrichment.flow.json
bsrikan: I will be pushing changes in a bit without the gradle.properties and let QS initialize the project. So that whenever some tries it they will have relevant project properties initialized at the time. This also avoids us having to update the gradle.properties with mlDHFVersion every release.

examples/insurance/README.md
```diff
@@ -30,6 +39,14 @@ You can finish configuring the flows and run the steps to complete the integrati
 10. Run the `CustomerMaster` flow to master the Advantage and Bedrock customer data. This merges documents for two matching customers in the final database. You can view the results in the Browse Data view.
 
 
+## How to enrich Customer Data using custom step
+
+1. Start the Data Hub and select the project folder: examples/insurance
```
rjrudin: I think it'd be better for this to say "Start QuickStart and select..." so that it's clear that the user needs to run QuickStart. 

Probably best to include a note saying that the user will need to install the application as well, which usually requires starting with a clean MarkLogic instance that doesn't have a data hub already installed.

examples/insurance/README.md
```diff
@@ -30,6 +39,14 @@ You can finish configuring the flows and run the steps to complete the integrati
...
+2. View the `customEnrichment` flow. Configure the `ingestAdvantage` step by setting the Source Directory Path to the `datasets/advantage` directory (exact path will depend on your filesystem).
+3. The scaffolded custom main.sjs has been edited to insert code that manipulates the instance to insert geo-spatial information corresponding to the postal code
```
rjrudin: I recommend including the full path here, as I think most readers won't know where to find this otherwise. And I think you should also explain how you generated the scaffolded main.sjs file, as I don't think we can assume that the reader knows how to do that part either.

examples/insurance/src/main/ml-modules/root/custom-modules/custom/enrichAdvantage/main.sjs
```diff
@@ -0,0 +1,78 @@
...
+   
+  //grab the 'doc' from the content value space
+  let doc = content.value;
+   
+  // let's just grab the root of the document if its a Document and not a type of Node (ObjectNode or XMLNode)
+  if (doc && (doc instanceof Document || doc instanceof XMLDocument)) {
+    doc = fn.head(doc.root);
+  }
+   
+  //get our instance, default shape of envelope is envelope/instance, else it'll return an empty object/array
```
rjrudin: I recommend removing most of these comments, as they don't add any value, they're just explaining what the code is clearly doing. They make the module seem more complicated than it really is.

examples/insurance/src/main/ml-modules/root/custom-modules/custom/enrichAdvantage/main.sjs
```diff
@@ -0,0 +1,78 @@
...
+  let instance = datahub.flow.flowUtils.getInstance(doc).toObject() || {};
+   
+  // get triples, return null if empty or cannot be found
+  let triples = datahub.flow.flowUtils.getTriples(doc) || [];
+   
+  //gets headers, return null if cannot be found
+  let headers = datahub.flow.flowUtils.getHeaders(doc) || {};
+   
+  //If you want to set attachments, uncomment here
```
rjrudin: Definitely remove these two commented lines since they don't apply.

examples/insurance/src/main/ml-modules/root/custom-modules/custom/enrichAdvantage/main.sjs
```diff
@@ -0,0 +1,78 @@
...
+  //instance['$attachments'] = doc;
+  instance['$attachments'] = {
+    envelope: {
+      headers: datahub.flow.flowUtils.getHeaders(doc) || {},
```
rjrudin: This can be simplified, since you've already obtained the headers/triples/instance - no need to call the flowUtils functions again.

bsrikan:
Thanks @wooldridge for the comments. Can you please make inline comments for the README updates with steps 2 and 3. I am not able to place them well.

bsrikan:
Can we get this PR approved now and merged to develop?

bsrikan:
> I'm checking the README on your develop branch here:
> 
> https://github.com/bsrikan/marklogic-data-hub/tree/develop/examples/insurance
> 
> Minor edits:
> 
>     * Add a period (.) to the end of the steps under "How to Enrich Customer Data Using Custom Step".
> 
>     * Change "2" to "two" in step #2.

Thanks @wooldridge for checking again, but these are really trivial and I have already spent more time on this than required.

PR #2775, opened 2019-07-12
DHFPROD-2356: Quickstart UI Automation on jobs page validation

timur-isangulov:
QS Jobs page validation tests.

### DIFF
--- a/web/e2e/page-objects/appPage.ts
+++ b/web/e2e/page-objects/appPage.ts
@@ -92,6 +92,14 @@ export class AppPage extends Page {
     await browser.waitForAngular();
   }
 
+  async clickJobsTab() {
+    return await browser.executeScript("arguments[0].click();", this.jobsTab);
+  }
+
+  async clickSettingsTab() {
+    await browser.executeScript("arguments[0].click();", this.settingsTab);
+    await browser.waitForAngular();
+  }
 }
 
 var appPage = new AppPage();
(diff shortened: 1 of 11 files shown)


web/e2e/page-objects/jobs/jobs.ts
```diff
@@ -148,6 +156,64 @@ export class JobsPage extends AppPage {
   jobOutputCloseButton() {
     return element(by.css('app-job-output mdl-button'));
   }
+
+
+  resetFiltersButton() {
```
ayuwono: some of these page objects like filtering are already exist on manageJobs.ts. Can you just delete jobs.ts and use manageJobs.ts, so we are consistent with manageFlows structure. Also, you can re-use the page objects on manageJobs.ts, we use these page objects on scenario tests. If you make some changes, make sure that it's not breaking the scenarios tests

web/e2e/page-objects/appPage.ts
```diff
@@ -92,6 +92,14 @@ export class AppPage extends Page {
     await browser.waitForAngular();
   }
 
+  async clickJobsTab() {
+    return await browser.executeScript("arguments[0].click();", this.jobsTab);
+  }
```
ayuwono: add await browser.waitForAngular() too here?

timur-isangulov: Added. Thanks.

web/e2e/page-objects/jobs/jobs.ts
timur-isangulov: I moved methods to manageJobs.ts page object, will review and remove unused classes later.

PR #2785, opened 2019-07-16
DHFPROD-2795:Page validation on MLCP Command

timur-isangulov:
1. Added new tests and verification in the existing tests for the MLCP command.
2. Fixed two ingestion tests that fail on Jenkins.

### DIFF
--- a/web/e2e/specs/flows/steps/ingestion.ts
+++ b/web/e2e/specs/flows/steps/ingestion.ts
@@ -113,8 +113,6 @@ export default function (qaProjectDir) {
       await manageFlowPage.createFlow(flowData.flow1);
       await editFlowPage.addStep(flowData.flow1, json);
       await browser.sleep(5000);
-      await expect(ingestStepPage.mlcpCommand.getText()).toContain("json");
-      await expect(ingestStepPage.mlcpCommand.getText()).toContain("-input_file_type \"documents\"");
       await editFlowPage.clickRunFlowButton();
       await browser.sleep(3000);
       await browser.wait(EC.visibilityOf(editFlowPage.runFlowHeader));
@@ -135,8 +133,6 @@ export default function (qaProjectDir) {
       await manageFlowPage.createFlow(flowData.flow2);
       await editFlowPage.addStep(flowData.flow2, xml);
       await browser.sleep(5000);
-      await expect(ingestStepPage.mlcpCommand.getText()).toContain("json");
-      await expect(ingestStepPage.mlcpCommand.getText()).toContain("-input_file_type \"documents\"");
       await editFlowPage.clickRunFlowButton();
       await browser.wait(EC.visibilityOf(editFlowPage.runFlowHeader));
       await editFlowPage.clickButtonRunCancel("flow");
@@ -253,8 +249,6 @@ export default function (qaProjectDir) {
       await manageFlowPage.createFlow(flowData.flow5);
       await editFlowPage.addStep(flowData.flow5, binary);
       await browser.sleep(5000);
-      await expect(ingestStepPage.mlcpCommand.getText()).toContain("json");
-      await expect(ingestStepPage.mlcpCommand.getText()).toContain("-input_file_type \"documents\"");
       await editFlowPage.clickRunFlowButton();
       await browser.wait(EC.visibilityOf(editFlowPage.runFlowHeader));
       await editFlowPage.clickButtonRunCancel("flow");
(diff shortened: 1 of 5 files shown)


web/e2e/specs/flows/steps/ingestion.ts
```diff
@@ -107,6 +113,8 @@ export default function (qaProjectDir) {
       await manageFlowPage.createFlow(flowData.flow1);
       await editFlowPage.addStep(flowData.flow1, json);
       await browser.sleep(5000);
+      await expect(ingestStepPage.mlcpCommand.getText()).toContain("json");
+      await expect(ingestStepPage.mlcpCommand.getText()).toContain("-input_file_type \"documents\"");
       await editFlowPage.clickRunFlowButton();
```
bsrikan: Previous test in line 36 already verifies this. Does it add value to have these expects here.  Other duplicates are line 138,139 and 256,257.
Line 160 is fine since the file type is different.

web/e2e/specs/flows/steps/ingestion.ts
```diff
@@ -247,6 +269,42 @@ export default function (qaProjectDir) {
...
+    it('Should verify target permissions', async function () {
+      await expect(ingestStepPage.mlcpCommand.getText()).toContain("-output_permissions \"rest-reader,read,rest-writer,update\"");
+      await ingestStepPage.targetPermissions.clear();
+      await ingestStepPage.targetPermissions.sendKeys("rest-reader,read,rest-writer,update,user1,execute");
+      await ingestStepPage.targetPermissions.sendKeys(protractor.Key.ENTER);
+      await expect(ingestStepPage.mlcpCommand.getText()).toContain("-output_permissions \"rest-reader,read,rest-writer,update,user1,execute\"");
+    });
+
+    it('Should target uri replacement', async function () {
+      await ingestStepPage.targetUriReplace.clear();
```
bsrikan: verify missing in the test description and in line 300.
Can you also please add tests that validate the error message for the URI preview when the target uri is wrong. There are 3 error messages

bsrikan:
@timur-isangulov the commit messages are same. I would rebase and convert them to a single commit

web/e2e/specs/flows/steps/ingestion.ts
```diff
@@ -247,6 +269,42 @@ export default function (qaProjectDir) {
...
+      await ingestStepPage.targetUriReplace.sendKeys("/web/e2e/qa-project/");
+      await expect(ingestStepPage.mlcpCommand.getText()).toContain("-output_uri_replace \"/web/e2e/qa-project/\"");
+      await ingestStepPage.targetUriReplace.clear();
+    });
+
+    it('Should uri preview', async function () {
+      await ingestStepPage.targetUriReplace.clear();
+      await ingestStepPage.targetUriReplace.sendKeys("/C/dev/fork/marklogic-data-hub/web/e2e/qa-project/input/flow-test/json/, '/web/e2e/qa-directory/'");
```
ayuwono: not sure why this is not failing on jenkins? Seems like a local directory path. It's better to use the variable qaProjectDir

examples: 
await ingestStepPage.setInputFilePath(qaProjectDir + "/input/advantage");
await ingestStepPage.setTargetUriReplace(qaProjectDir + "/input/advantage" + ", " + "\'\/advantage\'");

timur-isangulov: It won't fail on Jenkins because it's not testing the correctness of the path which will be different on Mac/Unix and Windows but verifies that the path parameter displays on MLCP.

web/e2e/specs/flows/steps/ingestion.ts
timur-isangulov: Fixed.

web/e2e/specs/flows/steps/ingestion.ts
ayuwono: I see, it's for the target uri preview

web/e2e/specs/flows/steps/ingestion.ts
timur-isangulov: Thanks, added missing verify. Protractor can't extract the text value of these fields. I tried to get it using JavaScript but it didn't work also.  I will discuss it with the developer and add tests in the future.

timur-isangulov:
> @timur-isangulov the commit messages are same. I would rebase and convert them to a single commit

Done. Thanks.

PR #2795, opened 2019-07-18
DHFPROD-2801: Upgrading to ml-gradle 3.15.2 and ml-app-deployer 3.15.1

rjrudin:
Now that the Gradle configurations are up-to-date for these libraries, we no longer need to explicitly depend on jdom2, as it is now properly depended on via the use of the "api" configuration in ml-app-deployer.

Releases notes for ml-gradle 3.15.2 are at https://github.com/marklogic-community/ml-gradle/releases/tag/3.15.2

### DIFF
--- a/marklogic-data-hub/build.gradle
+++ b/marklogic-data-hub/build.gradle
@@ -3,7 +3,7 @@ plugins {
     id 'java'
     id 'maven-publish'
     id 'com.jfrog.bintray' version '1.7.2'
-    id 'com.marklogic.ml-gradle' version '3.15.1'
+    id 'com.marklogic.ml-gradle' version '3.15.2'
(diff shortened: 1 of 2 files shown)


marklogic-data-hub/build.gradle
```diff
@@ -31,15 +31,15 @@ ext.junitJupiterVersion  = '5.3.1'
 
 dependencies {
```
srinathgit: In Line 6, in this file, I still see ml-gradle as 3.15.1. Should that be updated as well ?

rjrudin: Yes, it might as well - that ml-gradle is just used for deploying the DHF test application, but no reason not to update it. Thanks.

PR #2797, opened 2019-07-19
DHFPROD-2842 jobs endpoint no longer throws an error when no jobs are…

rjrudin:
… found

Also consolidated all Java code that talks to the jobs endpoint in JobDocManager to cut down on duplication. I didn't change any functionality in the clients to JobDocManager though - i.e. if they were throwing exceptions before, they're still throwing them now.


web/src/main/java/com/marklogic/hub/web/service/NewJobService.java
```diff
@@ -18,124 +15,26 @@
...
     private FlowManager flowManager;
 
     @Autowired
     private HubConfigImpl hubConfig;
 
-    private static final String ML_JOBS_NAME = "ml:jobs";
-
     private DatabaseClient client;
-
-    public static final String JOB_RESPONSE = "[{\n" +
```
rjrudin: Removed these two constants as they aren't used anywhere

PR #2798, opened 2019-07-19
DHFPROD-2845: QS UI jobs test suite update.

timur-isangulov:


aebadirad:
Why so many failures on the qs run?

timur-isangulov:
> Why so many failures on the qs run?
The installation failed.

PR #2799, opened 2019-07-19
DHFPROD-2798: Generate function metadata and OOTB date, dateTime functions

srinathgit:
1. This PR implements  DHFPROD-2798, DHFPROD-2576, DHFPROD-2157 as they are all related.
2. The function metadata generated in the today's 9.0-nightly and 10.0-nightly server would be incorrect. The issue will be fixed in tomorrow's nightly server (or the latest jenkins server)

### DIFF
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/GenerateFunctionMetadataCommand.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/GenerateFunctionMetadataCommand.java
@@ -1,0 +1,75 @@
+package com.marklogic.hub.deploy.commands;
+
+import com.marklogic.appdeployer.command.AbstractCommand;
+import com.marklogic.appdeployer.command.CommandContext;
+import com.marklogic.appdeployer.command.SortOrderConstants;
+import com.marklogic.client.DatabaseClient;
+import com.marklogic.client.datamovement.ApplyTransformListener;
+import com.marklogic.client.datamovement.DataMovementManager;
+import com.marklogic.client.datamovement.QueryBatcher;
+import com.marklogic.client.document.ServerTransform;
+import com.marklogic.client.query.StructuredQueryBuilder;
+import com.marklogic.hub.DatabaseKind;
+import com.marklogic.hub.HubConfig;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+import java.util.concurrent.TimeUnit;
+
+@Component
+public class GenerateFunctionMetadataCommand extends AbstractCommand {
+    @Autowired
+    private HubConfig hubConfig;
+    private Throwable caughtException;
+
+    public GenerateFunctionMetadataCommand() {
+        super();
+        // This command has to be executed after modules are loaded
+        setExecuteSortOrder(SortOrderConstants.LOAD_MODULES + 1);
+    }
+
+    @Override
+    public void execute(CommandContext context) {
+        DatabaseClient databaseClient = hubConfig.newStagingClient(hubConfig.getDbName(DatabaseKind.MODULES));
+        DataMovementManager dataMovementManager = databaseClient.newDataMovementManager();
+
+        StructuredQueryBuilder sb = databaseClient.newQueryManager().newStructuredQueryBuilder();
+
+
+        ServerTransform serverTransform = new ServerTransform("ml:generateFunctionMetadata");
+
+        ApplyTransformListener transformListener = new ApplyTransformListener()
+            .withTransform(serverTransform)
+            .withApplyResult(ApplyTransformListener.ApplyResult.IGNORE)
+            .onBatchFailure((batch, throwable) -> {
+                logger.error(throwable.getMessage());
+                // throw the first exception
+                if (caughtException == null) {
+                    caughtException = throwable;
+                }
+            });
+
+        /*
+        Query for uris "/data-hub/5/mapping-functions/" and "/custom-modules/mapping-functions/" which are reserved for mapping functions
+         */
+        QueryBatcher queryBatcher = dataMovementManager.newQueryBatcher(
+            new StructuredQueryBuilder().or(sb.directory(true, "/data-hub/5/mapping-functions/"),sb.directory(true, "/custom-modules/mapping-functions/")))
+            .withBatchSize(1)
+            .withThreadCount(4)
+            .onUrisReady(transformListener);
+
+        dataMovementManager.startJob(queryBatcher);
+        //Stop batcher if transform takes more than 2 minutes.
+        try {
+            queryBatcher.awaitCompletion(2L, TimeUnit.MINUTES);
+        } catch (InterruptedException e) {
+            logger.error("Loading function metadata timed out, took longer than 2 minutes");
+        }
+        dataMovementManager.stopJob(queryBatcher);
+
+        if (caughtException != null) {
+            throw new RuntimeException(caughtException);
+        }
+
+    }
+}
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/mapping-functions/parseDate.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/mapping-functions/parseDate.sjs
@@ -1,0 +1,70 @@
+/*
+  Copyright 2012-2019 MarkLogic Corporation
+
+  Licensed under the Apache License, Version 2.0 (the "License");
+  you may not use this file except in compliance with the License.
+  You may obtain a copy of the License at
+
+     http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+*/
+
+function parseDate(picture, value) {
+    let standardFormats = ["MM/DD/YYYY", "DD/MM/YYYY", "MM-DD-YYYY", "MM.DD.YYYY", "DD.MM.YYYY", "YYYYMMDD", "YYYY/MM/DD"];
+    let nonStandardFormats = ["Mon DD, YYYY", "DD Mon YYYY", "DD-Mon-YYYY" ];
+    let response;
+    if(standardFormats.includes(picture.trim())){
+        try{
+            let standardizedDate = xdmp.parseYymmdd(picture.replace("YYYY", "yyyy").replace("DD","dd"), value).toString().split("T")[0];
+            response = xs.date(standardizedDate);
+        }
+        catch (ex) {
+            fn.error(null, "Given value doesn't match with the specified pattern (" + picture + "," + value + ") for parsing date string.");
+        }
+    }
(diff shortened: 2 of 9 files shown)


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/mapping-functions/parseDate.sjs
```diff
@@ -0,0 +1,70 @@
...
+    let standardFormats = ["MM/DD/YYYY", "DD/MM/YYYY", "MM-DD-YYYY", "MM.DD.YYYY", "DD.MM.YYYY", "YYYYMMDD", "YYYY/MM/DD"];
+    let nonStandardFormats = ["Mon DD, YYYY", "DD Mon YYYY", "DD-Mon-YYYY" ];
+    let response;
+    if(standardFormats.includes(picture.trim())){
+        try{
+            let standardizedDate = xdmp.parseYymmdd(picture.replace("YYYY", "yyyy").replace("DD","dd"), value).toString().split("T")[0];
+            response = xs.date(standardizedDate);
+        }
+        catch (ex) {
+            response = "Given value doesn't match with the specified pattern for parsing date string.";
```
rjrudin: Is the intent that the mapping framework will know how to distinguish between an xs.date - which means a successful parse occurred - and an error message as a string? That seems to put a burden on the mapping framework to know how to interpret a string as an error. My guess is throwing an exception is the way to go here, and the mapping framework can then figure out the best way to handle exceptions across all mapping functions.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/mapping-functions/parseDate.sjs
```diff
@@ -0,0 +1,70 @@
...
+        }
+        else if(picture === nonStandardFormats[1]) {
+            pattern = "^([0-9]|[0-2][0-9]|[3][0-1]) (Jan|Feb|Mar|Apr|May|Jun|Jul|Aug|Sep|Oct|Nov|Dec) ([0-9]{4})$";
+        }
+        else {
+            pattern = "^([0-9]|[0-2][0-9]|[3][0-1])\-(Jan|Feb|Mar|Apr|May|Jun|Jul|Aug|Sep|Oct|Nov|Dec)\-([0-9]{4})$";
+        }
+        let match = new RegExp(pattern, 'i').exec(value);
+        if (match === null) {
+            return "Given value doesn't match with the specified pattern for parsing date string.";
```
rjrudin: For consistency, this should assign the string to the "response" variable.

marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/LoadFunctionMetaDataCommand.java
```diff
@@ -0,0 +1,76 @@
...
+            .onBatchFailure((batch, throwable) -> {
+                for(String item: batch.getItems()) {
+                    logger.error("Unable to load function metadata for: " + item);
+                }
+                logger.error(throwable.getMessage());
+                throwable.printStackTrace();
+            });
+
+        /*
+        Query for uris "/data-hub/5/mapping-functions/" and "/datahub/ext/mapping/" which are reserved for mapping functions
```
rjrudin: Would a user place their custom mapping functions under "/datahub/ext/mapping"? DHF is generating custom modules under "/custom-modules" - should mappings go under "/custom-modules" as well? I think these locations won't be perceived as consistent by developers. "/datahub/" can also be easily confused with "/data-hub/", which is where product modules go.

marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/LoadFunctionMetaDataCommand.java
```diff
@@ -0,0 +1,76 @@
...
+import com.marklogic.client.query.StructuredQueryBuilder;
+import com.marklogic.hub.DatabaseKind;
+import com.marklogic.hub.HubConfig;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+import java.util.concurrent.TimeUnit;
+
+@Component
+public class LoadFunctionMetaDataCommand extends AbstractCommand {
```
rjrudin: I recommend a naming that doesn't use "Load" because nothing is being loaded from a user's perspective. Instead, function metadata XML documents are being generated based on libraries of module functions that have already been loaded. 

How about "GenerateFunctionMetadataCommand"? (I would also use "Metadata" instead of "MetaData" as the former is consistent with how the word is spelled in other parts of MarkLogic).

marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/LoadFunctionMetaDataCommand.java
```diff
@@ -0,0 +1,76 @@
...
+        serverTransform.addParameter("forceLoad", String.valueOf(forceLoad));
+
+        ApplyTransformListener transformListener = new ApplyTransformListener()
+            .withTransform(serverTransform)
+            .withApplyResult(ApplyTransformListener.ApplyResult.IGNORE)
+            .onBatchFailure((batch, throwable) -> {
+                for(String item: batch.getItems()) {
+                    logger.error("Unable to load function metadata for: " + item);
+                }
+                logger.error(throwable.getMessage());
```
rjrudin: @ryanjdew @aebadirad  Thoughts on whether a failure to generate function metadata should result in the deployment failing? I lean towards making it fail. ml-gradle has a property to tell it to catch exceptions and keep going in case that's a problem. 

@srinathgit If we do agree on this throwing an error, then I think you'll need to declare a Throwable variable before you construct this listener. onBatchFailure can then assign the throwable that it catches to this declared variable. After DMSDK finishes, if the throwable is null, then throw it to cause a hard failure.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/mapping-functions/parseDate.sjs
ryanjdew: I agree that we probably want to throw an explicit error in the case of an invalid response.

marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/LoadFunctionMetaDataCommand.java
ryanjdew: I agree with making it fail. My experience is people tend to ignore logs and march on, only to have more issues down the road.

marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/LoadFunctionMetaDataCommand.java
srinathgit: @ryanjdew ,
Based on the https://wiki.marklogic.com/pages/viewpage.action?spaceKey=ENGINEERING&title=Declarative+Mapping+Plan, currently
1. OOTB functions go to "/data-hub/5/mapping-functions/*.sjs|mjs|xqy"
2. Custom functions "/datahub/ext/mapping/*.sjs|mjs|xqy"

Are you ok with changing the custom functions to "/custom-modules/mapping-functions/*.sjs|mjs|xqy ?

ryanjdew: @srinathgit I'm fine with that and think it is worth changing for consistency.

marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/GenerateFunctionMetadataCommand.java
```diff
@@ -0,0 +1,81 @@
...
+    private HubConfig hubConfig;
+    private boolean forceLoad = true;
+    private Throwable caughtException;
+
+    public GenerateFunctionMetadataCommand() {
+        super();
+        // This command has to be executed after modules are loaded
+        setExecuteSortOrder(SortOrderConstants.LOAD_MODULES + 1);
+    }
+    public void setForceLoad(boolean forceLoad) {
```
rjrudin: I recommend dropping this feature. I know we have it on commands for loading modules, but I don't see a use case for this for function metadata - for now, I think we always want to generate function metadata to ensure it's up-to-date.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/mapping-functions/parseDate.sjs
```diff
@@ -0,0 +1,70 @@
...
+    let standardFormats = ["MM/DD/YYYY", "DD/MM/YYYY", "MM-DD-YYYY", "MM.DD.YYYY", "DD.MM.YYYY", "YYYYMMDD", "YYYY/MM/DD"];
+    let nonStandardFormats = ["Mon DD, YYYY", "DD Mon YYYY", "DD-Mon-YYYY" ];
+    let response;
+    if(standardFormats.includes(picture.trim())){
+        try{
+            let standardizedDate = xdmp.parseYymmdd(picture.replace("YYYY", "yyyy").replace("DD","dd"), value).toString().split("T")[0];
+            response = xs.date(standardizedDate);
+        }
+        catch (ex) {
+            fn.error(null, "Given value doesn't match with the specified pattern for parsing date string.");
```
rjrudin: For each error message, I recommend adding the value that was inputted so the error message is more helpful.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/transforms/generate-function-metadata.sjs
```diff
@@ -0,0 +1,28 @@
...
+function transform(context, params, content) {
+
+  let forceLoad = params['forceLoad'] ? (xdmp.urlDecode(params['forceLoad']) === 'true'): true;
+  let uri = context.uri;
+
+  let pattern = '^(.*)\.(sjs|mjs|xqy)$';
+  let match = new RegExp(pattern).exec(uri);
+  if (match !== null) {
+    let uriVal = match[1];
+    /* TODO : Investigate hang when if 'forceLoad' is false and xslt doc is not present, the transform hangs at
```
rjrudin: Can remove this comment once forceLoad is removed as an option

marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/GenerateFunctionMetadataCommand.java
```diff
@@ -0,0 +1,75 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+import java.util.concurrent.TimeUnit;
+
+@Component
+public class GenerateFunctionMetadataCommand extends AbstractCommand {
+    @Autowired
+    private HubConfig hubConfig;
+    private Throwable caughtException;
```
rjrudin: Should just make this a local variable, but can address in a future PR. The downside to a class variable is if this command were invoked twice, an exception from the first run would carry over to the second run.

srinathgit: I agree.. I will make that change in a future PR.

PR #2805, opened 2019-07-22
DHFPROD-2844: Property name is not populated when editing a match option in mastering step

xnikhil08:
Fixed the issue. The property to match should now pre-populate in the edit-setting dialogue box for any match-option.

### DIFF
--- a/web/src/main/ui/app/components/flows-new/edit-flow/mastering/matching/ui/add-match-option-dialog.component.ts
+++ b/web/src/main/ui/app/components/flows-new/edit-flow/mastering/matching/ui/add-match-option-dialog.component.ts
@@ -36,7 +36,7 @@ export class AddMatchOptionDialogComponent {
   ngOnInit() {
     this.tooltips = FlowsTooltips.mastering.matching;
     this.form = this.fb.group({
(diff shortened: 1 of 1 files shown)


xnikhil08:
@wooldridge , @rjrudin - Please review.

web/src/main/ui/app/components/flows-new/edit-flow/mastering/matching/ui/add-match-option-dialog.component.ts
```diff
@@ -83,9 +83,7 @@ export class AddMatchOptionDialogComponent {
       return this.fb.array([this.createProp('')]);
     }
     const result = [];
-    this.data.option.propertyName.forEach(name => {
-      result.push(this.createProp(name))
-    })
+    result.push(this.createProp(this.data.option.propertyName));
```
rjrudin: Reduce is working when only one property name is specified, but not when two or more property names are specified.

xnikhil08: @rjrudin - Please review now.

PR #2812, opened 2019-07-23
DHFPROD-2799: Create mapping XML from JSON mapping

ryanjdew:


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mapping/entity-services/lib.sjs
```diff
@@ -0,0 +1,155 @@
...
+  return entitiesByTargetType[targetEntityType];
+}
+
+function retrieveFunctionImports() {
+  let customImports = [];
+  let shimURIs = datahub.hubUtils.queryLatest(function() {
+    return Sequence.from([
+      cts.uriMatch('/custom-modules/mapping-functions/*.xsl'),
+      cts.uriMatch('/data-hub/5/mapping-functions/*.xsl')
+    ]);
```
srinathgit: The function metadata generated have uri ending in *.xslt and not *.xsl

aebadirad: Can we tag these with a collection instead of relying on uri?

It'll make the search much faster, especially if have to do this a lot.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mapping/entity-services/lib.sjs
```diff
@@ -0,0 +1,155 @@
...
+      related.push(externalMapping);
+      related = getRelatedMappings(externalMapping, related);
+    }
+  }
+  return related;
+}
+
+function getTargetEntity(targetEntityType) {
+  if (!entitiesByTargetType[targetEntityType]) {
+    let targetArr = targetEntityType.split('/');
```
aebadirad: Can you check for $id in additional to this?

Just to future proof this.

srinathgit:
@rjrudin ,
This has to be run on 9.0-nightly or 10.0-nightly, it wouldn't work on ML 10.0-1 or 9.0-9

ryanjdew:
@rjrudin There is an ml-unit test added for testing the XML generation. Is there some specific test cases that you think should be expanded on?

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mapping/entity-services/lib.sjs
srinathgit: They are tagged with collection "http://marklogic.com/entity-services/function-metadata/compiled".

ryanjdew: Updated to use the collection. Submitted a bug that will need to be fixed in the nightly before custom functions will work, though. https://bugtrack.marklogic.com/52807

PR #2816, opened 2019-07-24
DHFPROD-2867: Defaulting more DHS properties per the 2.6.0 release

rjrudin:


marklogic-data-hub/src/main/java/com/marklogic/hub/cli/command/InstallIntoDhsCommand.java
```diff
@@ -99,12 +105,42 @@ protected Properties buildProjectProperties() {
     public void applyDhsSpecificProperties(Properties props) {
         props.setProperty("mlIsHostLoadBalancer", "true");
         props.setProperty("mlIsProvisionedEnvironment", "true");
+        props.setProperty("mlAppServicesPort", "8010");
 
         props.setProperty("mlFlowDeveloperRole", "flowDeveloper");
         props.setProperty("mlFlowOperatorRole", "flowOperator");
         // Mapping this to flowDeveloper for now,
         props.setProperty("mlDataHubAdminRole", "flowDeveloper");
```
aebadirad: Yeah, we'll need to address this differently.

FlowDevelopers shouldn't be doing higher level stuff beyond loading artifacts and custom modules to the database.

rjrudin: Note that mlDataHubAdminRole today is only used for setting permissions on trigger documents, and those trigger documents also have the same permissions with the flow developer role. So in effect, the role is not used today for anything.

bsrikan:
Changes required to the Verifier as well. Currently dhsVerify doesnt run.

bsrikan:
Still running into issue with the Verifier.

PR #2825, opened 2019-07-26
DHFPROD-2877: Fixing DHS-e2e example project to work with ssl

bsrikan:
The update also includes the addition of REST service extension to the example project.


examples/dhf4/DHS-e2e/DHS/gradle-dhs.properties
```diff
@@ -0,0 +1,54 @@
+mlDHFVersion=4.3.1
+mlHost=k1kwjmp86.zu4z1nsamcx.a.marklogicsvc.com
+mlIsHostLoadBalancer=true
+
+mlUsername=flowdev
+mlPassword=SomePass#123
```
rjrudin: Is it okay to have these passwords in here??

bsrikan: I think so. README instructions ask users to update gradle properties with the users they create. Besides, planning to use this project along with the core tests to verify DHS in our Jenkins jobs. DHS team create users for us with these credentials when they spin up a stack for testing.

PR #2829, opened 2019-07-26
DHFPROD-2905: Support 9.0-10 and above, 10.0 nightly > 06/01

srinathgit:
Support for 4.x-develop on 9.0-10 and above, 10.0 nightly > 06/01

### DIFF
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/impl/DataHubImpl.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/impl/DataHubImpl.java
@@ -250,8 +250,8 @@ public class DataHubImpl implements DataHub {
                 return false;
             }
             boolean isNightly = versionString.matches("[^-]+-(\\d{4})(\\d{2})(\\d{2})");
-            //Support any 9.0 version > 9.0-5 and all 10.0 versions.
 
+            //Support any 9.0 version >= 9.0-5.
             int minor = 0;
             int hotFixNum = 0;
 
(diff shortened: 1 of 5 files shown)


josvanroosmalen:
Can we get this also fixed in 4.2.2 (e.g. 4.2.3 or 4.2.2-1 or whatever)

josvanroosmalen:
I am very critical on this change. Why do you support 10.0 Nightlies, but not 10.0-2 once that hits the ground. This does not make any sense!?

sbayatpur:
4.x should work on 9.0-10 and future versions of 9, but users should upgrade to 5 if they want to use MarkLogic 10

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/DataHubImpl.java
```diff
@@ -246,53 +246,42 @@ public boolean isServerVersionValid(String versionString) {
                 versionString = versions.getMarkLogicVersion();
             }
             int major = Integer.parseInt(versionString.replaceAll("([^.]+)\\..*", "$1"));
-            if (major < 9) {
+            if (major != 9) {
                 return false;
             }
             boolean isNightly = versionString.matches("[^-]+-(\\d{4})(\\d{2})(\\d{2})");
             //Support any 9.0 version > 9.0-5 and all 10.0 versions.
-            if (major == 9) {
```
bsrikan: We probably have to update the comment as well?

srinathgit: Updated the comment

srinathgit:
Sorry for multiple review requests, all the tests seems to have passed finally

PR #2839, opened 2019-07-29
DHFPROD-2800: Step definition for declarative mapper

ryanjdew:


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mapping/entity-services/main.sjs
```diff
@@ -0,0 +1,187 @@
...
+    throw Error(e);
+  }
+
+  content.value = buildEnvelope(entity.info, doc, instance, outputFormat, options);
+  content.provenance = { [content.uri]: provenance };
+  return content;
+}
+
+// Extracted for unit testing purposes
+function buildEnvelope(entityInfo, doc, instance, outputFormat, options) {
```
rjrudin: This all looks good, but is it not yet possible to write a unit test for it? Or to setup an integration-level test that sets up an entity and a flow with an ES mapping step and runs it?

PR #2843, opened 2019-07-30
DHFPROD-2928: No longer throwing an error when flows directory is mis…

rjrudin:
…sing

This would occur in QS in a new project. If the flows directory hadn't been created yet, the "Flows" link would throw an error and the spinner would just keep spinning.


bsrikan:
@rjrudin hubInit from gradle or QS, always creates an empty flows directory. Wondering when or how, a "flows" directory does not get created

Issue #2811, opened 2019-07-23
DHF 5.0.1 Smart Mastering RFE: prohibited match option


crism:
#### The issue

Short description of the problem: It would be nice to note unique identifiers which should never be merged, regardless of other matches, or which count for negative scores.

What behavior are you expecting? The `reduce` option in merge options allows a penalty for things that match. It should also be possible to specify a penalty for things that do not match. For example, American people with Social Security Numbers; it may be permissible to merge people with incomplete data, but two individuals with present and different SSNs should not be merged (fraud investigations aside). More generally, if a dataset is partially curated, it should be possible to specify that items with distinct identifiers are not merge candidates.

Possible solution 1:
```json
"property": [
  {
    "localname": "givenName",
    "name": "givenName"
  }, {
    "localname": "id",
    "name": "id",
    "unique": true
  }
]
```
The `unique` identifier would forbid merging two items with non-null values for that property.

Possible solution 2:
```json
"scoring": {
  "add": [
    {
      "propertyName": "givenName",
      "weight": "15"
    }, {
      "propertyName": "id",
      "penalty": true,
      "weight": "1000"
    }
  ]
}
```
This would give 1,000 points if the `id` matches, but subtract 1,000 if present, non-null values do not match. Solution 2(a): the `penalty` property could take a number instead. Using an arbitrarily large number, relative to threshold values, would guarantee non-matching values never merged. I’m not thrilled with the property name here; it would be clearer as `penaltyOnDiscrepancy` or something like that.

Possible solution 3:
```json
"scoring": {
  "add": [
    {
      "propertyName": "givenName",
      "weight": "15"
    }
  ],
  "subtract": [
    {
      "propertyName": "id",
      "weight": "1000"
    }
  ]
}
```
This would add 15 points when the `givenName` matches, but subtract 1,000 if the `id` does not match.

I think possible solution 2 or 2(a) are my favorites.

#### Tech details

Which Operating System are you using? CentOS 7

Which version of MarkLogic are you using? 9.0-9

Which version of the Data Hub Framework are you using? 5.0.1

Is this a QuickStart UI bug? If so, which browser are you using? N/A

crism:
The `$filter-query` parameter in `process.xqy` does seem to point a way to this.

First, we can accept a `filterQuery` option in the match step, something like:

```javascript
  let filterQueryString = options.filterQuery || "cts.trueQuery()";
  let filterQuery = xdmp.eval(filterQueryString);

[...]

PR #2844, opened 2019-07-30
Adds unique-property checks and forbids merging.

crism:
Implements solution 1 from #2811. Forbids merging of items with declared-unique properites.

Probably could be optimized, but it works.


crism:
Gah, wrong branch. Sorry.

PR #2845, opened 2019-07-30
DHFPROD-1190: Added tests to verify that comparison operations work

rjrudin:
The story had called for new expressions, but I was informed that the regular comparison operators should work fine, and indeed they do. So I just added assertions to verify this.

### DIFF
--- a/marklogic-data-hub/src/test/ml-modules/root/test/suites/com/marklogic/hub/mapping/mapping-test.sjs
+++ b/marklogic-data-hub/src/test/ml-modules/root/test/suites/com/marklogic/hub/mapping/mapping-test.sjs
@@ -14,9 +14,16 @@ function mapsJSONasExpected() {
     },
     "targetEntity" : "Customer"
   }).value.envelope.instance.Customer;
+  const message = `Unexpected output: ${describe(mappedInstance)}`;
   return [
-    test.assertEqual("Bob", mappedInstance.firstname, `Unexpected output: ${describe(mappedInstance)}`),
-    test.assertEqual("Customer-1234", mappedInstance.id, `Unexpected output: ${describe(mappedInstance)}`)
+    test.assertEqual("Bob", mappedInstance.firstname, message),
+    test.assertEqual("Customer-1234", mappedInstance.id, message),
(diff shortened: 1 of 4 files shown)


marklogic-data-hub/src/test/ml-modules/root/test/suites/com/marklogic/hub/mapping/mapping-test.sjs
```diff
@@ -14,9 +14,16 @@ function mapsJSONasExpected() {
...
-    test.assertEqual("Bob", mappedInstance.firstname, `Unexpected output: ${describe(mappedInstance)}`),
-    test.assertEqual("Customer-1234", mappedInstance.id, `Unexpected output: ${describe(mappedInstance)}`)
+    test.assertEqual("Bob", mappedInstance.firstname, message),
+    test.assertEqual("Customer-1234", mappedInstance.id, message),
+    test.assertEqual(10, mappedInstance.firstNumber, message),
+    test.assertEqual(null, mappedInstance.secondNumber, message),
+    test.assertEqual(10, mappedInstance.thirdNumber, message),
+    test.assertEqual(10, mappedInstance.fourthNumber, message),
+    test.assertEqual(null, mappedInstance.fifthNumber, message),
+    test.assertEqual(null, mappedInstance.sixthNumber, message)
```
srinathgit: This test makes use of "/data-hub/5/builtins/steps/mapping/default/main.sjs" which is the main.sjs of the old mapping implementation (the one in 5.0). Shouldn't the tests exercise "/data-hub/5/builtins/steps/mapping/entity-services/main.sjs" which is in PR  https://github.com/marklogic/marklogic-data-hub/pull/2839 ?

rjrudin: Thanks, I hadn't realized there was a new module. I'll change this test over to use it and see what happens. 

@srinathgit  @ryanjdew  Should "/data-hub/5/builtins/steps/mapping/default/main.sjs" have some kind of deprecation warning in it then? If so, I can add that.

PR #2846, opened 2019-07-30
Forbid matches on declared-unique properties

crism:
Implements solution 1 from #2811. Forbids merging of items with declared-unique properites.

Probably could be optimized, but it works.

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/com.marklogic.smart-mastering/matcher-impl/matcher-impl.xqy
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/com.marklogic.smart-mastering/matcher-impl/matcher-impl.xqy
@@ -258,15 +258,29 @@ declare function match-impl:find-document-matches-by-options(
       xdmp:node-uri($document),
       blocks-impl:get-blocks(fn:base-uri($document))/node()
     )
+    let $excluded-docs-query :=
+      if (fn:exists($excluded-uris)) then
+        cts:document-query($excluded-uris)
+      else ()
+    let $excluded-matches-query := match-impl:build-uniqueness-queries(
+      $document,
+      $is-json,
+      $values-by-qname,
+      $options
+    )
+    let $exclusion-query :=
+      if ($excluded-docs-query or $excluded-matches-query) then
+        cts:or-query($excluded-docs-query, $excluded-matches-query)
+      else ()
     let $match-base-query := cts:and-query((
           $compiled-options => map:get("collectionQuery"),
           $minimum-threshold-combinations-query
         ))
     let $match-query :=
-      if (fn:exists($excluded-uris)) then
+      if (fn:exists($exclusion-query)) then
         cts:and-not-query(
           $match-base-query,
-          cts:document-query($excluded-uris)
+          $exclusion-query,
         )
       else
         $match-base-query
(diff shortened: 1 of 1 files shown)


marklogic-data-hub/src/main/resources/ml-modules/root/com.marklogic.smart-mastering/matcher-impl/matcher-impl.xqy
```diff
@@ -361,6 +375,55 @@ declare function match-impl:build-boost-query(
...
+  $match-options as item()
+)
+{
+  let $options :=
+    if ($match-options instance of object-node()) then
+      opt-impl:options-from-json($match-options)
+    else
+      $match-options
+  for $property-def in $options/matcher:property-defs/matcher:property
+    return if (xs:boolean($property-def/matcher:unique)) then
```
rjrudin: I like the idea of configuring this via a property, but it hurts my head to think through all the negatives here - i.e. this returns a query that is included in an or-query that is part of an and-not-query, and the returned query is an and-query that includes a not-query around an or-query. 

Abstracting up, it seems simpler and sufficient to construct the overall match query to be an AND query like this:

- An OR query on all the match actions
- (Optional) an AND query on what could be called "filter" or "inclusion" queries

That can then be tossed into a cts:and-not-query, with the not-query being as it is today. 

The "inclusion" query can then be expressed as an OR query (using a concrete example of SSN):

- Either the matching document has the same SSN
- Or it has an SSN element/property with no value
- Or it does not have an SSN element/property

This query wouldn't have a weight on it because weighting doesn't apply here - this is a hard and fast rule that all matches must meet. 

So the way I'd organize this is by creating a function to build up zero or more "inclusion" queries. One source of those could be the presence of the matcher:unique element.

marklogic-data-hub/src/main/resources/ml-modules/root/com.marklogic.smart-mastering/matcher-impl/matcher-impl.xqy
```diff
@@ -258,15 +258,29 @@ declare function match-impl:find-document-matches-by-options(
...
+      else ()
+    let $excluded-matches-query := match-impl:build-uniqueness-queries(
+      $document,
+      $is-json,
+      $values-by-qname,
+      $options
+    )
+    let $exclusion-query :=
+      if ($excluded-docs-query or $excluded-matches-query) then
+        cts:or-query($excluded-docs-query, $excluded-matches-query)
```
ryanjdew:
```suggestion
        cts:or-query(($excluded-docs-query, $excluded-matches-query))
```

marklogic-data-hub/src/main/resources/ml-modules/root/com.marklogic.smart-mastering/matcher-impl/matcher-impl.xqy
```diff
@@ -361,6 +375,55 @@ declare function match-impl:build-boost-query(
...
+    if ($match-options instance of object-node()) then
+      opt-impl:options-from-json($match-options)
+    else
+      $match-options
+  for $property-def in $options/matcher:property-defs/matcher:property
+    return if (xs:boolean($property-def/matcher:unique)) then
+      let $qname :=
+        fn:QName($property-def/@namespace, $property-def/@localname)
+      let $values := $values-by-qname => map:get(xdmp:key-from-QName($qname))
+      return if ($is-json) then
```
ryanjdew: In this section, we should adjust to make use of `helper:property-name-to-query` to make sure we pick up enhancements, like making use of range indexes, etc. An example of usage can be found here: https://github.com/marklogic/marklogic-data-hub/blob/v5.0.2/marklogic-data-hub/src/main/resources/ml-modules/root/com.marklogic.smart-mastering/algorithms/double-metaphone.xqy#L61 

Basically, passing the match options and the property name, it will return the proper cts:query for the property to values and weight to.

marklogic-data-hub/src/main/resources/ml-modules/root/com.marklogic.smart-mastering/matcher-impl/matcher-impl.xqy
ryanjdew: I agree especially with the comments on the double NOT logic. It should be pretty simple to avoid that by using the existing `$filter-query` parameter and extending with additional queries if the unique option is set on any of the values.

crism:
Moving discussion back to the original issue…

rjrudin:
@crism  mentioned that this can be closed, as his on-site implementation has varied quite a bit. So we won't merge this, but keeping it for reference purposes.

PR #2857, opened 2019-08-01
DHFPROD-2951: Added new triggers on init, back ported from 5.0.2

bsrikan:


bsrikan:
I guess these are similar test failures we saw in the PR for DHFPROD-2534. Working on test fix.

bsrikan:
> @bsrikan ,
> 
>     1. A bunch of tests require changes to the entities so that it has mandatory properties like baseURI etc.
> 
>     2. Also please run SSLTest to make sure if you have to update module count

Sure.

PR #2868, opened 2019-08-06
DHFPROD-2826: Custom steps in QuickStart

xnikhil08:
Updated the required fields and changed the logic of stepDefinitionType in UI for all the steps.


srinathgit:
One minor issue I see is that the "Source Directory Path" directory picker seems to be indented right, whereas the rest of them are on the left side of the window

xnikhil08:
> In the QuickStart UI, a regular Mapping or Mastering step requires: Name, Source, Entity
> 
> This is not being enforced for the new Custom Mapping or Custom Mastering steps, e.g.:
> 
> Screen Shot 2019-08-06 at 4 50 46 PM
> 
> I assume these same requirements should be enforced? @sbayatpur or @bsrikan?

@wooldridge - The custom mapping and mastering steps are in progress. This PR was only about Custom Ingestion.

xnikhil08:
> One minor issue I see is that the "Source Directory Path" directory picker seems to be indented right, whereas the rest of them are on the left side of the window. The rest seems ok as far as ingestion is concerned

@srinathgit - Could you please review again? Fixed that issue and also updated the custom hook related UI changes as discussed today.

bsrikan:
@aebadirad  adding you as a reviewer for backend changes.

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/FlowManagerImpl.java
```diff
@@ -264,11 +265,9 @@ protected void deleteRelatedStepArtifacts(Flow flow, Step removedStep) {
             return;
         }
 
-        Path custModulePath = this.hubConfig.getModulesDir().resolve("root").resolve("custom-modules")
-            .resolve(removedStep.getStepDefinitionType().toString().toLowerCase())
-            .resolve(removedStep.getStepDefinitionName());
+        StepDefinition.StepDefinitionType stepTypeOfRemovedStep = EnumUtils.getEnumIgnoreCase(StepDefinition.StepDefinitionType.class, removedStep.getStepDefinitionType().toString());
```
aebadirad: We can match stepdefintion type directly without invoking a third party library util via the getStepDefinitionType static method on the stepdefinition interface

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/FlowManagerImpl.java
```diff
@@ -261,10 +264,16 @@ protected void deleteRelatedStepArtifacts(Flow flow, Step removedStep) {
...
+                                                    removedStep.getStepDefinitionType().toString()
+                                                    );
+
+        if (removedStep.isMappingStep() && !mappingIsReferencedByAFlow(removedStep) && !isCustomMapping(removedStep)) {
             deleteMappingArtifacts(flow, removedStep);
+        } else if(!removedStep.isCustomStep() && !stepIsReferencedByAFlow(removedStep.getName(), stepTypeOfRemovedStep)) {
+            deleteStepDefinitionArtifacts(removedStep, stepTypeOfRemovedStep);
         } else if (removedStep.isCustomStep() && !stepIsReferencedByAFlow(removedStep.getName(), StepDefinition.StepDefinitionType.CUSTOM)) {
-            deleteStepDefinitionArtifacts(removedStep);
+            deleteStepDefinitionArtifacts(removedStep, stepTypeOfRemovedStep);
```
rjrudin: I recommend adding a test to FlowManagerServiceTest to cover the new scenario captured by the first "else if" block.

bsrikan: Thanks Rob, I have it in my stash. Will pop it in here sometime later today or next week.

bsrikan: @rjrudin I have pushed some tests. Pls review.

PR #2876, opened 2019-08-07
DHFPROD-2406: Declarative Mapper

ryanjdew:


srinathgit:
@ryanjdew ,
The tests will fail unless they are run on (9 and 10) nightly servers. Currently, tests on develop runs from 9.0-7. I think we have to talk to @SameeraPriyathamTadikonda to run develop only the 9.0 and 10.0 nightly servers

srinathgit:
@ryanjdew ,
When this is exposed from UI for 5.1 release, I am guessing, the legacy mapping (5.0 mapping) will be removed and the ES mapping will replace it ?

ryanjdew:
@rjrudin develop is 5.1-SNAPSHOT and 5.x-develop is 5.0-SNAPSHOT.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/transforms/generate-function-metadata.sjs
```diff
@@ -21,9 +21,17 @@ function transform(context, params, content) {
...
       let writeInfo = datahub.hubUtils.writeDocument(uriVal + ".xml", metaDataXml, permissionsExpression, [collection], datahub.config.MODULESDATABASE);
       if (writeInfo && fn.exists(writeInfo.transaction)) {
-        // Workaround to avoid XSLT-UNBPRFX error. See https://bugtrack.marklogic.com/52870
-        xdmp.moduleCacheClear();
-        es.functionMetadataPut(uriVal + ".xml");
+        // try/catch workaround to avoid XSLT-UNBPRFX error. See https://bugtrack.marklogic.com/52870
+        try {
+          es.functionMetadataPut(uriVal + ".xml");
+        } catch (e) {
+          if (/(prefix|XSLT-UNBPRFX)/ig.test(e.message)) {
```
rjrudin: I think it's worth adding a small unit test to verify this regular expression, with the check being pulled into a separate function. That function (and the test) become better places to document what's going on here as well, rather than an inline comment. 

Seems like it's worth veryfing that the result of "functionMetadataPut" still occurs as well.

srinathgit:
Based on the discussion we had in grooming meeting today my last comment can be ignored.

bsrikan:
Tests failing in 10_0 nightly:
https://jenkins.marklogic.com/job/Datahub_CI/job/PR-2876/15/console

rjrudin:
@bsrikan @ryanjdew I'm looking into those test failures.

rjrudin:
I'm putting together a PR with some test fixes

rjrudin:
The 3 failures in https://jenkins.marklogic.com/job/5.x/job/dhf-core-develop-rh7-PR/4759/testReport all seem to be due to ML not being the right version - e.g. "Error: TypeError: es.mapToCanonical is not a function".

PR #2878, opened 2019-08-07
DHFPROD-2912: Test and UI changes for 4.3.2

bsrikan:
This PR has:
Added some test for the back ported ES changes.
PageObject for baseUri field to add tests in the future.
Changes to QuickStart to make baseUri a required property


quick-start/src/main/ui/app/entities/entity.model.ts
```diff
@@ -160,7 +160,7 @@ export class Entity {
         height: 100
       },
       'info': {
-        'baseUri': null,
+        'baseUri': 'http://example.org',
```
aebadirad: http://example.org/, as we're doing this check:

readonly BASE_URI_REGEX = /^(?:http(s)?:\/\/)?[\w.-]+(?:\.[\w\.-]+)+[\w\-\._~:/?#[\]@!\$&'\(\)\*\+,;=.]/;

bsrikan: I changed it a little bit compared to what we have in 5x. I removed the last bit that expects trailing /
`+(\/)$`

aebadirad: oh, for some reason I put that in. Whoops.

aebadirad:
Thanks for fixing that :)

PR #2884, opened 2019-08-09
DHFPROD-2756: Enable unmerge mastering endpoint

ryanjdew:
Reverses the last set of merges made into the given merge URI. retainAuditTrail (default true) determines if provenance for the merge/unmerge is kept. blockFutureMerges (default true) ensures that the documents won't be merged together in the next mastering run.

REST API
DELETE /v1/resources/ml:smMerge

parameters

- rs:mergeURI=<URI> – the URI of the merged document
- rs:retainAuditTrail=<true|false> – optional; if true, the merged document will be moved to an archive collection; if false, the merged document will be deleted. Defaults to true.
- rs:blockFutureMerges=<true|false> – optional; if true, the unmerged documents will be blocked from merging together on a future match and merge run; if false, the unmerged documents will be allowed to match again. Defaults to true.

Gradle task

./gradlew hubUnmergeDocs

parameters

- -PmergeURI=<URI> – the URI of the merged document
- -PretainAuditTrail=<true|false> – optional; if true, the merged document will be moved to an archive collection; if false, the merged document will be deleted. Defaults to true.
- -PblockFutureMerges=<true|false> – optional; if true, the unmerged documents will be blocked from merging together on a future match and merge run; if false, the unmerged documents will be allowed to match again. Defaults to true.


ryanjdew:
@srinathgit The end-user shouldn't be creating an unmerge flow/step. We're only running through a flow/step for the PROV tracing. Implementing a UI to call the REST endpoint is outside the scope of this issue.

PR #2886, opened 2019-08-12
DHFPROD-2976: Add permissions to DHF 4 flow files

ryanjdew:


bsrikan:
This bug is marked for 5.0.3, so we should have a PR to 5.x-develop as well...

PR #2890, opened 2019-08-12
DHFPROD-2896: (5.x) Don't write Job docs from middleware if disableJobOutpu

ryanjdew:
…t = true

### DIFF
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/flow/impl/FlowRunnerImpl.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/flow/impl/FlowRunnerImpl.java
@@ -60,6 +60,7 @@ public class FlowRunnerImpl implements FlowRunner{
 
     private ThreadPoolExecutor threadPool;
     private JobDocManager jobDocManager;
+    private boolean disableJobOutput = false;
 
     @Override
     public FlowRunner onStatusChanged(FlowStatusListener listener) {
@@ -92,6 +93,11 @@ public class FlowRunnerImpl implements FlowRunner{
     }
 
     public RunFlowResponse runFlow(String flowName, List<String> stepNums, String jobId, Map<String, Object> options, Map<String, Object> stepConfig) {
+        if (options != null && options.containsKey("disableJobOutput")) {
+            disableJobOutput = Boolean.parseBoolean(options.get("disableJobOutput").toString());
+        } else {
+            disableJobOutput = false;
+        }
 
         Flow flow = flowManager.getFlow(flowName);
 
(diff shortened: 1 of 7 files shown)


marklogic-data-hub/src/main/java/com/marklogic/hub/step/impl/QueryStepRunner.java
```diff
@@ -129,7 +136,31 @@ public StepRunner withStopOnFailure(boolean stopOnFailure) {
 
     @Override
     public StepRunner withOptions(Map<String, Object> options) {
-        this.options = options;
+        if(flow == null){
```
rjrudin: Unit tests would be easy to write for this, and this seems important to test as well - and I guess the aggregation of options from the incoming map, the flow, and the step is needed for the "disable job output" flag to work? If so, the assertion messages in unit tests are a good place to document the "why" behind this code addition.

marklogic-data-hub/src/main/java/com/marklogic/hub/step/impl/StepRunnerUtil.java
```diff
@@ -56,4 +57,15 @@ protected static void initializeStepRun(JobDocManager jobDocManager, RunStepResp
         }
     }
 
+    protected static String objectToString(Object obj) {
```
rjrudin: Since this is a convenience function, I think it's worth checking for null here and likely returning null by default, as it should be up to the client to first throw an error if the object it's passing in is null.

marklogic-data-hub/src/main/java/com/marklogic/hub/flow/impl/FlowRunnerImpl.java
```diff
@@ -147,7 +153,7 @@ private void initializeFlow(String jobId) {
         jobStoppedOnError.set(false);
         runningJobId = jobId;
         runningFlow = flowMap.get(runningJobId);
-        if(jobDocManager == null) {
+        if(jobDocManager == null && !disableJobOutput) {
```
rjrudin: I think we need tests to verify that a) no job data is created, and b) nothing blows up when job data is disabled.

marklogic-data-hub/src/main/java/com/marklogic/hub/step/impl/StepRunnerUtil.java
ryanjdew: done

marklogic-data-hub/src/main/java/com/marklogic/hub/flow/impl/FlowRunnerImpl.java
ryanjdew: Added a test for this in the flow runner tests.

PR #2902, opened 2019-08-27
DHFPROD-3049: Adding fields and range field indexes to meta-data props of harmonized docs

rahulvudutala:


rahulvudutala:
@rjrudin I am working on the test.

PR #2903, opened 2019-08-29
DHFPROD-2784: Can now generate a NiFi template for a flow

rjrudin:
Have only tested this on the 3-step flow in the dh-5-example project, but it works fine there. Note the instructions printed out by the Gradle task for modifying the template after it's been loaded into NiFi.

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/nifi/database-client-service.xqy
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/nifi/database-client-service.xqy
@@ -1,0 +1,124 @@
+xquery version "1.0-ml";
+
+module namespace nifi = "http://marklogic.com/data-hub/nifi";
+
+declare function build-database-client-service($params as map:map)
+{
+  <controllerServices>
+    <id>{sem:uuid-string()}</id>
+    <parentGroupId>{map:get($params, "parent-group-id")}</parentGroupId>
+    <bundle>
+      <artifact>nifi-marklogic-nar</artifact>
+      <group>org.apache.nifi</group>
+      <version>1.8.0.3</version>
+    </bundle>
+    <comments></comments>
+    <descriptors>
+      <entry>
+        <key>Host</key>
+        <value>
+          <name>Host</name>
+        </value>
+      </entry>
+      <entry>
+        <key>Port</key>
+        <value>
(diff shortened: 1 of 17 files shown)


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/nifi/database-client-service.xqy
```diff
@@ -0,0 +1,124 @@
...
+    <name>DatabaseClient-STAGING</name>
+    <persistsState>false</persistsState>
+    <properties>
+      <entry>
+        <key>Host</key>
+        {element value{"${host}"}}
+      </entry>
+      <entry>
+        <key>Port</key>
+        <value>8010</value>
```
ryanjdew: Should this be a passed in value, since it can be configurable?

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/nifi/database-client-service.xqy
```diff
@@ -0,0 +1,124 @@
...
+      <entry>
+        <key>Username</key>
+        <value>{map:get($params, "username")}</value>
+      </entry>
+      <entry>
+        <key>Password</key>
+      </entry>
+      <entry>
+        <key>Database</key>
+        <value>data-hub-STAGING</value>
```
ryanjdew: Same as port. Maybe we should pull this value from /com.marklogic.hub/config.xqy?

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/nifi/database-client-service.xqy
rjrudin: That's a good call - I'll make it a process group variable with a default value of 8010.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/nifi/database-client-service.xqy
rjrudin: Same thing, I'll make it a process group variable with a default value.

bsrikan:
All tests seem to have passed. Will review again once hard coded values are updated per Ryan's comments

PR #2909, opened 2019-08-30
Feature/3041 null

rjrudin:


bsrikan:
@SameeraPriyathamTadikonda I think we can merge this PR. The 2 failures are a result of connection exception when mlClearModulesDatabase is called. Installed task test failed in the teardown step, so we can count that out. 
JobDeleteTaskTest failure in clearModulesDatabase isnt related to the fix anyway.

PR #2912, opened 2019-08-30
DHFPROD-3035: Custom ingestion step no longer has a default separator…

rjrudin:
… value

I couldn't remove "separator" because then I got a compilation error based on "separator" being required. So it at least doesn't have a default value anymore. I verified that if I switch the source format to "Delimited", then "Field Separator" appears with a default value of ",".


bsrikan:
Agree with @xnikhil08 . On creation the separator is empty string. As soon as you change it to any other type I see separator (,) assigned.

PR #2917, opened 2019-09-05
DHFPROD-2755: Initial manual merge work

ryanjdew:
Merge Documents
Manually merge documents together given a set of options.

REST API
POST /v1/resources/ml:smMerge

parameters

- rs:uri=\<URIn> – the URIs of the documents to merge. It is a repeating parameter for each document to merge together. 
- rs:flowName=\<nameOfFlowWithMastering> – The name of the flow that has the mastering settings.
- rs:step=\<masteringStepNumber> – optional; The number of the mastering step with settings. Defaults to 1.
- rs:preview=\<true|false> – optional; if true, the merge doc is returned in the response body and not committed to the database; if false, the merged document will be saved. Defaults to false.
- request body

JSON object with any step options to override.

Gradle task
./gradlew hubMergeEntities

parameters

- -PmergeURIs=\<URI1>,...,\<URIn> –  the URIs of the documents to merge, separated by commas.
- -PflowName=\<true|false> – optional; if true, the merged document will be moved to an archive collection; if false, the merged document will be deleted. Defaults to true.
- -Pstep=\<masteringStepNumber> – optional; The number of the mastering step with settings. Defaults to 1.
- -Ppreview=\<true|false> – optional; if true, the merge doc is returned in the response body and not committed to the database; if false, the merged document will be saved. Defaults to false.
- -Poptions=\<stepOptionOverrides> – optional; Any overrides to the mastering step options. Defaults to {}.

### DIFF
--- a/ml-data-hub-plugin/src/main/groovy/com/marklogic/gradle/task/MergeDocs.groovy
+++ b/ml-data-hub-plugin/src/main/groovy/com/marklogic/gradle/task/MergeDocs.groovy
@@ -1,0 +1,32 @@
+package com.marklogic.gradle.task
+
+import com.fasterxml.jackson.databind.JsonNode
+import com.fasterxml.jackson.databind.ObjectMapper
+import org.gradle.api.GradleException
+import org.gradle.api.tasks.TaskAction
+
+class MergeDocs extends HubTask  {
+    @TaskAction
+    void mergeDocs() {
+        def propName = "mergeURIs"
+        def mergeURIs = project.hasProperty(propName) ? project.property(propName).toString() : null
+        if (mergeURIs == null) {
+            throw new GradleException("mergeURIs is a required parameter. Supply the parameter with -PmergeURIs=URI1,URI2,...n")
+        }
+        def flowName = project.hasProperty('flowName') ? project.property('flowName').toString() : null
+        if (flowName == null) {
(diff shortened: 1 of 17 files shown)


ml-data-hub-plugin/src/main/groovy/com/marklogic/gradle/task/MergeDocs.groovy
```diff
@@ -0,0 +1,32 @@
+package com.marklogic.gradle.task
+
+import com.fasterxml.jackson.databind.JsonNode
+import com.fasterxml.jackson.databind.ObjectMapper
+import org.gradle.api.GradleException
+import org.gradle.api.tasks.TaskAction
+
+class MergeDocs extends HubTask  {
```
rjrudin: This should be MergeDocsTask for consistency with other tasks. 

I'm wondering too if we can be more specific and call it MergeEntitiesTask. Is it restricted to entities, or does smart mastering not depend on having entities defined?

ryanjdew: Changed to be consistent with other tasks and agree that "entities" is more accurate. Class name is now MergeEntitiesTask and the call is now hubMergeEntities

PR #2915, opened 2019-09-04
DHFPROD-2811: Added helper functions for mapping tests

rjrudin:
Also modified a couple tests to ensure they delete the documents that they create.

PR #2918, opened 2019-09-06
DHFPROD-2811: Mapping steps can now perform entity validation

rjrudin:
This also includes the changes sitting in PR - https://github.com/marklogic/marklogic-data-hub/pull/2915 . 

I'm pointing these at the 2406 mapping branch, but once that PR is merged, I'll modify this to point it at develop. 

Two of the tests I added for entity validation will fail on the trunk nightly build; I've opened up bug 53122 to address that.


rjrudin:
Will resubmit this later

PR #2936, opened 2019-09-11
dhfprod-3068: Don't merge just yet

aebadirad:


rjrudin:
Can the DHF extensions just get tossed under src/main/resources/ml-modules/services now? Or is it simpler for now to store them at their expected physical location?

rjrudin:
We're removing a lot of amps and modules as part of 5.0.3. I think we want those removed from user's applications as well:

1. The modules can be removed by recommending for users to run "gradle mlRedeploy" as part of upgrading to 5.0.3
1. The amps though are sitting in files in hub-internal-config. I think we want whatever Gradle task a user runs as part of upgrading to 5.0.3 to delete these as well. This could just be a new command that runs before amps are deployed and ensures that these amps are deleted. Thoughts?

ryanjdew:
@rjrudin The reason for the current location of the services/transforms in this PR is the "ml:" prefix needs to be added on load since OSes seem to have trouble filenames with ":" in them. The simplest solution to that seemed to be to add a DocumentFileProcessor that would add the proper prefix and REST permissions.

PR #2937, opened 2019-09-11
DHFPROD-2959: Throw HTTP 400 in case of plugin errors

srinathgit:
1. Throw HTTP 400 in case of plugin errors and handle error responses. 
2. I tested manually that 400 is thrown in case of plugin errors but would welcome suggestions on how to add automated tests in the current 4.x test infrastructure.


marklogic-data-hub/src/main/java/com/marklogic/hub/flow/impl/FlowRunnerImpl.java
```diff
@@ -400,8 +402,18 @@ public RunFlowResponse run(String jobId, String[] items, Map<String, Object> opt
                 }
             }
             catch(Exception e) {
-                e.printStackTrace();
-                throw new RuntimeException(e);
+                if(StringUtils.containsIgnoreCase(e.getMessage(), "Plugin error")){
```
rjrudin: I recommend putting this logic into a protected method so it can be easily unit tested - can pass an Exception into the method and verify that the correct logic is executed.

marklogic-data-hub/src/main/java/com/marklogic/hub/flow/impl/FlowRunnerImpl.java
```diff
@@ -400,8 +402,18 @@ public RunFlowResponse run(String jobId, String[] items, Map<String, Object> opt
...
+                    try {
+                        resp = objectMapper.readValue(((MarkLogicServerException)e).getFailedRequest().getMessage(), RunFlowResponse.class);
+                    } catch (IOException ex) {
+                        ex.printStackTrace();
+                        throw new RuntimeException(ex);
```
rjrudin: I think that if an IO error occurs while trying to parse the contents of the original exception, we should throw the original exception, not the IO error. Otherwise we'll lose the original error. Could include a message in the wrapped RuntimeException - i.e. "Unexpected IO error while parsing exception from running flow; original exception: " + e.getMessage().

marklogic-data-hub/src/main/java/com/marklogic/hub/flow/impl/FlowRunnerImpl.java
```diff
@@ -400,8 +402,18 @@ public RunFlowResponse run(String jobId, String[] items, Map<String, Object> opt
...
+                    }
+                }
+                else{
+                    e.printStackTrace();
```
rjrudin: I don't think it's necessary to call printStackTrace in either place here. Since a RuntimeException is being thrown, it will eventually be logged somewhere up the stack.

marklogic-data-hub/src/main/java/com/marklogic/hub/flow/impl/FlowRunnerImpl.java
srinathgit: I have done that and added a test

marklogic-data-hub/src/main/java/com/marklogic/hub/flow/impl/FlowRunnerImpl.java
srinathgit: Wrapped the original exception in RunTimeException and thrown that

marklogic-data-hub/src/main/java/com/marklogic/hub/flow/impl/FlowRunnerImpl.java
srinathgit: Removed it

PR #2938, opened 2019-09-11
DHFPROD-2989: Adding search:extract-path for headers in harmonized data and enabling search:transform-results

rahulvudutala:
This PR is to enable the explorer application to read the headers section from the extracted after search and also enable snippet information


rahulvudutala:
@rjrudin created PR https://github.com/marklogic/marklogic-data-hub/pull/2940

PR #2928, opened 2019-09-10
DHFPROD-2721:Validate the module for Custom step module is syntactically accurate

srinathgit:
This PR

1. Moves 4.x validation modules to legacy package
2. Backend/UI work for custom module validation

Created PR to this branch as custom step work is done in branch is done here

PR #2941, opened 2019-09-12
DHFPROD-2721:Validate the module for Custom step module is syntactically accurate

srinathgit:
This PR

1. Moves 4.x validation modules to legacy package
2. Backend/UI work for custom module validation

The review by @rjrudin on PR #2928 has been addressed in this PR

1. Should we always be instantiating a datahub in this fashion - i.e. with support for performance metrics? 
Removed support for perf metrics
2. Split the checking for permissions and static check into 2 different functions and added tests
3. Removed PUT and DELETE methods from the REST extension
4. Added clearer exception message in FlowManagerService when 'flow' is not found.


srinathgit:
@rjrudin ,
The delay in validating the new module is due to the delay in the updated flow document getting deployed to the server. The file watcher is quite slow and it takes a few seconds for the updated flow to be deployed. For validation, I use the uri of the "modulePath" from the flow document in server and hence the delay.

PR #2945, opened 2019-09-12
DHFPROD-2811: Mapping steps can now perform entity validation

rjrudin:
validateEntity:false is now included in scaffolded mapping steps.

Most of the changes are in test modules; just a handful are in main modules.

Commented out a couple lines in tests that were logging data.

Modified a couple tests to ensure they delete the documents that they create. 

Adding data-hub-test-helper.xqy to make it easier to create entities/flows/etc in unit tests. Rewrote a few tests to make use of this.


marklogic-data-hub/src/main/resources/hub-internal-artifacts/step-definitions/mapping/marklogic/default-mapping.step.json
```diff
@@ -14,6 +14,7 @@
     "collections": [
       "default-mapping",
       "mdm-content"
-    ]
+    ],
+    "validateEntity": false
```
bsrikan: Is this something that users should be able to set from GUI as well?

rjrudin: Eventually. This is just getting the backend in place.

rjrudin:
The last two test failures - https://jenkins.marklogic.com/job/Datahub_CI/job/PR-2945/6/testReport/ - are due to an ML bug - https://bugtrack.marklogic.com/53122 . I pushed a commit to disable those tests when running on a 10.0-2019* build.

PR #2952, opened 2019-09-14
DHFPROD-3068: Populate 'serverVersion' if it is null

srinathgit:
'serverVersion' gets populated only if dataHub.preInstallCheck() is called. bootstrap task doesn't call it. so adding a null check. This should fix the incorrect version in server rewriter


aebadirad:
Right, because the tests don't need to have it populated. I am a dummy.

PR #2953, opened 2019-09-16
DHFPROD-3137: Adding independent search options file for Explorer application

rahulvudutala:
* Created new search options file for Explorer Application so that it doesn't interfere with quick-start


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/hub-entities.xqy
```diff
@@ -177,12 +218,15 @@ declare %private function hent:fix-options($nodes as node()*)
       default return $n
 };
 
-declare function hent:dump-search-options($entities as json:array)
+declare function hent:dump-search-options($entities as json:array, $params as map:map)
```
rjrudin: This looks good, but one small tweak - unless we have a large number of params to pass around, it's better to explicitly declare the parameter here - i.e. "$for-explorer as xs:boolean?" (I would use "?" so it's fine to pass in an empty sequence).

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/hub-entities.xqy
```diff
@@ -105,6 +105,47 @@ declare function hent:wrap-duplicates(
   because of https://github.com/marklogic/entity-services/issues/359
 :)
 declare %private function hent:fix-options($nodes as node()*)
+{
```
rjrudin: I recommend adding unit tests - either to generateSearchOptions.sjs (which I added to address DHFPROD-2580), or to new modules in the same directory as generateSearchOptions. I would just add them to that module - you can put each test into a separate function.

rahulvudutala: Added the tests in the new commit

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/hub-entities.xqy
rahulvudutala: Changed the method signature to boolean in the new commits

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/impl/hub-entities/generateSearchOptions.sjs
```diff
@@ -11,13 +11,30 @@ const input =
...
   ),
   test.assertExists(options.xpath("/*:values[@name = 'Book']/*:range/*:element[@name = 'title']")),
+  test.assertExists(options.xpath("/*:transform-results[@apply = 'empty-snippet']")),
   test.assertNotExists(options.xpath("/*:values[@name = 'Book']/*:range/*:facet-option"),
     "A facet option does not need to be set on the values element")
 ];
+
+const expParams = true;
+const expOtions = hent.dumpSearchOptions(input, expParams);
+[
```
rjrudin: Just one tweak needed here - the assertions from above are being lost. To correct that, assign the first batch of assertions to an array - e.g. 

    let assertions = [the first batch of assertions];

Then add the others:

    assertions.concat(more assertions)

I tested this locally and verified that 9 assertions should be returned instead of just 5.

marklogic-data-hub/src/test/java/com/marklogic/hub/searchfacets/SearchFacetsTest.java
```diff
@@ -11,17 +11,19 @@
...
+import com.marklogic.hub.DataHub;
+import com.marklogic.hub.EntityManager;
 import com.marklogic.hub.HubConfig;
 import com.marklogic.hub.HubTestBase;
 import com.marklogic.hub.flow.impl.FlowRunnerImpl;
 import com.marklogic.hub.impl.HubConfigImpl;
 import com.marklogic.hub.step.StepDefinition;
 import org.apache.commons.io.FileUtils;
 import org.custommonkey.xmlunit.XMLUnit;
+import org.junit.Test;
```
rjrudin: This needs to be org.junit.jupiter.api.Test.   I ran the test locally, and both failed with a null-pointer on the "hubConfig" object because it wasn't initialized.

Also, after I fixed that locally, I ran the tests again but I got the following error:

om.marklogic.client.FailedRequestException: Local message: search failed: Bad Request. Server Message: XDMP-PATHRIDXNOTFOUND: cts:path-reference("//*:instance/e2eentity/name", ("type=string", "collation=http://marklogic.com/collation/codepoint")) -- No string path range index for //*:instance/e2eentity/name collation=http://marklogic.com/collation/codepoint

PR #2957, opened 2019-09-17
DHFPROD-3124:Adding a endpoint to fetch xpath and mapping functions

srinathgit:
Adding a endpoint to fetch xpath and mapping functions. It can be accesses using /api/current-project/mappings/functions endpoint

### DIFF
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/mapping/MappingFunctions.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/mapping/MappingFunctions.java
@@ -1,0 +1,27 @@
+package com.marklogic.hub.mapping;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.marklogic.client.DatabaseClient;
+import com.marklogic.client.extensions.ResourceManager;
+import com.marklogic.client.extensions.ResourceServices;
+import com.marklogic.client.io.JacksonHandle;
+import com.marklogic.client.util.RequestParameters;
+
+public class MappingFunctions extends ResourceManager {
+    private static final String NAME = "ml:mappingFunctions";
+
+    public MappingFunctions(DatabaseClient client) {
+        super();
+        client.init(NAME, this);
+    }
+
+    public JsonNode getMappingFunctions() {
+        ResourceServices.ServiceResultIterator resultItr = this.getServices().get(new RequestParameters());
+        if (resultItr == null || ! resultItr.hasNext()) {
+            return null;
+        }
+        ResourceServices.ServiceResult res = resultItr.next();
+        return res.getContent(new JacksonHandle()).get();
+    }
+}
+
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/mapping/MappingFunctions.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/mapping/MappingFunctions.java
@@ -3,7 +3,6 @@ package com.marklogic.hub.mapping;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.marklogic.client.DatabaseClient;
 import com.marklogic.client.extensions.ResourceManager;
-import com.marklogic.client.extensions.ResourceServices;
 import com.marklogic.client.io.JacksonHandle;
 import com.marklogic.client.util.RequestParameters;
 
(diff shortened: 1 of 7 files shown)


marklogic-data-hub/src/main/java/com/marklogic/hub/mapping/MappingFunctions.java
```diff
@@ -0,0 +1,27 @@
...
+public class MappingFunctions extends ResourceManager {
+    private static final String NAME = "ml:mappingFunctions";
+
+    public MappingFunctions(DatabaseClient client) {
+        super();
+        client.init(NAME, this);
+    }
+
+    public JsonNode getMappingFunctions() {
+        ResourceServices.ServiceResultIterator resultItr = this.getServices().get(new RequestParameters());
```
rjrudin: You should be able to simplify this to just the following:

    return getServices().get(new RequestParameters(), new JacksonHandle()).get();

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/extensions/mappingFunctions.xqy
```diff
@@ -0,0 +1,58 @@
...
+
+declare function mappingFunctions:source-format() as xs:string {
+  "javascript"
+};
+declare %rapi:transaction-mode("query") function mappingFunctions:get(
+  $context as map:map, $params as map:map
+) as map:map {
+  xdmp:apply($caller,$extName,$modPath,"GET",$context,$params)
+};
+declare %rapi:transaction-mode("query") function mappingFunctions:delete(
```
rjrudin: Because there's only a GET function, I think it's worth simplifying this by removing the delete/post/put functions.

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/mapping-functions/getMappingFunctions.sjs
```diff
@@ -0,0 +1,11 @@
+const mappingFunctions = require('/data-hub/5/extensions/mappingFunctions.sjs');
+const test = require("/test/test-helper.xqy");
+
+function testMappingFunctions() {
+return [
+    test.assertEqual(129, Object.keys(mappingFunctions.getXpathFunctions()).length),
```
rjrudin: I've done assertions like this before on other projects, but they tend to be brittle as this number can go up at any point. I recommend changing this to "assertTrue" that the number of functions is at least 129, and include an assertion message stating that as of ML version X.X, we expect at least 129 functions in the "fn:" namespace.

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/mapping-functions/getMappingFunctions.sjs
```diff
@@ -0,0 +1,11 @@
...
+    test.assertEqual(4, Object.keys(mappingFunctions.getMarkLogicFunctions()).length)
```
rjrudin: I think an assertion message is neeed here too to explain why 4 functions are expected - i.e. state that we expect there to be 4 functions in the "DHF core" namespace.

web/src/main/java/com/marklogic/hub/web/service/MappingManagerService.java
```diff
@@ -99,4 +100,8 @@ public MappingModel getMapping(String mappingName, boolean createIfNotExisted) t
         }
     }
 
+    public JsonNode getMappingFunctions() {
```
rjrudin: I don't see the point in these *Service objects. It seems simpler with less indirection for MappingController to autowire in the instance of HubConfig and then create MappingFunctions itself. It's one line of code then in the controller:

    return new MappingFunctions(hubConfig.newStagingClient()).getMappingFunctions();

srinathgit: I didn't want to break the precedence that we had . I will make that change

srinathgit:
@rjrudin ,
I have made all the changes you had suggested

web/src/main/java/com/marklogic/hub/web/service/MappingManagerService.java
rjrudin: I think next time we need code like this, it'd be worth cleaning up the other areas that use a results iterator instead of just using JacksonHandle.

PR #2959, opened 2019-09-17
DHFPROD-2778: Now have an endpoint for validating a mapping

rjrudin:
A separate story will handle having the UI utilize this endpoint, which is /v1/resources/ml:mappingValidator . Unit tests are included for this endpoint, as well as a JUnit test to verify that the Java code works correctly.


rjrudin:
I added a commit to catch and log errors that intermittently occur when trying to delete temporary build directories. Not related to this ticket at all, but the one failure in the last test run for this PR was due to that error.

rjrudin:
I rebased this on develop to fix a merge conflict in extensions-util.xqy.

rjrudin:
@SameeraPriyathamTadikonda - the test failure is not related and is environment-based - can you still merge this?

PR #2962, opened 2019-09-18
DHFPROD-2756: Manual merge should also create auditing doc

ryanjdew:


ryanjdew:
@rjrudin Added an assertion for the existence of an auditing doc

PR #2965, opened 2019-09-18
DHFPROD-2959:Check if error is a plugin error

srinathgit:
Throw HTTP 400 if errorCount > 0 and if the error is a "DATAHUB-PLUGIN-ERROR"


aebadirad:
Let's make sure quickstart handles these responses too

PR #2969, opened 2019-09-18
DHFPROD-3063: Declarative mapper tests

bsrikan:
Covers happy path and negative test scenarios for lookup functions, some xpath functions, date and dateTime functions and a custom function.
There are some tests disabled because of issues in reaching the part of the code that generates error.

### DIFF
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/step/RunStepResponse.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/step/RunStepResponse.java
@@ -16,12 +16,15 @@
 package com.marklogic.hub.step;
 
 import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import com.marklogic.hub.error.DataHubConfigurationException;
 import com.marklogic.hub.flow.Flow;
 import com.marklogic.hub.job.JobStatus;
 import com.marklogic.hub.step.impl.Step;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 
@@ -194,5 +197,10 @@ public class RunStepResponse {
                 "failedBatches: %d, stepStartTime: %s , stepEndTime: %s]", flowName, stepName, stepDefinitionName,
             stepDefinitionType, String.valueOf(success), status, totalEvents, successfulEvents, failedEvents,
             successfulBatches, failedBatches, stepStartTime, stepEndTime);
+    }
+
+    public JsonNode outputToJson(List<String> stepOutput, int index, String field) throws Exception{
+        JsonNode jsonOutput = new ObjectMapper().readTree(stepOutput.toString());
+        return jsonOutput.get(index).get(field);
     }
 }
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/step/RunStepResponse.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/step/RunStepResponse.java
@@ -16,15 +16,12 @@
 package com.marklogic.hub.step;
 
 import com.fasterxml.jackson.annotation.JsonIgnore;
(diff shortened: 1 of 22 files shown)


marklogic-data-hub/src/main/java/com/marklogic/hub/step/RunStepResponse.java
```diff
@@ -195,4 +198,9 @@ public String toString() {
             stepDefinitionType, String.valueOf(success), status, totalEvents, successfulEvents, failedEvents,
             successfulBatches, failedBatches, stepStartTime, stepEndTime);
     }
+
```
srinathgit: Can this method be moved to HubTestBase ? This doesn't have to be here

aebadirad: Also a single test needs to have one run with "fullOutput: true" on in options to capture the fulloutput to make sure we're passing that back cleaning post mapping.

bsrikan: @srinathgit sure. @aebadirad will we achieve the same results by testing with fullOutput set to true in other tests where we pass options?

bsrikan: @aebadirad I know from the jnj slack channel that there is an issue with fullOutput: true. So dont think adding that in the test will add any value unless that bug is fixed.

marklogic-data-hub/src/test/java/com/marklogic/hub/HubTestBase.java
```diff
@@ -964,6 +987,24 @@ protected void installHubArtifacts(HubConfig hubConfig, boolean force) {
...
+        commands.add(generateFunctionMetadataCommand);
+
+        SimpleAppDeployer deployer = new SimpleAppDeployer(((HubConfigImpl)hubConfig).getManageClient(), ((HubConfigImpl)hubConfig).getAdminManager());
+        deployer.setCommands(commands);
+        deployer.deploy(hubConfig.getAppConfig());
+
+        // Provides some time for post-commit triggers to complete
+        try {
+            Thread.sleep(3000);
+        } catch (InterruptedException e) {
```
bsrikan: I have a custom function test that is failing because the post commit trigger is taking longer than 3 sec to complete. It was fine locally, I tested it 5 times in a row. @rjrudin @srinathgit @aebadirad @ryanjdew, are you guys OK if I bump it up to 5sec?

rjrudin: Yup, fine with me. Though I think a nicer approach would be to check every second to see if the schemas database now has the expected artifact in it.

bsrikan:
Test is failing even after bumping up the sleep to 5sec. Runs fine locally. Nothing else to add here for this PR. Can someone else pls take a look. I have to focus on testing upgrade for 5.0.3.

rjrudin:
I'm looking into the test failure, I'll try to replace the "sleep" method with a check on the schemas being created. 

I'm also wondering how much overlap there is with this test and all the unit tests we have for mapping expressions. What gaps does this address, and could those gaps be addressed by just adding more unit tests, where we don't seem to have a need for "sleep" calls?

rjrudin:
I'm getting test failures on 2 tests - testDateANDDateTime and testCustomFunction. Both are timezone issues - the test expects "-07:00", but I'm running against ML on a Docker container, and "Z" is being returned instead.

rjrudin:
Submitted this PR to move the sleep code around - we only need it after "loadUserModules" (which should really be called "loadUserModulesAndArtifacts") since that's the only thing that loads entities, which is the only thing affected by post-commit triggers. 

https://github.com/bsrikan/marklogic-data-hub/pull/3

bsrikan:
installUserModules doesnt generate the function metadata and so moving the sleep there is not helping. Custom function test is still failing with undefined function because the xml and xslt are not generated in time for the test.

srinathgit:
Closing this as the commits in this PR have been brought in #2985

PR #2972, opened 2019-09-19
DHFPROD-3068: Be more responsible with our streams

ryanjdew:
See how this affects the stability of our pipeline


ryanjdew:
Also changed the CD pipeline test for bootstrap to check for options using the QueryOptionsManager, since it can no longer be read directly from modules.

PR #2975, opened 2019-09-19
DHFPROD-3162: Added command line option to not use SSL

rjrudin:
Can include --disableSsl=true

Also modifying CopyQueryOptionsCommand to not fail if a particular URI is not found


rjrudin:
@SameeraPriyathamTadikonda The test failure is not related to these changes and is environment-based - can you merge this?

PR #2979, opened 2019-09-20
DHFPROD-3068: Merge 3068 changes into develop for 5.1

ryanjdew:


rjrudin:
I pushed a commit to fix the one test that was failing, which was simply due to referencing a module for test purposes that no longer exists.

aebadirad:
Will need to update this again with the latest changes to 5.x-develop @ryanjdew

PR #2982, opened 2019-09-20
DHFPROD-3194: Fixing mlUpdateIndexesTest so it runs after saving the indexes

bsrikan:
The test here is enough to verify server side fix for "language":"zxx"
It is expected to fail in nightly builds until the bug is fixed.

I see a comment in the PR against develop and it can be enhanced later to verify generated json file.


SameeraPriyathamTadikonda:
Do we need these changes for 5.0.3?

srinathgit:
@SameeraPriyathamTadikonda ,
1. The failure in the pipeline is because the tests run against 10.0-2. It will pass against 10.0-nightly
2. Yes, this test is needed for 5.0.3. I will create a PR for 5.0.3 once this is merged to 5.x-develop

aebadirad:
@srinathgit make a commit, I want to make sure we get a weekend of regressions on 5.0.3 branch tonight

PR #2983, opened 2019-09-20
DHFPROD-3196: remove the variable assignment and return the evaluatio…

aebadirad:
…n response out raw

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/endpoints/collector.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/endpoints/collector.sjs
@@ -66,10 +66,8 @@ if(method === 'GET') {
           } else {
             urisEval = "cts.uris(null, null, " + query + ")";
           }
-          resp = xdmp.eval(urisEval, {options: options}, {database: xdmp.database(database)});
+          xdmp.eval(urisEval, {options: options}, {database: xdmp.database(database)});
         } catch (err) {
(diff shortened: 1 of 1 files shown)


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/endpoints/collector.sjs
```diff
@@ -66,10 +66,8 @@ if(method === 'GET') {
           } else {
             urisEval = "cts.uris(null, null, " + query + ")";
           }
-          resp = xdmp.eval(urisEval, {options: options}, {database: xdmp.database(database)});
+          xdmp.eval(urisEval, {options: options}, {database: xdmp.database(database)});
```
rjrudin: I think a comment is worthwhile to explain why this isn't assigned to anything. I looked at the code, and due to the nested if statements and the try/catch block, it's not immediately obvious that this could be the last statement in the function. 

Those nested if statements could be avoided by saying e.g. "If there's no step doc, throw an error; if there's no query, throw an error; then at the very end of the function, run the try/catch block". 

Also, is it necessary to assign the result of "fn.error" to "resp"? Doesn't the error get thrown right away?

aebadirad: This was actually a test to see if the future is resolved in the try catch, or if the var assignment was causing the future to wait.

Looks like the try/catch is the source of the issue.

PR #2985, opened 2019-09-23
DHFPROD-3192:Unable to use a custom mapping function

srinathgit:
This PR takes in commits from #2969 and fixes the failing tests. I will close #2969 once this gets merged.

### DIFF
--- a/marklogic-data-hub/src/test/java/com/marklogic/hub/HubTestBase.java
+++ b/marklogic-data-hub/src/test/java/com/marklogic/hub/HubTestBase.java
@@ -969,6 +969,13 @@ public class HubTestBase {
 
         LoadModulesCommand loadModulesCommand = new LoadModulesCommand();
         commands.add(loadModulesCommand);
+        //Call 'generateFunctionMetadataCommand' if mapping functions are deployed. If mapping is deployed first
+        // and it references custom functions, mapping xslt will not have reference to custom functions and tests
+        // would fail with XDMP-UNDFUN: (err:XPST0017) Undefined function customDateTime().
+
+        if(hubConfig.getHubProject().getCustomModulesDir().resolve("mapping-functions").toFile().exists()){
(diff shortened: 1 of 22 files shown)


marklogic-data-hub/src/test/java/com/marklogic/hub/HubTestBase.java
```diff
@@ -705,6 +707,22 @@ protected int getDocCountByQuery(String database, String query) {
         return count;
     }
 
+    protected JSONObject getQueryResults(String query, String database) {
```
rjrudin: Can this just return a JsonNode instead of JSONObject? What's the advantage of JSONObject over JsonNode?

marklogic-data-hub/src/test/java/com/marklogic/hub/HubTestBase.java
```diff
@@ -946,6 +969,13 @@ protected void installUserModules(HubConfig hubConfig, boolean force) {
 
         LoadModulesCommand loadModulesCommand = new LoadModulesCommand();
         commands.add(loadModulesCommand);
+        //Call 'generateFunctionMetadataCommand' if mapping functions are deployed. If mapping is deployed first
+        // and it references custom functions, mapping xslt will not have reference to custom functions and tests
+        // would fail with XDMP-UNDFUN: (err:XPST0017) Undefined function customDateTime().
+
+        if(hubConfig.getHubProject().getCustomModulesDir().resolve("mapping-functions").toFile().exists()){
+            commands.add(generateFunctionMetadataCommand);
```
rjrudin: Is there any downside to always adding this? I think it's fine to always add it. Otherwise, future tests may not work because they don't meet the condition of the "if" statement.

srinathgit: Ok, will include that.. no downside except a few a delay of few ms every time this method is called

PR #2987, opened 2019-09-24
DHFPROD-3110: Adding architect and developer roles for Explorer

rjrudin:
There's an open question about whether anything further should be done about the developer role, but I wanted to get what's in here out for review as I think it's stable.

There's some random white-space cleanup in here as well, probably from when I auto-formatted the code using Intellij.


rjrudin:
Removed explorer-architect-view-role

rjrudin:
Fixed the merge conflict, should be good to go now.

PR #2990, opened 2019-09-24
DHFPROD-3068: Take ...n on ml: prefix handling

ryanjdew:


marklogic-data-hub/src/main/resources/ml-modules/services/mlDbConfigs.xqy
```diff
@@ -15,7 +15,7 @@
 :)
 xquery version "1.0-ml";
 
-module namespace service = "http://marklogic.com/rest-api/resource/ml:dbConfigs";
+module namespace service = "http://marklogic.com/rest-api/resource/mlDbConfigs";
```
srinathgit: This would mean that we would be breaking the published REST API . I think there are customers using it as there was a support case yesterday about this.
https://marklogic.github.io/marklogic-data-hub/harmonize/rest/
https://marklogic.github.io/marklogic-data-hub/ingest/rest/
https://docs.marklogic.com/datahub//flows/run-flow-using-rest-client-api.html

ryanjdew: There is a hack to our REST library that will translate the "ml:" prefix to camelcase for us. The "ml:" prefix will need to go away eventually, though.

srinathgit: ok.. I missed that snippet while reviewing

rjrudin:
Agree, this looks great!

PR #2995, opened 2019-09-25
DHFPROD-3193: Test fixes

rjrudin:
Mostly related to mapping


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mapping/entity-services/lib.sjs
```diff
@@ -271,6 +271,10 @@ function validatePropertyMapping(targetEntityType, propertyName, sourcedFrom) {
 }
 
 function versionIsCompatibleWithES(version = xdmp.version()) {
+  // Turns out 9.0-10.2 is fine, but 9.0-10 is not
```
aebadirad: 9.0-10.1 isn't either, as n fyi. Nor is 10.01

rjrudin: I didn't see a 9.0-10.1 in mlregistry.marklogic.com

aebadirad: It's a released version. Trust me on that. Min versions for this stuff should be 10.02 and 9.10-2 or later.

rjrudin:
@aebadirad xdmp.effectiveVersion isn't giving me what we want (testing via Docker images of ML).

- On 9.0-10, it returns "9001000".
- On 9.0-10.2, it returns "9001000".
- On trunk build 9.0-20190911, it returns "9001000".

I agree that the current impl is hacky, but the scope of this PR is to just get the tests passing. I think if we want a better implementation, a separate JIRA ticket is needed which will involve some research into something less brittle. But xdmp.effectiveVersion doesn't seem to be the answer.

srinathgit:
@rjrudin ,
As part of DHFPROD-3102 which I am working on in this sprint, I have a logic that works fine with different hotfix, patch version as well as nightly. I think as you said for now, this makes all the tests to pass. When I submit the PR for DHFPROD-3102, I will update these tests as well.

aebadirad:
> But xdmp.effectiveVersion doesn't seem to be the answer.

Ah, right. I'm looking for xdmp.softwareVersion. Effective version gives us the 'rounded' version (eg 10.0-2.1 is 10.0-2).

If you want to make a new ticket, sure, but label it for 5.1.0. I'd like to get us using something other than string based matching.

aebadirad:
fixed the rebase issue

PR #3005, opened 2019-09-27
DHFPROD-2986: mastering summary document

ryanjdew:


marklogic-data-hub/src/main/resources/ml-modules/services/mlJobs.sjs
```diff
@@ -82,13 +82,36 @@ function post(context, params, input) {
...
+            if (jobReport) {
+              datahub.hubUtils.writeDocument(`/jobs/reports/${stepResp.flowName}/${step}/${jobId}.json`, jobReport, datahub.jobs.jobsPermissions, ['Jobs','JobReport'], datahub.config.JOBDATABASE);
+            }
+          }
         }
     }
 
     //Update the job doc
-    datahub.hubUtils.writeDocument("/jobs/"+ jobId +".json", jobDoc, "xdmp.defaultPermissions()", ['Jobs','Job'], datahub.config.JOBDATABASE);
+    datahub.hubUtils.writeDocument("/jobs/"+ jobId +".json", jobDoc, datahub.jobs.jobsPermissions, ['Jobs','Job'], datahub.config.JOBDATABASE);
```
rjrudin: Is this fixing a bug - the change from defaultPermissions to jobsPermissions?

ryanjdew: There wasn't a specific bug. It was inconsistent with how the permissions were set in when created/updated through runFlow.

PR #3006, opened 2019-09-27
DHFPROD-3010: Backend work for single mapping JSON artifact with nested entities

ryanjdew:


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mapping/entity-services/lib.sjs
```diff
@@ -119,6 +121,10 @@ function getRelatedMappings(mapping, related = [mapping]) {
   if (dhMappingTraceIsEnabled) {
     xdmp.trace(dhMappingTrace, `Getting related mappings for '${xdmp.describe(mapping)}'`);
   }
+  let internalMappings = mapping.xpath('/properties//object-node()[exists(targetEntityType) and exists(properties)]');
+  for (let internalMapping of internalMappings) {
+    related.push(internalMapping);
+  }
   let externalReferences = mapping.xpath('/properties/*/externalMapping');
   for (let mappingRef of externalReferences) {
```
srinathgit: @ryanjdew ,
Aren't you removing "externalMapping" property from the mapping artifact ? So, should this piece of code remain ?

ryanjdew: Good catch. Should restrict to just internal sub-mappings to reduce the amount code we maintain.

ryanjdew: Removed the externalMapping related code in the latest push.

PR #3010, opened 2019-09-28
DHFPROD-3231: Explorer Options- Adding word query for createdByJob field and allowing namespaces on search:extract-path

rahulvudutala:


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/hub-entities.xqy
```diff
@@ -172,6 +172,11 @@ declare %private function hent:fix-options-exp($nodes as node()*)
               <search:field name="datahubCreatedByJob"/>
             </search:range>
           </search:constraint>,
+          <search:constraint name="createdByJobWordConstraint">
```
rjrudin: I know there are a few existing constraints with names ending in "Constraint", but I think for consistency's sake it's better to not use that as a suffix. So do "createdByJobWord", and please remove "Constraint" from the existing constraint names. Otherwise, I think there'd be an expectation that all constraint names should end with "Constraint", and that's not the case and likely won't ever be the case.

rahulvudutala: Addressed the change

PR #3011, opened 2019-09-29
DHFPROD-3242: Add data-services endpoint for querying latest job info

akshaysonvane:
Does not include the Gradle task to generate the java class as DHF will not be using the endpoint for now.

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/jobInfo/jobInfo.api
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/jobInfo/jobInfo.api
@@ -1,0 +1,13 @@
+{
+    "functionName": "latestJobInfo",
+    "params": [
+        {
+            "name": "model",
+            "datatype": "string"
+        }
+    ],
+    "return": {
+        "datatype": "jsonDocument",
+        "$javaClass": "com.fasterxml.jackson.databind.JsonNode"
+    }
+}
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/jobInfo/jobInfo.api
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/jobInfo/jobInfo.api
@@ -1,8 +1,8 @@
 {
-    "functionName": "latestJobInfo",
+    "functionName": "getLatestJobData",
     "params": [
         {
-            "name": "model",
+            "name": "entityCollection",
             "datatype": "string"
         }
     ],
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/jobInfo/jobInfo.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/jobInfo/jobInfo.sjs
@@ -1,0 +1,41 @@
+/*
+  Copyright 2012-2019 MarkLogic Corporation
+
+  Licensed under the Apache License, Version 2.0 (the "License");
+  you may not use this file except in compliance with the License.
+  You may obtain a copy of the License at
+
+     http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+*/
+'use strict';
+
+var model;
+
+let res = {
+  "modelName": model,
+  "jobId": null,
+  "jobTime": null
+};
+
+let query = [cts.collectionQuery(model), cts.fieldRangeQuery("datahubCreatedOn", "<=", fn.currentDateTime(), "score-function=reciprocal")];
+let seq = fn.subsequence(cts.search(cts.andQuery(query)), 1, 1);
+
+if (fn.count(seq) > 0) {
+  let docUri = xdmp.nodeUri(seq);
+
+  res.jobTime = xdmp.documentGetMetadataValue(docUri, "datahubCreatedOn");
+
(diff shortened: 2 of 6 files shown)


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/jobInfo/service.json
```diff
@@ -0,0 +1,4 @@
+{
+  "endpointDirectory": "/data-hub/5/data-services/jobInfo/",
+  "$javaClass": "com.marklogic.hub.explorer.job.JobInfo"
```
akshaysonvane: The current path resolves in the explorer project.
Please suggest a path if DHF also wants to generate the endpoint Java class.

rjrudin: Similar to how we don't want "explorer" in the path of the ML module, I don't think we want it in the Java class either, as we don't expect our data services to be specific to the Explorer app. 

While I normally think that Java packages should be organized around features, I think the set of interfaces generated for data services is an exception to that. I think the organization would be better if all such interfaces were in a single package so that it's easy to quickly inspect all such interfaces to MarkLogic. 

For that reason, I think "com.marklogic.hub.dataservices" would be a good package to use. I don't think a child package under that is necessary for a data service unless we expect there to be multiple classes associated with a particular data service.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/jobInfo/jobInfo.sjs
```diff
@@ -0,0 +1,41 @@
...
+  res.jobTime = xdmp.documentGetMetadataValue(docUri, "datahubCreatedOn");
+
+  let jobIds = xdmp.documentGetMetadataValue(docUri, "datahubCreatedByJob");
+  if (jobIds) {
+    jobIds = jobIds.split(" ");
+    res.jobId = jobIds[jobIds.length - 1];
+  }
+}
+
+res;
```
rjrudin: Since we're adding this to the DHF project, I think we need to have tests for it as well in the project. I'll spend a little time with this on Monday to determine what the right mix of ML-based unit tests and Java-based integration tests would be appropriate (I'm currently leaning towards just unit tests).

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/jobInfo/jobInfo.api
```diff
@@ -0,0 +1,13 @@
+{
+    "functionName": "latestJobInfo",
```
rjrudin: I recommend calling this "getLatestJobData". I think we want data services to have names that are verb-oriented, like method names. I was initially thinking "getLatestJobDataForEntityCollection", but I can see this expanding to support additional parameters.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/jobInfo/jobInfo.api
```diff
@@ -0,0 +1,13 @@
...
+    "params": [
+        {
+            "name": "model",
```
rjrudin: I recommend calling this parameter "entityCollection" since it is assumed to be a collection value.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/jobInfo/jobInfo.sjs
```diff
@@ -0,0 +1,41 @@
...
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+*/
+'use strict';
+
+var model;
+
+let res = {
+  "modelName": model,
```
rjrudin: I would make the following changes:

- modelName = entityCollection
- jobId = latestJobId
- jobTime = latestJobDateTime

I like to use "DateTime" as the suffix to imply that a value is an xs.dateTime, and likewise for "Date" (xs.date) and "Time" (xs.time).

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/jobInfo/jobInfo.api
akshaysonvane: Included in https://github.com/marklogic/marklogic-data-hub/pull/3011/commits/e1d00a2bc5e0cc4d4afe9e00cfd0fc146f4f53aa

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/jobInfo/jobInfo.api
akshaysonvane: Included in https://github.com/marklogic/marklogic-data-hub/pull/3011/commits/e1d00a2bc5e0cc4d4afe9e00cfd0fc146f4f53aa

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/jobInfo/jobInfo.sjs
akshaysonvane: Included in https://github.com/marklogic/marklogic-data-hub/pull/3011/commits/e1d00a2bc5e0cc4d4afe9e00cfd0fc146f4f53aa

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/jobInfo/service.json
akshaysonvane: Included in https://github.com/marklogic/marklogic-data-hub/pull/3011/commits/1720b70725138f9db551e7fb8897364f408f8c76

PR #3013, opened 2019-09-30
Ensuring language is not submitted in a database payload

rjrudin:
This fixes the error in UpdateIndexesTaskTest, but the test still fails for me locally for what seems like the following reason - the number of path range indexes isn't increased because hubSaveIndexes is called before the count of current indexes is retrieved. 

I don't think this test is really testing much that isn't covered elsewhere, and for that reason, I think it would suffice for the test to just verify that no errors occur.


aebadirad:
Make sure this makes its way back to dev and 5.x dev

srinathgit:
@rjrudin ,
1. Are you using a nightly server > 10.0-20190924 ? If not, there will be "language" in the db payload
2. The test has issues in that it will run fine only if the db's don't have any indexes. So, in a fresh DHF installation, the test would run fine. Once you have run it once, the test fails because the indexes are already present

PR #3015, opened 2019-10-01
DHFPROD-2746: Enable read-only mastering endpoints

ryanjdew:
REST API details found here: https://wiki.marklogic.com/pages/viewpage.action?pageId=125345370


marklogic-data-hub/src/main/java/com/marklogic/hub/impl/MasteringManagerImpl.java
```diff
@@ -128,4 +145,47 @@ public MergeResponse merge(List<String> mergeURIs, String flowName, String stepN
...
+            ResourceServices.ServiceResultIterator resultItr = this.getServices().post(params, jsonOptions);
+            try {
+                if (resultItr == null || !resultItr.hasNext()) {
+                    resp = null;
+                } else {
+                    ResourceServices.ServiceResult res = resultItr.next();
+                    StringHandle handle = new StringHandle();
+                    ObjectMapper objectMapper = new ObjectMapper();
+                    objectMapper.disable(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES);
+                    objectMapper.enable(DeserializationFeature.UNWRAP_ROOT_VALUE);
```
rjrudin: Does using a JacksonHandle to capture the response not work because these features need to be set? i.e. does this not work?

    return getServices().post(params, jsonOptions, new JacksonHandle()).get();

That would give you a JsonNode instead of a MatchResponse, but I'm wondering if MatchResponse is really needed. Who's the client for this interface? If it's a GUI, odds are the MatchResponse class isn't needed and a JsonNode will work.

ryanjdew: Good point. Refactored to use more generic JacksonHandle/JsonNode.

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/MasteringManagerImpl.java
```diff
@@ -74,31 +89,30 @@ private DatabaseClient getSrcClient() {
...
             params.put("blockFutureMerges", blockFutureMerges.toString());
             params.put("targetDatabase", targetDatabase);
             params.put("sourceDatabase", targetDatabase);
-            StringHandle handle = new StringHandle();
+            JacksonHandle handle = new JacksonHandle();
             this.getServices().delete(params, handle);
             try {
-                ObjectMapper objectMapper = new ObjectMapper();
-                resp = objectMapper.readValue(handle.get(), UnmergeResponse.class);
+                resp = handle.get();
```
rjrudin: I don't think the try/catch is needed - I don't think handle.get() throws a checked exception.

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/MasteringManagerImpl.java
```diff
@@ -114,9 +128,48 @@ public MergeResponse merge(List<String> mergeURIs, String flowName, String stepN
...
                 } else {
                     ResourceServices.ServiceResult res = resultItr.next();
-                    StringHandle handle = new StringHandle();
-                    ObjectMapper objectMapper = new ObjectMapper();
-                    resp = objectMapper.readValue(res.getContent(handle).get(), MergeResponse.class);
+                    JacksonHandle handle = new JacksonHandle();
+                    resp = res.getContent(handle).get();
+                }
+            } catch (Exception e) {
+                throw new RuntimeException(e);
```
rjrudin: I don't think this try/catch is needed either, not the use of ServiceResult - should just be able to do:

    return getServices().post(params, jsonOptions, new JacksonHandle()).get();

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/MasteringManagerImpl.java
```diff
@@ -114,9 +128,48 @@ public MergeResponse merge(List<String> mergeURIs, String flowName, String stepN
...
+            params.put("uri", matchURI);
+            params.put("includeMatchDetails", includeMatchDetails.toString());
+            params.put("flowName", flowName);
+            params.put("step", stepNumber);
+            params.put("targetDatabase", targetDatabase);
+            params.put("sourceDatabase", targetDatabase);
+            JacksonHandle jsonOptions = new JacksonHandle().with(options);
+            ResourceServices.ServiceResultIterator resultItr = this.getServices().post(params, jsonOptions);
+            try {
+                if (resultItr == null || !resultItr.hasNext()) {
```
rjrudin: Same thing, I think you just need post(params, jsonOptions, new JacksonHandle()).get()

marklogic-data-hub/src/test/java/com/marklogic/hub/master/MasterTest.java
```diff
@@ -97,18 +109,20 @@ private HubModuleManager getPropsMgr() {
...
-        assertEquals(0, getDocCount(HubConfig.DEFAULT_FINAL_SCHEMAS_DB_NAME, "http://marklogic.com/xdmp/tde"));
-
-        getDataHub().clearDatabase(HubConfig.DEFAULT_STAGING_SCHEMAS_DB_NAME);
-        assertEquals(0, getDocCount(HubConfig.DEFAULT_STAGING_SCHEMAS_DB_NAME, "http://marklogic.com/xdmp/tde"));
-
-        installHubArtifacts(getDataHubAdminConfig(), true);
-        installUserModules(getDataHubAdminConfig(), true);
+    public void testMatchEndpoint() throws Exception {
+        Flow flow = flowManager.getFlow("myNewFlow");
+        if (flow == null) {
```
rjrudin: This doesn't require a change, but in a test, I think it's nicer to just do "assertNotNull(flow)" - less code, and now the method doesn't need to declare that it throws a checked exception.

PR #3024, opened 2019-10-02
Feature/dhfprod 3010

xnikhil08:
Some changes for DHFPROD-3021 and fix for fontAwesome PRO icons.

PR #3025, opened 2019-10-02
DHFPROD-3010: ES Mapping in UI and related stories

srinathgit:
This PR includes

1. DHFPROD-3010 
2. DHFPROD-3019
3. DHFPROD-3102

Once PR #3024 gets merged to feature branch, it will include  DHFPROD-3022


marklogic-data-hub/src/test/java/com/marklogic/hub/LoadTestModules.java
```diff
@@ -48,11 +47,10 @@ public static void loadTestModules(String host, int finalPort, String username,
         config.getModulePaths().add("../build/mlBundle/marklogic-unit-test-modules/ml-modules");
         config.getModulePaths().add("../src/test/ml-modules");
 
-        // Need to run GenerateFunctionMetadataCommand as well so that function metadata is generated both for
-        // core mapping functions and custom functions under src/test/ml-modules/root/custom-modules.
+        // Removing GenerateFunctionMetaDataCommand as the command has spring related dependencies (hubConfig, versions
```
rjrudin: Doesn't this cause some of the unit tests to fail then? 

Also, HubConfig and Versions can be passed to LoadTestModules by TestConfig, which has access to every Spring-managed bean.

srinathgit: yeah, I notice that  validateMapping.sjs test is failing. So in order for HubConfig and Versions to be passed to LoadTestModules , should I include anything ?

rjrudin: I submitted a PR to the branch to fix this

PR #3029, opened 2019-10-03
DHFPROD-3010: Fixing the bug with functionslist and Field list

xnikhil08:
Fixed the bugs in functions list icon and field list icon


xnikhil08:
> The mapping is not saved once you select a input field or add an expression

@srinathgit - This is how it was needed that mapping should be saved once you edit the expression after adding the required fields or function. I.e User will have to edit the function after adding it anyway to provide required parameters.

xnikhil08:
@srinathgit - Please verify it now and merge the PR if everything looks good.

srinathgit:
@xnikhil08 ,
I still see the issue, the behavior is inconsistent when I select a input field or type a value myself. Sometimes step gets saved , sometimes it doesn't.

PR #3031, opened 2019-10-03
Fix failing test on feature/DHFPROD-3010 and other minor issues

srinathgit:
Fix failing test on feature/DHFPROD-3010 and other minor issues


marklogic-data-hub/src/main/java/com/marklogic/hub/step/impl/MappingStepDefinitionImpl.java
```diff
@@ -16,15 +16,21 @@
...
+import com.marklogic.hub.util.BeanUtil;
 
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 
 public class MappingStepDefinitionImpl extends AbstractStepDefinition {
 
+    @JsonIgnore
+    Versions versions = BeanUtil.getBean(Versions.class);
```
rjrudin: Oof. This is rough, but I understand the need for it. I definitely recommend not declaring this as a class attribute - there's no real performance again in that, as getBean is just hitting an in-memory hash map. Then you don't need JsonIgnore.

marklogic-data-hub/src/main/java/com/marklogic/hub/step/impl/MappingStepDefinitionImpl.java
```diff
@@ -40,7 +46,11 @@ public MappingStepDefinitionImpl(String name) {
 
         options.put("sourceDatabase", HubConfig.DEFAULT_STAGING_NAME);
         options.put("targetDatabase", HubConfig.DEFAULT_FINAL_NAME);
-
-        setModulePath("/data-hub/5/builtins/steps/mapping/default/main.sjs");
+        if(versions.isVersionCompatibleWithES()){
```
rjrudin: I recommend changing this to do the following:

1. See if BeanUtil has an instance of Versions. Don't throw an error if it doesn't, as there might not be a Spring container in the context that this class is being used (such as a unit test).
1. If it can't get an instance of Versions, use the ES path.
1. Otherwise, use Versions to choose the path. 

I also recommend adding tests on this logic.

marklogic-data-hub/src/main/java/com/marklogic/hub/util/BeanUtil.java
```diff
@@ -0,0 +1,19 @@
+package com.marklogic.hub.util;
+
+
+import org.springframework.beans.BeansException;
+import org.springframework.context.ApplicationContext;
+import org.springframework.context.ApplicationContextAware;
+import org.springframework.stereotype.Service;
+
+@Service
+public class BeanUtil implements ApplicationContextAware {
```
rjrudin: While Spring's method is called "getBean", I think a better name for this is "ApplicationContextReference". I'd also just put a "getApplicationContext" method on this so a client can do what it needs with the Spring context. "BeanUtil" sounds like a class that performs operations on JavaBeans.

marklogic-data-hub/src/main/java/com/marklogic/hub/step/impl/MappingStepDefinitionImpl.java
srinathgit: Fixed in latest commit 69488d8bda278aa2ab8b428bab6f587ad18f8b89

marklogic-data-hub/src/main/java/com/marklogic/hub/step/impl/MappingStepDefinitionImpl.java
srinathgit: Fixed in 69488d8bda278aa2ab8b428bab6f587ad18f8b89

marklogic-data-hub/src/main/java/com/marklogic/hub/util/BeanUtil.java
srinathgit: Fixed in 69488d8bda278aa2ab8b428bab6f587ad18f8b89

PR #3036, opened 2019-10-07
DHFPROD-2744: separate match step

ryanjdew:
Separate match step and appropriate refactors to separate match/merge logic.

### DIFF
--- a/marklogic-data-hub/src/main/resources/hub-internal-artifacts/step-definitions/mastering/marklogic/matching-mastering.step.json
+++ b/marklogic-data-hub/src/main/resources/hub-internal-artifacts/step-definitions/mastering/marklogic/matching-mastering.step.json
@@ -1,0 +1,20 @@
+{
+  "lang": "zxx",
+  "name": "matching-mastering",
+  "type": "mastering",
+  "version": 1,
+  "modulePath": "/data-hub/5/builtins/steps/mastering/default/matching.sjs",
+  "retryLimit" : 0,
+  "batchSize" : 100,
+  "threadCount" : 4,
+  "options": {
+    "acceptsBatch": true,
+    "stepUpdate": false,
+    "sourceDatabase": "data-hub-FINAL",
+    "targetDatabase": "data-hub-FINAL",
+    "sourceQuery": "cts.collectionQuery('default-mapping')",
+    "collections": [
+      "mastering-summary"
+    ]
+  }
+}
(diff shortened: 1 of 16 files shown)


marklogic-data-hub/src/main/resources/ml-modules/root/com.marklogic.smart-mastering/survivorship/merging/base.xqy
```diff
@@ -2317,12 +2317,15 @@ declare function merge-impl:is-uri-locked($uri as xs:string) as xs:boolean {
   $uri = merge-impl:locked-uris()
 };
 
+(: Don't want to trigger static analysis for our sepatate read-only operations :)
+declare variable $lock-for-update-fun := fn:function-lookup(xs:QName('xdmp:lock-for-update'), 1);
```
rjrudin: Interesting - so without this, there's a chance that the matching code may hit this function and thus ML will make it an update function?

I'm wondering if the entrance to the matching function should fail if xdmp:request-timestamp() doesn't return a timestamp so that it fails fast and there's no chance of it running in update mode?

(Also, tiny thing - "sepatate" should be "separate")

marklogic-data-hub/src/main/resources/hub-internal-artifacts/step-definitions/mastering/marklogic/matching-mastering.step.json
```diff
@@ -0,0 +1,20 @@
+{
+  "lang": "zxx",
+  "name": "matching-mastering",
+  "type": "mastering",
```
srinathgit: We have the format "default-*" for all the OOTB step definitions . Can the step def names be "default-matching" and "default-merging" ?

ryanjdew: Changed to reflect the "default-*" convention with the latest commit.

marklogic-data-hub/src/main/resources/ml-modules/root/com.marklogic.smart-mastering/survivorship/merging/base.xqy
ryanjdew: Fixed typo and we call lock-for-update early on with our mastering step already.

ryanjdew:
Closing this in favor of the PR with the additional merging changes and all tests passing.  https://github.com/marklogic/marklogic-data-hub/pull/3038

PR #3037, opened 2019-10-07
DHFPROD-3146: Function metadata now generated when loading modules

rjrudin:
This includes during mlWatch, mlLoadModules, mlReloadModules, mlDeploy, and mlRedeploy.


marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/LoadUserModulesCommand.java
```diff
@@ -176,7 +176,6 @@ public void execute(CommandContext context) {
         Path userModulesPath = hubConfig.getHubPluginsDir();
         String baseDir = userModulesPath.normalize().toAbsolutePath().toString();
         Path startPath = userModulesPath.resolve("entities");
-        Path mappingPath = userModulesPath.resolve("mappings");
```
rjrudin: Just did a little cleanup in this class, no functional changes.

PR #3038, opened 2019-10-08
DHFPROD-2923: Separate merge step (includes match step changes)

ryanjdew:


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mastering/default/matching.sjs
```diff
@@ -0,0 +1,47 @@
...
+function main(content, options) {
+  masteringMainStep.checkOptions(null, options, null, requiredOptionProperties);
+  let matchOptions = new NodeBuilder().addNode({ options: options.matchOptions }).toNode();
+  let matchSummaryJson = mastering.buildMatchSummary(
+    content,
+    matchOptions,
+    options.filterQuery ? cts.query(options.filterQuery) : cts.trueQuery(),
+    datahub.prov.granularityLevel() === datahub.prov.FINE_LEVEL
+  );
+  let collections = ['datahubMasteringMatchSummary'];
```
rjrudin: Just a naming question - should this be "dataHubMasteringMatchSummary"? Or possibly "data-hub-mastering-match-summary"? I had asked in slack yesterday about "datahub" vs "data-hub/dataHub", and it seems like the latter is preferred? Granted though, we have hundreds of references to "datahub" already in the codebase - e.g. datahub.sjs. - so this ship may have already sailed.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mastering/default/matching.sjs
```diff
@@ -0,0 +1,47 @@
...
+  if (options.targetEntity) {
+    collections.push(`datahubMasteringMatchSummary-${options.targetEntity}`);
+  }
+  return {
+    uri: `/datahub/5/mastering/match-summary/${sem.uuidString()}.json`,
```
rjrudin: Same question here - in the DHF modules, we do "/data-hub/5/". But we also have "/data-hub/5/datahub-singleton.sjs". 

I think the opportunity for consistency here is gone, unless we invest heavily in a refactor, which seems unlikely. Though at least in this instance, "/data-hub/5/" seems easy to do and is consistent with the modules, FWTW.

marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/DeployDatabaseFieldCommand.java
```diff
@@ -80,7 +80,7 @@ protected String addExistingFieldsAndRangeFieldIndexes(String payload, DatabaseM
 
         addExistingFields(newProps, existingProps);
         addExistingRangeFieldIndexes(newProps, existingProps);
-
+        addExistingRangePathIndexes(newProps, existingProps);
```
rjrudin: Good catch - there's an existing unit test that you can easily add setup and assertions to as well for this.

ryanjdew: Updated the unit test to check the path range indexes as well.

PR #3043, opened 2019-10-08
DHFPROD:- 3021 Support mapping expressions for namespaces and nested XML and JSON files in the UI

xnikhil08:


rjrudin:
It seems like this has changes in it that aren't just for DHFPROD-3021 - i.e. the Versions check, which I thought had been in a separate PR? Would be easier to review this if it's slimmed down to just new stuff for 3021.

xnikhil08:
> @xnikhil08 ,
> This needs a rebase with develop

@srinathgit - Rebased with Develop.

PR #3050, opened 2019-10-10
DHFPROD-3159: No longer overwriting ml-config directory during upgrade

rjrudin:
Instead, upgradeHub only backs up the hub-internal-config directory. It then checks to see if range-element-index has an empty array in final-database.json, and if so, it's removed without overwriting the file. 

Also removed the support for replacing the DHF version in build.gradle via regex, as our instructions for an upgrade tell the user to do this themselves first. 

Also did a little renaming while I was in DataHubImpl so that "install into DHS" is now "deploy to DHS".

### DIFF
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/DataHub.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/DataHub.java
@@ -97,7 +97,7 @@ public interface DataHub {
      *
      * @param listener
      */
-    void dhsInstall(HubDeployStatusListener listener);
+    void deployToDhs(HubDeployStatusListener listener);
 
     /**
      * Updates the indexes in the database based on the project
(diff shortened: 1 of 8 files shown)


marklogic-data-hub/src/main/java/com/marklogic/hub/DataHub.java
```diff
@@ -178,16 +178,6 @@
...
 
-    /**
-     * Upgrades the hub based on list of provided updated flows. All flows SHOULD be provided.
-     * The method without params will handle this automatically.
-     * Must be run as a user with sufficient privileges to install a data hub.
-     * @param updatedFlows - the list of the name of the flows you want to update
-     * @return boolean - false if upgrade fails for a reason other than an upgrade exception
-     * @throws CantUpgradeException - should the hub fail to upgrade for incompatibility reasons
-     */
-    boolean upgradeHub(List<String> updatedFlows) throws CantUpgradeException;
```
rjrudin: I removed this, as updatedFlows was never used. I know this is part of the public interface for DataHub, but I can't imagine anyone was calling this directly since updatedFlows was ignored.

PR #3051, opened 2019-10-10
DHFPROD-3288: Limiting facets to 25 to improve Explorer performance 

rahulvudutala:
This PR has 2 commits addressing the following issues
* Limiting facets to 25 to improve Explorer performance in case of a large number of facets
* Adding default query options file for the explorer

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules-final/options/exp-default.xml
+++ b/marklogic-data-hub/src/main/resources/ml-modules-final/options/exp-default.xml
@@ -3,18 +3,13 @@
     <collection/>
   </constraint>
   <constraint name="createdByJob">
-    <value>
+    <range facet="false">
       <field name="datahubCreatedByJob"/>
-    </value>
+    </range>
   </constraint>
   <constraint name="createdByStep">
-    <value>
-      <field name="datahubCreatedByStep"/>
-    </value>
-  </constraint>
-  <constraint name="createdByJobRange">
     <range>
-      <field name="datahubCreatedByJob"/>
+      <field name="datahubCreatedByStep"/>
       <facet-option>limit=25</facet-option>
     </range>
   </constraint>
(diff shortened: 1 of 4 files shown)


marklogic-data-hub/src/main/resources/ml-modules-final/options/exp-default.xml
```diff
@@ -0,0 +1,43 @@
...
+      <field name="datahubCreatedByJob"/>
+    </value>
+  </constraint>
+  <constraint name="createdByStep">
+    <value>
+      <field name="datahubCreatedByStep"/>
+    </value>
+  </constraint>
+  <constraint name="createdByJobRange">
+    <range>
```
rjrudin: Note that range defaults to facet=true. Unless you need a facet on job ID - seems unlikely? - you should add facet=false.

rahulvudutala: @rjrudin This is a requirement for the explorer to enable faceting on JobId's. For datahubCreatedOn we are using a date-picker. So I am setting the facet="false" for the createdOn range constraint

rahulvudutala:
@rjrudin Regarding explorer search options in DHF, We need the search constraints that are generated by entity-services on the entity property indexes when a new entity is added or when a new index is added on the entity property. So we need  the exp-final-entity-options.xml to be deployed from datahub.
 
As for the explorer default-options file, we can use the RawCombinedQueryBuilder to pass the options. But since explorer doesn’t know the version of the datahub, we will be searching on the fields or indexes that doesn’t exist with pre 5.1 versions of datahub. (For example, if a user is having 4.x datahub and trying to use explorer, and we query using the default query options using RawCombinedQueryBuilder, we get error saying field doesn’t exist or index doesn’t exist.)
 
The reason for adding the default explorer options file is to make sure that these options files are used starting and also when a user upgrades to 5.1, the user is not blocked from using explorer because of missing options file.

marklogic-data-hub/src/main/resources/ml-modules-final/options/exp-default.xml
rahulvudutala: @rjrudin Removed faceting on createByJobRange but retained the constraint and created a new commit

PR #3052, opened 2019-10-10
smooth-mastering example flow

crism:
As suggested, a PR with the README for a proposed example flow.


crism:
By request, a more detailed explanation.

My current customer has a large amount of information from disparate sources ingested nightly. There is little or no change control in the original sources. The ingestion process relies on the preceding ingestion being present, so a hash calculation can determine whether the new document is different or not, and update the timestamp in the envelope accordingly.

Also, given the amount of data being ingested, reprocessing it all every night would be extremely time-consuming, unnecessarily, as most of it has not changed from one day to the next.

By default, a DHF mapping step destroys the original, putting the mapping result at the URI of the original. That would disrupt the change detection process and result in repetitive unnecessary processing. (In addition, when XML is mapped to JSON, the result URI still has `.xml` at the end, which leads to user confusion.)

Our solution is to create non-destructive, idempotent mapping and mastering stages.

The `translation-mapping` step uses the default mapping behavior, but puts the result at a predictable URI based on primary keys in the output document.

The `smooth-mastering` step uses the default mastering behavior, but again puts the result at a predictable URI, squelches attempts to mark source documents as archived, and ensures that singletons are copied forward to the next stage even if not merged with any other documents.

The example flow planned for this will also demonstrate using time-windowed `sourceQuery` parameters to avoid redundant processing.

aebadirad:
Let us know when this is ready to review :)

aebadirad:
@crism ack, can we get a rebase on this develop so it can go out with the next release?

crism:
Rebased, @aebadirad. Core code is good to go. Still trying to create a clean example of how to use it.

crism:
This is mostly ready to go, @aebadirad, though I get conflicting update errors on the merge step. I’m not sure why; something is different from 5.0.2 behavior, but I haven’t narrowed it down.

crism:
This is ready to go modulo the requested changes on #3088, @aebadirad.

aebadirad:
@crism are we good here?

crism:
@aebadirad, yes, freshly rebased, and includes the #3087 fix needed to make it work. May want to wait for #3088 to be accepted, but that’s between you and @ryanjdew. :smiley:

crism:
Freshly rebased once again, @aebadirad, now that #3088 has been accepted.

aebadirad:
@rjrudin @ryanjdew can you guys peek at this?

rjrudin:
@crism @sbayatpur Because 5.3.0 will include step processors, which will allow for altering the URI before any document is persisted, I am going to close this. I think it serves as a useful example, but it's doing what step processors are intended to avoid - forcing a develop to duplicate an existing DHF step and then modify it.

PR #3056, opened 2019-10-11
DHFPROD-3025 Test, Clear button in mapping UI for nested entities

srinathgit:
This PR in addition to DHFPROD-3025 , also includes DHFPROD-3286 which is adding a REST endpoint for mapping validation included in commit https://github.com/marklogic/marklogic-data-hub/commit/80799856dc22ee9b1fa3466a094d5eecf0ace7cc


marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/builtins/steps/mapping/entity-services/suite-setup.xqy
```diff
@@ -11,4 +11,5 @@ xquery version "1.0-ml";
 import module namespace hub-test = "http://marklogic.com/data-hub/test" at "/test/data-hub-test-helper.xqy";
 import module namespace test = "http://marklogic.com/test" at "/test/test-helper.xqy";
 
-hub-test:load-non-entities($test:__CALLER_FILE__)
+hub-test:load-non-entities($test:__CALLER_FILE__),
+xdmp:document-insert("/mapTest.json",xdmp:unquote('{"id":1,"transactionDate":"12/7/2019","customer":{"firstName":"Oralia","lastName":"Dinesen","gender":"F"},"items":[{"name":"Voltsillam","price":2.0,"quantity":7},{"name":"Latlux","price":7.17,"quantity":10},{"name":"Biodex","price":5.01,"quantity":2},{"name":"Fixflex","price":8.77,"quantity":6},{"name":"Keylex","price":5.57,"quantity":3}]}'))
```
rjrudin: You can put mapTest.json under "test-data/content" and this JSON will be loaded for you.

marklogic-data-hub/src/main/java/com/marklogic/hub/mapping/TestMapping.java
```diff
@@ -0,0 +1,28 @@
+package com.marklogic.hub.mapping;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.marklogic.client.DatabaseClient;
+import com.marklogic.client.extensions.ResourceManager;
+import com.marklogic.client.io.JacksonHandle;
+import com.marklogic.client.util.RequestParameters;
+
+public class TestMapping extends ResourceManager {
```
rjrudin: Let's use data services here instead (I was just looking at MappingValidator and thinking how easy that would be to switch over). 

There's now a data service on the develop branch. See around line 160 for the Gradle task to generate a Java interface based on an API file. 

I recommend generating a "com.marklogic.hub.dataservices.MappingService" and calling this function "testMapping". We can add validateMapping at a later date. 

Let me know if you have any questions, I can help out and/or submit a PR into this to do the work.

rjrudin: I'm actually going to submit a PR for the data services change, as I'd like to fold MappingValidator into it as well. Should be ready shortly.

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/builtins/steps/mapping/entity-services/suite-setup.xqy
srinathgit: Added mapTest.json under "test-data/content"

PR #3059, opened 2019-10-11
DHFPROD-3310: Adding "date" as an accepted datatype

bsrikan:
This PR also updates unit-test data for validate-entity to have test coverage for the code change


bsrikan:
Included a small update to entity and mapping artifact to test for null check of parseDate - DHFPROD-3121

PR #3062, opened 2019-10-14
DHFPROD-3019: Remove "fn:" from display name as well

srinathgit:
Remove "fn:" from display name as well

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/services/mlMappingFunctions.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/services/mlMappingFunctions.sjs
@@ -59,7 +59,7 @@ function getXpathFunctions() {
       let signature = xdmp.functionSignature(xpathFunctions[i]).replace("function", xdmp.functionName(xpathFunctions[i]));
       signature = signature.match(/fn:(.*?) as.*?/)[1];
       xpathFunction["signature"] = signature;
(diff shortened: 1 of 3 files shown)


marklogic-data-hub/src/main/resources/ml-modules/services/mlMappingFunctions.sjs
```diff
@@ -59,7 +59,7 @@ function getXpathFunctions() {
       let signature = xdmp.functionSignature(xpathFunctions[i]).replace("function", xdmp.functionName(xpathFunctions[i]));
       signature = signature.match(/fn:(.*?) as.*?/)[1];
       xpathFunction["signature"] = signature;
-      response[xdmp.functionName(xpathFunctions[i])] = xpathFunction;
+      response[String(xdmp.functionName(xpathFunctions[i])).replace("fn:","")] = xpathFunction;
```
rjrudin: Can you update the getMappingFunctions.sjs test to do a verification on this - I'm thinking it would suffice to verify that "sum" exists  but not "fn:sum", with an assertion message stating why "sum" is there but not "fn:sum"

srinathgit: Fixed in latest commit

srinathgit:
@rjrudin ,
In addition to the change you suggested, I have also removed redundant code

PR #3066, opened 2019-10-14
DHFPROD-3275: Update management of Merge Collections

wooldridge:
This is a Smart Mastering feature under Merging tab in UI:

- Include default collections in Merge Collections table.
- Retrieve defaults from new backend endpoint, see DHFPROD-3303
- Allow user to add additional collections for each event.

### DIFF
--- a/web/src/main/ui/app/components/flows-new/services/manage-flows.service.ts
+++ b/web/src/main/ui/app/components/flows-new/services/manage-flows.service.ts
@@ -100,4 +100,10 @@ ManageFlowsService {
     console.log('GET api/current-project/mappings/functions');
     return this.http.get('api/current-project/mappings/functions');
   }
+
+  getDefaultCollections(entityName) {
+    console.log('GET api/mastering/defaultCollections/' + entityName);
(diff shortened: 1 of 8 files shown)


web/src/main/ui/app/components/flows-new/services/manage-flows.service.ts
```diff
@@ -100,4 +100,10 @@ ManageFlowsService {
     console.log('GET api/current-project/mappings/functions');
     return this.http.get('api/current-project/mappings/functions');
   }
+
+  getDefaultCollections(entityName) {
```
rjrudin: Is manage-flows.service kind of a broad facade across DHF? Just curious, as I don't think of "default merge collections" as belonging to a "manage flows" service, but if it's intended to be a pretty broad facade, that's fine.

wooldridge: Its surface area is broad right now. That's where all the services supporting step-based flows live. I agree that it should subdivided at some point.

PR #3088, opened 2019-10-17
Allow matching on non-string values.

### LINKED ISSUE #3087: Smart Mastering: match property values are all cast to strings
#### The issue

Short description of the problem: When matching on integer-valued properties, no match is ever found.

What behavior are you expecting? Matches should work on integers and other non-string types as well as on strings.

#### Tech details

Which Operating System are you using? CentOS 7.

Which version of MarkLogic are you using? 10.0-2

Which version of the Data Hub Framework are you using? 5.1-SNAPSHOT (from the `develop` branch)

Is this a QuickStart UI bug? If so, which browser are you using? N/A

It looks like this arose with commit 9959a55; before that, the compiled query checked the castability of the value being compared. (Although I note this change is present in DHF 5.0.2, which seems to be working.) This could probably be corrected by putting the castability tests back into `match-impl:query-map-to-query()`.

crism:
Fixes #3087 by exploding values if castable to other types.

(This is based on previous behavior, and does not handle all types. It could probably be expanded to `date`, `dateTime`, etc.)


crism:
> I think it would be better to remove the original cast to string (but still filter out nodes empty string values). This would reduce the amount of processing done and encourage consistent modeling, as there shouldn't be mixed types for properties in a document.

That’s reasonable. Let me give that a shot.

marklogic-data-hub/src/main/resources/ml-modules/root/com.marklogic.smart-mastering/matcher-impl/matcher-impl.xqy
```diff
@@ -427,10 +427,12 @@ declare function match-impl:values-by-qname(
 {
   map:new(
     for $qname in fn:distinct-values((map:get($compiled-options, "queries") ! map:get(.,"qname")))
-    let $values := fn:distinct-values($document//*[fn:node-name(.) eq $qname] ! fn:normalize-space(fn:string(.))[.])
+    let $values := fn:distinct-values(
+      $document//*[fn:node-name(.) eq $qname][fn:normalize-space(fn:string(.))]/fn:data(.)
```
rjrudin: Is the addition of "fn:data" the key change here? 

@ryanjdew  I'm good with this if you are.

crism:
This is freshly rebased but no new changes. Should be good to merge any time.

marklogic-data-hub/src/main/resources/ml-modules/root/com.marklogic.smart-mastering/matcher-impl/matcher-impl.xqy
crism: Yes. We now filter on blank values, but ignore the string cast and get the raw data if it passes the test.

There is a possible change in behavior here: a string-valued node with extra whitespace would have been trimmed, previously (_e.g._ `  this   has spaces ` ⇒ `this has spaces`), and now won’t be.

crism: Note that only the value to search for is changed—the targets are not changed in any way. Under the old behavior, a document might not even match itself. I would argue this change is for the better; data sanitization should be intentional and explicit, not automagical.

ryanjdew: I'm good with this change.

aebadirad:
Conflict, need to get that resolved. Can someone address it? I think we probably want the newer change right?

ryanjdew:
Freshly rebased and squashed the commits. I added a second commit to fix one ml-unit-test that wasn't accounting for differing data types in the query by values. Will wait for CI unit test to complete before reviewing.

PR #3092, opened 2019-10-18
DHFPROD-3023: Test button checks for errors in UI mapping

xnikhil08:
Contains changes for the UI portion of story DHFPROD-3023

### DIFF
--- a/web/src/main/ui/app/components/flows-new/edit-flow/mapping/mapping.component.ts
+++ b/web/src/main/ui/app/components/flows-new/edit-flow/mapping/mapping.component.ts
@@ -29,6 +29,7 @@ import { Flow } from "../../models/flow.model";
       [entityName]="this.entityName"
       [entityNested] = "entityNested"
       [nmspace] = "nmspace"
+      [mapValidationResult]="this.mapValidationResult"
       (updateURI)="this.updateURI($event)"
       (updateMap)="this.updateMap($event)"
     ></app-mapping-ui>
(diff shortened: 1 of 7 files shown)


web/src/main/ui/app/components/flows-new/edit-flow/mapping/mapping.component.ts
```diff
@@ -77,6 +78,48 @@ export class MappingComponent implements OnInit {
...
+            "sourcedFrom": "Shrivastava",
+            "errorMessage": "Invalid XPath expression: lastname"
+          },
+          "gender": {
+            "sourcedFrom": "Male"
+          }
+        }
+      }
+    }
+  };
```
srinathgit: Mapping validation works for nested entities as well . The code was merged to develop. This hardcoded response  can be removed and you can check if it works for nested entities .

xnikhil08: @srinathgit - Please review now. That code is removed.

wooldridge:
There will most likely be tweaks required to mapping tests, but let's merge this so we can get what's current in mapping into develop.

ryanjdew:
Looks like there is an error in the UI build.

`ERROR in src/main/ui/app/components/flows-new/edit-flow/mapping/mapping.component.ts.MappingComponent.html(15,7): : Property 'mapValidationResult' does not exist on type 'MappingComponent'.`

wooldridge:
@xnikhil08 can you fix the build issue and push to this branch?

xnikhil08:
> @xnikhil08 can you fix the build issue and push to this branch?

Fixed it.

PR #3093, opened 2019-10-18
DHFPROD-3313:Keep xml/xslt when reinstallUserModules() is called

srinathgit:
Keep xml/xslt when reinstallUserModules() is called

### DIFF
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/impl/DataHubImpl.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/impl/DataHubImpl.java
@@ -376,7 +376,8 @@ public class DataHubImpl implements DataHub {
             );
 
             String query =
-                "cts:uris((),(),cts:not-query(cts:collection-query('hub-core-module')))[\n" +
+                "cts:uris((),(),cts:not-query(cts:collection-query(('hub-core-module','http://marklogic.com/entity-services/mapping'" +
+                    ",'http://marklogic.com/entity-services/function-metadata'))))[\n" +
                     "  fn:not(\n" +
                     "    fn:matches(., \"^.+options/(" + String.join("|", options) + ").xml$\") or\n" +
                     "    fn:starts-with(., \"/marklogic.rest.\")\n" +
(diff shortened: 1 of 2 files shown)


web/src/main/java/com/marklogic/hub/web/service/DataHubService.java
```diff
@@ -190,7 +190,6 @@ private void installUserModules(HubConfig hubConfig, boolean forceLoad, DeployUs
         loadHubArtifactsCommand.setHubConfig(hubConfig);
         loadHubArtifactsCommand.setForceLoad(forceLoad);
 
-        commands.add(generateFunctionMetadataCommand);
```
rjrudin: Isn't this still needed for when e.g. a user's project is first installed?

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/DataHubImpl.java
```diff
@@ -376,7 +376,8 @@ public void clearUserModules() {
             );
 
             String query =
-                "cts:uris((),(),cts:not-query(cts:collection-query('hub-core-module')))[\n" +
+                "cts:uris((),(),cts:not-query(cts:collection-query(('hub-core-module','http://marklogic.com/entity-services/mapping'" +
```
rjrudin: This doesn't really seem safe... I think if a user runs "mlClearModulesDatabase", the user would expect the xml/xslt files generated for their own custom functions to be deleted as well. 

This seems to be addressing a timing issue in QS, right? Where between the time it takes for QS to save the mapping and then reinstall modules, the user may choose to run their mapping? 

In which case - isn't their always a chance that the user's custom mapping functions are deleted and not yet reloaded by the time the user chooses to run the mapping?

I'll play around with this locally. My hunch is that in the QS context, we don't need to clear the modules database in the first place. Clearing modules makes sense when you e.g. start your day and pull from git and you want to ensure any deletions in git are made in your modules database as well. I don't think it's necessary to do while in QS though.

rjrudin:
Take a look at https://github.com/marklogic/marklogic-data-hub/pull/3094 - I think the better fix is to not call "reinstallUserModules" when saving entities/flows/mappings but rather to just call "installUserModules".

I am thinking the call to "reinstall" is at the heart of these intermittent problems people see with modules/mappings disappearing.

srinathgit:
Closing this as #3094 has fix for it.

PR #3094, opened 2019-10-18
DHFPROD-3313: Don't clear modules unless needed in QS

rjrudin:
When saving an artifact - e.g. an entity, flow, or mapping - I don't see any reason to clear user modules first. The dataHub.clearUserModules method isn't getting rid of old artifacts from the staging/final databases. 

Clearing user modules just runs the risk of bugs like this due to timing conditions where a user may then run a flow before the user modules have been reloaded.

### DIFF
--- a/web/src/main/java/com/marklogic/hub/web/service/DataHubService.java
+++ b/web/src/main/java/com/marklogic/hub/web/service/DataHubService.java
@@ -102,11 +102,13 @@ public class DataHubService {
 
     @Async
     public void installUserModules(HubConfig config, boolean forceLoad, DeployUserModulesListener deployListener, ValidateListener validateListener) {
+        logger.info("Installing user modules");
         long startTime = PerformanceLogger.monitorTimeInsideMethod();
-
         try {
             installUserModules(config, forceLoad, deployListener);
-            validateUserModules(config, validateListener);
+            if (validateListener != null) {
+                validateUserModules(config, validateListener);
+            }
         } catch (Throwable e) {
             throw new DataHubException(e.getMessage(), e);
         }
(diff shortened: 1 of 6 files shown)


web/src/main/java/com/marklogic/hub/web/service/DataHubService.java
```diff
@@ -187,6 +184,7 @@ private void installUserModules(HubConfig hubConfig, boolean forceLoad, DeployUs
         loadUserArtifactsCommand.setHubConfig(hubConfig);
         loadUserArtifactsCommand.setForceLoad(forceLoad);
 
+        // TODO Why load hub artifacts when this method is for loading user modules/artifacts?
```
srinathgit: There is an option to clear all dbs in the QS dashboard. If someone clicks that, the default flows and step definitions will be deleted as well. This is meant to reload them

web/src/main/java/com/marklogic/hub/web/service/EntityManagerService.java
```diff
@@ -134,9 +134,7 @@ public EntityModel saveEntity(EntityModel entity) throws IOException {
         else {
             HubEntity renamedEntity = em.saveEntity(hubEntity, true);
             entity.setFilename(renamedEntity.getFilename());
-
-            // Redeploy the flows
-            dataHubService.reinstallUserModules(hubConfig, null, null);
+            dataHubService.installUserModules(hubConfig, true, null, null);
```
srinathgit: There is a FileWatcherService which looks into changes in entities, mappings and flows dir and does installUserModules (with 'forceLoad' set to false). Do we even have to call installUserModules() explicitly here and also when saving mapping and flows ?

PR #3098, opened 2019-10-18
DHFPROD-3245:Add 'urisOnly' option in ml:Collections and use it in UI

srinathgit:
DHFPROD-3245:Add 'urisOnly' option in ml:Collections and use it in UI

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/services/mlCollections.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/services/mlCollections.sjs
@@ -21,6 +21,7 @@ function get(context, params) {
   let resp;
   let database = params.database ? params.database.trim() : null;
   let sourceQuery = params.sourceQuery ? params.sourceQuery.trim() : null;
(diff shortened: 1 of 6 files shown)


marklogic-data-hub/src/main/resources/ml-modules/services/mlCollections.sjs
```diff
@@ -31,10 +32,19 @@ function get(context, params) {
       docs = docs.toArray();
       let i;
       for (i in docs) {
-        let obj = {
-          "uri": xdmp.nodeUri(docs[i]),
-          "docs": docs[i]
-        };
+        let obj;
+        if(urisOnly === 'true') {
```
aebadirad: if(urisOnly) {

srinathgit: From SearchService.Collections , 'urisOnly' is passed as a string. So, if urisOnly is set to "false" (or any other string),if(urisOnly) will be true, hence this check if(urisOnly === 'true')

PR #3121, opened 2019-10-22
DHFPROD-3023:(and DHFPROD-3421) Load mapping when it changes, popup only if mapping nor present

srinathgit:
This PR solves 2 issues

DHFPROD-3023: Loads only mapping when it changes, don't force load artifacts when not required, introduce a delay of 750 ms after "Test" button is hit
In my mac, if mapping changes, it gets loaded in less than < 0.2 secs usually

DHFPROD-3421 - Show popup only if error shows that mapping is not present

### DIFF
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/LoadUserArtifactsCommand.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/LoadUserArtifactsCommand.java
@@ -18,7 +18,6 @@ package com.marklogic.hub.deploy.commands;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.node.ObjectNode;
-import com.marklogic.appdeployer.AppConfig;
 import com.marklogic.appdeployer.command.AbstractCommand;
 import com.marklogic.appdeployer.command.CommandContext;
 import com.marklogic.appdeployer.command.SortOrderConstants;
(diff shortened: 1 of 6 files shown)


web/src/main/java/com/marklogic/hub/web/service/EntityManagerService.java
```diff
@@ -134,10 +134,7 @@ public EntityModel saveEntity(EntityModel entity) throws IOException {
         else {
             HubEntity renamedEntity = em.saveEntity(hubEntity, true);
             entity.setFilename(renamedEntity.getFilename());
-            dataHubService.installUserModules(hubConfig, true, null, null);
```
rjrudin: Isn't this still needed for loading the entity model file into staging/final? Granted, it should only load that as opposed to loading all user modules.

web/src/main/ui/app/components/flows-new/edit-flow/mapping/ui/mapping-ui.component.ts
```diff
@@ -145,58 +145,58 @@ export class MappingUiComponent implements OnChanges {
...
+                delete self.mapResults["info"];
+                self.mapResults = self.mapResults[self.entityName];
+                console.log(self.mapResults);
+              },
+                err => {
+                  if (err.hasOwnProperty('error') && String(err['error']['message']).indexOf("Could not find mapping") >= 0) {
+                    console.log("found error");
+                    if (err['error']['code'] == 500) {
+                      let result = self.dialogService.alert(
+                        'Please try again in a few seconds. Thanks for your patience.',
```
rjrudin: I recommend a more specific message here so the user isn't left wondering what's going on - perhaps:

"QuickStart hasn't finished loading the updated mapping. Please try again."

Though if we detect this error, couldn't QS wait a few times itself - or is that what the 750ms delay below is for?

marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/LoadUserArtifactsCommand.java
```diff
@@ -101,7 +98,7 @@ private PropertiesModuleManager getModulesManager() {
                 }
             }
         }
-
+        pmm.initialize();
```
rjrudin: This is a critical change - can you add a comment here to explain why this is being done, which I believe is so that in the QuickStart context, the command only loads what needs to be loaded when the user makes a change?

web/src/main/ui/app/components/flows-new/edit-flow/mapping/ui/mapping-ui.component.ts
srinathgit: @rjrudin ,
1. Whenever someone hits "Test" button, I add a delay for 750 ms (regardless of whether the mapping has changed or not, the assumption is more often than not the user will hit the button after changing the mapping) and then make the request. I don't retry in case if mapping is not found. Is that logic ok ?
2. I will reword the message in the pop up

PR #3126, opened 2019-10-22
DHFPROD-3326: Explorer roles

akshaysonvane:


marklogic-data-hub/src/main/resources/hub-internal-config/security/roles/explorer-user.json
```diff
@@ -0,0 +1,7 @@
+{
+  "role-name": "explorer-user",
+  "description": "A role that can access entity models and instances.",
+  "role": [
+    "rest-reader", "entity-model-reader"
```
rjrudin: I emailed you and a few others about this role; I just don't think it should be created, because with ML 10, it's not likely that entity instances will have rest-reader/read on them any longer.

akshaysonvane: We have decided to remove the "explorer-user" role as in the case of DHS customers can use "flow-operator" or whatever derived role that they have used to run the flows and in case of on-prem they can always use their custom roles.

Changes addressed here: https://github.com/marklogic/marklogic-data-hub/pull/3126/commits/e973a7bf14be018c67ea78db85b77b5f38799b30

PR #3139, opened 2019-10-24
DHFPROD-3451: Mapping steps created in QS now include validateEntity

rjrudin:


web/src/main/java/com/marklogic/hub/web/service/FlowManagerService.java
```diff
@@ -259,13 +259,18 @@ public StepModel createStep(String flowName, Integer stepOrder, String stepId, S
         return transformStepToWebModel(step);
     }
 
+    protected Step mergeDefaultStepDefinitionIntoStep(StepModel stepModel, Step step) {
```
rjrudin: Extracted this to make it easy to write a unit test

PR #3146, opened 2019-10-24
DHFPROD-3347: XDMP-NOTXN could occur

ryanjdew:


ryanjdew:
@aebadirad If the transaction closes before make this call we don't care about it, since it isn't stalling mastering. What should we be doing with the exception?

marklogic-data-hub/src/main/resources/ml-modules/root/com.marklogic.smart-mastering/survivorship/merging/base.xqy
```diff
@@ -2383,7 +2383,10 @@ declare function merge-impl:locked-uris() {
  :)
 declare function merge-impl:locked-uris($host-id, $transaction-id) {
   fn:distinct-values(
-    xdmp:transaction-locks($host-id, $transaction-id)/(host:waiting|host:write)[fn:starts-with(., $lock-task-prefix)]/fn:substring-after(., $lock-task-prefix)
+    (: transaction may have closed between getting it from the host status and now looking for locks :)
+    try {
+      xdmp:transaction-locks($host-id, $transaction-id)/(host:waiting|host:write)[fn:starts-with(., $lock-task-prefix)]/fn:substring-after(., $lock-task-prefix)
+    } catch * {()}
```
rjrudin: This seems like an error we don't care about that much, so maybe log it at the debug level, or info-level at best? I could see "warn" as well I guess.

aebadirad:
Record a debug or info notification that the lock was not found with the tx id, host id, and lock name you were attempting to find.

While you're not caring if it throws an error, it could be indicative of another systemic problem if documents are not being mastered, yet transactions don't exist that would be expected to be resolved. Let's not swallow that, it could be useful for debugging cluster issues.

ryanjdew:
@aebadirad It isn't a systematic bug. It has to call xdmp:transaction-locks to determine if there is a deadlock by mastering and if the transaction completes before this call it is actually is a good thing because that means the transaction was able to completed and wasn't deadlocked. If it is deadlocked by another mastering transaction, we grab the URI and exclude it from the current transaction since it is already being handled.

ryanjdew:
Alao adjusted the jobReport check in MasterTest to use the actual mastered count

bsrikan:
Should the CR be dismissed? @ryanjdew can you pls do that if the change requested is not required. It is blocking the PR from getting merged.

PR #3152, opened 2019-10-27
DHFPROD-3023: Create per property mapping, run and display the result

srinathgit:
DHFPROD-3023:Create per property mapping, run and display the result

This PR removes the old implementation . Both validation and running of the mapping takes place in the same request. Since per property mappings are created, this implementation doesn't depend on the overall mapping xslt ( hence the a sleep of 750 ms which was there earlier has been removed as well)


web/src/main/ui/app/components/flows-new/edit-flow/mapping/ui/entity-table-ui.component.ts
```diff
@@ -61,24 +59,30 @@ export class EntityTableUiComponent implements OnChanges {
...
+    if(this.mapResp['properties'][field] && this.mapResp['properties'][field]['errorMessage']) {
+      return true;
     } else {
       return false;
     }
+  } else {
+    return false;
   }
+  
+}
```
bsrikan: Indentation is off

srinathgit: Fixed the indentation

srinathgit:
@bsrikan ,
boolean and all other data types should be visible now in the UI

rjrudin:
I just tried this branch out and a clean install of trunk_nightly, but now "Test" isn't showing any results in the GUI. The Firefox network inspector shows a response of the mapping, but without any values in it.

rjrudin:
I tried this out locally while testing a separate issue, and it seems to be working correctly. I only did a test of a flat entity though, I didn't try nested properties.

PR #3153, opened 2019-10-28
DHFPROD-3366: Make mapping compatible with local targetEntity

ryanjdew:


srinathgit:
@rjrudin , @ryanjdew 
Mapping validation still doesn't accept local target entity. You see "errorMessage" in the response  with  "Could not find a target entity: #/definitions/Name" if local target entity is present

PR #3157, opened 2019-10-28
DHFPROD-3409: Trim mapping context

wooldridge:
When inserting info from the prop menu in mapping, take into account the parent context and trim if needed.


SameeraPriyathamTadikonda:
@wooldridge There are conflicts when I try to rebase and merge. So can I  do squash and merge manually instead?

wooldridge:
@SameeraPriyathamTadikonda I'm fine with that. If you need me to help from me, let me know.

PR #3158, opened 2019-10-28
DHFPROD-3476: DHF installer now deploys the new 5.1.0 roles

rjrudin:


marklogic-data-hub/src/main/java/com/marklogic/hub/cli/command/InstallIntoDhsCommand.java
```diff
@@ -72,6 +74,12 @@ else if (c instanceof LoadUserModulesCommand) {
 
         commands.add(new CopyQueryOptionsCommand(hubConfig));
 
+        DeployRolesCommand deployRolesCommand = new DeployRolesCommand();
+        deployRolesCommand.setResourceFilenamesIncludePattern(
+            Pattern.compile("(entity-model-reader|explorer-architect).*")
+        );
```
aebadirad: This will mean users cannot deploy their own roles made of other roles permitted by DHS.

A whitelist strategy is not something I'm comfortable with.

rjrudin: This isn't for users, it's for the DHS team. Users use Gradle tasks - i.e. mlDeployRoles will continue to deploy all of the user's custom roles.

PR #3163, opened 2019-10-29
DHFPROD-3353: DHFPROD-3355:Retain map after source doc uris/type

srinathgit:
This PR fixes DHFPROD-3353, DHFPROD-3355 and retains mapped properties after source doc uris or source type is changed

### DIFF
--- a/web/src/main/ui/app/components/flows-new/edit-flow/mapping/mapping.component.ts
+++ b/web/src/main/ui/app/components/flows-new/edit-flow/mapping/mapping.component.ts
@@ -150,18 +150,30 @@ export class MappingComponent implements OnInit {
         this.sampleDocURI = map.sourceURI;
         this.editURIVal = this.sampleDocURI;
       }
-      this.loadSampleDoc()
+      this.loadSampleDoc();
       if (map && map.properties) {
+        this.isNestedMap();
+        console.log("isNested: " + this.connsNested);
         self.conns = {};
-        _.forEach(map.properties, function(srcObj, entityPropName) {
-          self.conns[entityPropName] = srcObj.sourcedFrom;
-        });
+        if(!self.connsNested ) {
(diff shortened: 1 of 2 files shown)


web/src/main/ui/app/components/flows-new/edit-flow/mapping/mapping.component.ts
```diff
@@ -148,19 +148,38 @@ export class MappingComponent implements OnInit {
...
         self.connsOrig = _.clone(self.conns);
       }
     },
     () => {},
     () => {});
   }
 
+  isNestedMap() {
+    let props = Object.keys(this.mapping.properties)
+    for(let i = 0; i < props.length ; i++){
```
rjrudin: Just a note, I think this can be simplified by using an array function, such as "find" or "findIndex":

    return this.mapping.properties.findIndex(prop => prop.hasOwnProperty('targetEntityType') > -1;

srinathgit: @rjrudin ,
'findIndex' is available for Arrays only. When I use it, I get  this.mapping.properties.findIndex is not a function

rjrudin: Oh sorry - I mistyped that, it would be Object.keys(this.mapping.properties).findIndex(key => this.mapping.properties[key].hasOwnProperty("targetEntityType"). Or something like that - just pointing out that any time you write a for loop with an index counter, there's likely an array function that gets it done in one line.

srinathgit: Fixed in latest commit

srinathgit:
@wooldridge ,
It should be fixed now

PR #3167, opened 2019-10-29
DHFPROD-3509: Missing items in QS mapping source data table in case of array of xml/json elements

xnikhil08:
1. Fix for the bug DHFPROD-3509.
2. Fixed the issue of Type not showing as Array and Object in source table for a JSON doc, if both XML and JSON Docs exists in same collection.

### DIFF
--- a/web/src/main/ui/app/components/flows-new/edit-flow/mapping/mapping.component.ts
+++ b/web/src/main/ui/app/components/flows-new/edit-flow/mapping/mapping.component.ts
@@ -236,7 +236,7 @@ export class MappingComponent implements OnInit {
         
       });
       self.sampleDocNestedProps = this.updateNestedDataSourceNew(startRoot,ParentKeyValuePair);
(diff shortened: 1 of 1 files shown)


web/src/main/ui/app/components/flows-new/edit-flow/mapping/mapping.component.ts
```diff
@@ -300,18 +301,20 @@ export class MappingComponent implements OnInit {
...
                 }
               };
 
               //loading attributes
               self.loadAttributes(childNode, obj, 'singleNode');
 
               nodeToJSON(tempObj, childNode);
-              if (obj[childNode.nodeName].constructor.name === 'Array') {
-                obj[childNode.nodeName].push(tempObj);
+              let ob = {[`${childNode.nodeName}`] : tempObj}
```
wooldridge: Should "let ob" be "let obj" here?

PR #3176, opened 2019-10-31
Test users now have an easy-to-remember password

rjrudin:
This is just to make life easier for when I want to run the unit tests via host:8011/test/default.xqy as flow-developer instead of as admin user. It has not impact on DHF application code.


SameeraPriyathamTadikonda:
Merging this manually as all the unit tests passed.

PR #3178, opened 2019-10-31
DHFPROD-3534: Adding collection name with entity-model-name when running a flow using gradle task

rahulvudutala:


marklogic-data-hub/src/main/resources/ml-modules/services/mlRunFlow.sjs
```diff
@@ -39,6 +39,10 @@ function post(context, params, input) {
     let stepDetailsOptions = stepDetails.options || {};
     // build combined options
     let combinedOptions = Object.assign({}, stepDetailsOptions, flowOptions, stepRefOptions, options);
+    let collections = combinedOptions.collections;
+    if(combinedOptions.targetEntity && !collections.includes(combinedOptions.targetEntity)) {
+      combinedOptions.collections.push(combinedOptions.targetEntity);
```
rjrudin: @ryanjdew @srinathgit @bsrikan  Thoughts on this? This change is saying - if a step or flow has a targetEntity defined, then any inserted documents will be added to that collection, regardless of whether it's configured or not. 

I can see how that is generally good, but I'm not sure if we should force this behavior. 

Additionally, DHFPROD-3534 identifies an issue - that the way steps are initially defined differs between QS and Gradle - but this does not address that issue. I think it's best to address that issue, rather than forcing behavior into the running of *any* step. 

When hubRunFlow is run, the task sets targetEntity to "entity-name" and then prints out the following: 

"IMPORTANT: Your new flow configuration file "flows/SomeFlow.flow.json" contains step templates with example values, such as 'inputFilePath' and 'entity-name'. The flow will not run as is. You MUST customize the steps for your project before running the flow."

However, when hubCreateStepDefinition is run, targetEntity is not stubbed out to "entity-name", and no such message is displayed.

I think since DHfPROD-3534 is identifying an issue with hubCreateStepDefinition, we should change its behavior, and the logical change seems to be that that task should include targetEntity:"entity-name" and print a message similar to the one that hubRunFlow prints out.

ryanjdew: @rjrudin @rahulvudutala  That is fair. Also, I missed that this change would apply an entity collection to all of the mastering documents, even those that aren't actual entity instances.

bsrikan: @rjrudin I think you are talking about hubCreateFlow, that is when we get the msg. hubCreateFlow does not create any step definitions, only a json flow artifact with 3 steps. So you can technically run "hubRunFlow" after running "hubCreateFlow". 
"hubRunFlow" will then run ingestion,mapping and mastering step. There will be some errors.

Ingest step: gives a Runtime exception for input path
Mapping step: gives an error for not having found mapping name
Mastering step: is a pass thru

I think the requirement for Explorer is to have a collection with entity name in the flow artifact, so the harmonized document has entity name as a collection. 

Fixing hubCreateStepDefinition might not achieve what Explorer wants/needs.

hao1st: @bsrikan What you said is what I understood. This is why @rahulvudutala tried to add the entity name as an additional collection to harmonized documents if missing during hubRunflow.

rjrudin: @hao1st @rahulvudutala "the harmonized document has entity name as a collection" is an implementation detail. What is the user-facing requirement?

rjrudin: I mistakenly said "hubRunFlow" above - as was pointed out, I meant "hubCreateFlow".

The difference with creating a step via Gradle is that hubCreateStepDefinition does not require an entity type to be specified the way e.g. QuickStart does when creating a mapping/matching/merging/mastering step. Thus, the step created by hubCreateStepDefinition does not have targetEntity set in it, while QS does set targetEntity, since it's a required input.

The solution I'm proposing is to follow the pattern shown by hubCreateFlow - when running hubCreateStepDefinition, set targetEntity to "entity-name" and print a message similar to the one printed by hubCreateFlow. That avoids forcing the inclusion of a collection regardless of the step, which may not always be desired by the user. 

Also, for "Explorer expects that each document has a related entity model." - why? What user-facing problem is being solved here?

rjrudin: Also, re: what @bsrikan stated - "I think the requirement for Explorer is to have a collection with entity name in the flow artifact, so the harmonized document has entity name as a collection."

There's no guarantee that a flow will have targetEntity set. Many users are making use of custom steps today because of gaps with mapping and mastering steps. It is not likely that they will have set targetEntity at the flow level. So this change is not going to address that scenario.

akshaysonvane: > Also, for "Explorer expects that each document has a related entity model." - why? What user-facing problem is being solved here?

The use-case is it to find all harmonized documents for a particular entity model. 
Explorer app performs a collection query on the assumption that all documents harmonized for a particular entity model are also tagged with the entity model's name.

browda: To a great extent, we need to understand exactly why we would expect that a collection would necessarily be tagged with the name of an entity model.  If the assumption is that the user will be led down the garden path as it were and be forced to do that association, that's one thing, but it appears as if the reality is something different.  This is analogous to the bromide that a new version of a program should never remove a feature that's in widespread use without going through a deprecation process.  With databases, this is still more problematic -- it's an archive where some data is used more often than other data -- since forcing compliance with what would appear to be an arbitrary standard that doesn't appear to be useful in day-to-day activity by users of that datybase.

rjrudin:
I'm closing this per the meeting today, where we agreed to have @ryanjdew look into a custom constraint that performs a query on /envelope/instance/(entity type).

PR #3179, opened 2019-10-31
DHFPROD-3539: Smart Mastering cleanup items

ryanjdew:
Since we're changing the path for the merge range, this will require a re-deploy to get all the changes.

Items addressed:

 

- When I run a match step on a set of documents, a match summary JSON document is created (no change from previous behavior). This match summary JSON document is changed in the following ways:

  - The URIs are displayed so the user understands which URIs will be processed (instead of just hashes)
  - The name 'URIsToActOn' is changed to 'URIsToProcess'
  -  In the "actionDetails" section, "merge" actions should have the relevant threshold included similar to "notify" actions
  - Merge step should honor source collection/query instead of ignoring it for its own method. (e.g., the merge step should restrict the range index values to the collection/query specified)

- When I run a mastering step or a merge step, an audit file is created for each merge. These audit files should be added to the sm-entityname-audit collection (instead of the mdm-audit collection)


bsrikan:
Thanks @wooldridge for verifying.

PR #3181, opened 2019-11-01
DHFPROD-3547: Use proper document root for declarative mapping

ryanjdew:


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mapping/entity-services/main.sjs
```diff
@@ -60,7 +60,7 @@ function main(content, options) {
   }
   let entityName = lib.getEntityName(targetEntityType);
   let entity = lib.getTargetEntity(targetEntityType);
-  let instance = doc.xpath('head((/*:envelope/(*:instance[count(* except *:info) gt 1]|*:instance/(* except *:info)),./object-node(),./*))');
+  let instance = doc.xpath('head((/*:envelope/(object-node("instance")|*:instance/(element() except *:info)),./object-node(),./*))');
```
srinathgit: Can you make the change here as well https://github.com/marklogic/marklogic-data-hub/blob/63a520f683302d0336747548d6a2bcce89d0f5b7/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mapping/entity-services/lib.sjs#L324 ?

ryanjdew: Updated it with the change there as well with latest push

bsrikan:
@ryanjdew is this expected. If I set the complete path for a nested property(no context) it gets me null. 
image

With source context set I get right value.
image

bsrikan:
Same is true for xml as well:
with context
image

with no context
image

PR #3196, opened 2019-11-06
DHFPROD-3603: Don't assume json dataFormat when creating match option…

ryanjdew:
…s via QS

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/com.marklogic.smart-mastering/impl/process.xqy
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/com.marklogic.smart-mastering/impl/process.xqy
@@ -712,14 +712,7 @@ declare function proc-impl:build-content-objects-from-match-summary(
                 let $uris-to-merge := $action-details => map:get("uris") => json:array-values()
                 let $merged-doc-def := merge-impl:build-merge-models-by-uri($uris-to-merge, $merge-options, $uri)
                 let $merged-doc := $merged-doc-def => map:get("value")
-                let $merge-uri := merge-impl:build-merge-uri(
-                  $uri,
-                  if ($merged-doc instance of element() or
-                    $merged-doc instance of document-node(element())) then
-                    $const:FORMAT-XML
-                  else
-                    $const:FORMAT-JSON
-                )
+                let $merge-uri := $uri
                 let $prov-entry :=
                       let $merge-provenance-info as map:map? := if ($fine-grain-provenance) then $merged-doc-def => map:get("provenance") else ()
                       let $match-provenance-info as map:map? := $action-details => map:get("provenance")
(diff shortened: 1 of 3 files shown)


marklogic-data-hub/src/main/resources/ml-modules/root/com.marklogic.smart-mastering/impl/process.xqy
```diff
@@ -712,14 +712,7 @@ declare function proc-impl:build-content-objects-from-match-summary(
...
                 let $merged-doc := $merged-doc-def => map:get("value")
-                let $merge-uri := merge-impl:build-merge-uri(
-                  $uri,
-                  if ($merged-doc instance of element() or
-                    $merged-doc instance of document-node(element())) then
-                    $const:FORMAT-XML
-                  else
-                    $const:FORMAT-JSON
-                )
+                let $merge-uri := $uri
```
ryanjdew: This change is also for the merge URIs looking off.

bsrikan:
There is a build error. Can you pls look into it @ryanjdew 

Date: 2019-11-06T21:06:18.063Z
Hash: ecca7871949c2b5089ab
Time: 49495ms
chunk {0} runtime.26209474bfa8dc87a77c.js (runtime) 1.41 kB [entry] [rendered]
chunk {1} main.d88adad4ce8420d95d4e.js (main) 130 bytes [initial] [rendered]
chunk {2} polyfills.8d1e9ff8a9397209762d.js (polyfills) 128 bytes [initial] [rendered]
chunk {3} styles.ff4e1db7be9b7df89b8f.css (styles) 130 kB [initial] [rendered]

ERROR in src/main/ui/app/components/flows-new/edit-flow/mastering/matching/matching.model.ts(28,14): error TS2339: Property 'dataFormat' does not exist on type 'Matching'.

bsrikan:
I still continue to get empty actionDetails when mapped document is in XML format. JSON seems to work.

ryanjdew:
@bsrikan If the match options already sets dataFormat as "json", this PR won't change existing match options. This change makes it so we don't set it at all when we create new steps. Not having it set will cause matching/mastering check the input documents to determine data format.

bsrikan:
@ryanjdew matchOptions.dataFormat in my match step was json and options.outputFormat is also json. Created a new match step which didnt have dataFormat. Still I see the same issue with this PR even when I set both dataFormat and outputFormat to xml

ryanjdew:
@bsrikan Running the MatchFix step and setting the source collection to MatchFix on the merge step seems to give me the merges I'd expect to see.

bsrikan:
I didnt run the merge step. There were 500 mapped documents. So running the match step should give me more than 5 match summary documents. I always get 5 and when I open the summary report the actionDetails is empty. I will try again on a clean install and fresh snapshot build.

PR #3198, opened 2019-11-07
DHFPROD-3606:Generate function metadata only if forceLoad is set to true

srinathgit:
Generate function metadata only if forceLoad is set to true

### DIFF
--- a/web/src/main/java/com/marklogic/hub/web/service/DataHubService.java
+++ b/web/src/main/java/com/marklogic/hub/web/service/DataHubService.java
@@ -188,7 +188,13 @@ public class DataHubService {
         loadHubArtifactsCommand.setHubConfig(hubConfig);
         loadHubArtifactsCommand.setForceLoad(forceLoad);
 
-        commands.add(generateFunctionMetadataCommand);
+        // Generating function metadata xslt causes running existing mapping (xslts) step to fail with undefined function
+        // for any mappings that use these functions. So, we have to generate function metadata xslt only when 'forceLoad'
+        // is set to true which would ensure that mappings are inserted and compiled to xslts as well.
+        // Added as a temporary fix for DHFPROD-3606.
+        if(forceLoad){
+            commands.add(generateFunctionMetadataCommand);
+        }
         commands.add(loadUserModulesCommand);
         commands.add(loadUserArtifactsCommand);
         commands.add(loadHubArtifactsCommand);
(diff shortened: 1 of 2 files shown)


web/src/main/java/com/marklogic/hub/web/service/DataHubService.java
```diff
@@ -188,7 +188,9 @@ private void installUserModules(HubConfig hubConfig, boolean forceLoad, DeployUs
         loadHubArtifactsCommand.setHubConfig(hubConfig);
         loadHubArtifactsCommand.setForceLoad(forceLoad);
 
-        commands.add(generateFunctionMetadataCommand);
+        if(forceLoad){
+            commands.add(generateFunctionMetadataCommand);
+        }
```
rjrudin: Looking at the clients of this method - I don't think this can be done for CurrentProjectController::onWatchEvent. If the file change that was detected was for a mapping functions module, forceLoad will be false, and then function metadata will not be generated. 

Now how likely is it that the person using QS is also going to make a change to a mapping functions module? Today, that seems likely because the typical QS user is an ML SE/consultant. 

The root problem seems to be that generating function metadata for an unmodified mapping functions module causes the compiled XSLTs to fail. As you noted in 3606, this will still be a problem when using the Gradle tasks. I think we need that problem fixed, as this patch here only addresses QS and I think introduces the problem I pointed out above as well. 

@ryanjdew Perhaps the command outside of  ML isn't the way to go here, but rather a pre-commit trigger in the modules database? Then function metadata is only generated when the mapping module is created/modified, and it can be removed when the module is deleted. 

Though @srinathgit - what happens if I add a step to your steps in 3606 - a 3.5 where I modify a mapping functions module and then run the step again? Does it work, or does it fail then too?

srinathgit: If mapping function is added/edited, users should click on "REDEPLOY" button in the Manage Flows page which will force load , thereby generating function metadata again and recompiling the mappings

rjrudin: I did some testing locally with your change, and I verified that it at least is an improvement. 

My concern about changes to mapping function modules is unfounded, which isn't great - I noticed that if I change a custom mapping function module, QS doesn't detect that and thus doesn't load it. That then forces me to run mlLoadModules, which then causes the bug. 

So I think this change is worthwhile, but I'll add other comments to 3606.

rjrudin:
This now seems worth merging in. It's a patch on what appears to be a server bug. So the bug isn't completely solved, but it's an improvement.

web/src/main/java/com/marklogic/hub/web/service/DataHubService.java
rjrudin: @srinathgit  One request - please add comments here to explain what is going on here. Ideally we can remove this in the future soon.

PR #3199, opened 2019-11-07
DHFPROD-3376: mapping UI- If there is no ingested data, there should be an informational message explaining that the query didn't return any source documents

xnikhil08:
Contains the changes for the story DHFPROD-3376.

### DIFF
--- a/web/src/main/ui/app/components/flows-new/edit-flow/mapping/ui/mapping-ui.component.html
+++ b/web/src/main/ui/app/components/flows-new/edit-flow/mapping/ui/mapping-ui.component.html
@@ -1,7 +1,9 @@
 <!-- START NEW MAPPING VIEW -->
 <div *ngIf="isVersionCompatibleWithES" class="source-prop-container-map-exp">
-
-  <div>
+  <div *ngIf="targetEntity && !targetEntity.hasDocs">
+    <em>Unable to find documents that match your query.</em>
+  </div>
+  <div [ngClass]="{ hidden: !(targetEntity && targetEntity.hasDocs)}">
   
     <div class="Source_Tbl">
       <div class="item-title">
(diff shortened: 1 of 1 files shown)


web/src/main/ui/app/components/flows-new/edit-flow/mapping/ui/mapping-ui.component.html
```diff
@@ -1,7 +1,9 @@
 <!-- START NEW MAPPING VIEW -->
 <div *ngIf="isVersionCompatibleWithES" class="source-prop-container-map-exp">
-
-  <div>
+  <div *ngIf="targetEntity && !targetEntity.hasDocs">
+    <em>Unable to find documents that match your query.</em>
+  </div>
```
srinathgit: In the old mapping  UI, if no docs are found, the info we see is "Before configuring this mapping you must ingest source documents." . Can we continue to have that or is there any reason for the change ?

bsrikan: I checked the message suggested in the user story and it matched the implementation here.

web/src/main/ui/app/components/flows-new/edit-flow/mapping/ui/mapping-ui.component.html
```diff
@@ -1,7 +1,9 @@
 <!-- START NEW MAPPING VIEW -->
 <div *ngIf="isVersionCompatibleWithES" class="source-prop-container-map-exp">
-
-  <div>
+  <div *ngIf="targetEntity && !targetEntity.hasDocs">
+    <em>Unable to find documents that match your query.</em>
```
wooldridge: @eltesoro-ml has updated the messaging, see the Jira ticket for the new version.

xnikhil08: Updated the message as per the Jira tifcket.

PR #3202, opened 2019-11-08
DHFPROD-3616:Don't remove mapping functions when "REDEPLOY"button is hit

srinathgit:
Don't remove mapping functions when "REDEPLOY"button is hit

### DIFF
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/impl/DataHubImpl.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/impl/DataHubImpl.java
@@ -27,7 +27,6 @@ import com.marklogic.appdeployer.command.modules.DeleteTestModulesCommand;
 import com.marklogic.appdeployer.command.modules.LoadModulesCommand;
 import com.marklogic.appdeployer.command.security.*;
 import com.marklogic.appdeployer.impl.SimpleAppDeployer;
-import com.marklogic.client.FailedRequestException;
 import com.marklogic.client.admin.QueryOptionsManager;
 import com.marklogic.client.admin.ResourceExtensionsManager;
 import com.marklogic.client.admin.ServerConfigurationManager;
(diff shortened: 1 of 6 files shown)


marklogic-data-hub/src/main/java/com/marklogic/hub/impl/DataHubImpl.java
```diff
@@ -385,7 +384,8 @@ public void clearUserModules() {
                 "cts:uris((),(),cts:not-query(cts:collection-query('hub-core-module')))[\n" +
                     "  fn:not(\n" +
                     "    fn:matches(., \"^.+options/(" + String.join("|", options) + ").xml$\") or\n" +
-                    "    fn:starts-with(., \"/marklogic.rest.\")\n" +
+                    "    fn:starts-with(., \"/marklogic.rest.\") or\n" +
+                    "    fn:matches(.,\"/custom-modules/mapping-functions/.*.sjs|mjs|xqy\") "+
```
rjrudin: A comment is needed here because we have a method called "clearUserModules", but we're then saying, "Well, but don't delete these modules", which leaves someone (like me) thinking - well why not?

Additionally, an easy way to get at least some sort of unit test in place here is to move construction of this string into a protected method. Then, a unit test can invoke the protected method and at least verify that the query string includes the fn:matches clause that's been added here - and the assertion can also document why we're not deleting these modules.

srinathgit: We already have a DataHubInstallTest.testClearUserModules(). I will add a test there where I can confirm that mapping functions aren't removed.

srinathgit: Added test

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/DataHubImpl.java
```diff
@@ -380,12 +379,13 @@ public void clearUserModules() {
                     }
                 }
             );
-
+            //The custom mapping functions, ootb rest extensions and search options are not deleted by this method
```
rjrudin: Sorry for being stubborn, but why are we not deleting these?

srinathgit: Hitting "REDEPLOY" doesn't load these and they will be lost unless users run mlDeploy again

srinathgit: One thing I don't understand is why LoadUserModulesCommand which is called when DataHubService.installUserModules() is run, doesn't load the custom mapping functions ? Could this also be the issue that @bsrikan mentioned in the curation slack room ?

bsrikan:
Looks ok to me.

srinathgit:
The latest commit should fix the test failures

rjrudin:
Is this still needed, or is #3205 the complete solution?

srinathgit:
I think this can be closed. #3205 addresses the issue

PR #3205, opened 2019-11-09
DHFPROD-3616:Fix forceLoad in LoadUserModulesCommand

srinathgit:
This PR does 2 things

1. Fixes 'forceLoad' in LoadUserModulesCommand so that property file manager is initialized and timestamp file is deleted before loading from standard ml-gradle location when forceLoad is set to true
2. Load all user artifacts so that mappings get recompiled so that we don't run into XDMP-UNDFUN error (DHFPROD-3606)

### DIFF
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/LoadUserArtifactsCommand.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/LoadUserArtifactsCommand.java
@@ -98,9 +98,6 @@ public class LoadUserArtifactsCommand extends AbstractCommand {
                 }
             }
         }
-        //Initializing the 'PropertiesModuleManager' instance so that timestamp file (local-user-modules-deploy-timestamps.properties)
-        // is read and only the modified files are deployed to server
-        pmm.initialize();
         return pmm;
     }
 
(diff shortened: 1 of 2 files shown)


marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/LoadUserArtifactsCommand.java
```diff
@@ -100,7 +100,7 @@ private PropertiesModuleManager getModulesManager() {
         }
         //Initializing the 'PropertiesModuleManager' instance so that timestamp file (local-user-modules-deploy-timestamps.properties)
         // is read and only the modified files are deployed to server
-        pmm.initialize();
+        //pmm.initialize();
```
rjrudin: I recommend deleting this line and the comments above it - keeping it will otherwise be confusing.

marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/LoadUserModulesCommand.java
```diff
@@ -180,6 +177,12 @@ public void execute(CommandContext context) {
         // load any user files under plugins/* int the modules database.
         // this will ignore REST folders under entities
         DefaultModulesLoader modulesLoader = getStagingModulesLoader(config);
+        // Load modules from standard ml-gradle location after 'PropertiesModuleManager' is initialized. This will ensure
+        // that in case of 'forceLoad' the ml-javaclient-utils timestamp file is deleted first and we the old timestamp file
```
rjrudin: Need to fix the grammar in the comment "and we the old timestamp file is not used when loading" - actually, I think you can end the comment right after "is deleted first".

PR #3220, opened 2019-11-11
DHFPROD-2806: Added json-mapping-example project

rjrudin:
Also updated the insurance project to use the new mapping system

This is for 5.1-develop


bsrikan:
@rjrudin When I run the flows they work as expected.
For jsonToJson flow, when I hit Test button I get 500 error in the Network tab of the browser as so:
`Local message: failed to apply resource at resources/ml:mappingValidator: Bad Request. Server Message: XDMP-URI: cts.doc("") -- Invalid URI format: ""`

bsrikan:
Looks like there is no sourceURI in the jsonToJson flow. With that fixed, we can merge the example project.

PR #3223, opened 2019-11-11
DHFPROD-3523: No longer depending on com.sun class for file watching

rjrudin:
This class isn't in OpenJDK, which means you can't compile/run QS on OpenJDK


rjrudin:
@srinathgit Can you test that locally? I only have OpenJDK and thus can't run it with that class included.

srinathgit:
@rjrudin ,
I just moved to openJDK 11 in my local machine. I will coordinate with @bsrikan  to test it

PR #3224, opened 2019-11-11
DHFPROD-3608: Don't break when ingesting JSON to XML (5.2)

ryanjdew:


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/flow-utils.sjs
```diff
@@ -111,18 +111,17 @@ class FlowUtils {
       }
     } else if (inputFormat === dataFormat) {
       if(content instanceof Element &&  content.nodeName.toLowerCase() === 'root' && content.namespaceURI.toLowerCase() === ""){
-        instance = Sequence.from(content.xpath('/root/node()'));
+        instance = Sequence.from(content.xpath('node()'));
```
aebadirad: this handles xml to xml - MCLP will insert <root> for csv splits, this to get 'under' that.

ryanjdew: The if statement checks that content is already at the root element. I suspect it was working before because the first "/" would travel back up to the document root, but since it is already at the root element it should only need to get the child nodes.

aebadirad: I was just confirming the intent of that block added for the record. not necessarily disputing the node traversal

PR #3250, opened 2019-11-14
DHFPROD-3191:Create prov doc in ingestion step

srinathgit:
Create prov doc in ingestion step

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/flow.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/flow.sjs
@@ -249,6 +249,9 @@ class Flow {
         }
       }
     }
+    if (prov.granularityLevel() !== prov.OFF_LEVEL) {
+      this.datahub.prov.commit();
+    }
     if (!combinedOptions.noBatchWrite) {
       let batchStatus = "finished";
       if (this.globalContext.failedItems.length) {
@@ -260,9 +263,6 @@ class Flow {
       }
       if (!combinedOptions.disableJobOutput) {
         this.datahub.jobs.updateBatch(this.globalContext.jobId, this.globalContext.batchId, batchStatus, uris, writeTransactionInfo, this.globalContext.batchErrors[0]);
-      }
-      if (prov.granularityLevel() !== prov.OFF_LEVEL) {
-        this.datahub.prov.commit();
       }
     }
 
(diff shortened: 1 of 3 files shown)


marklogic-data-hub/src/test/java/com/marklogic/hub/flow/FlowRunnerTest.java
```diff
@@ -133,7 +140,16 @@ public void testIngestCSVasXML(){
         EvalResult res = resultItr.next();
         long count = Math.toIntExact((long) res.getNumber());
         Assertions.assertEquals(count, 25);
-    }
+        if(! isCertAuth() ) {
+           EvalResultIterator itr = client.newServerEval().xquery("xdmp:estimate(fn:collection('http://marklogic.com/provenance-services/record'))").eval();
+           if(itr != null && itr.hasNext()) {
+               Assertions.assertTrue(itr.next().getNumber().intValue() == 25);
```
rjrudin: It's better to use assertEquals with an expected value of 25 - that way if the test fails, the fail message states the expected and actual values instead of just saying "I expected true but it was false".

marklogic-data-hub/src/test/java/com/marklogic/hub/flow/FlowRunnerTest.java
```diff
@@ -116,7 +117,13 @@ public void testRunFlow(){
     }
 
     @Test
-    public void testIngestCSVasXML(){
+    public void testIngestCSVasXML() throws Exception {
+        //prov docs cannot be read by "flow-developer-user", so creating a client using 'secUser' which is 'admin'
```
rjrudin: Just a note that this will change after 5.1 when we introduce the data-hub-job-reader role.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/flow.sjs
```diff
@@ -249,6 +249,9 @@ class Flow {
         }
       }
     }
+    if (prov.granularityLevel() !== prov.OFF_LEVEL) {
```
rjrudin: Is the reason for this that mlRunIngest.sjs sets "noBatchWrite" to true?

srinathgit: Yeah and because of that the prov docs aren't getting written

PR #3281, opened 2019-11-20
DHFPROD-3722:Make mapping validation/test work with final db

srinathgit:
Make mapping validation/test work with final db

### DIFF
--- a/web/src/main/java/com/marklogic/hub/web/service/MappingManagerService.java
+++ b/web/src/main/java/com/marklogic/hub/web/service/MappingManagerService.java
@@ -18,6 +18,7 @@ package com.marklogic.hub.web.service;
 
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.marklogic.hub.DatabaseKind;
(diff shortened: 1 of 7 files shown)


web/src/main/java/com/marklogic/hub/web/service/MappingManagerService.java
```diff
@@ -55,7 +58,7 @@
     @Autowired
     HubConfigImpl hubConfig;
 
-    private MappingValidator mappingValidator = null;
+    private Map<String, MappingValidator> mappingValidators = new HashMap<>();
```
rjrudin: Does the Spring container get reloaded when a user logs out of QS to switch to another project? I'm wondering if this map would persist through that process, such that the wrong DatabaseClient objects would be used when testing mapping validations in the other project.

srinathgit: The objects do persist in the map after a user logs out. I will add a method that will set the map to null in CurrentProjectController.onLogoutSuccess() method

srinathgit: Updated the PR

PR #3284, opened 2019-11-20
DHFPROD-3630: Reduce list cache hits when determining if we can delet…

ryanjdew:
…e match summary documents

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mastering/default/merging.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mastering/default/merging.sjs
@@ -49,10 +49,13 @@ function main(content, options) {
   for (let matchSummary of relatedMatchSummaries) {
     let matchSummaryObj = matchSummary.toObject().matchSummary;
     let URIsToProcess = matchSummaryObj.URIsToProcess;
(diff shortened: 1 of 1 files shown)


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mastering/default/merging.sjs
```diff
@@ -49,10 +49,13 @@ function main(content, options) {
   for (let matchSummary of relatedMatchSummaries) {
     let matchSummaryObj = matchSummary.toObject().matchSummary;
     let URIsToProcess = matchSummaryObj.URIsToProcess;
-    let allURIsProcessed = URIsToProcess.every((uri) => processedURIs.includes(uri) || cts.exists(cts.andQuery([
-      cts.documentQuery(uri),
+    // use one cts.uris call per match summary to reduce list cache hits
```
rjrudin: This change makes sense - do one query instead of N queries, where N = number of URIsToProcess - but I thought the reason would be for better performance, as one query is almost always faster than N queries? I'm confused by the "reduce list cache hits" comment.

ryanjdew: It came from Silvano's testing. https://project.marklogic.com/jira/browse/DHFPROD-3630?focusedCommentId=163485&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-163485

PR #3289, opened 2019-11-20
DHFPROD-3690: Proper mapping in JSON to XML scenario

ryanjdew:


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mapping/entity-services/lib.sjs
```diff
@@ -408,11 +413,30 @@ function versionIsCompatibleWithES(version = xdmp.version()) {
   return false;
 }
 
+function extractInstance(docNode) {
+  let instance = docNode.xpath('/*:envelope/(object-node("instance")|*:instance/(element() except *:info))');
```
rjrudin: I think this warrants multiple unit tests directly on this function (in addition to the test you added already, which has a broader scope). The XPath expression is supporting multiple scenarios; we should have a unit test for each scenario that it supports to ensure that none of them break in the future, in case someone tries to change this.

ryanjdew: Added an additional test specifically for extractInstance. In the future I think it would be great to get set criteria for out unit tests. For example, are going for code coverage? Are going for direct calls to each function with a cyclomatic complexity of X or more? etc.

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/builtins/steps/mapping/entity-services/test-data/content/json-to-xml.xml
```diff
@@ -0,0 +1,15 @@
...
+    <personId>222</personId>
+    <theNickname>Nicky</theNickname>
+    <theName>
+      <middleName>Middle</middleName>
+      <lastName>Last</lastName>
+      <firstName>
+        <theValue>First</theValue>
+        <thePrefix>SomePrefix</thePrefix>
+      </firstName>
+    </theName>
```
bsrikan: Can we please add an array example to the test.

ryanjdew: @bsrikan Are you talking about the scenario for https://project.marklogic.com/jira/browse/DHFPROD-3731? I believe that is a separate issue.

bsrikan: No, just to get more coverage just to make sure extractInstance works as expected for array of elements

PR #3300, opened 2019-11-22
WIP Prototype of running a flow with no project dependency

rjrudin:
This is a WIP, looking for feedback on this approach of making a new implementation of HubConfig to avoid HubConfigImpl's hard dependency on a HubProject. 

SimpleHubConfig is a new impl of HubConfig that does not have a hard dependency on a HubProject. It extends the new AbstractHubConfig class, to which I moved as much of HubConfigImpl as possible for reuse purposes. 

Modified a few clients of HubConfig that were casting to HubConfigImpl to call getMlUsername and getMlPassword - those are now part of the HubConfig interface so that those casts aren't needed. 

Added a new method to FlowRunnerImpl so that a Flow is retrieved from ML instead of the filesystem. 

Added StepDefinitionProvider so that StepRunnerFactory can get a StepDefinition from ML instead of from the filesystem. 

RunStepTest isn't yet a real test, it's just a sample program to show the new stuff in action

### DIFF
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/collector/impl/CollectorImpl.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/collector/impl/CollectorImpl.java
@@ -23,8 +23,6 @@ import com.marklogic.client.MarkLogicIOException;
 import com.marklogic.hub.HubConfig;
 import com.marklogic.hub.collector.DiskQueue;
 import com.marklogic.hub.collector.Collector;
-import com.marklogic.hub.flow.Flow;
-import com.marklogic.hub.impl.HubConfigImpl;
 import com.marklogic.rest.util.MgmtResponseErrorHandler;
 import org.apache.http.auth.AuthScope;
 import org.apache.http.auth.UsernamePasswordCredentials;
@@ -32,8 +30,6 @@ import org.apache.http.client.HttpClient;
 import org.apache.http.conn.ssl.X509HostnameVerifier;
 import org.apache.http.impl.client.BasicCredentialsProvider;
 import org.apache.http.impl.client.HttpClientBuilder;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 import org.springframework.http.HttpMethod;
 import org.springframework.http.MediaType;
 import org.springframework.http.client.HttpComponentsClientHttpRequestFactory;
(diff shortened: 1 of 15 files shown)


marklogic-data-hub/src/main/java/com/marklogic/hub/collector/impl/CollectorImpl.java
```diff
@@ -59,18 +55,16 @@
 import java.util.*;
 
 public class CollectorImpl implements Collector {
-    private DatabaseClient client = null;
-    private HubConfig hubConfig = null;
 
+    private DatabaseClient client;
+    private HubConfig hubConfig;
 
-    private Flow flow = null;
```
rjrudin: This actually wasn't being used at all, so removed it.

marklogic-data-hub/src/main/java/com/marklogic/hub/flow/impl/FlowRunnerImpl.java
```diff
@@ -93,21 +102,28 @@ public RunFlowResponse runFlow(String flowName, List<String> stepNums,  String j
...
 
     public RunFlowResponse runFlow(String flowName, List<String> stepNums, String jobId, Map<String, Object> options, Map<String, Object> stepConfig) {
+        Flow flow = flowManager.getFlow(flowName);
+        if (flow == null) {
+            throw new RuntimeException("Flow " + flowName + " not found");
+        }
+        return runFlow(flow, stepNums, jobId, options, stepConfig);
+    }
+
+    public RunFlowResponse runFlowDefinedInMarkLogic(String flowName, List<String> stepNums, String jobId, Map<String, Object> options, Map<String, Object> stepConfig) {
```
rjrudin: This will eventually have a better name.

marklogic-data-hub/src/main/java/com/marklogic/hub/step/MarkLogicStepDefinitionProvider.java
```diff
@@ -0,0 +1,24 @@
...
+
+    private DatabaseClient databaseClient;
+
+    public MarkLogicStepDefinitionProvider(DatabaseClient databaseClient) {
+        this.databaseClient = databaseClient;
+    }
+
+    @Override
+    public StepDefinition getStepDefinition(String name, StepDefinition.StepDefinitionType type) {
+        // TODO Support custom step definitions
```
rjrudin: This won't work yet for custom step definitions, but this is just a prototype. The real impl will support them.

PR #3304, opened 2019-11-22
DHFPROD-3620: refactor method signature and names for dateparse and d…

aebadirad:
…atetimeparse functions in core mapping functions


aebadirad:
Srinath bailed me out, this should fix it all :)

PR #3305, opened 2019-11-22
DHFPROD-3620: refactor method signature and names for dateparse and d…

aebadirad:
…atetimeparse functions in core mapping functions


marklogic-data-hub/src/test/java/com/marklogic/hub/deploy/commands/GenerateFunctionMetadataCommandTest.java
```diff
@@ -50,8 +50,8 @@ public void setupEach() throws IOException {
         DocumentWriteSet writeSet = modMgr.newWriteSet();
         StringHandle handle = new StringHandle("'use strict';\n" +
             "\n" +
-            "function testModule(picture, value) {\n" +
-            "  return xdmp.parseDateTime(picture, value);\n" +
+            "function testModule(value, pattern) {\n" +
+            "  return xdmp.parseDateTime(value, pattern);\n" +
```
srinathgit: This is a xdmp function, so I think , it has to be changed to xdmp.parseDateTime(pattern, value)

PR #3319, opened 2019-11-25
DHFPROD-3499: Stubbing out mapping-functions directory

rjrudin:


marklogic-data-hub/src/main/java/com/marklogic/hub/impl/HubProjectImpl.java
```diff
@@ -870,22 +877,6 @@ private void deleteObsoleteServerFilesFromHubInternalConfig() {
         }
     }
 
-    private void deleteObsoleteDirsFromHubInternalConfig() {
```
rjrudin: Unused function

PR #3324, opened 2019-11-25
DHFPROD-3103: xml mapping project

bsrikan:
This PR includes updates to existing json-mapping-project with scenarios for xml as output format as described in DHFPROD-3103


examples/json-mapping-example/README.md
```diff
@@ -87,3 +89,19 @@ occur multiple times under "*:/person/*:names".
 1. As shown in the "prefix" mapping, XML attributes can be referenced via the "@" symbol. 
 
 Otherwise, all of the features used in the jsonToJsonFlow can be used in this mapping as well. 
+
+### Mapping features in the jsonToXml flow
+
+From the "Flows" view in QuickStart, open the jsonToXml flow and select the mapping step. QuickStart will display the
+JSON document in staging in the "Source Data" panel and the mapping in the "Entity" panel. The mapping displays all of
+the properties of the Person entity along with its nested Name and Address entities. Each row in the mapping is similar to the jsonToJson flows
```
rjrudin: This sentence should end with "jsonToJson flow." as opposed to "flows" with no period.

examples/json-mapping-example/README.md
```diff
@@ -87,3 +89,19 @@ occur multiple times under "*:/person/*:names".
...
+
+### Mapping features in the xmlToXml flow
+
+From the "Flows" view in QuickStart, open the xmlToXml flow and select the mapping step. QuickStart will display the XML
+document in staging in the "Source Data" panel and the mapping in the "Entity" panel.
+The xmlToXml mapping is intended to highlight usage of namespace and namespace prefix in the mapping
```
rjrudin: Need a period at the end of this sentence.

examples/json-mapping-example/README.md
```diff
@@ -87,3 +89,19 @@ occur multiple times under "*:/person/*:names".
...
+
+1. The properties in the expression have the namespace defined for "person" element. The mapping also shows usage of namespace prefix. The source document has namespace prefix and the same is reflected in the source table and element selector.
+1. The "state" property demonstrates usage of the "if then else" expression.
+1. The "email" property is an array of string and can hold multiple values. Click on the "Test" button to verify multiple values returned with an ellipsis showing how many more such elements were extracted
```
rjrudin: A few tweaks:

- In the first statement, should be 'defined for the "person" element." And then "usage of a namespace prefix". And "source document has a namespace prefix".
- In the second statement, "usage of an" instead of "usage of the". 
- In the third statement, "array of strings" instead of "string". And then "verify multiple values are returned". And then need a period at the end.

bsrikan: @rjrudin 
Too many 'the' in the first statement, so I decided against. Changed it like so:
The properties in the expression have namespace defined for the "person" element.
In the second statement I used "the" because "if then else" was a definite item

Agree with other changes. Pls let me know if you still prefer 
the "if then else" expression OR
an "if then else" expression

rjrudin: Sounds good for the first item. I think "an" is right for the second item.

examples/mapping-example/README.md
```diff
@@ -21,15 +21,17 @@ Next, start up QuickStart and browse to this project folder and login to QuickSt
...
 1. The xmlToJson flow ingests and maps a single XML document to a JSON entity instance
+1. The xmlToXml flow ingests and maps a single XML document to an XML entity instance
 
-The jsonToJson flow demonstrates a wide array of mapping features, while the xmlToJson flow is more focused on a few 
-features specific to XML.
+The jsonToJson and jsonToXml flows demonstrate a wide array of mapping features, while the xmlToJson and xmlToXml flows are more focused on a few
+features specific to XML like attributes and namespaces.
 
-Both flows can be run from the "Flows" page in QuickStart. Each will ingest one document into the staging database and then
+All the flows can be run from the "Flows" page in QuickStart. Each will ingest one document into the staging database and then
```
rjrudin: Can you add one note here, which I should have added originally? Where it says "Each of the mapping expressions is then described below. ", add - "Please run each of the flows first to ensure that the mapping step in each flow can be tested."

rjrudin: Can you add one note here, which I should have added originally? Where it says "Each of the mapping expressions is then described below. ", add - "Please run each of the flows first to ensure that the mapping step in each flow can be tested."

PR #3329, opened 2019-11-27
DHFPROD-3584: test for custom ingestion/mapping/mastering

bsrikan:
…pping/mastering

Updated build.gradle with a junit jupiter version that supports TestMethodOrderer api.


bsrikan:
This is strange. 2 of the 3 tests I added are failing in 5.1 but they pass locally against 5.1 as well as in the Jenkins develop branch tests

PR #3331, opened 2019-11-27
DHFPROD-3812: Fixing hubInit

rjrudin:
This avoids problems when the user runs hubInit and no username/password is set yet.

### DIFF
--- a/ml-data-hub-plugin/src/main/groovy/com/marklogic/gradle/DataHubPlugin.groovy
+++ b/ml-data-hub-plugin/src/main/groovy/com/marklogic/gradle/DataHubPlugin.groovy
@@ -44,7 +44,6 @@ import org.gradle.api.Project
 import org.slf4j.Logger
 import org.slf4j.LoggerFactory
 import org.springframework.boot.autoconfigure.EnableAutoConfiguration
-import org.springframework.context.ApplicationContext
 import org.springframework.context.annotation.AnnotationConfigApplicationContext
 
 @EnableAutoConfiguration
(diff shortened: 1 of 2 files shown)


marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/util/ModuleWatchingConsumer.java
```diff
@@ -15,15 +18,36 @@
...
+        this.commandContext = commandContext;
+        this.hubConfig = hubConfig;
+        this.versions = versions;
+    }
+
     @Override
     public void accept(Set<Resource> resources) {
+        if (generateFunctionMetadataCommand == null && hubConfig != null && versions != null) {
+            generateFunctionMetadataCommand = new GenerateFunctionMetadataCommand(hubConfig.newModulesDbClient(), versions);
+        }
```
srinathgit: 1. This 'generateFunctionMetadataCommand' object is not managed by spring container whereas all other objects are. Will there be any issues because of that ?

ml-data-hub-plugin/src/main/groovy/com/marklogic/gradle/DataHubPlugin.groovy
```diff
@@ -140,7 +140,7 @@ class DataHubPlugin implements Plugin<Project> {
         WatchTask watchTask = project.tasks.getByName("mlWatch")
         CommandContext commandContext = new CommandContext(hubConfig.getAppConfig(), hubConfig.getManageClient(), hubConfig.getAdminManager())
         Versions versions = ((ApplicationContext)project.property("dataHubApplicationContext")).getBean(Versions.class)
-        watchTask.onModulesLoaded = new ModuleWatchingConsumer(commandContext, new GenerateFunctionMetadataCommand(hubConfig.newModulesDbClient(), versions))
```
srinathgit: 1. Instead of these changes, can 'GenerateFunctionMetadataCommand'   instance be obtained from the bean and used ? Something like     
GenerateFunctionMetadataCommand generateFunctionMetadataCommand = ((ApplicationContext)project.property("dataHubApplicationContext")).getBean(GenerateFunctionMetadataCommand.class)

PR #3335, opened 2019-11-27
DHFPROD-3769: Set appropriate URI extension when mapping across formats

ryanjdew:


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mapping/default/main.sjs
```diff
@@ -87,6 +87,10 @@ function main(content, options) {
 
   //now let's make our attachments, if it's xml, it'll be passed as string
   instance['$attachments'] = doc;
+  // fix the document URI if the format changes
+  if (content.uri && !content.uri.endsWith(outputFormat)) {
```
rjrudin: Is outputFormat known to be lowercase here? I don't know if it matters if "XML" or "JSON" is used?

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mapping/entity-services/main.sjs
```diff
@@ -164,7 +168,14 @@ function buildEnvelope(entityInfo, doc, instance, outputFormat, options) {
       if (attachments instanceof Document && attachments.documentFormat === 'JSON') {
         nb.addText(xdmp.quote(attachments));
       } else {
-        nb.addNode(attachments);
+        // can get sequence of nodes in JSON to XML scenario
```
rjrudin: Is this being included to address a separate issue?

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mapping/entity-services/main.sjs
```diff
@@ -70,6 +70,10 @@ function main(content, options) {
     datahub.debug.log({message: e, type: 'error'});
     throw Error(e);
   }
+  // fix the document URI if the format changes
+  if (content.uri && !content.uri.endsWith(outputFormat)) {
```
rjrudin: Because this is duplicated, I recommend adding a function to a library that accepts a URI and an output format and if the "if" condition is met, a modified URI is returned; else, the original URI is returned. That function can then be covered via very small unit tests to address all the scenarios we can think of.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mapping/entity-services/main.sjs
ryanjdew: There isn't a separate issue filed that I've seen, but the unit test would fail without either this change or manipulating the JSON object to have a single root property.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mapping/entity-services/main.sjs
ryanjdew: Done. Moved it to a function in flowUtils and added unit test specifically for it.

bsrikan:
Fails to run both json flows in the mapping-example project

image

xml-json seems ok

xml-xml doesnt extract all the array items, only first item is returned by both Test and Run button 

<names datatype="array">
	<Name><first xsi:type="xs:string">Jane</first><prefix xsi:type="xs:string">Mrs</prefix><middle xsi:type="xs:string">Joan</middle><last xsi:type="xs:string">Smith</last></Name>
      </names>

rjrudin:
@bsrikan I'm double-checking, but that's not related to this - I'm guessing that project just needs to be updated based on the change to parseDate and parseDateTime.

rjrudin:
@bsrikan Yup, that's the problem - I'm submitting a separate PR related to 3620 to fix the mapping-example project.

SameeraPriyathamTadikonda:
Merging it manually as the test passes locally.

PR #3336, opened 2019-11-27
DHFPROD-3769: Set appropriate URI extension when mapping across formats

ryanjdew:


SameeraPriyathamTadikonda:
Merging it manually as the test passes locally

PR #3344, opened 2019-12-02
DHFPROD-3814: Removing sorting feature in columns of Entity table

wooldridge:


bsrikan:
@wooldridge I think we need the sort arrow removed from mapping table as well. Name and Value column in mapping-ui component have sort currently.

wooldridge:
@bsrikan the sorting features on Source Data side work fine. Sorting on the Entity did not work, implementing sorting there is more challenging because of the recursive structure of the table. For this issue we are removing the non-functioning sort on the Entity side and leaving the Source Data sorting, which has always worked fine. Unless @sbayatpur thinks otherwise.

sbayatpur:
Yes, the sorting feature in the source table should stay, but the sorting feature in the entity table should be removed

PR #3351, opened 2019-12-03
DHFPROD-3617:Insert artifacts with new roles

srinathgit:
Insert artifacts with new roles.

@rjrudin ,
I don't think anything needs to be done for upgrade scenario as hubInit would initialize the new roles and deploying them would take care of writing them with appropriate roles.

### DIFF
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/LoadHubArtifactsCommand.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/LoadHubArtifactsCommand.java
@@ -40,7 +40,7 @@ import java.io.InputStream;
 import java.util.Date;
 
 /**
- * Loads user artifacts like mappings and entities. This will be deployed after triggers
+ * Loads hub artifacts (ootb flows and step defs). This will be deployed after triggers
  */
 @Component
 public class LoadHubArtifactsCommand extends AbstractCommand {
@@ -100,7 +100,7 @@ public class LoadHubArtifactsCommand extends AbstractCommand {
                 InputStream inputStream = r.getInputStream();
                 StringHandle handle = new StringHandle(IOUtils.toString(inputStream));
                 inputStream.close();
-                DocumentMetadataHandle meta = buildDocumentMetadata("http://marklogic.com/data-hub/flow");
+                DocumentMetadataHandle meta = buildMetadata(hubConfig.getFlowPermissions(), "http://marklogic.com/data-hub/flow");
 
                 if (forceLoad || propertiesModuleManager.hasFileBeenModifiedSinceLastLoaded(flowFile)) {
                     stagingFlowDocumentWriteSet.add("/flows/" + flowFile.getName(), meta, handle);
@@ -117,7 +117,7 @@ public class LoadHubArtifactsCommand extends AbstractCommand {
                 InputStream inputStream = r.getInputStream();
                 StringHandle handle = new StringHandle(IOUtils.toString(inputStream));
                 inputStream.close();
-                DocumentMetadataHandle meta = buildDocumentMetadata("http://marklogic.com/data-hub/step-definition");
+                DocumentMetadataHandle meta = buildMetadata(hubConfig.getStepDefinitionPermissions(), "http://marklogic.com/data-hub/step-definition");
 
                 if (forceLoad || propertiesModuleManager.hasFileBeenModifiedSinceLastLoaded(flowFile)) {
                     stagingStepDocumentWriteSet.add("/step-definitions/" + flowFile.getParentFile().getParentFile().getName() + "/" + flowFile.getParentFile().getName() + "/" + flowFile.getName(), meta, handle);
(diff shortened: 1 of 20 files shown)


marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/LoadHubArtifactsCommand.java
```diff
@@ -142,17 +142,28 @@ public void execute(CommandContext context) {
...
+        String permissions = null;
+        switch (collection) {
+            case "http://marklogic.com/data-hub/flow":
+                permissions = config.getFlowPermissions();
+                break;
+            case "http://marklogic.com/data-hub/step-definition":
+                permissions = config.getStepDefinitionPermissions();
+                break;
+            default:
+                throw new RuntimeException("The provided collection name for the artifact is incorrect.");
```
rjrudin: I don't think this is necessary. If neither is a match, it's a programming error - just return null. An automated test should verify that we get the expected permissions for flows and step definitions.

marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/LoadUserArtifactsCommand.java
```diff
@@ -281,8 +261,33 @@ private void executeWalk(
...
+                permissions = config.getMappingPermissions();
+                break;
+            case "http://marklogic.com/data-hub/flow":
+                permissions = config.getFlowPermissions();
+                break;
+            case "http://marklogic.com/data-hub/step-definition":
+                permissions = config.getStepDefinitionPermissions();
+                break;
+            default:
+                throw new RuntimeException("The provided collection name for the artifact is incorrect.");
```
rjrudin: Same thing, just keep permissions as null.

marklogic-data-hub/src/main/resources/hub-internal-config/security/roles/data-hub-operator.json
```diff
@@ -5,7 +5,13 @@
     "rest-extension-user",
     "rest-reader",
     "rest-writer",
-    "data-hub-job-reader"
+    "data-hub-job-reader",
```
rjrudin: The operator should only have reader roles

srinathgit: Changed the roles so that operator has reader roles and developer has writer roles

marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/LoadHubArtifactsCommand.java
```diff
@@ -142,17 +142,28 @@ public void execute(CommandContext context) {
...
-        documentPermissionsParser.parsePermissions(hubConfig.getModulePermissions(), meta.getPermissions());
+        String permissions = null;
+        switch (collection) {
+            case "http://marklogic.com/data-hub/flow":
+                permissions = config.getFlowPermissions();
+                break;
+            case "http://marklogic.com/data-hub/step-definition":
+                permissions = config.getStepDefinitionPermissions();
+                break;
+            default:
```
rjrudin: It looks like HubConfig object is only need to get permissions. So to avoid a switch statement altogether (and the questions of what to return for "default"), I recommend just passing in the permissions string, which the caller knows, instead of a HubConfig.

marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/LoadUserArtifactsCommand.java
```diff
@@ -147,7 +146,7 @@ public FileVisitResult preVisitDirectory(Path dir, BasicFileAttributes attrs) th
                             entityDefModulesFinder,
                             propertiesModuleManager,
                             entityResourceToURI,
-                            buildMetadataForEntityModels(hubConfig),
+                            buildMetadata(hubConfig, "http://marklogic.com/entity-services/models"),
```
rjrudin: So for example, this would be buildMetadata(hubConfig.getEntityModelPermissions(), "the collection")

marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/LoadUserArtifactsCommand.java
```diff
@@ -281,8 +261,33 @@ private void executeWalk(
...
+     *
+     * @param collection
+     * @param config
+     * @return
+     */
+    protected DocumentMetadataHandle buildMetadata(HubConfig config, String collection) {
         DocumentMetadataHandle meta = new DocumentMetadataHandle();
+        String permissions = null;
+        switch (collection) {
+            case "http://marklogic.com/entity-services/models":
```
rjrudin: Same thing - avoid a switch statement by having the caller pass in the permissions string instead of a HubConfig.

marklogic-data-hub/src/main/resources/hub-internal-config/security/roles/data-hub-developer.json
```diff
@@ -6,7 +6,12 @@
     "manage-user",
     "ps-user",
     "rest-admin",
-    "tde-admin"
+    "tde-admin",
```
rjrudin: @srinathgit I just added the requirement to 3617 for flow-developer-role to get these roles as well, and for flow-operator-role to get all the reader roles.

srinathgit: Added reader roles to "flow-operator-role" and writer roles to "flow-developer-role"

marklogic-data-hub/src/main/resources/hub-internal-config/security/roles/flow-operator-role.json
```diff
@@ -4,7 +4,12 @@
   "role": [
     "rest-reader",
     "rest-extension-user",
-    "rest-writer"
+    "rest-writer",
```
rjrudin: Thanks Srinath!

@aebadirad flow-operator-role has "rest-writer" today - does that it mean should have all the new *-writer roles as well - i.e. the ones inherited above by flow-developer-role?

PR #3365, opened 2019-12-05
DHFPROD-3092: Set accurate counts when writing documents from step fail

ryanjdew:


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/flow.sjs
```diff
@@ -223,7 +223,28 @@ class Flow {
     let writeTransactionInfo = {};
     //let's update our jobdoc now
     if (!combinedOptions.noWrite) {
-      writeTransactionInfo = this.datahub.hubUtils.writeDocuments(this.writeQueue, 'xdmp.defaultPermissions()', collections, this.globalContext.targetDatabase);
+      try {
+        writeTransactionInfo = this.datahub.hubUtils.writeDocuments(this.writeQueue, 'xdmp.defaultPermissions()', collections, this.globalContext.targetDatabase);
+      } catch (e) {
+        this.globalContext.completedItems = [];
```
rjrudin: Thoughts on moving the contents of the catch block to a separate function? I'd like to see that as a general move for DHF code - runFlow is about 140 LOC, and runStep is about 90 LOC, and understanding the context of several lines of code can be a bit difficult. Function names can improve the readability of our code too. 

For example, this could be handleErrorFromWritingWriteQueue (a somewhat awkward name, but consistent with the user of "write" documents in the "writeQueue").

ryanjdew: Refactored the error handling logic with the latest changes.

PR #3367, opened 2019-12-06
DHFPROD-3874:Setting document permissions correctly during ingestion

srinathgit:
Setting document permissions correctly during ingestion

### DIFF
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/step/impl/WriteStepRunner.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/step/impl/WriteStepRunner.java
@@ -23,6 +23,8 @@ import com.fasterxml.jackson.dataformat.csv.CsvSchema;
 import com.marklogic.client.DatabaseClient;
 import com.marklogic.client.datamovement.*;
 import com.marklogic.client.document.ServerTransform;
+import com.marklogic.client.ext.util.DefaultDocumentPermissionsParser;
+import com.marklogic.client.ext.util.DocumentPermissionsParser;
 import com.marklogic.client.io.DocumentMetadataHandle;
 import com.marklogic.client.io.Format;
 import com.marklogic.client.io.InputStreamHandle;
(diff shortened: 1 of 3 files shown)


marklogic-data-hub/src/test/java/com/marklogic/hub/flow/FlowRunnerTest.java
```diff
@@ -99,6 +101,11 @@ public void testRunFlow(){
         Assertions.assertTrue(getDocCount(HubConfig.DEFAULT_FINAL_NAME, "json-map") == 1);
         Assertions.assertTrue(getDocCount(HubConfig.DEFAULT_FINAL_NAME, "xml-map") == 1);
         Assertions.assertTrue(JobStatus.FINISHED.toString().equalsIgnoreCase(resp.getJobStatus()));
+        XMLDocumentManager docMgr = stagingClient.newXMLDocumentManager();
```
rjrudin: The change looks good, but can you change this into a unit test that hits applyPermissions directly? You'll need to make that method protected, which I think is fine for the sake of unit testing.

I say this because I'm about to dig into why this test method is failing in Jenkins but nowhere else, so I don't want to see anything added to this test as something is fishy with it already. A unit test is much faster to run as well.

srinathgit: @rjrudin ,
I can change it to a unit test but this PR is for 5.1-develop where all tests are passing. I haven't made a PR for develop.

marklogic-data-hub/src/main/java/com/marklogic/hub/step/impl/WriteStepRunner.java
```diff
@@ -709,6 +709,7 @@ private String generateAndEncodeURI(String path) throws  URISyntaxException {
 
     private void applyPermissions(String permissions, DocumentMetadataHandle metadataHandle) {
         String[] perms = permissions.split(",");
+        Map<String, List<DocumentMetadataHandle.Capability>> permMap = new HashMap<>();
```
rjrudin: @srinathgit I just remembered that there's a class in ml-javaclient-util that does this already - DefaultDocumentPermissionsParser.  It's being used in DHF already. You give it the permissions string and the DocumentPermissions object from DocumentMetadataHandle, and it handles the rest. I recommend using that.

aebadirad: Yes, please reuse this method, it keeps us static in how we handle the permissions string across the datahub.

PR #3374, opened 2019-12-07
DHFPROD-3618:Insert/update job/batch docs with permissions using new roles

srinathgit:
Insert/update job/batch docs with permissions using new roles


srinathgit:
@rjrudin ,
yeah, I have changed the couple of references to 'datahub.jobs.jobsPermissions' to 'datahub.jobs.jobPermissionsScript'

PR #3380, opened 2019-12-10
DHFPROD-3619:Set default permissions to step/step defs based on new roles

srinathgit:
Set default permissions to step/step defs based on new roles


marklogic-data-hub/src/test/java/com/marklogic/hub/HubTestBase.java
```diff
@@ -1195,8 +1195,9 @@ protected String getTimezoneString() {
     protected void setupProjectForRunningTestFlow() {
         basicSetup();
         getDataHubAdminConfig();
-        clearDatabases(HubConfig.DEFAULT_STAGING_NAME, HubConfig.DEFAULT_FINAL_NAME, HubConfig.DEFAULT_JOB_NAME);
-
+        for(String database : new String[] {HubConfig.DEFAULT_STAGING_NAME, HubConfig.DEFAULT_FINAL_NAME, HubConfig.DEFAULT_JOB_NAME}) {
+            new DatabaseManager(adminHubConfig.getManageClient()).clearDatabase(database,true);
```
rjrudin: This definitely makes sense, I'll be curious if truly clearing a database causes any tests to fail. If so, I think those tests would need updating.

srinathgit: This was taking a lot more time, so I changed the query to not obtain the prov collection uris

PR #3383, opened 2019-12-10
DHFPROD-3510: Use data-hub-operator and data-hub-developer roles

ryanjdew:


marklogic-data-hub/gradle.properties
```diff
@@ -59,7 +59,9 @@ mlFlowDeveloperRole=flow-developer-role
 mlFlowDeveloperUserName=flow-developer
 mlFlowDeveloperPassword=password
 mlDataHubAdminRole=data-hub-admin-role
+mlDataHubOperatorRole=data-hub-operator
+mlDataHubDeveloperRole=data-hub-developer
 
-mlModulePermissions=rest-reader,read,rest-writer,insert,rest-writer,update,rest-extension-user,execute,flow-developer-role,read,flow-developer-role,execute,flow-developer-role,insert,flow-operator-role,read,flow-operator-role,execute
+mlModulePermissions=rest-reader,read,rest-writer,insert,rest-writer,update,rest-extension-user,execute,flow-developer-role,read,flow-developer-role,execute,flow-developer-role,insert,flow-operator-role,read,flow-operator-role,execute,data-hub-developer,read,data-hub-developer,execute,data-hub-developer,insert,data-hub-operator,read,data-hub-operator,execute
```
rjrudin: There's a separate story for modifying mlModulePermissions - it actually specifies setting the value to just the new data-hub roles and not using any of the existing ones.

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/HubConfigImpl.java
```diff
@@ -155,6 +155,8 @@
     private String flowDeveloperUserName;
 
     private String dataHubAdminRoleName;
+    private String dataHubOperatorRoleName;
```
rjrudin: We actually don't want these properties - we don't want to provide any ability to override these role names, as they're the agreed upon names that DHS will be expecting to use as well.

marklogic-data-hub/src/main/resources/hub-internal-config/triggers/ml-dh-entity-create.json
```diff
@@ -30,6 +30,14 @@
     {
       "role-name": "%%mlFlowOperatorRole%%",
       "capability": "read"
+    },
+    {
+      "role-name": "%%mlDataHubDeveloperRole%%",
```
rjrudin: These should be hard-coded to the role name, we don't want to allow for any overrides.

marklogic-data-hub/src/main/resources/ml-modules/root/com.marklogic.hub/config.sjs
```diff
@@ -8,5 +8,7 @@ module.exports = {
   HUBLOGLEVEL: "%%mlHubLogLevel%%",
   FLOWOPERATORROLE: "%%mlFlowOperatorRole%%",
   FLOWDEVELOPERROLE: "%%mlFlowDeveloperRole%%",
+  DATAHUBOPERATORROLE: "%%mlDataHubOperatorRole%%",
```
rjrudin: I think it's better not to define these here, as it implies that they could vary per environment. But we don't want them to vary. I think constants defined elsewhere is fine to reduce duplication of the string value, but I wouldn't consider that necessary either. These role names are intended to be written in stone and never to change.

ryanjdew: moved to the constants library.

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/HubConfigImpl.java
ryanjdew: now hardcoded

marklogic-data-hub/src/main/resources/hub-internal-config/triggers/ml-dh-entity-create.json
ryanjdew: now hardcoded

marklogic-data-hub/src/main/resources/scaffolding/gradle_properties
```diff
@@ -1,127 +1,127 @@
...
-# When "mlIsHostLoadBalancer" is set to "true", the value specified for "mlHost" will be used as the load balancer.
-# You do not need to explicitly set the value of "mlLoadBalancerHosts" but if you do it must match the value of the property "mlHost"
-# mlLoadBalancerHosts=your-load-balancer-hostname
-
-# Default module permissions which allow flow-operator-role to execute flows
-mlModulePermissions=rest-reader,read,rest-writer,insert,rest-writer,update,rest-extension-user,execute,%%mlFlowDeveloperRole%%,read,%%mlFlowDeveloperRole%%,execute,%%mlFlowDeveloperRole%%,insert,%%mlFlowOperatorRole%%,read,%%mlFlowOperatorRole%%,execute
-
-# If DHF is running in a provisioned environment please specify it here
-# mlIsProvisionedEnvironment=false
+# These settings are used by Data Hub when
```
rjrudin: Did anything change in this file? I can't tell because the entire file is marked as different.

marklogic-data-hub/src/main/resources/ml-modules/root/com.marklogic.hub/config.xqy
```diff
@@ -13,3 +13,5 @@ declare variable $HUB-VERSION := "%%mlHubVersion%%";
 declare variable $HUB-LOG-LEVEL := "%%mlHubLogLevel%%";
 declare variable $FLOW-OPERATOR-ROLE := "%%mlFlowOperatorRole%%";
 declare variable $FLOW-DEVELOPER-ROLE := "%%mlFlowDeveloperRole%%";
+declare variable $DATA-HUB-OPERATOR-ROLE := "data-hub-operator";
```
rjrudin: I think these can be removed too, right?

PR #3391, opened 2019-12-12
DHFPROD-3504:Configure DH resources requiring new granular privileges

srinathgit:
Configure DH resources requiring new granular privileges.

@SameeraPriyathamTadikonda ,
This PR should be run on 10.0-nightly, this will fail on 10.0-2.1 against which all PRs usually run.

### DIFF
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/CreateGranularPrivilegesCommand.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/CreateGranularPrivilegesCommand.java
@@ -7,9 +7,7 @@ import com.marklogic.appdeployer.command.UndoableCommand;
 import com.marklogic.hub.DatabaseKind;
 import com.marklogic.hub.HubConfig;
 import com.marklogic.mgmt.api.security.Privilege;
-import com.marklogic.mgmt.resource.databases.DatabaseManager;
 import com.marklogic.mgmt.resource.security.PrivilegeManager;
-import com.marklogic.rest.util.ResourcesFragment;
 
 import java.util.Arrays;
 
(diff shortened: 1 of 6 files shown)


marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/CreateGranularPrivilegesCommand.java
```diff
@@ -44,44 +42,84 @@ public Integer getUndoSortOrder() {
...
-        final String finalDbId = databasesXml.getIdForNameOrId(finalDbName);
-        final String stagingDbId = databasesXml.getIdForNameOrId(stagingDbName);
+        final String finalTriggersDbName = hubConfig.getDbName(DatabaseKind.FINAL_TRIGGERS);
+        final String stagingTriggersDbName = hubConfig.getDbName(DatabaseKind.STAGING_TRIGGERS);
 
         PrivilegeManager mgr = new PrivilegeManager(context.getManageClient());
         Privilege p = new Privilege(null, "admin-database-clear-" + finalDbName);
         p.setKind("execute");
-        p.setAction("http://marklogic.com/xdmp/privileges/admin/database/clear/" + finalDbId);
+        p.setAction("http://marklogic.com/xdmp/privileges/admin/database/clear/$$database-id(" + finalDbName+ ")");
```
rjrudin: What just the database ID by itself not working?

marklogic-data-hub/src/test/java/com/marklogic/hub/security/DataHubDeveloperTest.java
```diff
@@ -49,17 +51,33 @@ public void task9ConfigureBitemporal() throws IOException {
             "  }\n" +
             "}";
 
-        new TemporalAxesManager(userWithRoleBeingTestedClient, FINAL_DB).save(temporalAxis);
+        TemporalAxesManager mgr = new TemporalAxesManager(userWithRoleBeingTestedClient, FINAL_DB);
```
rjrudin: For these tests that will require 10.0-3, can we adopt the same thing we did for mapping tests, where we require a certain nightly version or greater, or 10.0-3 or greater? Then @SameeraPriyathamTadikonda  doesn't need to do anything additional.

marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/CreateGranularPrivilegesCommand.java
srinathgit: It does work, I saw this comment "Once RMA supports pseudo functions - https://bugtrack.marklogic.com/53540 - then we won't need to get the	 resource IDs here and can use resource names instead" and the bug had been shipped, so changed them to use pseudo functions

marklogic-data-hub/src/test/java/com/marklogic/hub/security/DataHubDeveloperTest.java
srinathgit: Ok, will do

marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/CreateGranularPrivilegesCommand.java
rjrudin: Oh interesting - since that's the case, we could externalize this config into files, as we don't need code anymore to look up the resource IDs. But I think we can hold off on that for now.

PR #3393, opened 2019-12-13
DHFPROD-3617: More tests for DHFPROD-3617 and DHFPROD-3618

bsrikan:


marklogic-data-hub/src/test/java/com/marklogic/hub/core/HubProjectTest.java
```diff
@@ -162,6 +168,23 @@ public void testInit() throws IOException {
...
 
+        //per DHFPROD-3617,DHFPROD-3618 following properties shouldn't be there in gradle.properties after hubInit is run. Users can adjust these if needed
+        assertNull(props.getProperty("mlEntityPermissions"));
+        assertNull(props.getProperty("mlFlowPermissions"));
+        assertNull(props.getProperty("mlMappingPermissions"));
+        assertNull(props.getProperty("mlStepDefinitionPermissions"));
+        assertNull(props.getProperty("mlJobPermissions"));
+
+        //per DHFPROD-3617,DHFPROD-3618 below roles will only be used for permissions so they will not have privileges or inherited any roles
+        ObjectMapper mapper = new ObjectMapper();
```
rjrudin: The text I wrote in those tickets was just to make it clear that the roles, as defined today, don't have any reason to inherit any other roles or privileges. But that's not a hard requirement. If six months from now, we need to have any of these roles inherit other roles or privileges, that's perfectly fine. 

So I recommend removing these assertions, as they could cause confusion for someone in the future who then has to wonder - Why am I not allowed to add any roles or privileges to these roles?

bsrikan: It is a requirement now right and if it changes in the future the tests can be updated. The comment points to the tickets responsible for this behavior.
I can remove these assertions, but if they are relevant now, is there any harm in asserting them?

bsrikan: I have made the changes requested, but I dont understand the part about "confusion in the future"

PR #3394, opened 2019-12-13
DHFPROD-3729: data-hub-security-admin can create privileges

rjrudin:


marklogic-data-hub/src/test/java/com/marklogic/hub/security/DataHubSecurityAdminTest.java
```diff
@@ -95,4 +98,114 @@ public void task5CannotCreateCustomRoleInheritingExistingRole() {
             new Role(adminUserApi, roleName).delete();
         }
     }
+
```
srinathgit: All these tests would fail in versions < 10.0-3 . I think these tests have to be disabled based on HubTestBase. isVersionCompatibleWithGranularPrivilege() like PR #3391 .

marklogic-data-hub/src/main/resources/hub-internal-config/security/privileges/data-hub-create-custom-privilege.json
```diff
@@ -0,0 +1,5 @@
+{
```
srinathgit: Typo: The issue # is DHFPROD-3729 (not 3279, has to be changed in the commit )

marklogic-data-hub/src/test/java/com/marklogic/hub/security/DataHubSecurityAdminTest.java
rjrudin: Good catch, I copied that method over from the 3504 PR.

PR #3395, opened 2019-12-13
DHFPROD-3522:Initialize new DH project without data-hub-admin-role

srinathgit:
Initialize new DH project without data-hub-admin-role

### DIFF
--- a/marklogic-data-hub/src/main/resources/hub-internal-config/security/roles/data-hub-admin-role.json
+++ b/marklogic-data-hub/src/main/resources/hub-internal-config/security/roles/data-hub-admin-role.json
@@ -1,6 +1,6 @@
 {
-  "role-name": "%%mlDataHubAdminRole%%",
-  "description": "A role can deploy modules to a data hub instance.",
+  "role-name": "data-hub-admin-role",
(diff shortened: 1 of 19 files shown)


srinathgit:
@ryanjdew ,
Thanks Ryan. The story mentions that data-hub-admin-role.json has to be retained,but that would mean that there will be reference to %%mlDataHubAdminRole%% in a bunch of places. I have added a comment in the story asking if it is fine to remove the role altogether

marklogic-data-hub/src/main/resources/hub-internal-config/security/roles/data-hub-admin-role.json
```diff
@@ -1,5 +1,5 @@
 {
-  "role-name": "%%mlDataHubAdminRole%%",
+  "role-name": "data-hub-admin-role",
   "description": "A role can deploy modules to a data hub instance.",
```
rjrudin: Can you change this description as noted in the JIRA ticket too? This current description isn't really correct to start with, and I think we now want it to note that the role is deprecated and shouldn't be used.

PR #3396, opened 2019-12-14
DHFPROD-3672: Escape XML characters in XPath expressions when validating

ryanjdew:


web/src/main/ui/app/components/flows-new/edit-flow/mapping/ui/entity-table-ui.component.ts
```diff
@@ -211,8 +211,8 @@ export class EntityTableUiComponent implements OnChanges {
 
   insertField(fieldName, index, prop) {
     let field = fieldName;//.replace(/[^\/]+\:/g, '');
-    if(String(field).includes(" ")){
-      field = "*[local-name(.)='" + field + "']";
+    if (/(&|>|<|'|"|}|{|\s)/g.test(String(field))) {
```
rjrudin: This is optional, no action needed, but when there's a pattern like this, I think it's usually worth introducing a local variable with a name that documents the intent of the pattern - e.g. "let hasXmlPattern = ...". The intent is implied by the call to escapeXml on the next line, but I always like the explicit self-documenting nature of a good variable name.

PR #3397, opened 2019-12-14
DHFPROD-3672: Escape XML characters in XPath expressions when validating

ryanjdew:


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mapping/entity-services/lib.sjs
```diff
@@ -307,7 +307,7 @@ function validatePropertyMapping(fullMapping, propertyName, sourcedFrom) {
 function runMapping(mapping, uri, propMapping={"targetEntityType":mapping.targetEntityType, "namespaces": mapping.namespaces,"properties": {}}, paths=['properties']) {
   Object.keys(mapping.properties).forEach(propertyName => {
     let mappedProperty = mapping.properties[propertyName];
-    let sourcedFrom = mappedProperty.sourcedFrom;
+    let sourcedFrom = escapeXML(mappedProperty.sourcedFrom);
```
srinathgit: Since xml characters have already been escaped in the mapping, isn't this redundant ?

ryanjdew: The front-end code only escapes the string portion in the XPath expression and then this escapes the whole XPath expression to go into the XML. So ultimately the string portion of that expression gets escaped twice when it gets into the mapping XML, but that is needed and intentional.

PR #3399, opened 2019-12-16
DHFPROD-3943: Set the right undo sort order for 'DeployDatabaseFieldCommand'

srinathgit:
Set the right undo sort order for 'DeployDatabaseFieldCommand'

### DIFF
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/DeployDatabaseFieldCommand.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/DeployDatabaseFieldCommand.java
@@ -61,6 +61,13 @@ public class DeployDatabaseFieldCommand extends AbstractResourceCommand {
         return addExistingFieldsAndRangeFieldIndexes(payload, new DatabaseManager(context.getManageClient()));
     }
 
+    @Override
+    public void undo(CommandContext context) {
+        //no-op as deleting database will delete all the fields and indexes associated with it.
+        logger.info("No action required on undeploy, as the command for deleting databases on undeploy will also delete " +
+            "the fields and indexes created by this command.");
+    }
+
     protected String addExistingFieldsAndRangeFieldIndexes(String payload, ResourceManager dbManager) {
         Fragment newProps = new Fragment(payload);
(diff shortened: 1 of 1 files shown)


marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/DeployDatabaseFieldCommand.java
```diff
@@ -43,6 +43,7 @@
 
     public DeployDatabaseFieldCommand() {
         setExecuteSortOrder(SortOrderConstants.DEPLOY_OTHER_DATABASES + 1);
+        setUndoSortOrder(SortOrderConstants.DELETE_OTHER_DATABASES - 1);
```
rjrudin: @srinathgit This command actually shouldn't do anything on undeploy. Because it's extending AbstractResourceCommand (which simplified the implementation a great deal), it implements UndoableCommand, but the right fix is to override the "undo" method and not do anything. The command for undeploying databases will then handle undeploying.

rjrudin: Also, can you submit a PR for fixing this on 5.1-develop as well?

marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/DeployDatabaseFieldCommand.java
```diff
@@ -61,6 +61,11 @@ protected String adjustPayloadBeforeSavingResource(CommandContext context, File
         return addExistingFieldsAndRangeFieldIndexes(payload, new DatabaseManager(context.getManageClient()));
     }
 
+    @Override
+    public void undo(CommandContext context) {
+        //no-op
```
rjrudin: I should have mentioned this before - I recommend a log statement here just to answer "why" this method is overridden to not do anything - e.g. logger.info("No action required on undeploy, as the command for deleting databases on undeploy will also delete the fields and indexes created by this command.");

aebadirad: A comment too, in the code, would be nice so we know why that is a no-op in case we need to change it for whatever reason. (or just a link to the jira ticket #)

PR #3407, opened 2019-12-18
DHFPROD-3112: handle deletion of job uris

bsrikan:


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/4/impl/job-lib.xqy
```diff
@@ -72,9 +72,11 @@ declare private function job:delete-jobs($job-ids as xs:string*) as map:map
   let $results := map:map()
   let $_ :=
     for $id in $job-ids
-    let $uri := cts:uris((), (),
+    let $uris := cts:uris((), (),
```
rjrudin: This will be more efficient if a single cts:uris query is done for all job-ids - i.e. 1 query instead of N queries, where N = count($job-ids). 

However, that impacts the if/else block ahead. Though I think that if/else block has a bug in it. It's saying - if the URI is available, then delete it and add the job-id to the "success key". Else add the job-id to the "failed-key". Doesn't it seem like URI should be added instead of job-id? In which case, doing a single cts:uris query won't impact the block.

bsrikan: I didnt change any of that and it has been around since 4.0 maybe.
We are trying to delete the batch documents as well associated with a job id. Earlier commit to 3112 had a bug and so this PR only fixes that. 
Having said that, I did try removing for loop and running cts:uris with "$job-ids" instead and adding URI in the if/else block.
Currently hubDeleteJobs output looks like below. I dont want to spend more time on this than I already have and if we are OK to have "jobs deleted" show URIs instead of job ids, I will go ahead with the changes. Please confirm!

```
Deleting jobs: 251d13a4-1c08-4629-94fc-cde186f60e4d,3f0a0ffd-4d8b-4a73-96e4-5bd246ca346c
JobDeleteResponse:
        total jobs deleted: 4
        total errors: 0
        jobs deleted: [251d13a4-1c08-4629-94fc-cde186f60e4d, 251d13a4-1c08-4629-94fc-cde186f60e4d, 3f0a0ffd-4d8b-4a73-96e4-5bd246ca346c, 3f0a0ffd-4d8b-4a73-96e4-5bd246ca346c]
        total traces deleted: 0
        errors: []
```

rjrudin: Okay - it's fine to leave the for-loop as-is then.

ml-data-hub-plugin/src/test/resources/job-test/batch2.json
```diff
@@ -0,0 +1,14 @@
+{
+  "batch": {
+    "jobId": "10584668255644629399",
+    "batchId": "11345653515268525918",
+    "processor": "Mapping",
+    "stepNumber": "1",
+    "batchStatus": "failed",
+    "timeStarted": "2019-02-05T21:39:37.284127-08:00",
+    "timeEnded": "N/A",
```
rjrudin: Just a note - we really shouldn't be putting "N/A" here, as there's a range index on this field. If we don't have a value, the right thing to do is not define the field. It's something we can clean up later, but in the future, I think it's better to not keep adding "N/A" in additional places.

PR #3408, opened 2019-12-18
DHFPROD-3112: handle deletion of job uris(5.1-develop)

bsrikan:


bsrikan:
For job-ids that are invalid, there is no errorCount returned. Possible inputs to "jobIds" parameter is 1. valid jobid(s)
2. invalid jobid(s)
3. valid + invalid job ids
looping over list of uris doesnt get into else loop to map:put failed key

bsrikan:
For invalid job-ids there is no uri returned so else loop is never invoked.

bsrikan:
Below code works. Can I get a review on this here in the comment?

```
  let $results := map:map()
  let $_ :=
    for $id in $job-ids
    let $uris := cts:uris((), (),
                cts:and-query(
                  ( cts:directory-query("/jobs/", "infinity"), cts:json-property-value-query("jobId", $id))))
    return if (fn:exists($uris) = fn:false()) then map:put($results, $FAILED-KEY, (map:get($results, $FAILED-KEY), $id))
      else
        for $uri in $uris
        return
        if (fn:doc-available($uri)) then (
          xdmp:document-delete($uri),
          map:put($results, $SUCCESS-KEY, (map:get($results, $SUCCESS-KEY), $id))
        )
        else ()
  return $results
```

bsrikan:
Thanks for the review on this Ryan and Rob. Have pushed the changes, hopefully the tests will pass and this can be merged.

PR #3417, opened 2019-12-30
DHFPROD-3984: Add data services for artifacts (load data configuration API)

ryanjdew:
Also adjusted the hub utils to invoke javascript modules, because I was finding that the way external variables were recognized in the evals was changing depending on whether called via mjs or sjs.

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/artifacts/core.mjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/artifacts/core.mjs
@@ -1,0 +1,129 @@
+/**
+ Copyright 2012-2019 MarkLogic Corporation
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+ http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+'use strict';
+
+import * as LoadData from './loadData';
+
+const DataHubSingleton = require('/data-hub/5/datahub-singleton.sjs');
+const dataHub = DataHubSingleton.instance();
+// define constants for caching expensive operations
+const cachedArtifacts = {};
+const registeredArtifactTypes = {
+    loadData: LoadData
+};
+
+export function getTypesInfo() {
+    const typesInfo = [];
+    for (const artifactType of Object.keys(registeredArtifactTypes)) {
+        if (registeredArtifactTypes.hasOwnProperty(artifactType)) {
+            const artifactLibrary = registeredArtifactTypes[artifactType];
+            typesInfo.push({
+                type: artifactType,
+                storageDatabases: artifactLibrary.getStorageDatabases(),
+                collections: artifactLibrary.getCollections(),
+                fileExtension: getArtifactFileExtension(artifactType),
+                directory: getArtifactDirectory(artifactType),
+                nameProperty: artifactLibrary.getNameProperty(),
+                versionProperty: artifactLibrary.getVersionProperty()
+            });
+        }
+    }
+    return typesInfo;
+}
+
+export function getArtifacts(artifactType) {
+    const artifacts = [];
+    const queriesForAnd = [];
+    const artifactLibrary =  getArtifactTypeLibrary(artifactType);
+    for (const coll of artifactLibrary.getCollections()) {
+        queriesForAnd.push(cts.collectionQuery(coll));
+    }
+    if (queriesForAnd.length) {
+        const results = cts.search(queriesForAnd.length > 1 ? cts.andQuery(queriesForAnd) : queriesForAnd[0]);
+        for (const result of results) {
+            artifacts.push(result.toObject());
+        }
+    }
+    return artifacts;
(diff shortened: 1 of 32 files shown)


web/src/main/java/com/marklogic/hub/web/service/ArtifactService.java
```diff
@@ -0,0 +1,36 @@
+package com.marklogic.hub.web.service;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.marklogic.hub.HubConfig;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Service;
+
+@Service
+public class ArtifactService {
```
rjrudin: Is there any reason to have this class other than similar classes were created in the past? To me, it seems sufficient for the web controller - or whatever class receives the HTTP request - to talk directly to  the DataService. And that controller can be responsible for instantiating the DS - i.e. it would depend on HubConfig, and then it would use that to create a new DS instance on the fly.

rjrudin: I think not having this web/service class also moves us away from everything being a singleton based on a single HubConfig instance. In the above pattern, the controller would obtain a HubConfig based on the current user (i.e. from the user's HTTP session) and then instantiate the appropriate DS on the fly.

web/src/main/java/com/marklogic/hub/web/web/LoadDataController.java
```diff
@@ -0,0 +1,68 @@
...
+
+    @Autowired
+    private MlRestErrorHandlerService mlRestErrorHandlerService;
+
+    private String artifactType = "loadData";
+
+    @RequestMapping(method = RequestMethod.GET)
+    @ResponseBody
+    public ResponseEntity<JsonNode> getLoadDataConfigurations() {
+        return mlRestErrorHandlerService.handleRestException(() ->
```
rjrudin: I think this is better handled via Spring MVC's HandlerExceptionResolver mechanism - here's one article on it - https://spring.io/blog/2013/11/01/exception-handling-in-spring-mvc#handlerexceptionresolver . You can register a global error handler for a FailedRequestException. Then controllers don't need to worry about this, they can just invoke the appropriate data service interface.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/artifacts/core.mjs
```diff
@@ -0,0 +1,129 @@
...
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+'use strict';
+
+import * as LoadData from './loadData';
+
+const DataHubSingleton = require('/data-hub/5/datahub-singleton.sjs');
```
srinathgit: Shouldn't we retain support for 9.x servers as well ? If so, wouldn't there be issues running mjs in 9.x servers ?

ryanjdew: This is for 5.3, which I believe we said that we'll be requiring 10.0-3

web/src/main/java/com/marklogic/hub/web/service/ArtifactService.java
ryanjdew: Thinned it down by calling the Data Services class directly. I did add an abstract artifact controller to simplify implementations for future artifacts.

web/src/main/java/com/marklogic/hub/web/web/LoadDataController.java
ryanjdew: Registered FailedRequestException with our custom exception handler instead.

web/src/main/java/com/marklogic/hub/web/web/AbstractArtifactController.java
```diff
@@ -0,0 +1,39 @@
...
+import com.marklogic.hub.dataservices.ArtifactService;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.http.HttpStatus;
+import org.springframework.http.ResponseEntity;
+
+public abstract class AbstractArtifactController {
+    @Autowired
+    private HubConfig hubConfig;
+
+    public ResponseEntity<JsonNode> getArtifacts() {
```
rjrudin: I recommend making all of these methods protected so that the reader knows they're only going to be used by subclasses.

web/src/main/java/com/marklogic/hub/web/web/LoadDataController.java
```diff
@@ -0,0 +1,51 @@
...
+import org.springframework.scheduling.annotation.EnableAsync;
+import org.springframework.stereotype.Controller;
+import org.springframework.web.bind.annotation.PathVariable;
+import org.springframework.web.bind.annotation.RequestBody;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestMethod;
+import org.springframework.web.bind.annotation.ResponseBody;
+
+@Controller
+@RequestMapping("/api/current-project/loadData")
```
rjrudin: @wooldridge @xnikhil08 Curious about your thoughts here. "current-project" doesn't seem meaningful to me, or rather it seems redundant. There's no concept of e.g. "mapping artifacts" that aren't part of a Data Hub application/project. So I recommend removing "current-project" from the APIs, which also seems like a nice way of defining a new set of APIs with no overlap with those used for QuickStart?

I'm also thinking "artifact" would be a useful term to include here - e.g. "/api/artifact/loadData" - thoughts on that?

PR #3418, opened 2019-12-30
DHFPROD-3782:Insert modules with datahub specific roles

srinathgit:
This PR addresses DHFPROD-3782, DHFPROD-4022 (each of them in specific commits)

### DIFF
--- a/marklogic-data-hub/src/main/resources/dhf-defaults.properties
+++ b/marklogic-data-hub/src/main/resources/dhf-defaults.properties
@@ -59,8 +59,8 @@ mlFlowDeveloperUserName=flow-developer
 mlFlowDeveloperPassword=flow-developer
 mlHubLogLevel=default
 
-# Default module permissions which allow flow-operator-role to execute flows
-mlModulePermissions=rest-reader,read,rest-writer,insert,rest-writer,update,rest-extension-user,execute
+# Default module permissions
+mlModulePermissions=data-hub-module-reader,read,data-hub-module-reader,execute,data-hub-module-writer,update,rest-extension-user,execute
 
 # Default roles for reading and updating user artifacts
 mlEntityModelPermissions=data-hub-entity-model-reader,read,data-hub-entity-model-writer,update
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mapping/entity-services/lib.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mapping/entity-services/lib.sjs
@@ -300,12 +300,18 @@ function validatePropertyMapping(fullMapping, propertyName, sourcedFrom) {
     // As of trunk 10.0-20190916, mappings are being validated against entity schemas in the schema database.
     // This doesn't seem expected, as the validation will almost always fail.
     // Thus, this is not using es.mappingCompile, which does validation, and just invokes the transform instead.
-    let stylesheet = xdmp.xsltInvoke("/MarkLogic/entity-services/mapping-compile.xsl", xmlMapping);
-    xdmp.xsltEval(stylesheet, [], {staticCheck: true});
+    let stylesheet = fn.head(xdmp.xsltInvoke("/MarkLogic/entity-services/mapping-compile.xsl", xmlMapping));
+    xdmp.xsltEval(removeStandardFunction(stylesheet), [], {staticCheck: true});
   } catch (e) {
     // TODO Move this into a separate function for easier testing?
     return getErrorMessage(e);
   }
+}
+
+//This function removes import of "standard-library.xqy" from mapping xslt. Server Bug 54051
+function removeStandardFunction(stylesheet) {
+  return fn.head(xdmp.xqueryEval(' declare variable $stylesheet external; let $node := $stylesheet/node() return element {fn:node-name($node)} {$node/namespace::*, $node/attribute(), $node/element()[fn:not(@href = "/MarkLogic/entity-services/standard-library.xqy")]}'
+  ,{stylesheet:stylesheet}, null));
 }
 
 function runMapping(mapping, uri, propMapping={"targetEntityType":mapping.targetEntityType, "namespaces": mapping.namespaces,"properties": {}}, paths=['properties']) {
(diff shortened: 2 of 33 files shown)


marklogic-data-hub/src/main/java/com/marklogic/hub/cli/command/InstallIntoDhsCommand.java
```diff
@@ -120,8 +120,6 @@ public void applyDhsSpecificProperties(Properties props, boolean disableSsl) {
 
         props.setProperty("mlFlowDeveloperRole", "flowDeveloper");
         props.setProperty("mlFlowOperatorRole", "flowOperator");
-        props.setProperty("mlModulePermissions",
-            "flowDeveloper,read,flowDeveloper,execute,flowDeveloper,insert,flowOperator,read,flowOperator,execute,flowOperator,insert");
```
rjrudin: I think this needs to remain for now, as these are defined via DHS-specific roles.

marklogic-data-hub/src/main/resources/dhf-defaults.properties
```diff
@@ -59,8 +59,8 @@ mlFlowDeveloperUserName=flow-developer
 mlFlowDeveloperPassword=flow-developer
 mlHubLogLevel=default
 
-# Default module permissions which allow flow-operator-role to execute flows
-mlModulePermissions=rest-reader,read,rest-writer,insert,rest-writer,update,rest-extension-user,execute
+# Default module permissions
+mlModulePermissions=data-hub-module-reader,read,data-hub-module-reader,execute,data-hub-module-writer,insert,rest-extension-user,execute
```
rjrudin: Just a note that the PR for DHFPROD-3779 defines these in the new AbstractHubConfig class, so they'll need to be updated there as well.

rjrudin: Also, data-hub-module-writer,insert should be data-hub-module-writer/update.

srinathgit: Changed to data-hub-module-writer/update

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/triggers/mapping/mappingJSONtoXML.sjs
```diff
@@ -24,6 +24,7 @@ if (esMappingLib.versionIsCompatibleWithES()) {
   xdmp.invokeFunction(function () {
       const es = require('/MarkLogic/entity-services/entity-services');
       es.mappingPut(xmlURI);
+      xdmp.nodeDelete(cts.doc(xmlURI+'.xslt').xpath('//xdmp:import-module[@href="/MarkLogic/entity-services/standard-library.xqy"]'));
```
rjrudin: Does this work?? I'm surprised it would, because the xmlURI doc doesn't exist until the es.mappingPut call commits.

marklogic-data-hub/src/test/java/com/marklogic/hub/flow/FlowRunWithDataHubOperatorTest.java
```diff
@@ -167,5 +191,9 @@ public void testRunFlow() throws Exception {
         String file = sh.get();
         Assertions.assertNotNull(file);
         Assertions.assertTrue(file.contains("ingest.csv"));
+        //Flow with custom xqy mapping function
+        Assertions.assertTrue(JobStatus.FINISHED.toString().equalsIgnoreCase(resp1.getJobStatus()));
```
rjrudin: Side note - should use assertEquals here, with the expected value first. The benefit is that if the assertion fails, JUnit will print the expected and actual values, which is more helpful than just seeing "assertion failed".

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mapping/entity-services/lib.sjs
```diff
@@ -300,14 +300,20 @@ function validatePropertyMapping(fullMapping, propertyName, sourcedFrom) {
...
+    let stylesheet = fn.head(xdmp.xsltInvoke("/MarkLogic/entity-services/mapping-compile.xsl", xmlMapping));
+    xdmp.xsltEval(removeStandardFunction(stylesheet), [], {staticCheck: true});
   } catch (e) {
     // TODO Move this into a separate function for easier testing?
     return getErrorMessage(e);
   }
 }
 
+//This function removes import of "standard-library.xqy" from mapping xslt. Server Bug 54051
+function removeStandardFunction(stylesheet) {
```
rjrudin: We need a unit test for this function to ensure it's behaving correctly. One risk here is that if the "to-delete" node doesn't exist, this will return an empty element. So we need at least two unit tests - a happy path where the node does exist, and then one where the node doesn't exist.

For the implementation, I recommend just doing this:

```
element {fn:node-name($node)} {
  $node/namespace::*,
  $node/attribute(),
  element()[fn:not(@href = '/MarkLogic/entity-services/standard-library.xqy')]
}
```

That will work regardless of whether the node to delete exists or not.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/triggers/mapping/mappingJSONtoXML.sjs
srinathgit: It works, I have added a test to ensure that it gets removed.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mapping/entity-services/lib.sjs
srinathgit: Changed the implementation and added tests

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mapping/entity-services/lib.sjs
```diff
@@ -300,14 +300,20 @@ function validatePropertyMapping(fullMapping, propertyName, sourcedFrom) {
...
-    xdmp.xsltEval(stylesheet, [], {staticCheck: true});
+    let stylesheet = fn.head(xdmp.xsltInvoke("/MarkLogic/entity-services/mapping-compile.xsl", xmlMapping));
+    xdmp.xsltEval(removeStandardFunction(stylesheet), [], {staticCheck: true});
   } catch (e) {
     // TODO Move this into a separate function for easier testing?
     return getErrorMessage(e);
   }
 }
 
+//This function removes import of "standard-library.xqy" from mapping xslt. Server Bug 54051
```
rjrudin: Nice, I was just going to say that we probably need a comment here to explain what's going on.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/triggers/mapping/mappingJSONtoXML.sjs
rjrudin: I guess es.mappingPut must be running in a separation transaction then?? Otherwise the ".xslt" file wouldn't exist yet when the nodeDelete is performed.

srinathgit: es.mappingPut does an eval to insert the xslt, so I think the doc insert is committed before the node is deleted

PR #3419, opened 2019-12-30
DHFPROD-3779: Can now run flows without Spring or a HubProject

rjrudin:
Summary of changes:

- Extracted AbstractHubConfig from HubConfigImpl so that SimpleHubConfig can be used without depending on Spring
- Introduced StepDefinitionProvider so that StepRunnerFactory is no longer bound to retrieving step definitions from files. An ML-based implementation of this now retrieves step definitions via a new data service.
- Added FlowRunner.runFlowWithoutProject, along with new FlowInputs class to encapsulate all the possible inputs
- Subclassed FlowRunnerTest and MappingTest so that all of the tests in those two classes are now also run via runFlowWithoutProject

### DIFF
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/job/JobDocManager.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/job/JobDocManager.java
@@ -46,7 +46,7 @@ public class JobDocManager extends ResourceManager {
         try {
             resultItr = this.getServices().post(params, new StringHandle("{}").withFormat(Format.JSON));
         } catch (Exception e) {
-            throw new RuntimeException("Unable to update the job document");
+            throw new RuntimeException("Unable to update the job document; cause: " + e.getMessage(), e);
         }
         if (resultItr == null || !resultItr.hasNext()) {
             return null;
--- a/marklogic-data-hub/src/test/resources/flow-runner-test/flows/testFlow.flow.json
+++ b/marklogic-data-hub/src/test/resources/flow-runner-test/flows/testFlow.flow.json
@@ -29,6 +29,10 @@
       "customHook": {},
       "options": {
         "permissions": "data-hub-operator,read,data-hub-operator,update"
+      },
+      "fileLocations": {
+        "inputFilePath": "input",
+        "outputURIReplacement": ".*/input,''"
       }
     },
     "3": {
--- a/marklogic-data-hub/build.gradle
+++ b/marklogic-data-hub/build.gradle
@@ -162,6 +162,10 @@ task generateSourceDataInterface(type: com.marklogic.client.tools.gradle.Endpoin
     serviceDeclarationFile = dataServicesPath + "/mastering/service.json"
 }
 
+task generateStepDefinitionInterface(type: com.marklogic.client.tools.gradle.EndpointProxiesGenTask, group: dataServicesGroup) {
+    serviceDeclarationFile = dataServicesPath + "/stepDefinition/service.json"
+}
+
 task generateDataServiceInterfaces {
     description = "Generate Java interfaces for all Data Services. Must run this from the ./marklogic-data-hub directory"
     dependsOn {
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/collector/impl/CollectorImpl.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/collector/impl/CollectorImpl.java
@@ -23,7 +23,6 @@ import com.marklogic.client.MarkLogicIOException;
 import com.marklogic.hub.HubConfig;
 import com.marklogic.hub.collector.DiskQueue;
 import com.marklogic.hub.collector.Collector;
-import com.marklogic.hub.flow.Flow;
 import com.marklogic.hub.impl.HubConfigImpl;
 import com.marklogic.rest.util.MgmtResponseErrorHandler;
 import org.apache.http.auth.AuthScope;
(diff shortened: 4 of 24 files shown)


marklogic-data-hub/src/main/java/com/marklogic/hub/job/JobDocManager.java
```diff
@@ -46,7 +46,7 @@ public JsonNode postJobs(String jobId, String status, String step, String lastCo
         try {
             resultItr = this.getServices().post(params, new StringHandle("{}").withFormat(Format.JSON));
         } catch (Exception e) {
-            throw new RuntimeException("Unable to update the job document");
+            throw new RuntimeException("Unable to update the job document; cause: " + e.getMessage(), e);
```
rjrudin: This isn't related to running flows, but I ran into this a few times while testing out flow running, and the lack of an error message or the original exception made debugging more difficult, so figured I'd improve that.

marklogic-data-hub/src/main/java/com/marklogic/hub/HubConfig.java
```diff
@@ -642,4 +642,8 @@
      * and then call this method.
      */
     void refreshProject();
+
+    String getMlUsername();
```
rjrudin: These likely should always have been exposed, as multiple clients were casting HubConfig to HubConfigImpl to access these methods.

marklogic-data-hub/src/test/resources/flow-runner-test/flows/testFlow.flow.json
```diff
@@ -29,6 +29,10 @@
       "customHook": {},
       "options": {
         "permissions": "data-hub-operator,read,data-hub-operator,update"
+      },
+      "fileLocations": {
```
rjrudin: Had to add this so that when the flow is run without a project, the input path can be made absolute.

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/SimpleHubConfig.java
```diff
@@ -0,0 +1,183 @@
...
+
+    /**
+     * Provides a minimally-configured instance of SimpleHubConfig that is suitable for e.g. running a flow.
+     *
+     * @param host
+     * @param mlUsername
+     * @param mlPassword
+     */
+    public SimpleHubConfig(String host, String mlUsername, String mlPassword) {
+        AppConfig appConfig = new AppConfig();
```
srinathgit: SimpleHubConfig by design will work only with basic and digest authentication schemes and is not expected to work with SSL / certificate based authentication ?

rjrudin: This is just a convenience constructor - you can use the no-arg constructor and set any inputs you want.

But a dirty secret of DHF is that if you want to run a collector for any step, you must use basic/digest authentication - that's because CollectorImpl uses a RestTemplate to talk to the collector and endpoint and only supports username/password.

srinathgit: Collector does support SSL/cert auth. We have those tests running in pipeline. It gets the context from client object (which is hubConfig.newStagingClient())
https://github.com/marklogic/marklogic-data-hub/blob/develop/marklogic-data-hub/src/main/java/com/marklogic/hub/collector/impl/CollectorImpl.java#L150

rjrudin: SSL is separate from the authentication strategy, and SimpleHubConfig doesn't have any impact on that. "Certificate authentication" is a different kind of authentication that DatabaseClientFactory supports which does not require a username/password. CollectorImpl requires a username/password for authentication, and SimpleHubConfig doesn't have any impact on that.

rjrudin:
Going to modify this - I realized that AbstractHubConfig can't default values, as that breaks loadConfigurationFromProperties in HubConfigImpl. 

Instead, I'll have SimpleHubConfig read from dhf-defaults.properties to obtain default values. That will ensure that dhf-defaults.properties is the only place that defines default values (I would have preferred this to be done in Java code instead of in a properties file, but it makes sense to stick with the properties file).

rjrudin:
@ryanjdew @srinathgit I moved the initialization of default values from AbstractHubConfig (which was a mistake) into the SimpleHubConfig constructor. Default values are now still only defined by dhf-defaults.properties.

rjrudin:
Ack, going to change this one more time so that the application of properties is done via functions instead of direct assignments - that will allow the code to be reused with any source of properties.

rjrudin:
Thanks @srinathgit  - I made a fix to HubTestBase.resetProperties so that it would reset properties on the adminHubConfig object properly - it wasn't resetting "projectProperties" in AbstractHubConfig. That may fix all the problems, as a number of them were connection issues due to invalid ports being used.

rjrudin:
Hmm, the Legacy* tests are still failing on my branch, while they pass on the develop branch. Investigating those.

rjrudin:
@srinathgit I figured out the problem with the legacy tests - I had modified newStagingClient() to call newStagingClient(null) so that the DatabaseClient would work with Data Services endpoints (which does not allow for a database to be set). But that broke LegacyCollectorImpl. So now newStagingClient() is back to calling newStagingClient(stagingDbName), and the clients using DataServices will have to invoke newStagingClient(null) themselves.

rjrudin:
Also, I believe the long job run time was due to the invalid ports being used, as the tests failing due to invalid ports would have been waiting a long time for a connection to timeout.

rjrudin:
Had one failing test - ProjectsControllerTest. Fixed it by marking each of the SSL-related protected fields with the JsonIgnore annotation. They didn't need it before because they were private fields, but they're protected now so that subclasses can easily access them.

PR #3420, opened 2019-12-31
DHFPROD-3781:Export project files as a zip

srinathgit:
Export project files as a zip

### DIFF
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/HubProject.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/HubProject.java
@@ -18,6 +18,7 @@ package com.marklogic.hub;
 
 import com.marklogic.hub.step.StepDefinition;
 
+import java.io.File;
 import java.io.IOException;
 import java.nio.file.Path;
 import java.util.Map;
@@ -246,6 +247,11 @@ public interface HubProject {
      * @throws IOException if problem happens with the on-disk project.
      */
     void upgradeProject() throws IOException;
+
+    /**
+     * Exports the project content to disk
+     */
+    void exportProject(File location);
 
     String getHubModulesDeployTimestampFile();
 
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/impl/HubProjectImpl.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/impl/HubProjectImpl.java
@@ -35,10 +35,7 @@ import org.springframework.core.io.Resource;
 import org.springframework.core.io.support.PathMatchingResourcePatternResolver;
 import org.springframework.stereotype.Component;
 
-import java.io.File;
-import java.io.FileWriter;
-import java.io.IOException;
-import java.io.InputStream;
+import java.io.*;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
@@ -46,6 +43,8 @@ import java.nio.file.attribute.PosixFilePermission;
 import java.util.*;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipOutputStream;
 
 import static com.marklogic.hub.HubConfig.HUB_MODULES_DEPLOY_TIMESTAMPS_PROPERTIES;
 import static com.marklogic.hub.HubConfig.USER_MODULES_DEPLOY_TIMESTAMPS_PROPERTIES;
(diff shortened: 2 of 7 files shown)


marklogic-data-hub/src/main/java/com/marklogic/hub/HubProject.java
```diff
@@ -247,6 +247,13 @@
      */
     void upgradeProject() throws IOException;
 
+    /**
+     * Exports the project content to disk
+     *
+     * @throws IOException if there are issues with exporting project to disk.
+     */
+    void exportProject() throws IOException;
```
rjrudin: While the story calls for a path, it's a Gradle-specific path, and I don't think our Java layer should know what it is. Therefore, I think this method should accept a File (or Path) as an input.

marklogic-data-hub/src/test/java/com/marklogic/bootstrap/DataHubInstallTest.java
```diff
@@ -122,9 +123,27 @@ public void setup()
             } catch (IOException e) {
                 throw new RuntimeException(e);
             }
+            copyFlowArtifactsToProject();
+            Path hubProjDir = getDataHubAdminConfig().getHubProject().getProjectDir();
+            Assertions.assertTrue(hubProjDir.resolve("input").toFile().exists());
+            project.exportProject();
+            //copy file to build dir from project dir
+            FileUtils.copyFileToDirectory(hubProjDir.resolve("build").resolve("datahub-project.zip").toFile(), new File("build"));
```
rjrudin: One benefit of exportProject requiring an input is that the client no longer needs to assume what the File is - the client will now determine that.

marklogic-data-hub/src/main/java/com/marklogic/hub/HubProject.java
rjrudin: Also, let's not throw an IOException here - IMO, IOException shouldn't be a checked exception because 99% of the time, there's nothing the caller can do to recover. No checked exception should be thrown, we'll rethrow checked ones in the implementation.

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/HubProjectImpl.java
```diff
@@ -543,6 +542,60 @@ public void upgradeProject() throws IOException {
         removeEmptyRangeElementIndexArrayFromFinalDatabaseFile();
     }
 
+    @Override
+    public void exportProject() throws IOException {
```
rjrudin: So instead of throwing an IOException, let's catch any exception and rethrow it as a RuntimeException, e.g.

    catch(Exception ex) {
      throw new RuntimeException("Unable to export project, cause: " + ex.getMessage(), ex);
    }

If any error occurs, there's nothing the caller - in this case, a Gradle task - will be able to do other than pass the error onto the user. So we might as well do that in here.

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/HubProjectImpl.java
```diff
@@ -543,6 +542,60 @@ public void upgradeProject() throws IOException {
...
+                if(fileToZip.isDirectory()) {
+                    zout.putNextEntry(new ZipEntry(file + File.separator));
+                    zipSubDirectory(file + File.separator, fileToZip, zout);
+                }
+                else {
+                    zipSubDirectory("" , fileToZip, zout);
+                }
+            }
+            catch (Exception e){
+                logger.error("Unable to export project", e.getMessage() );
```
rjrudin: We definitely don't want to catch and swallow any exception - this could result in some of the zip being written but not all of it. We instead want every exception to be handled as described above.

marklogic-data-hub/src/main/java/com/marklogic/hub/HubProject.java
srinathgit: The story doesn't mention a path to be passed to "hubExportProject" task . It says that the exported file should be available at "./build/datahub-project.zip". So the story should be changed  so that "hubExportProject" accepts a path (something like hubExportProject -Ppath=<path_export_file>)

rjrudin: What I mean is that the Gradle task should know the path of "build/datahub-project.zip" and pass that in as a File to the Java class. The Java class shouldn't know about a "build" directory, since that's a Gradle-specific construct. It's an implementation detail that isn't relevant to the user story, but it matters in that we don't want the Java class knowing about Gradle-specific stuff.

PR #3422, opened 2020-01-03
DHFPROD-4023: Added examples for running a flow without a project

rjrudin:
This is intended to be linked to from new docs that will be added to  https://docs.marklogic.com/datahub/flows/run-flow-using-datahub-api.html

### DIFF
--- a/examples/dh-5-example/src/main/entity-config/exp-final-entity-options.xml
+++ b/examples/dh-5-example/src/main/entity-config/exp-final-entity-options.xml
@@ -3,6 +3,8 @@
   <search:constraint name="Collection">
     <search:collection>
       <search:facet-option>limit=25</search:facet-option>
+      <search:facet-option>frequency-order</search:facet-option>
+      <search:facet-option>descending</search:facet-option>
     </search:collection>
   </search:constraint>
   <search:constraint name="createdByJob">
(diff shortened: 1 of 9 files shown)


examples/dh-5-example/src/main/entity-config/exp-final-entity-options.xml
```diff
@@ -3,6 +3,8 @@
   <search:constraint name="Collection">
     <search:collection>
       <search:facet-option>limit=25</search:facet-option>
+      <search:facet-option>frequency-order</search:facet-option>
```
rjrudin: These were automatically added when I deployed the app based on the latest on develop, so checking them in.

rjrudin:
@SameeraPriyathamTadikonda The Jenkins job for this has one failing test - DeployProtectedPathsWhenUpdatingIndexesTest . I ran that locally on this branch, including rebased off develop, and the test passes. The changes in here have no impact on that test, nor on the feature it's testing, which involves deploying DHF.

SameeraPriyathamTadikonda:
@rjrudin  I need one more approval for merging this Pull request manually.

rjrudin:
@SameeraPriyathamTadikonda  The latest set of 13 failures are not impacted by this PR. This PR is only modifying files in the dh-5-example project and then the new FlowInputs class, and all it's doing is adding a method to that class. I think it can be safely merged to develop.

SameeraPriyathamTadikonda:
Merging it manually as the fix is in the PR 3439

PR #3427, opened 2020-01-06
DHFPROD-2691:Prohibit DHS users from updating DHF modules

srinathgit:
Prohibit DHS users from updating DHF modules

### DIFF
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/cli/command/InstallIntoDhsCommand.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/cli/command/InstallIntoDhsCommand.java
@@ -67,6 +67,7 @@ public class InstallIntoDhsCommand extends AbstractInstallerCommand {
         commands.addAll(commandMap.get("mlModuleCommands"));
 
         commands.add(new CopyQueryOptionsCommand(hubConfig));
+        commands.add(new UpdateDhsModulesPermissionsCommand(hubConfig));
 
         DeployRolesCommand deployRolesCommand = new DeployRolesCommand();
         deployRolesCommand.setResourceFilenamesIncludePattern(
@@ -121,7 +122,7 @@ public class InstallIntoDhsCommand extends AbstractInstallerCommand {
         props.setProperty("mlFlowDeveloperRole", "flowDeveloper");
         props.setProperty("mlFlowOperatorRole", "flowOperator");
         props.setProperty("mlModulePermissions",
-            "flowDeveloper,read,flowDeveloper,execute,flowDeveloper,insert,flowOperator,read,flowOperator,execute,flowOperator,insert");
+            "data-hub-module-reader,read,data-hub-module-reader,execute,data-hub-environment-manager,update,rest-extension-user,execute");
 
         props.setProperty("mlAppServicesAuthentication", "basic");
         props.setProperty("mlFinalAuth", "basic");
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/cli/command/UpdateDhsModulesPermissionsCommand.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/cli/command/UpdateDhsModulesPermissionsCommand.java
@@ -1,0 +1,90 @@
+/*
+ * Copyright 2012-2019 MarkLogic Corporation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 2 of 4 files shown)


marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/UpdateDhsModulesPermissionsCommand.java
```diff
@@ -0,0 +1,64 @@
...
+@Component
+public class UpdateDhsModulesPermissionsCommand extends AbstractCommand {
+
+    @Autowired
+    private HubConfig hubConfig;
+    private UpdateDhsResourcePermissions service;
+
+    //This command has to be run after LoadHubModulesCommand is run.
+    public UpdateDhsModulesPermissionsCommand() {
+        setExecuteSortOrder(SortOrderConstants.LOAD_MODULES);
```
rjrudin: This should be SortOrderConstants.LOAD_MODULES + 1 to ensure that the modules have all been loaded already.

marklogic-data-hub/src/main/java/com/marklogic/hub/flow/impl/FlowRunnerImpl.java
```diff
@@ -5,6 +5,7 @@
 import com.marklogic.client.io.JacksonHandle;
 import com.marklogic.hub.FlowManager;
 import com.marklogic.hub.HubConfig;
+import com.marklogic.hub.flow.Flow;
```
rjrudin: Can you remove this so there's one less modification to FlowRunnerImpl? Keeps the version history cleaner.

marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/UpdateDhsModulesPermissionsCommand.java
```diff
@@ -0,0 +1,64 @@
...
+import com.marklogic.hub.dataservices.UpdateDhsResourcePermissions;
+import org.apache.commons.io.FilenameUtils;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.core.io.Resource;
+import org.springframework.stereotype.Component;
+
+import java.util.List;
+
+@Component
+public class UpdateDhsModulesPermissionsCommand extends AbstractCommand {
```
rjrudin: This command is specific to installing DHF into DHS, so a few notes on it:

- It should be in the com.marklogic.hub.cli.deploy package along with the other 2 commands specific to installing DHF into DHS
- It shouldn't be a Spring component - InstallIntoDhsCommand should just pass a HubConfig into its constructor (similar to CopyQueryOptionsCommand)

And then regarding the use of data services - I don't think we should do that for a deployment concern. CopyQueryOptionsCommand is using the /v1/eval endpoint to mess around with query options as part of installing DHF into DHS. For consistency and simplicity, I think this command should follow the same approach - i.e. build a script and send it to /v1/eval. Building a script inside of Java code isn't the cleanest thing ever, but I think it's suitable for a deployment concern - whereas Data Services is absolutely the best approach for any interface that we want a customer to be able to invoke (but we definitely don't want a customer invoking this endpoint).

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/dhs/updateResourcePermissions.sjs
```diff
@@ -0,0 +1,39 @@
...
+  See the License for the specific language governing permissions and
+  limitations under the License.
+*/
+'use strict';
+
+var extensionType;
+var fileName;
+
+
+xdmp.eval('let partUri = "/marklogic.rest.'+ extensionType + '/' + fileName + '/assets/"; '+
```
rjrudin: This code can all be reused in the Java command - though I don't think you need an xdmp.eval here. Just do what CopyQueryOptionsCommand is doing - call hubConfig.newModulesDbClient().

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/DataHubImpl.java
```diff
@@ -830,6 +796,9 @@ private void updateModuleCommandList(Map<String, List<Command>> commandsMap) {
         commands.add(loadUserArtifactsCommand);
         commands.add(loadHubArtifactsCommand);
         commands.add(generateFunctionMetadataCommand);
+        if(hubConfig.getIsProvisionedEnvironment()){
```
rjrudin: One small tweak here - this new command is specific to InstallIntoDhsCommand, there's no need for DataHubImpl to know about it. Just add it to the list of commands in InstallIntoDhsCommand, just like CopyQueryOptionsCommand.

PR #3428, opened 2020-01-06
DHFPROD-3089:Ensure absence of some dirs doesn't affect authentication

srinathgit:
Ensure absence of some dirs doesn't affect authentication


web/src/main/java/com/marklogic/hub/web/web/CurrentProjectController.java
```diff
@@ -212,24 +217,33 @@ public String preInstallCheck() {
...
             watcherService.addListener(this);
         }
     }
 
+    private void enableWatcherService(File dir) {
+        try{
+            watcherService.watch(dir.toString());
+        }
+        catch (IOException ex){
+            logger.warn("Unable to enable watcher service for "+ dir.toString(), ex);
```
rjrudin: I think for an exception like this - which is often true of warnings - we should just log the exception message, but not the full stacktrace. If we do the stacktrace for several exceptions, the user may thing something has gone horribly wrong when she sees several stacktraces being logged. 

So for this, and generally for warnings, I recommend doing the following:

    logger.warn("Unable to enable watcher server for directory: " + dir.toString() + "; cause: " + ex.getMessage());

PR #3430, opened 2020-01-07
DHFPROD-3911:Replace jsonPropertyValueQuery with pathRangeQuery

srinathgit:
Replace jsonPropertyValueQuery with pathRangeQuery

### DIFF
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/impl/HubProjectImpl.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/impl/HubProjectImpl.java
@@ -34,7 +34,21 @@ import org.springframework.context.annotation.Lazy;
 import org.springframework.core.io.Resource;
 import org.springframework.core.io.support.PathMatchingResourcePatternResolver;
 import org.springframework.stereotype.Component;
+import org.w3c.dom.Document;
+import org.w3c.dom.Element;
+import org.w3c.dom.Node;
 
+import javax.xml.parsers.DocumentBuilder;
+import javax.xml.parsers.DocumentBuilderFactory;
+import javax.xml.transform.OutputKeys;
+import javax.xml.transform.Transformer;
+import javax.xml.transform.TransformerFactory;
+import javax.xml.transform.dom.DOMSource;
+import javax.xml.transform.stream.StreamResult;
+import javax.xml.xpath.XPath;
+import javax.xml.xpath.XPathConstants;
+import javax.xml.xpath.XPathExpression;
+import javax.xml.xpath.XPathFactory;
 import java.io.*;
 import java.nio.file.Files;
 import java.nio.file.Path;
@@ -288,10 +302,6 @@ public class HubProjectImpl implements HubProject {
         writeResourceFile("hub-internal-config/database-fields/job-database.xml",
             hubDatabaseFieldsDir.resolve("job-database.xml"), true);
 
-        Path userDatabaseFieldsDir = getUserConfigDir().resolve("database-fields");
-        userDatabaseFieldsDir.toFile().mkdirs();
-        writeResourceFile("ml-config/database-fields/final-database.xml",
-            userDatabaseFieldsDir.resolve("final-database.xml"), true);
 
         // Per DHFPROD-3159, we no longer overwrite user config (ml-config) files. These are rarely updated by DHF,
         // while users may update them at any time. If DHF ever needs to update one of these files in the future, it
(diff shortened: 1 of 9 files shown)


ml-data-hub-plugin/src/test/resources/update-indexes/final-database.xml
```diff
@@ -129,5 +129,12 @@
       <range-value-positions>true</range-value-positions>
       <invalid-values>reject</invalid-values>
     </range-path-index>
+    <range-path-index>
```
rjrudin: It looks like these duplicate files are only needed for UpdateIndexesTaskTest. But after looking at that test, I don't think it's testing anything specific to DHF - it's just re-testing ml-app-deployer functionality.  DHF can safely assume that if it defines indexes in a database file, then ml-app-deployer will deploy them correctly.

I think we should get rid of these duplicate files. My first choice is to delete them and delete UpdateIndexesTaskTest. My second choice is to delete them and modify UpdateIndexesTaskTest to copy the real files from ../marklogic-data-hub/src/main/resources/

marklogic-data-hub/src/main/resources/ml-config/database-fields/final-database.xml
```diff
@@ -114,5 +114,12 @@
       <range-value-positions>true</range-value-positions>
       <invalid-values>reject</invalid-values>
     </range-path-index>
+    <range-path-index>
```
rjrudin: HubProjectImpl.upgradeProject will need to be modified to inject this index into the user's existing final-database.xml file. Odds are, the user hasn't modified it, but we don't want to overwrite the file in case the user has made modifications. The contents of the file can be read in as an XML document, and then you can just manually add this new index to it, and then write it back to disk.

ml-data-hub-plugin/src/test/groovy/com/marklogic/gradle/task/UpdateIndexesTaskTest.groovy
```diff
@@ -92,8 +92,8 @@ class UpdateIndexesTaskTest extends BaseTest {
 		result.task(":mlUpdateIndexes").outcome == SUCCESS
 
         //Test to verify range-path-index for mlUpdateIndex
-		assert (getStagingIndexValuesSize('//m:range-path-index') == 2)
-		assert (getFinalIndexValuesSize('//m:range-path-index') == 2)
+		assert (getStagingIndexValuesSize('//m:range-path-index') == 3)
```
rjrudin: I think the better test to enhance is DeployDatabaseFieldCommandTest. See my comments below about what to do with UpdateIndexesTaskTest, which I don't think is proving much of anything.

ml-data-hub-plugin/src/test/resources/update-indexes/final-database.xml
srinathgit: UpdateIndexesTaskTest ensures that the payload generated by 'hubSaveIndexes' task is correct and can be deployed without issues. Also the files (final-database.xml and staging-database.xml) are not duplicates of each other, there are additional indexes in them

rjrudin: Okay, sounds good.

marklogic-data-hub/src/main/resources/ml-config/database-fields/final-database.xml
srinathgit: Modified HubProjectImpl.upgradeProject to inject the new index

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/HubProjectImpl.java
```diff
@@ -542,6 +555,67 @@ public void upgradeProject() throws IOException {
...
+                transformer.setOutputProperty(OutputKeys.INDENT, "yes");
+                transformer.setOutputProperty(OutputKeys.ENCODING, "UTF-8");
+                transformer.setOutputProperty("{http://xml.apache.org/xslt}indent-amount", "2");
+
+                transformer.transform(new DOMSource(document),
+                    new StreamResult(new OutputStreamWriter(new FileOutputStream(finalDbFile), "UTF-8")));
+            }
+        }
+        catch (Exception e) {
+            throw new DataHubProjectException("Unable to upgrade the project: Failed to add index to final database", e);
```
rjrudin: I think this is okay. I was initially worried that by throwing the exception, the user would think that hubUpgrade failed. But in reality, almost all of it succeeded except the last step. I think the error message could do a better job in explaining what failed - e.g. "Error while upgrading project; was not able to add //actionDetails/*/uris path range index to final-database.xml file; cause: " + ex.getMessage(). 

@srinathgit  Thoughts on throwing the error, or just logging it?

srinathgit: I like throwing the exception and will change the error message to the one you had suggested

PR #3437, opened 2020-01-08
DHFPROD-4073: dhsVerify now correctly verifies roles

rjrudin:
I also realized that we still don't want flow-developer-role, flow-operator-role, or data-hub-admin-role deployed to DHS, as those grant too many privileges and thus shouldn't be present.

### DIFF
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/cli/command/VerifyDhfInDhsCommand.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/cli/command/VerifyDhfInDhsCommand.java
@@ -14,6 +14,7 @@ import org.springframework.context.ApplicationContext;
 
 import java.util.Properties;
 import java.util.Set;
+import java.util.stream.Stream;
 
 @Parameters(commandDescription = "Verify a DHF installation in a DHS environment")
 public class VerifyDhfInDhsCommand extends AbstractVerifyCommand {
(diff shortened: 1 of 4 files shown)


marklogic-data-hub/src/main/java/com/marklogic/hub/cli/command/VerifyDhfInDhsCommand.java
```diff
@@ -92,16 +99,17 @@ private void verifyModules() {
             DocumentMetadataHandle metadata = documentManager.readMetadata("/com.marklogic.hub/config.sjs", new DocumentMetadataHandle());
             DocumentMetadataHandle.DocumentPermissions perms = metadata.getPermissions();
 
-            Set<DocumentMetadataHandle.Capability> capabilities = perms.get("flowDeveloper");
-            verify(capabilities.contains(DocumentMetadataHandle.Capability.READ), "Flow developer should be able to read modules");
-            verify(capabilities.contains(DocumentMetadataHandle.Capability.INSERT), "Flow developer should be able to insert modules");
-            verify(capabilities.contains(DocumentMetadataHandle.Capability.EXECUTE), "Flow developer should be able to execute modules");
+            Set<DocumentMetadataHandle.Capability> capabilities = perms.get("data-hub-module-reader");
```
bsrikan: @rjrudin can you please explain the change here and for flowOperator?

rjrudin: The command is now using a permissions string of data-hub-module-reader,read,data-hub-module-reader,execute,data-hub-environment-manager,update,rest-extension-user,execute .

rjrudin: We need a meeting with the DHS team to determine if they want to keep the flowOperator and flowDeveloper roles. If so, they'll need to adapt to inherit these new roles.

PR #3439, opened 2020-01-08
DHFPROD-3935: Fixing ValidateMappingTest

rjrudin:
It started failing because it was accessing adminHubConfig before ensuring it was initialized to use flow-developer as a user.

Perhaps HubTestBase should be doing this before every test?? Otherwise there's the chance that one test modified adminHubConfig, and that bleeds over into the next test, causing an error.

I also modified setupProjectForRunningTestFlow, as it was ignoring the HubConfig object being passed into it.

### DIFF
--- a/marklogic-data-hub/src/test/java/com/marklogic/hub/flow/FlowRunWithDataHubOperatorTest.java
+++ b/marklogic-data-hub/src/test/java/com/marklogic/hub/flow/FlowRunWithDataHubOperatorTest.java
@@ -105,7 +105,7 @@ public class FlowRunWithDataHubOperatorTest extends HubTestBase {
         //Copy mapping functions
         FileUtils.copyFileToDirectory(getResourceFile("flow-runner-test/mapping-functions/add-function.xqy"),
             adminHubConfig.getHubProject().getCustomMappingFunctionsDir().toFile());
-        setupProjectForRunningTestFlow(getHubFlowRunnerConfig("test-data-hub-developer", "test-data-hub-developer"));
+        setupProjectForRunningTestFlow();
 
         //Run flow as "data-hub-operator"
         getHubFlowRunnerConfig("test-data-hub-operator", "test-data-hub-operator");
--- a/marklogic-data-hub/src/test/java/com/marklogic/hub/HubTestBase.java
+++ b/marklogic-data-hub/src/test/java/com/marklogic/hub/HubTestBase.java
@@ -1217,7 +1217,7 @@ public class HubTestBase {
         }
     }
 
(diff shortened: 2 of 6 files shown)


marklogic-data-hub/src/test/java/com/marklogic/hub/flow/FlowRunWithDataHubOperatorTest.java
```diff
@@ -105,7 +105,7 @@ public void setupEach() throws Exception {
         //Copy mapping functions
         FileUtils.copyFileToDirectory(getResourceFile("flow-runner-test/mapping-functions/add-function.xqy"),
             adminHubConfig.getHubProject().getCustomMappingFunctionsDir().toFile());
-        setupProjectForRunningTestFlow(getHubFlowRunnerConfig("test-data-hub-developer", "test-data-hub-developer"));
+        setupProjectForRunningTestFlow();
```
srinathgit: 1. I wanted to ensure that "data-hub-developer" role can deploy artifacts and modules but missed to use this config in HubTestBase.setupProjectForRunningTestFlow().I think the method setupProjectForRunningTestFlow() can be changed to use that config so that we can test "data-hub-developer" role can deploy artifacts and modules 
2. Just an FYI, I added "any-uri" privilege to "data-hub-developer" so that it can deploy mappings and entities which are protected by uri privileges

rjrudin: I don't believe calling getHubFlowRunnerConfig("test-data-hub-developer", "test-data-hub-developer") was having effect though, because the second line of setupProjectForRunningTestFlow is a call to "getDataHubAdminConfig();", where therefore overrides the effect of getHubFlowRunnerConfig. 

FlowRunWithDataHubOperatorTest then calls getHubFlowRunnerConfig after calling setupProjectForRunningTestFlow, which ensures the flow is run as that user.

If your goal is to ensure that data-hub-developer can load artifacts and modules, I don't think this is the right test for that (and that wasn't happening either). DataHubDeveloperTest has a task19LoadModulesAndArtifacts test in it - perhaps that needs to be enhanced?

srinathgit: Yeah, I missed to use getHubFlowRunnerConfig("test-data-hub-developer", "test-data-hub-developer") to deploy artifacts, modules in HubTestBase.setupProjectForRunningTestFlow(). I will modify the task19LoadModulesAndArtifacts later

Issue #3440, opened 2020-01-08
[BUG] DHF 5.1: Inconsistent field index declaration and use for datahubCreatedByJob


crism:
#### The issue

Short description of the problem: the `datahubCreatedByJob` field has an index defined with an English collation, but a field reference with root collation. This causes intermittent (and difficult to reproduce) errors when deploying.

What behavior are you expecting? Clean deployment.

What are the steps to reproduce the issue:
1) Install DHF 5.1.
2) Run `./gradlew mlDeploy`.

#### Tech details

Which Operating System are you using? CentOS

Which version of MarkLogic are you using? 9.0-11

Which version of the Data Hub are you using? 5.1.0

crism:
This is already being tracked internally. but I’m patching it locally and may as well submit a PR.

rjrudin:
@crism We've got the JIRA ticket to track this internally, so no need for another ticket. Thanks for submitting the PR though.

PR #3441, opened 2020-01-08
Fix datahubCreatedByJob collation inconsistency (#3440)

crism:
Fixes intermittent deployment bug #3440

### DIFF
--- a/examples/dh-5-example/README.md
+++ b/examples/dh-5-example/README.md
@@ -50,3 +50,26 @@ In addition, a subset of steps can be run when running a flow:
 
     ./gradlew hubRunFlow -PflowName=ingestion_only-flow -Psteps=1
     ./gradlew hubRunFlow -PflowName=ingestion_only-flow -Psteps=2,3
+
+## Running flows without depending on a project
+
+As of 5.2.0, this project includes examples of constructing a FlowRunnerImpl that does not depend on project files, nor
+on a Spring container, for running flows. See the "runFlowWithoutProject" Gradle task for more information.
+
+To author your own Java code that uses FlowRunnerImpl, you can start with the Gradle configuration below for depending 
+on the Data Hub Java API in your own project:
+
+```
+plugins {
+    id "java"
+}
+
+repositories {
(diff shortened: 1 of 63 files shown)


rjrudin:
Let's not merge this yet - I posted on the internal JIRA ticket that the appropriate solution may be to adjust hub-entities.xqy when the search options are generated.

@crism If you have time, could you try that solution out - i.e. bring your own patched copy of hub-entities.xqy into your codebase, and then deploy your app and see if having constraints with the "en" collation avoids the error? 

Also, do you have steps that are guaranteed to produce the error every time?

crism:
@rjrudin, it seems like a first-time `mlDeploy` or `mlRedeploy` will trigger the error, but after that, it’s uncertain. I’m not sure why it doesn’t crop up reliably.

I’ll try it the other way ’round in the patch. It seems to me that `hubCreatedByJob` is unlikely to be a linguistically-valued property, so root collation is appropriate, but your point about triggering reindexing for existing installations is well-taken.

crism:
Part of the reproducibility problem has to do with reindexing. When the range index definition is different than previously, it triggers reindexing. Then the search options come in hot on its heels, and because indexing isn’t done, the index isn’t found, and so it fails. After waiting for reindexing, the next assertion isn’t different, doesn’t trigger reindexing, and succeeds.

rjrudin:
Modifying the search options seems like the right choice, if the "en" collation is intended. If it is intended, we should have a comment documenting why it's intended, as it's guaranteed to raise eyebrows in the future. @rahulvudutala Can you provide that comment? Would be nice to include it in this PR.

crism:
(It’s worth noting that this bug actually arises whether or not the index and usage are consistent… if it’s _different_ than previously installed, the rapidity of deployment will trigger an error. This is a broader problem probably worth looking into.)

rjrudin:
Closing this, as the root cause isn't the mismatch between search options and the index, but rather that the search options are deployed before the index is built. That's a server issue we'll need to figure out.

PR #3446, opened 2020-01-09
DHFPROD-4046:Ensure that watcher service is disabled after user logs out

srinathgit:
Ensure that watcher service is disabled after user logs out

### DIFF
--- a/web/src/main/java/com/marklogic/hub/web/web/CurrentProjectController.java
+++ b/web/src/main/java/com/marklogic/hub/web/web/CurrentProjectController.java
@@ -299,15 +299,26 @@ public class CurrentProjectController implements FileSystemEventListener, Valida
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


web/src/main/java/com/marklogic/hub/web/web/CurrentProjectController.java
```diff
@@ -301,13 +301,22 @@ private void clearAuthenticationAttributes(HttpServletRequest request) {
...
+        disableWatcherService(hubConfig.getFlowsDir().toFile());
+        disableWatcherService(hubConfig.getStepDefinitionsDir().toFile());
+        disableWatcherService(hubProject.getCustomModulesDir().toFile());
+        watcherService.removeListener(this);
         mappingManagerService.unsetMappingValidators();
     }
+
+    private void disableWatcherService(File dir) {
+        try{
+            watcherService.unwatch(dir.toString());
```
rjrudin: One way to make this a little nicer is to verify that "dir" is not null, and that dir.exists() as well. If either is not true, no need to unwatch.

PR #3450, opened 2020-01-10
DHFPROD-3974: Changes for new UI install

ryanjdew:


marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/LoadUserModulesCommand.java
```diff
@@ -256,6 +257,10 @@ public FileVisitResult visitFile(Path file, BasicFileAttributes attrs) {
 
     public void setHubConfig(HubConfig hubConfig) {
         this.hubConfig = hubConfig;
+        if (this.entityManager == null) {
```
rjrudin: Would this ever happen? I don't think it's necessary, as long as this command class is assumed to be instantiated via Spring. I don't think just instantiating EntityManagerImpl and then setting hubConfig on it will often work, as EntityManagerImpl has other dependencies (HubProject, could be more in the future) that aren't being set. 

I thinking asserting that entityManager is not null may be worthwhile, although a user couldn't do anything with the error - it would reflect a system misconfiguration (Spring has a handy Assert class for such purposes).

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/HubConfigImpl.java
```diff
@@ -2096,4 +2102,32 @@ public void resetHubConfigs() {
...
+        CmaConfig cmaConfig = null;
+        if (this.appConfig != null) {
+            cmaConfig = this.appConfig.getCmaConfig();
+        }
+        if (cmaConfig == null) {
+            cmaConfig = new CmaConfig(true);
+        }
+        appConfig.setCmaConfig(cmaConfig);
+        hubConfig.setAppConfig(appConfig, true);
+        hubConfig.setManageConfig(manageConfig);
```
rjrudin: I don't think this will work, because none of the manageConfig properties are being set.

ryanjdew: The necessary values are set as part of the session creation in the one-ui project.

ryanjdew:
@rjrudin The idea here is to clone a HubConfig and set the credentials, etc. specifically to a user for their session. 

I'll investigate the two approaches that you mentioned. I am a little skeptical about the first approach working for us. I believe it would need to be on more than just than HubConfigImpl since there is a number of interconnected classes with configuration and I'm not also sure how that annotation would affect things outside the context of the Spring Boot application. Would it just be ignored?

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/DataHubImpl.java
```diff
@@ -1040,10 +1040,41 @@ protected void prepareProjectBeforeUpgrading(HubProject hubProject, String curre
...
+    public HubConfigImpl getHubConfig() {
+        return this.hubConfig;
     }
 
     // only used in test
     public void setVersions(Versions versions) {
         this.versions = versions;
     }
+
+    public Object clone() throws CloneNotSupportedException {
```
rjrudin: Is there any benefit to implementing this vs one-ui just constructing a DataHubImpl itself and then populating the fields necessary? 

The first 3 commands in the list only require a HubConfig. I am wondering if a DataHubImpl(HubConfig) constructor would be better here. That constructor would instantiate the first 3 commands. LoadUserModulesCommand has 2 other dependencies, so the caller of that constructor would need to provide an instance of LoadUserModulesCommand itself. 

I think the fact that our Java classes have private fields that depend on Autowired and don't expose dependencies via constructors makes life really hard for anyone that wants to instantiate our classes without Spring. That's the problem you're running into here. I think the way we solve that is by creating those constructors so that it's much easier for a project like one-ui to instantiate these classes itself, because the constructors will make the dependencies plainly visible.

rjrudin: I just noticed DataHubImpl has a couple dependencies that aren't used - I'm going to submit a PR to clean that up.

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/HubConfigImpl.java
rjrudin: What if when one-ui needs to create a new HubConfigImpl, it first creates one based on properties in the project, and then it overlays the user-specific stuff? 

I basically really want to avoid having implementations of "clone" in marklogic-data-hub that are based on one-ui requirements. I think "clone" is really only appropriate for objects with primitive fields and complex fields that are either stateless or have no user-specific state, and neither DataHubImpl nor HubConfigImpl fit those descriptions (particularly since we have circular dependencies between them, which is something else I'd like to resolve - HubConfigImpl should not have any knowledge of those 4 Spring-managed beans).

marklogic-data-hub/src/main/java/com/marklogic/hub/job/impl/JobMonitorImpl.java
```diff
@@ -54,6 +54,10 @@ public void setupClient() {
         batches = new Batches(client);
     }
 
+    public void setHubConfig(HubConfig hubConfig) {
```
rjrudin: Instead of a setter, let's add a no-arg constructor and then a constructor that requires a HubConfig. The client can then create a new JobMonitorImpl based on a HubConfig.

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/EntityManagerImpl.java
```diff
@@ -659,4 +659,8 @@ public boolean savePii() {
         return true;
     }
 
+    public void setHubConfig(HubConfig hubConfig) {
```
rjrudin: Same thing here, let's use a constructor here instead of a setter so it's clear that a new object is being created that's specific to this HubConfig.

marklogic-data-hub/src/main/java/com/marklogic/hub/legacy/LegacyFlowManager.java
```diff
@@ -37,6 +37,12 @@
      */
     void setHubConfig(HubConfig hubConfig);
 
+    /**
+     * Get HubConfig
+     * @return hubConfig a hubConfig object
+     */
+    HubConfig getHubConfig();
```
rjrudin: I couldn't find where this is needed?

marklogic-data-hub/src/main/java/com/marklogic/hub/HubConfig.java
```diff
@@ -599,6 +601,10 @@
      */
     DatabaseClient newStagingClient(String dbName);
 
+    String getStagingTriggersDbName();
+
+    AdminManager getAdminManager();
+    ManageClient getManageClient();
```
rjrudin: These seem like reasonable additions.

marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/LoadUserModulesCommand.java
rjrudin: Note that on develop, there's now an EntityManagerImpl(HubConfig) constructor. I recommend using that because it also assigns a value to the HubProject field.

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/DataHubImpl.java
```diff
@@ -895,10 +895,31 @@ protected void prepareProjectBeforeUpgrading(HubProject hubProject, String curre
     // only used in test
     public void setHubConfig(HubConfigImpl hubConfig) {
         this.hubConfig = hubConfig;
+        if (this.loadUserModulesCommand != null) {
```
rjrudin: All of these have constructors that take a HubConfig - those should be used instead, particularly for LoadHubModulesCommand, which will then construct its other field using the given HubConfig.

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/EntityManagerImpl.java
```diff
@@ -35,32 +35,44 @@
...
 @Component
 public class EntityManagerImpl extends LoggingObject implements EntityManager {
     private static final String ENTITY_FILE_EXTENSION = ".entity.json";
 
     @Autowired
     private HubConfig hubConfig;
 
-    @Autowired
-    private HubProject hubProject;
+    private HubProject hubProject = null;
```
rjrudin: I think the better thing to do is not track this as a field at all - there's no reason to, as we always want to use the HubProject associated with the HubConfig.

PR #3453, opened 2020-01-10
DHFPROD-4160: Fixes #3451 by changing and query to or.

### LINKED ISSUE #3451: [BUG] DHF 5.1: Match query options get and-ed instead of or-ed
#### The issue

Short description of the problem: After upgrading to DHF 5.1, matches that succeded previously began to fail.

What behavior are you expecting? Successful match.

What are the steps to reproduce the issue:
1) Create a match rule with both add and expand options.
2) Log the match query.
3) Observe that the add and expand sub-queries are joined in an `and-query()`; typically, they will not both be true, and so matching will fail.

#### Tech details

Which Operating System are you using? CentOS 7

Which version of MarkLogic are you using? 9.0-11

Which version of the Data Hub are you using? 5.1.0

The problem appears to stem for the change to use `group-queries-by-scope()`, the results of which are then anded. This is a change from DHF 5.0.3, in which minimum threshold queries were grouped in an `or-query()`.

crism:
Fixes #3451.


marklogic-data-hub/src/main/resources/ml-modules/root/com.marklogic.smart-mastering/matcher-impl/matcher-impl.xqy
```diff
@@ -303,7 +303,7 @@ declare function match-impl:find-document-matches-by-options(
           $queries
         return
           if (fn:count($queries) gt 1) then
-            cts:and-query($queries)
+            cts:or-query($queries)
```
rjrudin: @ryanjdew I don't know enough to know the impact here. Thoughts?

I see that the code prior to the latest change was this:

```
return
        if (fn:count($queries) gt 1) then
          helper-impl:group-queries-by-scope($queries, cts:and-query#1)
        else
          $queries
```

And is now this:

```
return
        let $queries := if (fn:count($queries) gt 1) then
          helper-impl:group-queries-by-scope($queries, cts:and-query#1)
        else
          $queries
        return
          if (fn:count($queries) gt 1) then
            cts:and-query($queries)
          else
            $queries
```

So this looks like a new line of code. I just don't know whether "or" or "and" is the right choice here.

ryanjdew:
@crism changed where the OR was getting applied to ensure other items don't break. It is applying the OR at the output of the custom query function. That takes care of your issue, correct?

crism:
@ryanjdew, this modified patch still works for our usage. Thank you!

SameeraPriyathamTadikonda:
Please rebase the source branch, there are some merge conflicts.

PR #3455, opened 2020-01-11
DHFPROD-3625:Ingest correctly when xml doc root node name is 'root'

srinathgit:
Ingest correctly when xml doc root node name is 'root'


marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/builtins/steps/ingestion/default/testXmlIngestWIthRoot.sjs
```diff
@@ -0,0 +1,11 @@
+const ing = require('/data-hub/5/builtins/steps/ingestion/default/main.sjs');
+
+let assertions = [];
+
+function runJsonToXmlIngest() {
+  let input = fn.head(xdmp.unquote('<root><name>foo</name></root>'));
+  let options = {"outputFormat":"xml"}
+  let content = {"uri":"/test/test.xml", "value":input}
+  assertions.push(test.assertEqual('<root xmlns=""><name>foo</name></root>', ing.main(content,options).value.xpath("//*:root")));
```
rjrudin: This is a good test, but I think the assertion should have an explicit xpath without "//" - i.e. "/*:envelope/*:instance/*:root". That does a better job of documenting what's expected.

PR #3457, opened 2020-01-13
DHFPROD-3974: HubConfigImpl no longer has to refresh Spring-managed b…

rjrudin:
…eans

This allows for the new UI project to not deal with HubConfigImpl, which won't be a singleton, having dependencies on singletons managed by Spring. 

In addition, the classes that were extending ResourceManager and being managed by Spring no longer extend ResourceManager. Instead, they create a subclass of ResourceManager when needed, based on the current settings of HubConfig.

### DIFF
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/impl/DataHubImpl.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/impl/DataHubImpl.java
@@ -46,7 +46,6 @@ import com.marklogic.client.io.JacksonHandle;
 import com.marklogic.client.io.QueryOptionsListHandle;
 import com.marklogic.hub.DataHub;
 import com.marklogic.hub.DatabaseKind;
-import com.marklogic.hub.FlowManager;
 import com.marklogic.hub.HubConfig;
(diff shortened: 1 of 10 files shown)


marklogic-data-hub/src/main/java/com/marklogic/hub/impl/DataHubImpl.java
```diff
@@ -133,69 +126,39 @@
     @Autowired
     private Versions versions;
 
-    @Autowired
-    private LegacyFlowManagerImpl legacyFlowManager;
-
-    @Autowired
-    private FlowManager flowManager;
```
rjrudin: Fun fact - these dependencies were both unused.

marklogic-data-hub/src/test/java/com/marklogic/hub/HubTestBase.java
```diff
@@ -156,16 +155,6 @@
     @Autowired
     protected LegacyFlowManagerImpl fm;
 
-    @Autowired
```
rjrudin: All unused stuff here

PR #3459, opened 2020-01-13
DHFPROD-4071: Failure to generate search options no longer fails load…

rjrudin:
…ing modules

The ticket for this didn't specify how search options failed to be generated, only that "{}" was written to an XML file, which of course caused a problem when the options were deployed.

Now, any error occurring while generating search options is logged with a stacktrace. And if an error occurs, the search options files won't be updated.


rjrudin:
@srinathgit @rahulvudutala Just to be clear on the change here:

Currently, if search options fail to generate, the following happens:

1. A stack trace is written to stderr with no context
1. "{}" is written to the options XML files
1. Eventually, then options files are loaded, ML throws an error because "{}" is invalid for XML files (and then ml-gradle usually gets blamed, as that's how things go)

With this change, if search options fail to generate, the following happens:

1. The exception is logged at the "error" level with context 
1. The options files are not updated with invalid content

My point is that we absolutely do not want to write "{}" to XML files, as that just compounds the problem and misleads the user into thinking that module-loading is broken. If search options cannot be generated, then we don't want to update the options files. The best we can do is log the error with some description of what went wrong.

PR #3464, opened 2020-01-14
DHFPROD-4079: dhsDeploy now deploys additional permitted resources

rjrudin:
Moved the guts of this from DataHubImpl to DhsDeployer. The "dhsDeploy" method in the DataHub interface was only used by the DhsDeployTask and thus is safe to remove (it is really an internal API that never should have been exposed in the DataHub interface). 

Removed LoadHubSchemasCommand, which has been OBE for multiple releases. 

Added some constructors to DHF objects so that they can more easily be reused in a non-Spring context, which is the case for DhsDeployer.


marklogic-data-hub/src/main/java/com/marklogic/hub/dhs/DhsDeployer.java
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    /**
+     * This list of commands is based on what a user with the data-hub-developer role is permitted to deploy.
+     *
+     * @param hubConfig
+     * @return
+     */
+    protected List<Command> buildCommandListForDeployingToDhs(HubConfig hubConfig) {
+        List<Command> commands = new ArrayList<>();
```
rjrudin: Most of this class was just moved here from DataHubImpl - the contents of this method are new though.

marklogic-data-hub/src/main/java/com/marklogic/hub/dhs/DhsDeployer.java
```diff
@@ -0,0 +1,182 @@
...
+        if (authMethod != null) {
+            logger.info("Setting security context type for App-Services to: " + authMethod);
+            appConfig.setAppServicesSecurityContextType(SecurityContextType.valueOf(authMethod.toUpperCase()));
+        }
+    }
+
+    /**
+     * The contents of hub-internal-config should have been deployed to DHS already via the DHF installer. And
+     * it is not expected that a user will not have permission to deploy all of the resources in this directory.
+     *
```
bsrikan: can we make the comment simpler. "not expected" followed by "not have" is confusing and makes me re-read it.

bsrikan: Also the method name here implies that we are removing hub-internal-config. But we only are ignoring it for dhsDeploy.

rjrudin: Good points, I made the updates.

PR #3468, opened 2020-01-15
DHFPROD-3755: Changing casing of flow arguments to match hubRunFlow

rjrudin:
I updated the JIRA ticket, as I think it's better for the argument casing to be consistent with hubRunFlow instead of with mlcp.


rjrudin:
Merging this, as the test failure is an intermittent one that is not impacted by these changes.

PR #3470, opened 2020-01-15
DHFPROD-3764: Initial work for new curation UI

ryanjdew:


marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/GenerateFunctionMetadataCommand.java
```diff
@@ -91,4 +91,9 @@ public void execute(CommandContext context) {
             logger.warn("GenerateFunctionMetadataCommand is not supported on this MarkLogic server version ");
         }
     }
+
+    public void setHubConfig(HubConfig hubConfig) {
```
rjrudin: I recommend adding a constructor instead

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/DataHubImpl.java
```diff
@@ -895,10 +895,31 @@ protected void prepareProjectBeforeUpgrading(HubProject hubProject, String curre
     // only used in test
     public void setHubConfig(HubConfigImpl hubConfig) {
         this.hubConfig = hubConfig;
+        if (this.loadUserModulesCommand != null) {
```
rjrudin: I recommend using the constructor for each of these

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/EntityManagerImpl.java
```diff
@@ -35,32 +35,44 @@
...
 @Component
 public class EntityManagerImpl extends LoggingObject implements EntityManager {
     private static final String ENTITY_FILE_EXTENSION = ".entity.json";
 
     @Autowired
     private HubConfig hubConfig;
 
-    @Autowired
-    private HubProject hubProject;
+    private HubProject hubProject = null;
```
rjrudin: I recommend just ditching this, it only complicates things. Just change each reference to hubConfig.getHubProject().

marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/LoadUserModulesCommand.java
```diff
@@ -270,6 +270,12 @@ public FileVisitResult visitFile(Path file, BasicFileAttributes attrs) {
 
     public void setHubConfig(HubConfig hubConfig) {
         this.hubConfig = hubConfig;
+        if (this.entityManager == null) {
+            this.entityManager = new EntityManagerImpl();
+        }
+        if (this.entityManager instanceof EntityManagerImpl) {
```
hao1st: EntityManagerImpl is the only implementation of EntityManager. Maybe we should add the setHubConfig as an new interface method instead of checking the instance type and do the cast...

rjrudin: Setters for dependencies shouldn't be part of an interface, as that exposes the implementation. EntityManagerImpl(HubConfig) should be used instead when setHubConfig is called (though I don't think there should be a setHubConfig on LoadUserModulesCommand either - it should be LoadUserModulesCommand(HubConfig) instead).

PR #3473, opened 2020-01-16
DHFPROD-3156: Increase performance of flow run

ryanjdew:


marklogic-data-hub/src/main/java/com/marklogic/hub/flow/impl/FlowRunnerImpl.java
```diff
@@ -214,7 +214,8 @@ private void initializeFlow(String jobId) {
             jobDocManager = new JobDocManager(hubConfig.newJobDbClient());
         }
         if(threadPool == null || threadPool.isTerminated()) {
-            threadPool = new CustomPoolExecutor(1, 1, 0L, TimeUnit.MILLISECONDS
+            int cores = Runtime.getRuntime().availableProcessors();
+            threadPool = new CustomPoolExecutor(cores/2, cores, 0L, TimeUnit.MILLISECONDS
```
srinathgit: Flowrunner in our implementation, always uses only single thread. I am not sure if increasing threadpool would help. We have to see if the dir that he has chosen for ingestion has a lot of files(or if it is a network storage)  which may cause FileCollector to take a longer time or if there are issues in reporting the state back to the UI.

ryanjdew: @srinathgit Good point on the flow runner only using a single thread, but my profiling/testing still indicated that the threadpool was the point of contention. I was encountering the issue when running a single document through multiple steps. After you pointing out that the flow runner only uses one thread, I wonder if simply setting the thread pool size to 2 would be enough. I think what was happening was that afterExecute would add the next step to the threadPool, but the current step execution was occupying the space.

PR #3476, opened 2020-01-20
DHFPROD-4104:Use DS to perform flow/step CRUD operations

srinathgit:
As per requirement for this task, CRUD operations must be performed first on db and then to project.  

The corresponding PR for CRUD APIs  in one UI repo is
https://project.marklogic.com/repo/projects/PROD/repos/one-ui/pull-requests/23/diff

### DIFF
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/impl/FlowManagerImpl.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/impl/FlowManagerImpl.java
@@ -20,13 +20,13 @@ import com.fasterxml.jackson.core.JsonParseException;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.JsonMappingException;
 import com.fasterxml.jackson.databind.JsonNode;
-import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.marklogic.client.DatabaseClient;
 import com.marklogic.client.ext.helper.LoggingObject;
 import com.marklogic.hub.FlowManager;
 import com.marklogic.hub.HubConfig;
 import com.marklogic.hub.MappingManager;
 import com.marklogic.hub.StepDefinitionManager;
+import com.marklogic.hub.dataservices.ArtifactService;
 import com.marklogic.hub.error.DataHubProjectException;
 import com.marklogic.hub.flow.Flow;
 import com.marklogic.hub.flow.impl.FlowImpl;
@@ -34,12 +34,16 @@ import com.marklogic.hub.step.StepDefinition;
 import com.marklogic.hub.step.impl.Step;
 import com.marklogic.hub.util.json.JSONObject;
 import com.marklogic.hub.util.json.JSONStreamWriter;
+import com.marklogic.hub.util.json.JSONUtils;
 import org.apache.commons.io.FileUtils;
+import org.apache.commons.lang3.StringUtils;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 
-import java.io.*;
-import java.nio.file.Path;
+import java.io.File;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.InputStream;
 import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.LinkedHashMap;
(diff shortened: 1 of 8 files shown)


aebadirad:
Uh. When did this become a requirement?

Write to project first, attempt to deploy, report back errors if they exist. Disk is still where the project 'lives' and should represent the 'latest' saved state.

ryanjdew:
@aebadirad If we go the project first route, we don't have MarkLogic handling the security around the artifacts. For example, if the data service returns Forbidden do we revert the project file or do we just leave them out of sync? 

If we leave them out of sync, the artifacts may appear to work properly to the end user because the server has a different copy. Then the user may decide to add their project copy to version control leading to the infamous "works on my machine" response when things break.

aebadirad:
Leave them out of sync and toss an error. The error maybe simply that their VM housing the server was shut off or they forgot to give themselves a permission on their dev box.

The code/artifacts they alter have to be merged back and is part of a release by the customer.

srinathgit:
@rjrudin ,
I have used DS to insert artifacts. As per requirement, the artifacts have to be written to project so didn't do the second item you mentioned in the comment

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/FlowManagerImpl.java
```diff
@@ -59,37 +65,27 @@
     @Autowired
     private StepDefinitionManager stepDefinitionManager;
 
+    private ArtifactService artifactService;
```
rjrudin: I don't think we want this as a field of the class. Since ArtifactService is bound to a DatabaseClient, storing this as a field binds this class to a DatabaseClient, and thus to a single user. That's the main problem with all the DHF Service/Manager classes - they have HubConfig as a field, and thus they're bound to a single user. 

If the intent was for Service/Manager classes to be Spring-managed singletons, they should be stateless / threadsafe classes. That's unfortunately not the case, due to the HubConfig field. Ideally, HubConfig - or some subset of it - would be passed in as an argument to methods that need it. In the absence of that, an ArtifactService should be instantiated each time it's needed and not stored as a field.

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/FlowManagerImpl.java
```diff
@@ -344,21 +368,14 @@ protected void deleteDocumentsInDirectory(String directory) {
             logger.info(format("Deleting documents in directory '%s' in staging database", directory));
         }
         DatabaseClient stagingClient = hubConfig.newStagingClient();
-        try {
-            stagingClient.newServerEval().javascript(query).evalAs(String.class);
-        } finally {
-            stagingClient.release();
-        }
+        //Since client is getting reused in one UI, client.release() has to be removed
```
rjrudin: @srinathgit  @ryanjdew @aebadirad  I think this is an argument here for a new "getStagingClient()" method instead of "newStagingClient()".  "new" implies that the caller should own the lifecycle of the DatabaseClient, and thus that the caller is free to call "release" on it - i.e. "Give me a new DatabaseClient, and don't worry, I'll release it when I decide I'm done with it". Whereas "get" implies - "Give me your staging DatabaseClient, instantiating it as needed. I recognize it as 'yours', and thus I won't call release on it."

Arguably, "newStagingClient" should have been called "getStagingClient" in a single-user context as well. But my main point is that "new" implies that a new object is being created and thus the caller owns it, whereas "get" implies that HubConfig owns the object being returned.

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/StepDefinitionManagerImpl.java
```diff
@@ -42,13 +44,16 @@
     @Autowired
     private HubConfig hubConfig;
 
+    private ArtifactService artifactService;
```
rjrudin: Same thing - unless this class is intended to be bound to a single user/thread, it should not sure ArtifactService as a field.

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/StepDefinitionManagerImpl.java
```diff
@@ -68,6 +73,7 @@ public void saveStepDefinition(StepDefinition stepDefinition, boolean autoIncrem
         } catch (IOException e) {
             throw new DataHubProjectException("Could not write Step to disk for project.");
         }
+        getArtifactService().setArtifact("stepDefinitions", stepDefinition.getName(),mapper.valueToTree(stepDefinition));
```
rjrudin: The duplication of calling "valueToTree" here to convert an object into a JsonNode suggests the concept of an abstract Artifact class that Mapping, StepDefinition, EntityModel, and Flow all extend. Artifact would have an e.g. "JsonNode toJsonNode()" method on it. That encapsulates the logic of converting an object to a JsonNode into the object itself, as opposed to exposing and duplicating it across N Manager classes.

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/ScaffoldingImpl.java
```diff
@@ -34,19 +34,21 @@
...
 import java.util.HashMap;
 import java.util.Map;
 
+@SessionScope
 @Component
 public class ScaffoldingImpl implements Scaffolding {
 
-    @Autowired
-    private HubProject project;
+    private HubProject project = null;
```
rjrudin: Why even store this as a field? The class already has a dependency on HubConfig. When a HubProject is needed, why not just call hubConfig.getHubProject? Then we don't need a field, and we don't need a PostConstruct method either.

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/StepDefinitionManagerImpl.java
srinathgit: Is it ok to move the convertArtifactToJson method to JSONUtils so that all the Manager class can use it ?

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/FlowManagerImpl.java
srinathgit: I agree with you. "newStagingClient()" in HubConfigImpl returns a new client object every time. In HubConfigSession, it looks like the decision was made to reuse the client. I will leave this in place for now

ryanjdew: Yes, we may want to rename the function in the hubconfig interface (as part of another ticket) and leave it to the implementation to determine if it should create a new client every time. The reasoning behind not creating a new client every time in the session implementation is to avoid holding onto plain text credentials by build all the required clients at time of login.

PR #3478, opened 2020-01-20
DHFPROD-4086: Added example projects to pipeline

SameeraPriyathamTadikonda:


SameeraPriyathamTadikonda:
Yes, Goal is to do a smoke test on the latest snapshots to check if the flows run as expected.

PR #3489, opened 2020-01-22
DHFPROD-4099: Support configurable sessio timeout (5 min as default)

hao1st:
https://project.marklogic.com/jira/projects/DHFPROD/issues/DHFPROD-4099

1. Let the oneui app deal with configurable session timeout (5 min as defauly)
2. Reclaim reclaim hub session resouces once session timeout or user logout
3. Session cookie name is "DHOneUISession", and it can be changed in application.properties
4. Remove unnessary codes and format codes

### DIFF
--- a/one-ui/src/main/java/com/marklogic/hub/oneui/auth/AuthConfig.java
+++ b/one-ui/src/main/java/com/marklogic/hub/oneui/auth/AuthConfig.java
@@ -17,55 +17,42 @@ package com.marklogic.hub.oneui.auth;
 
 import com.marklogic.hub.oneui.models.EnvironmentInfo;
 import com.marklogic.hub.oneui.services.EnvironmentService;
+
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.Lazy;
 import org.springframework.security.config.annotation.authentication.builders.AuthenticationManagerBuilder;
 import org.springframework.security.config.annotation.web.builders.HttpSecurity;
-import org.springframework.security.config.annotation.web.builders.WebSecurity;
 import org.springframework.security.config.annotation.web.configuration.EnableWebSecurity;
 import org.springframework.security.config.annotation.web.configuration.WebSecurityConfigurerAdapter;
 import org.springframework.security.web.authentication.UsernamePasswordAuthenticationFilter;
 
 /**
- * Extends Spring Boot's default web security configuration class and hooks in MarkLogic-specific classes from
- * marklogic-spring-web. Feel free to customize as needed.
+ * Extends Spring Boot's default web security configuration class and hooks in MarkLogic-specific
+ * classes from marklogic-spring-web. Feel free to customize as needed.
  */
 @Configuration
 @EnableWebSecurity
 public class AuthConfig extends WebSecurityConfigurerAdapter {
+    @Autowired private EnvironmentService environmentService;
 
-    @Autowired
-    EnvironmentService environmentService;
+    @Autowired private LoginLogoutHandler loginLogoutHandler;
 
-    private LoginLogoutHandler loginLogoutHandler = new LoginLogoutHandler();
+    @Autowired private RestAuthenticationEntryPoint restAuthenticationEntryPoint;
 
-    /**
-     * @return a config class with ML connection properties
-     */
+    @Lazy @Autowired private ConnectionAuthenticationFilter authFilter;
+
+    @Autowired public MarkLogicAuthenticationManager markLogicAuthenticationManager;
+
+    /** @return a config class with ML connection properties */
     @Bean
     public EnvironmentInfo environmentInfo() {
         return environmentService.getEnvironment();
     }
 
-    @Autowired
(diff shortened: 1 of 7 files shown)


one-ui/src/main/java/com/marklogic/hub/oneui/models/HubConfigSession.java
```diff
@@ -990,7 +1007,18 @@ private DatabaseClient getClientByDatabaseKindAndName(DatabaseKind kind, String
         return client;
     }
 
-    public DataHub getDataHub() {
-        return this.dataHub;
+    /**
```
srinathgit: Is this code required ? Just configuring session timeout property not sufficient ? I would assume as HubConfigSession is a session scoped bean , just invalidating session would eventually destroy these objects ?

one-ui/src/main/java/com/marklogic/hub/oneui/auth/AuthConfig.java
```diff
@@ -17,55 +17,42 @@
...
 @EnableWebSecurity
 public class AuthConfig extends WebSecurityConfigurerAdapter {
+    @Autowired private EnvironmentService environmentService;
 
-    @Autowired
-    EnvironmentService environmentService;
+    @Autowired private LoginLogoutHandler loginLogoutHandler;
 
-    private LoginLogoutHandler loginLogoutHandler = new LoginLogoutHandler();
+    @Autowired private RestAuthenticationEntryPoint restAuthenticationEntryPoint;
```
rjrudin: Do you have your IDE formatter configured to put annotations on the same line as the field they refer to? I have not seen this convention before, and it seems most of the DHF code has annotations on lines above the field/method that they refer to - the "@Test" annotation being the most obvious one. Unless you have strong feelings about this, can you ensure that your IDE will keep annotations above the fields? Otherwise, we'll have a lot of changes in PRs just due to a different formatter being used, which requires reviewers to spend more time trying to understand what changed.

one-ui/src/main/java/com/marklogic/hub/oneui/auth/LoginLogoutHandler.java
```diff
@@ -19,36 +19,47 @@
...
 
+@Controller
 public class LoginLogoutHandler implements AuthenticationSuccessHandler, LogoutSuccessHandler {
 
-    @Override
-    public void onAuthenticationSuccess(HttpServletRequest request, HttpServletResponse response, Authentication authentication) throws IOException {
-        ConnectionAuthenticationToken authenticationToken = (ConnectionAuthenticationToken)authentication;
+  @Override
+  public void onAuthenticationSuccess(
+      HttpServletRequest request, HttpServletResponse response, Authentication authentication)
```
rjrudin: It seems your formatter may also have a fairly low limit for how long a line can be? I think 120 is the default. Again, there are a lot of changes here that are due just to formatting changes, making it more difficult to understand what the real changes are.

one-ui/src/main/java/com/marklogic/hub/oneui/auth/MarkLogicAuthenticationManager.java
```diff
@@ -90,62 +85,41 @@ public Authentication authenticate(Authentication authentication) throws Authent
...
+        */
                 throw ex;
             }
         }
         DatabaseClient dataServicesClient = hubConfig.newStagingClient(null);
         boolean stagingServerAccessible = false;
         try {
-            stagingServerAccessible = dataServicesClient.checkConnection().isConnected();
+            stagingServerAccessible =
+                dataServicesClient.newServerConfigManager().getQueryValidation() != null;
```
rjrudin: Why is this being used instead of "isConnected"? I believe checkConnection was added for that exact purpose.

one-ui/src/main/java/com/marklogic/hub/oneui/models/HubConfigSession.java
rjrudin: I think code like this is a great argument for a HubClient class that only contains DatabaseClient objects and none of the original state. "reset" methods to me are an anti-pattern, particularly because it means that any time more fields are added to an object, someone has to remember to consider whether the new fields need to be reset or not.

hao1st: Actually, I do not think we need to add "reset" here after I checked in...

one-ui/src/main/java/com/marklogic/hub/oneui/auth/AuthConfig.java
hao1st: I used the google style and not sure why only this file annotation was formatted with the same line. I will turn off the format and redo what I have done.

one-ui/src/main/java/com/marklogic/hub/oneui/auth/MarkLogicAuthenticationManager.java
hao1st: Oh, my change was based on an older version. I will delete the PR and recreate a new one.

one-ui/src/main/java/com/marklogic/hub/oneui/auth/AuthConfig.java
hao1st: I will add some tests too.

PR #3490, opened 2020-01-22
DHFPROD-4104:Middle tier for flow/step CRUD

srinathgit:
Middle tier for flow/step CRUD


one-ui/src/main/java/com/marklogic/hub/curation/controllers/FlowController.java
```diff
@@ -0,0 +1,115 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.http.HttpStatus;
+import org.springframework.http.ResponseEntity;
+import org.springframework.scheduling.annotation.EnableAsync;
+import org.springframework.stereotype.Controller;
+import org.springframework.web.bind.annotation.*;
+
+@Controller
+@RequestMapping("/api/flows")
+@EnableAsync
```
hao1st: we can remove @EnableAsync for the implmentation.

rjrudin: That sounds right. Also, if we did need it, I think a comment would be very helpful. Spring annotations can be powerful, but they often beg the question - why is this annotation here?

one-ui/src/main/java/com/marklogic/hub/curation/controllers/FlowController.java
```diff
@@ -0,0 +1,115 @@
...
+
+    @Autowired
+    private FlowManagerService flowManagerService;
+
+    @RequestMapping(method = RequestMethod.POST)
+    @ResponseBody
+    public ResponseEntity<?> createFlow(@RequestBody String flowJson) {
+        Flow flow;
+        try {
+            flow = flowManagerService.createFlow(flowJson, false);
```
rjrudin: Shouldn't the service throw a useful exception here if the payload is invalid? The controller should be as thin as possible, just passing inputs and outputs around. I think all of these methods should roughly be:

    return new ResponseEntity<>(service.createFlow(flowJson, false), HttpStatus.OK);

one-ui/src/main/java/com/marklogic/hub/curation/controllers/FlowController.java
```diff
@@ -0,0 +1,115 @@
...
+        }
+        return new ResponseEntity<>(flow, HttpStatus.OK);
+    }
+
+    @RequestMapping(value = "/{flowName}", method = RequestMethod.PUT)
+    @ResponseBody
+    public ResponseEntity<?> updateFlow(@PathVariable String flowName, @RequestBody String flowJson) {
+        Flow flow;
+        try {
+            flow = flowManagerService.createFlow(flowJson, true);
```
rjrudin: Same thing here, createFlow should throw a useful exception. There shouldn't be any need for the controller to do anything other than call the service method.

one-ui/src/main/java/com/marklogic/hub/curation/services/FlowManagerService.java
```diff
@@ -0,0 +1,288 @@
...
+    @Autowired
+    private StepDefinitionManager stepDefinitionManager;
+
+    @Autowired
+    private Scaffolding scaffolding;
+
+    public Flow createFlow(String flowJson, boolean isFlowPresent) {
+        JSONObject jsonObject;
+        try {
+            jsonObject = new JSONObject(flowJson);
```
rjrudin: What's the point of using JSONObject? Why not just accept a JsonNode instead? The Spring controller can convert the RequestBody automatically into a JsonNode as well. That will cut out a lot of code and make for a richer method interface, as createFlow(JsonNode) is better than createFlow(String).

one-ui/src/main/java/com/marklogic/hub/curation/services/FlowManagerService.java
```diff
@@ -0,0 +1,288 @@
...
+            throw new DataHubException("Unable to parse flow json string : " + e.getMessage());
+        }
+
+        if (!jsonObject.isExist("name") || StringUtils.isEmpty(jsonObject.getString("name"))) {
+            throw new BadRequestException("Flow Name not provided. Flow Name is required.");
+        }
+
+        String flowName = jsonObject.getString("name");
+        Flow flow;
+        if (!isFlowPresent) {
```
rjrudin: This all seems logic that would go into the DS endpoint. @aebadirad What are your thoughts on that? That would require calling ML first. But it would get rid of a ton of code in Java and move it into a more interesting DS endpoint that can see if the flow exists already, and create it if not or update it otherwise.

one-ui/src/main/java/com/marklogic/hub/curation/services/FlowManagerService.java
```diff
@@ -0,0 +1,288 @@
...
+    @Autowired
+    private FlowManager flowManager;
+
+    @Autowired
+    private StepDefinitionManager stepDefinitionManager;
+
+    @Autowired
+    private Scaffolding scaffolding;
+
+    public Flow createFlow(String flowJson, boolean isFlowPresent) {
```
rjrudin: I would think the client shouldn't have to tell the service whether the flow exists yet or not - the service should know how to figure that out. The client just wants to say - here's my JSON, please save it.

one-ui/src/main/java/com/marklogic/hub/oneui/models/StepModel.java
```diff
@@ -0,0 +1,307 @@
...
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.node.JsonNodeFactory;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import com.marklogic.hub.step.StepDefinition;
+import com.marklogic.hub.step.impl.Step;
+import com.marklogic.hub.util.json.JSONObject;
+
+import java.util.Iterator;
+
+public class StepModel {
```
rjrudin: What's the benefit of this class? Can't we just pass a JsonNode between the browser and ML? That would significantly reduce the code in the Java middle tier and put a lot more smarts into ML. And we wouldn't need to convert a JsonNode into some other class and then back to a JsonNode.

one-ui/src/main/java/com/marklogic/hub/curation/services/FlowManagerService.java
srinathgit: JSONObject (com.marklogic.hub.util.json) is a utility class that is composed of JsonNode  and has a bunch of other utility methods

one-ui/src/main/java/com/marklogic/hub/oneui/models/StepModel.java
srinathgit: The UI doesn't know anything about step definition. So, when some one creates a custom step, we have to create a custom step definition, main.sjs in addition to the step itself that goes inside the flow. So, the core and web models for steps are different, webModel (called StepModel.java) has 'modulePath' which provides the uri of the main.sjs
whereas it is not present in the core step model (Step.java). This is all inherited from 5.0 code codebase .

PR #3495, opened 2020-01-23
DHFPROD-3990:Display project information in UI

srinathgit:
DHFPROD-3990:Display project information in UI

### DIFF
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/impl/HubProjectImpl.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/impl/HubProjectImpl.java
@@ -646,6 +646,11 @@ public class HubProjectImpl implements HubProject {
         }
     }
 
+    @Override
+    public String getProjectName() {
+        return getProjectDir().toFile().getName();
+    }
+
     private void zipSubDirectory(String basePath, File fileToZip, ZipOutputStream zout) throws IOException {
         File[] files = fileToZip.listFiles();
         if(files != null) {
--- a/one-ui/src/main/java/com/marklogic/hub/oneui/auth/LoginLogoutHandler.java
+++ b/one-ui/src/main/java/com/marklogic/hub/oneui/auth/LoginLogoutHandler.java
@@ -15,6 +15,8 @@
  */
 package com.marklogic.hub.oneui.auth;
 
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ObjectNode;
 import org.springframework.security.core.Authentication;
 import org.springframework.security.web.WebAttributes;
 import org.springframework.security.web.authentication.AuthenticationSuccessHandler;
@@ -31,10 +33,15 @@ public class LoginLogoutHandler implements AuthenticationSuccessHandler, LogoutS
     @Override
     public void onAuthenticationSuccess(HttpServletRequest request, HttpServletResponse response, Authentication authentication) throws IOException {
         ConnectionAuthenticationToken authenticationToken = (ConnectionAuthenticationToken)authentication;
+        ObjectMapper mapper = new ObjectMapper();
+        ObjectNode resp = mapper.createObjectNode();
+        resp.put("isInstalled", authenticationToken.stagingIsAccessible());
+        resp.put("hasManagePrivileges", authenticationToken.hasManagePrivileges());
+        resp.put("projectName", (String) request.getSession().getAttribute("projectName"));
 
         clearAuthenticationAttributes(request);
         response.setContentType("application/json");
-        response.getOutputStream().write(("{\"isInstalled\": " + authenticationToken.stagingIsAccessible() + ", \"hasManagePrivileges\": " + authenticationToken.hasManagePrivileges() + " }").getBytes());
+        response.getOutputStream().write(mapper.writeValueAsBytes(resp));
     }
 
     private void clearAuthenticationAttributes(HttpServletRequest request) {
--- a/one-ui/src/main/java/com/marklogic/hub/oneui/controllers/EnvironmentController.java
+++ b/one-ui/src/main/java/com/marklogic/hub/oneui/controllers/EnvironmentController.java
@@ -122,4 +122,10 @@ public class EnvironmentController {
         obj.put("reset", true);
         return obj;
     }
+
+    @RequestMapping(value = "/project-info", method = RequestMethod.GET)
+    @ResponseBody
+    public JsonNode getProjectInfo() {
+        return environmentService.getProjectInfo();
+    }
 }
--- a/one-ui/src/main/java/com/marklogic/hub/oneui/services/EnvironmentService.java
+++ b/one-ui/src/main/java/com/marklogic/hub/oneui/services/EnvironmentService.java
@@ -15,20 +15,28 @@
  */
 package com.marklogic.hub.oneui.services;
 
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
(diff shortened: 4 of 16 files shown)


one-ui/src/main/java/com/marklogic/hub/oneui/controllers/EnvironmentController.java
```diff
@@ -122,4 +127,18 @@ public JsonNode reset() {
...
     }
+
+    @RequestMapping(value = "/project-info", method = RequestMethod.GET)
+    @ResponseBody
+    public JsonNode getProjectInfo(HttpSession session) {
+        ObjectMapper objectMapper = new ObjectMapper();
+        ObjectNode resp = objectMapper.createObjectNode();
+        resp.put("dhfVersion", versions.getHubVersion());
+        String projDir = hubConfig.getHubProject().getProjectDir().toString();
+        projDir = projDir.substring(projDir.lastIndexOf(File.separator)+1);
```
rjrudin: This logic is repeated above. Seems like HubProject should have a "getProjectName()" method, assuming that project name should always equal the name of the directory? 

Also, since getProjectDir returns a Path, it seems better to call toFile().getName() instead of messing around with Strings.

one-ui/src/main/java/com/marklogic/hub/oneui/auth/MarkLogicAuthenticationManager.java
```diff
@@ -113,8 +114,12 @@ public Authentication authenticate(Authentication authentication) throws Authent
             stagingServerAccessible = dataServicesClient.checkConnection().isConnected();
         } catch (Exception e) {
         }
+
+        String projDir = environmentService.getProjectDirectory();
+        projDir = projDir.substring(projDir.lastIndexOf(File.separator) + 1);
```
rjrudin: This is really "project name", right? Not the project directory?

srinathgit: Yeah, but we don't have a "project name" currently. I talked to @ryanjdew  and he suggested that we use the project directory for now.

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/HubProjectImpl.java
```diff
@@ -646,6 +646,11 @@ public void exportProject(File location) {
         }
     }
 
+    @Override
```
rjrudin: I commented on the JIRA ticket that I think we should get rid of this from the mockup. We don't have a concept of a "Project Name", and we shouldn't invent one during development. I think the requirement needs to be removed from the story. 

I can see an argument for the project directory, though I don't think that really needs to be there either.

one-ui/src/main/java/com/marklogic/hub/oneui/auth/LoginLogoutHandler.java
```diff
@@ -28,13 +32,21 @@
...
+    HubConfigSession hubConfig;
+
     @Override
     public void onAuthenticationSuccess(HttpServletRequest request, HttpServletResponse response, Authentication authentication) throws IOException {
         ConnectionAuthenticationToken authenticationToken = (ConnectionAuthenticationToken)authentication;
+        ObjectMapper mapper = new ObjectMapper();
+        ObjectNode resp = mapper.createObjectNode();
+        resp.put("isInstalled", authenticationToken.stagingIsAccessible());
+        resp.put("hasManagePrivileges", authenticationToken.hasManagePrivileges());
+        resp.put("projDir", hubConfig.getHubProject().getProjectName());
```
rjrudin: If we keep this, should call this "projectName", not "projDir"

one-ui/src/main/java/com/marklogic/hub/oneui/controllers/EnvironmentController.java
```diff
@@ -122,4 +126,16 @@ public JsonNode reset() {
...
         return obj;
     }
+
+    @RequestMapping(value = "/project-info", method = RequestMethod.GET)
+    @ResponseBody
+    public JsonNode getProjectInfo(HttpSession session) {
+        ObjectMapper objectMapper = new ObjectMapper();
+        ObjectNode resp = objectMapper.createObjectNode();
+        resp.put("dhfVersion", versions.getHubVersion());
+        resp.put("projName", hubConfig.getHubProject().getProjectName());
```
rjrudin: Let's spell these out, there's no benefit in shortening them - so "projectName" and "projectPath" (I think "path" is better than "dir" here, particularly because the output of getProjectDir is a Path (which means that method should really be called getProjectPath, but oh well)).

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/HubProjectImpl.java
srinathgit: I saw your comment and response from @sbayatpur on it. The suggestion from her is to use proj dir name for now. I think we can change it once she decides on 'Project Name'.

SameeraPriyathamTadikonda:
Please rebase this branch to resolve merge conflicts

one-ui/src/main/java/com/marklogic/hub/oneui/controllers/EnvironmentController.java
```diff
@@ -122,4 +126,16 @@ public JsonNode reset() {
         obj.put("reset", true);
         return obj;
     }
+
+    @RequestMapping(value = "/project-info", method = RequestMethod.GET)
+    @ResponseBody
+    public JsonNode getProjectInfo(HttpSession session) {
```
rjrudin: What about tests for this? We'd at least want a smoke test to verify that there's not an e.g. obvious null-pointer exception. 

Also, I think @hao1st  mentioned this too - based on the current design of oneui, couldn't HubProject be part of EnvironmentService? That would start removing one concern from HubConfig, which is overloaded with concerns right now. I think this would be a good move, as the HubProject is a global object that is part of the environment - it's not specific to a user.

srinathgit: 1. I will add a test .
2. I don't think we can autowire HubProject in EnvironmentService. The first object created in one-ui is HubConfigSession and that is what we are using through out and that's why I had to access HubProject from HubConfigSession

one-ui/src/main/java/com/marklogic/hub/oneui/services/EnvironmentService.java
```diff
@@ -15,21 +15,29 @@
...
+import java.io.*;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.prefs.Preferences;
 
 @Service
 public class EnvironmentService {
 
+    @Autowired
+    private HubConfigSession hubConfig;
```
rjrudin: Couldn't you depend on HubProject instead, since that's not specific to a user?

one-ui/src/main/java/com/marklogic/hub/oneui/controllers/EnvironmentController.java
rjrudin: If HubProjectImpl is autowired in, then it's always available, right?

I'll approve the PR, I just feel strongly that we need to stop using HubConfig, which is doing far too much, and use HubProject and a doesnt-yet-exist HubClient when possible.

one-ui/src/main/java/com/marklogic/hub/oneui/services/EnvironmentService.java
srinathgit: No, I tried autowiring HubProject object here but its fields (projDir etc) are not populated and all are null and had to get  the project from HubConfigSession

PR #3497, opened 2020-01-24
DHFPROD-4158: Can now deploy to DHS via multiple roles

rjrudin:
Most of the files in here are for a new example project - dhs-example. 

This is now simpler as we're no longer in the business of eagerly setting DHS-specific properties, like basic auth and SSL. It'll still be up to the user to set those in their gradle-dhs.properties file.

### DIFF
--- a/examples/dhs-example/README.md
+++ b/examples/dhs-example/README.md
@@ -1,0 +1,103 @@
+This example project demonstrates where to store resources that can be deployed to a DHS instance using a user with the 
+data-hub-developer role, or the data-hub-security-admin role, or both. 
+
+This project is not meant to be deployed. It is intended only to serve as a reference for where resource files
+should be stored. Because of this, it also does not define any Data Hub artifacts, such as flows or entities.
+
+## General deployment guidelines
+
+The tasks for deploying resources to DHS will never read from the src/main/hub-internal-config directory. The 
+resources in that directory are deployed automatically within DHS. For example, if you have defined your configuration
+directories via the following property:
+
+    mlConfigPaths=src/main/hub-internal-config,src/main/ml-config,src/main/other-config
+    
+The ml-config and other-config directories will still be processed, but hub-internal-config will not be.
+
+## Deploying resources as a data-hub-security-admin user
+
+As of 5.2.0, a user with the data-hub-security-admin role is permitted to deploy roles that grant privileges that are
+inherited by the user performing the deployment. As an example of this, the 
+src/main/ml-config/security/roles/custom-role1.json file defines a new role with the "manage" privilege, which is 
+inherited by the data-hub-security-admin role. 
+
+Permitted resources can be deployed via the following task (assuming that gradle-dhs.properties defines the host and
+credentials for the DHS instance):
+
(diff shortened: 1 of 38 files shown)


examples/dhs-example/README.md
```diff
@@ -0,0 +1,97 @@
...
+
+As of 5.2.0, a user with the data-hub-security-admin role is permitted to deploy roles that grant privileges that are
+inherited by the user performing the deployment. As an example of this, the 
+src/main/ml-config/security/roles/custom-role1.json file defines a new role with the "manage" privilege, which is 
+inherited by the data-hub-security-admin role. 
+
+Permitted resources can be deployed via the following task (assuming that gradle-dhs.properties defines the host and
+credentials for the DHS instance):
+
+    ./gradlew -PenvironmentName=dhs dhsDeployAsSecurityAdmin
```
bsrikan: Should this be changed to hubDeployAsSecurityAdmin?

examples/dhs-example/README.md
```diff
@@ -0,0 +1,97 @@
...
+
+## Scheduled tasks
+
+Because a DHS instance uses multiple groups for application servers, and because a scheduled task is associated with a
+group, scheduled task files are stored in the directory "./tasks/(name of group)". See 
+src/main/ml-config/tasks for an example in this project. Also see
+[the ml-gradle docs](https://github.com/marklogic-community/ml-gradle/wiki/Resource-reference#scheduled-tasks) for more information.
+
+## Temporal axes and collections
+
```
bsrikan: Is it worth mentioning here that lsqt is not supported?

ml-data-hub-plugin/src/main/groovy/com/marklogic/gradle/DataHubPlugin.groovy
```diff
@@ -190,10 +184,18 @@ class DataHubPlugin implements Plugin<Project> {
...
+
+        project.task("hubDeployAsDeveloper", group: deployGroup, type: DeployAsDeveloperTask,
+            description: "Deploy project configuration as a user with the data-hub-developer role"
+        ).finalizedBy(["hubGenerateExplorerOptions"]).mustRunAfter("hubDeployAsSecurityAdmin")
+
+        project.task("hubDeploy", group: deployGroup, dependsOn: ["hubDeployAsDeveloper", "hubDeployAsSecurityAdmin"],
+            description: "Deploy project configuration as a user with the data-hub-security-admin and data-hub-developer roles")
+
+        project.task("dhsDeploy", dependsOn: ["hubDeploy"],
+            description: "dhsDeploy has been replaced in 5.2.0 by hubDeploy and similar tasks. It is now simply an alias for hubDeploy.")
```
bsrikan: nice addition.

examples/dhs-example/README.md
rjrudin: Good catch, made the fix.

examples/dhs-example/README.md
rjrudin: That's a good idea,  I added mention of that.

PR #3504, opened 2020-01-24
DHFPROD-4099: Support configurable session timeout (5 min as default)

hao1st:


one-ui/src/main/java/com/marklogic/hub/oneui/auth/LoginLogoutHandler.java
```diff
@@ -19,13 +19,15 @@
...
 import org.springframework.security.web.authentication.logout.LogoutSuccessHandler;
+import org.springframework.stereotype.Controller;
 
 import javax.servlet.ServletException;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import javax.servlet.http.HttpSession;
 import java.io.IOException;
 
+@Controller
```
rjrudin: Is there are more appropriate annotation to use here? Perhaps just "Component"? If that has the same effect, I think that'd be better because this class of course is not a controller.

one-ui/src/main/java/com/marklogic/hub/oneui/models/HubConfigSession.java
```diff
@@ -1002,4 +1008,19 @@ public void afterPropertiesSet() throws Exception {
         hubConfigImpl.hydrateConfigs();
         this.dataHub = new DataHubImpl(this);
     }
+
+    @Override
+    public void destroy() {
```
rjrudin: There needs to be a test for this method. Otherwise, someone could add code that causes a null pointer exception or something similar, and we'd never know it. 

I think a suitable test would be to populate that map with two DatabaseClient objects, and then call destroy. Then verify that the map is empty (I can't think of a good way to verify that the clients were released). That's a decent smoke test.

one-ui/src/main/java/com/marklogic/hub/oneui/auth/LoginLogoutHandler.java
hao1st: Actually, i think the handler should be a controller. It can be refactor into the com.marklogic.hub.oneui.controllers package to make it more clear.

one-ui/src/main/java/com/marklogic/hub/oneui/models/HubConfigSession.java
hao1st: @rjrudin A new test class was created for checking the function.

one-ui/src/test/java/com/marklogic/hub/oneui/TestHelper.java
```diff
@@ -23,7 +23,7 @@
     public String dataHubDeveloperPassword;
 
     @Autowired
-    private HubConfigSession hubConfig;
+    public HubConfigSession hubConfig;
```
srinathgit: I think we can avoid using public class variables. You can autowire HubConfigSession or use a getter

hao1st: @srinathgit I just changed it based on your suggestion.

one-ui/src/main/java/com/marklogic/hub/oneui/models/HubConfigSession.java
```diff
@@ -1002,4 +1008,24 @@ public void afterPropertiesSet() throws Exception {
...
+                        logger.debug(String.format("release %s (%s)", e.getDatabase(), e.toString()));
+                        e.release();
+                        e = null;
+                    });
+            clientsByKindAndDatabaseName.clear();
+        }
+    }
+
+    //only for test purpose
+    public Map<DatabaseKind, Map<String, DatabaseClient>> getAllDatabaseClients() {
```
srinathgit: This method can be "protected" and test can be added to same package

hao1st: @srinathgit good suggestion...

one-ui/src/main/java/com/marklogic/hub/oneui/auth/LoginLogoutHandler.java
srinathgit: I agree with @rjrudin . I think it should be a "Component" and not a "Controller".

hao1st: OK, i do not have strong opinion with which one we should use for the spring annotation. I will check it because of majority takes over.

one-ui/src/main/java/com/marklogic/hub/oneui/models/HubConfigSession.java
rjrudin: Just a note - this logic is the sort of thing that should eventually move into a HubClient class, where it'd be a lot easier to test - i.e. create a HubClient with 3 DatabaseClients, then call destroy, then verify that each of the 3 DatabaseClients throws an error if you try to do something with it.

PR #3505, opened 2020-01-24
WIP DHFPROD-4152: Matching and merging steps now support permissions

rjrudin:
@ryanjdew The initial scope of this was to ensure that the options.permissions property is honored by the matching/merging steps. I believe that's fixed.

Then I tried to run MasterTest as a data-hub-operator - i.e. someone without any default permissions. I ran into a number of problems where the SM code is inserting a document and only calling "xdmp:default-permissions()". 

I started to add a hack of "default" perms of data-hub-operator read/update, but I wanted to get your thoughts first. 

Is it correct that SM requires a user to have default permissions? In my experience, that's not common for users to have - normally the application code determines what permissions to add. That would leave two options:

1. Use a "default" set of perms in addition to default permissions (easy enough to do)
1. Add support for configuring the perms for various documents inserted by SM (a lot more effort)

Thoughts?

### DIFF
--- a/examples/dh-5-example/flows/ingestion_mapping_mastering-flow.flow.json
+++ b/examples/dh-5-example/flows/ingestion_mapping_mastering-flow.flow.json
@@ -17,7 +17,7 @@
         "collections": [
           "mastering-flow-ingest-json"
         ],
-        "permissions": "rest-reader,read,rest-writer,update",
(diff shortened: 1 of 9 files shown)


ryanjdew:
@rjrudin I don't see an issue with the default set of permissions approach.

PR #3508, opened 2020-01-27
DHFPROD-4222:Create endpoints for getting all flows, all steps in a flow

srinathgit:
Create endpoints for getting all flows, all steps in a flow

### DIFF
--- a/one-ui/src/main/java/com/marklogic/hub/curation/services/FlowManagerService.java
+++ b/one-ui/src/main/java/com/marklogic/hub/curation/services/FlowManagerService.java
@@ -4,9 +4,9 @@ import com.fasterxml.jackson.databind.JsonNode;
 import com.marklogic.hub.FlowManager;
 import com.marklogic.hub.StepDefinitionManager;
 import com.marklogic.hub.error.DataHubProjectException;
-import com.marklogic.hub.oneui.exceptions.DataHubException;
 import com.marklogic.hub.flow.Flow;
 import com.marklogic.hub.flow.impl.FlowImpl;
+import com.marklogic.hub.oneui.exceptions.DataHubException;
 import com.marklogic.hub.oneui.models.StepModel;
 import com.marklogic.hub.scaffold.Scaffolding;
 import com.marklogic.hub.step.StepDefinition;
@@ -23,8 +23,7 @@ import javax.ws.rs.BadRequestException;
 import javax.ws.rs.NotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
-import java.util.LinkedHashMap;
-import java.util.Map;
+import java.util.*;
 
 @Service
 public class FlowManagerService {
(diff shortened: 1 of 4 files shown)


one-ui/src/main/java/com/marklogic/hub/curation/services/FlowManagerService.java
```diff
@@ -100,6 +99,30 @@ public void deleteFlow(String flowName) {
         flowManager.deleteFlow(flowName);
     }
 
+    public List<Flow> getFlows() {
```
rjrudin: Shouldn't this method be on FlowManager? FlowManager has knowledge of how to read files and get flow names, so I think it should know how to return a list of flows as well (and sort them too).

one-ui/src/main/java/com/marklogic/hub/curation/services/FlowManagerService.java
```diff
@@ -100,6 +99,30 @@ public void deleteFlow(String flowName) {
...
+        List<String> flowNames = flowManager.getFlowNames();
+        List<Flow> flows = new ArrayList<>();
+        flowNames.forEach((flowName)->{
+            flows.add(getFlow(flowName));
+        });
+        Collections.sort(flows, (a, b) -> a.getName().compareTo(b.getName()));
+        return flows;
+    }
+
+    public List<StepModel> getSteps(String flowName) {
```
rjrudin: What's the point of having separate Step and StepModel classes? This seems confusing and is unnecessary work of having to translate between the two. 

Also, couldn't a "getFlow" method just return a JsonNode, in which case this method isn't even needed?

Isn't life a lot easier if we just pass JSON around??

srinathgit: The UI doesn't know anything about step definition. So, when some one creates a custom step, we have to create a custom step definition, main.sjs in addition to the step itself that goes inside the flow. So, the core and web models for steps are different, webModel (called StepModel.java) has 'modulePath' which provides the uri of the main.sjs
whereas it is not present in the core step model (Step.java). This is all inherited from 5.0 code codebase .

ryanjdew: @rjrudin As to the comment about using JsonNode, the goal is to move to generic JSON in most cases. I think Flows/Steps are unique given the need for Java to really understand the values for ingestion, so I think that is an outside case where it is better to have solid classes with a more rigid schema.

rjrudin: I don't understand why the UI doesn't know anything about a step definition. "Step" and "Step Definition" are distinct concepts in the DHF domain model; a client should understand both of them. 

Is "StepModel" really a synonym for "StepDefinition"? If so, can StepModel be renamed to StepDefinition?

I think the core problem here is if someone looks at the Step and StepModel classes, there's no way to know why they both exist. Perhaps the immediate solution is class comments to explain what concept each class represents, and in the StepModel case, why it differs from Step. Those can be in a separate PR; I'll stand down on this point for now.

PR #3509, opened 2020-01-27
DHFPROD-4095: Provide role information and better error output on login to UI

ryanjdew:


one-ui/src/main/java/com/marklogic/hub/oneui/auth/LoginFailureHandler.java
```diff
@@ -40,7 +41,13 @@ public void onAuthenticationFailure(HttpServletRequest request, HttpServletRespo
         String json = new Gson().toJson(output);
         response.setContentType("application/json");
         response.setCharacterEncoding("UTF-8");
-        response.setStatus(HttpServletResponse.SC_UNAUTHORIZED);
+        if (exception.getMessage().contains("Cannot connect to MarkLogic")) {
```
rjrudin: This seems brittle - an "instanceof" check like below would be a lot better - more reliable and self-documenting. Can the custom code that creates this message throw a better exception for this class to catch?

one-ui/src/main/java/com/marklogic/hub/oneui/auth/LoginLogoutHandler.java
```diff
@@ -27,14 +33,24 @@
...
     @Override
     public void onAuthenticationSuccess(HttpServletRequest request, HttpServletResponse response, Authentication authentication) throws IOException {
         ConnectionAuthenticationToken authenticationToken = (ConnectionAuthenticationToken)authentication;
-
+        ObjectNode output = mapper.createObjectNode();
         clearAuthenticationAttributes(request);
+        output.put("isInstalled", authenticationToken.stagingIsAccessible());
+        output.put("hasManagePrivileges", authenticationToken.hasManagePrivileges());
+        if (authenticationToken.stagingIsAccessible()) {
+            RolesService rolesService = RolesService.on(authenticationToken.getHubConfigSession().newStagingClient());
```
hao1st: https://github.com/marklogic/marklogic-data-hub/blob/f1305eb6fb34b989023422a53aca5523360ffaf7/one-ui/src/main/java/com/marklogic/hub/oneui/auth/LoginLogoutHandler.java#L49

Since we autowired the hubConfigSession, we do not need to access it through authentication token.

ryanjdew: I should remove the autowired hubConfig. I was getting a null hubConfig when I tried to use the autowired version.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/roles/getRoles.sjs
```diff
@@ -0,0 +1,45 @@
...
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+'use strict';
+
+const dataHubRoles = [
+    'data-hub-admin',
```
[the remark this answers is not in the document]
ryanjdew: changed to look for startsWith "data-hub".

one-ui/src/main/java/com/marklogic/hub/oneui/auth/LoginFailureHandler.java
ryanjdew: changed to use instanceof BadRequestException

one-ui/src/main/java/com/marklogic/hub/oneui/auth/MarkLogicAuthenticationManager.java
```diff
@@ -88,7 +90,7 @@ public Authentication authenticate(Authentication authentication) throws Authent
             hasManagePrivileges = true;
         }
         catch(ResourceAccessException ex) {
-            throw new BadCredentialsException("Cannot connect to MarkLogic at " + environmentInfo.mlHost + ". Are you sure MarkLogic is running?");
+            throw new BadRequestException("Cannot connect to MarkLogic at " + environmentInfo.mlHost + ". Are you sure MarkLogic is running?");
```
rjrudin: I recommend tucking the original exception in here (as a 2nd argument to the constructor), just in case. It may be useful to at least see the message.

one-ui/src/main/java/com/marklogic/hub/oneui/auth/MarkLogicAuthenticationManager.java
```diff
@@ -107,10 +109,13 @@ public Authentication authenticate(Authentication authentication) throws Authent
         boolean stagingServerAccessible = false;
         try {
             stagingServerAccessible = dataServicesClient.checkConnection().isConnected();
-        } catch (Exception e) {
+        } catch (Exception ignored) {
```
hao1st: i think we should not ignore the exception after checking the internal implementation of checkConnection().

```suggestion
    DatabaseClient.ConnectionResult connResult;
    try {
      connResult = dataServicesClient.checkConnection();
    } catch (Exception e) {
      throw new InternalAuthenticationServiceException(e.getMessage());
    }

    if (connResult != null && !connResult.isConnected()) {
      if (connResult.getStatusCode() != null && connResult.getStatusCode() == 401) {
        throw new BadCredentialsException(connectionResult.getErrorMessage());
      } else {
        throw new InternalAuthenticationServiceException(connResult.getErrorMessage());
      }
    }
```

one-ui/src/main/java/com/marklogic/hub/oneui/auth/MarkLogicAuthenticationManager.java
```diff
@@ -107,10 +109,13 @@ public Authentication authenticate(Authentication authentication) throws Authent
...
+        }
+        if (!(hasManagePrivileges || stagingServerAccessible)) {
```
hao1st: we do not need to check stagingServerAccessible because it is a flag to indicate connection status.

PR #3512, opened 2020-01-27
DHFPROD-3788:Download project

srinathgit:
Download project

### DIFF
--- a/one-ui/src/main/java/com/marklogic/hub/oneui/controllers/EnvironmentController.java
+++ b/one-ui/src/main/java/com/marklogic/hub/oneui/controllers/EnvironmentController.java
@@ -18,6 +18,7 @@ package com.marklogic.hub.oneui.controllers;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.node.ObjectNode;
+import com.marklogic.hub.HubProject;
 import com.marklogic.hub.deploy.util.HubDeployStatusListener;
 import com.marklogic.hub.error.DataHubConfigurationException;
 import com.marklogic.hub.oneui.models.HubConfigSession;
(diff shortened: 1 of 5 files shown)


one-ui/src/main/java/com/marklogic/hub/oneui/controllers/EnvironmentController.java
```diff
@@ -114,6 +121,23 @@ public void onError(String commandName, Exception exception) {
         return payload;
     }
 
+    @RequestMapping(value = "/project-download", produces = "application/zip")
+    public void downloadProject(HttpServletRequest request, HttpServletResponse response) {
+        HubProject project = hubConfig.getHubProject();
+        File exportFile = project.getProjectDir().resolve("build").resolve("datahub-project.zip").toFile();
+        project.exportProject(exportFile);
```
rjrudin: We can get this done in one write - instead of calling this method, add an overloaded method to HubProject - e.g. exportProject(OutputStream) - and then pass in response.getOutputStream. That method will wrap a ZipOutputStream around it and write the contents directly to the HTTP response. That avoids messing around with files.

one-ui/src/main/java/com/marklogic/hub/oneui/controllers/EnvironmentController.java
```diff
@@ -114,6 +121,23 @@ public void onError(String commandName, Exception exception) {
...
+
+        response.setContentType(MediaType.APPLICATION_OCTET_STREAM_VALUE);
+        response.addHeader("Content-Disposition", "attachment; filename=" + exportFile.getName());
+        response.setContentLength((int) exportFile.length());
```
rjrudin: Hmm - I see this, as I don't think this is necessary to call - at least I'm hoping it isn't, as it would force us to write the data out to a file first. But I'm pretty certain it doesn't need to be called, and thus shouldn't if we can write the zip directly to the response.

one-ui/src/main/java/com/marklogic/hub/oneui/controllers/EnvironmentController.java
```diff
@@ -114,6 +121,23 @@ public void onError(String commandName, Exception exception) {
         return payload;
     }
 
+    @RequestMapping(value = "/project-download", produces = "application/zip")
+    public void downloadProject(HttpServletRequest request, HttpServletResponse response) {
```
rjrudin: You can test this by using Spring's MockHttpServletRequest (which can be empty) and MockHttpServletResponse. After calling the method, you can get the response bytes from the Mock response. And you should be able to open a ZipInputStream on that and verify a couple entries (no need to verify every single entry - I think verifying 2 of them is enough of a test).

PR #3516, opened 2020-01-28
DHFPROD-3619: Applying default permissions in QS for every step type

rjrudin:
I'm not sure why these steps didn't gain default permissions the first time around, but I think we're better off doing it for all step types.

I know that 3619 was closed back in sprint 22, so I wasn't sure what JIRA ticket to do this against. But I think it's important that we do this across the board in QS.


rjrudin:
Never mind, going to open a new JIRA bug for this.

PR #3520, opened 2020-01-29
DHFPROD-4158: Changing MasterTest assertion

rjrudin:
See the comment.

@ryanjdew  Let me know what you think about this. I don't know where to start to try to figure out why only 205 are in the collection sometimes. I think it'd help if the test were only try to e.g. match and merge 2 documents, with one additional document that is expected not to be merged. With over 200 documents being ingested, it's very difficult to know which ones are supposed to be merged together.


bsrikan:
I like the idea of trimming down the number of documents so its easier to determine why it fails.. Recently this test seems to have been failing on 1.0.x clusters.

PR #3530, opened 2020-01-30
DHFPROD-4255: Fix example projects to run in pipeline

SameeraPriyathamTadikonda:
Added gradlew to the project

Fixing errors to run mapping-example project

Fixing errors with custom-hook example

Fixing errors with dh-5-example project

Updated the input path


Jenkinsfile
```diff
@@ -534,11 +534,25 @@ pipeline{
...
+                            export M2_HOME=$MAVEN_HOME/bin; \
+                            export PATH=$GRADLE_USER_HOME:$PATH:$MAVEN_HOME/bin; \
+                            cd $WORKSPACE/data-hub/examples/dh-5-example; \
+                            rm -rf $GRADLE_USER_HOME/caches; \
+                            ./gradlew -i hubInit -Ptesting=true; \
+                            ./gradlew -i mlDeploy -Ptesting=true -PmlUsername=admin -PmlPassword=admin; \
+                            ./gradlew hubRunFlow -PflowName=ingestion_only-flow -Ptesting=true -PmlUsername=admin -PmlPassword=admin; \
+                            ./gradlew hubRunFlow -PflowName=ingestion_mapping-flow -Ptesting=true -PmlUsername=admin -PmlPassword=admin; \
+                            ./gradlew hubRunFlow -PflowName=ingestion_mapping_mastering-flow -Ptesting=true -PmlUsername=admin -PmlPassword=admin;
+                            "
```
bsrikan: Can we use anything other than admin/admin to run the flow.

SameeraPriyathamTadikonda: how can I do that? Instead of editing the gradle properties file. I'm passing the params.

bsrikan: We can maintain a copy of gradle.properties to be used across examples. Paste that over to the project directory after hubinit, then mlDeploy as admin by passing credentials via params and run the flow, so mlUsername credentials in the gradle props file will be used.

SameeraPriyathamTadikonda: Copying gradle.properties file from marklogic-datahub and running the flows as flow developer.

PR #3531, opened 2020-01-30
DHFPROD-4255: Can run hubInit on dh-5-example again

rjrudin:
runFlowWithoutProject was causing hubInit to bomb since the ml* properties weren't defined yet


SameeraPriyathamTadikonda:
Can I merge this manually?

PR #3535, opened 2020-01-31
DHFPROD-4096: Middle-tier Update load data API to store settings

hao1st:
https://project.marklogic.com/jira/browse/DHFPROD-4096


ryanjdew:
@hao1st This looks good. Can we also get a server-side unit test similar to these? https://github.com/marklogic/marklogic-data-hub/tree/feature/5.3.0/marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/data-services/artifacts

hao1st:
> @hao1st This looks good. Can we also get a server-side unit test similar to these? https://github.com/marklogic/marklogic-data-hub/tree/feature/5.3.0/marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/data-services/artifacts

@ryanjdew sure.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/artifacts/core.mjs
```diff
@@ -26,7 +27,8 @@ const cachedArtifacts = {};
 const registeredArtifactTypes = {
     loadData: LoadData,
```
rjrudin: What's the difference between a "loadData" artifact and a "loadDataSettings" artifact? I'm reading DHFPROD-4096, and the purpose of the new endpoints is to support only modifying the "settings" of a "loadData" config. That doesn't seem like it would mean that "settings" is a separate artifact type though, right?

one-ui/src/main/java/com/marklogic/hub/curation/controllers/LoadDataController.java
```diff
@@ -108,6 +108,19 @@ public String getArtifactType() {
         return "loadData";
     }
 
+    @RequestMapping(value = "/{artifactName}/settings", method = RequestMethod.GET)
+    @ResponseBody
+    public ResponseEntity<JsonNode> getArtifactSettings(@PathVariable String artifactName) {
+        ResponseEntity<JsonNode> resp = super.getArtifactSettings(artifactName);
```
rjrudin: Tiny nit - I think it's better just to return the result, there's no value in defining a variable here and then returning it.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/artifacts/core.mjs
hao1st: I understand, Conceptionally, the settings is not artififact type. One loadData artifact can be associated with only one setting. I just wanted to separate different implementation details because of different validation (different required fields, document properties) and query.  I can certainly add different methods for settings in the loadData.mjs instead of have a new type implmetation (loadDataSetting.mjs).

one-ui/src/main/java/com/marklogic/hub/curation/controllers/LoadDataController.java
hao1st: Yeah, i forgot simplifying after testing.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/artifacts/loadDataSettings.mjs
```diff
@@ -0,0 +1,68 @@
...
+'use strict';
+
+const DataHubSingleton = require('/data-hub/5/datahub-singleton.sjs');
+
+// define constants for caching expensive operations
+const dataHub = DataHubSingleton.instance();
+
+const collections = ['http://marklogic.com/data-hub/load-data-artifact/settings'];
+const databases = [dataHub.config.STAGINGDATABASE, dataHub.config.FINALDATABASE];
+const permissions = [xdmp.permission(dataHub.consts.DATA_HUB_DEVELOPER_ROLE, 'update'), xdmp.permission(dataHub.consts.DATA_HUB_OPERATOR_ROLE, 'read')];
```
srinathgit: @ryanjdew ,
Should the permissions here be "data-hub-load-data-reader,read,data-hub-load-data-reader,update" ? If so, this change has to be made for all the artifacts ?

ryanjdew: I believe @bsrikan will be addressing that in another PR (https://github.com/marklogic/marklogic-data-hub/pull/3536). If not, we can create a task to adjust it after the roles have been added.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/artifacts/core.mjs
rjrudin: If the settings is not an artifact type, then this will permanently create confusion by including it in a variable called "registeredArtifactTypes". 

Is the goal for the web client to have an endpoint where it only needs to pass in a portion of a LoadData object? If so, couldn't the middle tier wrap it in an otherwise-empty LoadData object? And the LoadData module would be able to figure out that only the settings are being passed in. That seems much easier to understand vs corrupting the meaning of an "artifact" by including both "LoadData" and "LoadDataSettings" in it.

hao1st: @rjrudin i removed the confusion in the latest check-in...

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/artifacts/core.mjs
```diff
@@ -67,10 +67,30 @@ export function getArtifacts(artifactType) {
...
     const artifactKey = generateArtifactKey(artifactType, artifactName, artifactVersion);
     const artifactLibrary =  getArtifactTypeLibrary(artifactType);
+
     // Currently there is no versioning for loadData artifacts
     const node = getArtifactNode(artifactType, artifactName, artifactVersion);
+
+    let settingNode = {};
+    if (artifactType == 'loadData') {
+        //delete related config file if existed
+        settingNode = artifactLibrary.getArtifactSettingNode(artifactName,
```
rjrudin: Is the issue that loadData has a separate settings document, but other artifacts do not? If so, why is it a separate document? Conceptually, I would think that the "settings" of a "loadData" would be part of the "loadData" document.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/artifacts/core.mjs
```diff
@@ -106,6 +126,42 @@ export function setArtifact(artifactType, artifactName, artifact) {
...
+
+        if (fn.empty(settingNode)) {
+            return {};
+        }
+        cachedArtifacts[artifactSettingKey] = settingNode.toObject();
+    }
+    return cachedArtifacts[artifactSettingKey];
+}
+
+export function setArtifactSettings(artifactType, artifactName, settings) {
```
rjrudin: Given that this is setting collections that are specific to loadData, shouldn't this have a loadData-specific name?

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/artifacts/core.mjs
hao1st: @rjrudin The implementation was based on https://project.marklogic.com/jira/browse/DHFPROD-4096
I assumed we could stored them as two different kinds of documents for flexibility and easier write/read. @ryanjdew What do you think?

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/artifacts/core.mjs
hao1st: you mean we can use 

http://marklogic.com/data-hub/load-data-artifact/settings/{loadArtifactName} insteads of http://marklogic.com/data-hub/load-data-artifact/settings ?

It should be better for query.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/artifacts/core.mjs
rjrudin: I don't see anything in 4096 that suggests any benefits from having a separate document. The middle tier of course won't care. But in ML, I think life is a lot simpler if it's a single document. You may need a separate endpoint for just changing settings on loadData specifically, as opposed to trying to toss that into the generic artifact endpoint. That would be good in the sense that the generic artifact endpoint wouldn't end up with multiple "If this is loadData, then do this loadData-specific code".

hao1st: https://project.marklogic.com/jira/browse/DHFPROD-4173
Data mapping settings only have three more properties than data loading settings (Source database,  Provenance granularity and Target Format). I think we should consider them...

ryanjdew: I personally like the separate documents. We're treating the writing of the settings and the artifact separately in the UI. Also, the settings serve a different purpose. With the exception of ingest/load-data, the artifacts themselves will be primarily used server-side, while the settings will be used by the Data Hub orchestration code.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/artifacts/core.mjs
ryanjdew: @hao1st  It think what is meant is the collection more like `http://marklogic.com/data-hub/${artifactType}/settings`. Perhaps we could get the collection from the artifactLibrary and append "/settings" to them to stay consistent.

hao1st: @ryanjdew  make sense.

PR #3536, opened 2020-01-31
DHFPROD-3916: Adding load data roles

bsrikan:
DHFPROD-3916 talks about data-hub-load-data-reader and data-hub-load-data-writer roles to be able to update/view the load configuration in one-ui.
This PR is an attempt to create these new 5.3 roles.

### DIFF
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/impl/HubProjectImpl.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/impl/HubProjectImpl.java
@@ -362,6 +362,11 @@ public class HubProjectImpl implements HubProject {
             IOUtils.closeQuietly(is);
         }
 
+        //New 5.3.0 roles
+        writeRoleFile(rolesDir, "data-hub-load-data-reader.json");
+        writeRoleFile(rolesDir, "data-hub-load-data-writer.json");
+
+
         // New 5.2.0 roles
         writeRoleFile(rolesDir, "data-hub-admin.json");
         writeRoleFile(rolesDir, "data-hub-developer.json");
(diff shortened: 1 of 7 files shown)


marklogic-data-hub/src/main/java/com/marklogic/hub/impl/HubProjectImpl.java
```diff
@@ -362,6 +362,11 @@ public Path getCustomMappingFunctionsDir() {
             IOUtils.closeQuietly(is);
         }
 
+        //New 5.3.0 roles
+        writeRoleFile(rolesDir, "data-hub-load-data-reader.json");
+        writeRoleFile(rolesDir, "data-hub-load-data-writer.json");
+
```
srinathgit: This would create roles but it seems incomplete as "data-hub-operator/developer" are not inheriting these roles. The load data artifact is not written with appropriate permissions.

ryanjdew:
Rebased the branch, added the proper load-data roles to the developer/operator roles, and adjusted to write the the documents with the load-data permissions.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/artifacts/loadData.mjs
```diff
@@ -22,7 +22,7 @@ const dataHub = DataHubSingleton.instance();
 
 const collections = ['http://marklogic.com/data-hub/load-data-artifact'];
 const databases = [dataHub.config.STAGINGDATABASE, dataHub.config.FINALDATABASE];
-const permissions = [xdmp.permission(dataHub.consts.DATA_HUB_DEVELOPER_ROLE, 'update'), xdmp.permission(dataHub.consts.DATA_HUB_OPERATOR_ROLE, 'read')];
+const permissions = [xdmp.permission(dataHub.consts.DATA_HUB_LOAD_DATA_WRITE_ROLE, 'update'), xdmp.permission(dataHub.consts.DATA_HUB_LOAD_DATA_READ_ROLE, 'read')];
 const requiredProperties = ['name', 'sourceFormat', 'targetFormat'];
```
bsrikan: Is there a downside to use the roles directly as a string than as a constant?

ryanjdew: Just the benefit that if we ever make the role configurable in the future there would only be one pace to make that change on the server side.

PR #3538, opened 2020-01-31
DHFPROD-4258:Provide host, stagingPort info to UI and some refactoring

srinathgit:
Provide host, stagingPort info to UI and some refactoring

### DIFF
--- a/one-ui/src/main/java/com/marklogic/hub/oneui/services/EnvironmentConfig.java
+++ b/one-ui/src/main/java/com/marklogic/hub/oneui/services/EnvironmentConfig.java
@@ -1,0 +1,85 @@
+/*
+ * Copyright 2012-2019 MarkLogic Corporation
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ */
+package com.marklogic.hub.oneui.services;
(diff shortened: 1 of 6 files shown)


one-ui/src/main/java/com/marklogic/hub/oneui/services/EnvironmentConfig.java
```diff
@@ -0,0 +1,85 @@
...
+    private String projectDir;
+    private String projectName;
+    private String dataHubVersion;
+    private String marklogicVersion;
+    private int stagingPort;
+    private String host;
+
+    @Autowired
+    @JsonIgnore
+    private HubConfigSession hubConfig;
```
rjrudin: I know this has been merged already, but I really encourage trying to minimize dependencies on HubConfig. Note that some of it is user-specific, but most of it is not. HubProject is not user-specific, and ideally that can be depended on without depending on HubConfig. Staging port is not user-specific. It's awkward because this is depending on a user-specific object for data that isn't user-specific.

srinathgit: @rjrudin ,
In one-ui, HubConfigSession is the first object created and hence every object depends on it. If I autowire, HubProject object, it's value is  is null ,so it has to be obtained from HubConfigSession and hence the dependency on it.

rjrudin: Isn't there a Spring container running by the time the user logs in (which I assume results in the creation of a HubConfigSession), and couldn't HubProject just be obtained from the Spring container?

srinathgit: Yeah, the HubProject is created prior to the user logging in by the Spring container but it's fields are all null. That's why I had to get the project from HubConfigSession.

rjrudin: That seems strange that the HubProject fields are null, as it's not a user-specific object, and thus I wouldn't think that its state depends on a user being logged in. I would think that when the application starts up, a HubProjectImpl can be fully initialized and stored as a singleton in the Spring container.

PR #3539, opened 2020-02-01
DHFPROD-3916: Added support for Role based privileges

xnikhil08:
This contains the role based access related changes for stories 3915, 3916 and 3917.

### DIFF
--- a/one-ui/ui/src/components/load-data/load-data-card.module.scss
+++ b/one-ui/ui/src/components/load-data/load-data-card.module.scss
@@ -73,3 +73,10 @@
     width: 700px;
     text-align: left;
 }
+
+.disabledDeleteIcon {
+    font-size: 21px;
+    opacity: 0.5;
+    pointer-events: none;
+    padding-bottom: 2px;
+}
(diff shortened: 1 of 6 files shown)


xnikhil08:
> As data-hub-data-load-writer user (perms: data-hub-load-data-writer, data-hub-module-reader, data-hub-operator, and a bunch of inherited):
> 
> * Output URI Replacement feature doesn't work for me (the following doesn't result in a change to the sample URI):
> 
> DHFPROD-3915-target-uri-replace
> 
> * After I create some loads, I cannot delete them. Then when I refresh the page, no data loads appear:
> 
> https://drive.google.com/file/d/1sqjphQ-SbJ59nfcGuehlK-Tk8CshUtEK/view
> 
> I'm seeing the following 400 upon delete:
> 
> {"code":400,"message":"Bad Request","details":"SEC-PERMDENIED: xdmp.documentDelete("/loadData/Test1.loadData.json") -- Permission denied: update /loadData/Test1.loadData.json"}
> 
> I'm seeing a 200 upon refresh.

@wooldridge & @ryanjdew - I tested this by manually creating the roles 'data-hub-load-data-writer' and 'data-hub-load-data-reader' in my local marklogic database, and assigned few other roles like data-hub-operator and data-hub-module-reader to the '--writer' role. 
**Are those the only privileges required ?**
Looks like the the writer role still needs enhanced privileges to update the load data artifact.

ryanjdew:
@wooldridge @xnikhil08 PR https://github.com/marklogic/marklogic-data-hub/pull/3536 was merged into feature/5.3.0 a moment ago. That should resolve the permissions/missing roles issue that you're seeing.

PR #3543, opened 2020-02-03
Updated CONTRIBUTING

rjrudin:
Added section on how to build and use the DH Java library.


aajacobs:
I just went through this process this morning, and these modifications answered the questions I had.  I think it's worth adding a mention of this process to the README as well, as I would not have looked in the CONTRIBUTING document to learn how to deploy DH from source without first being pointed there.

PR #3556, opened 2020-02-06
DHFPROD-4350: Ensure config directories in AppConfig get changed prop…

ryanjdew:
…erly on install

### DIFF
--- a/one-ui/src/main/java/com/marklogic/hub/oneui/controllers/EnvironmentController.java
+++ b/one-ui/src/main/java/com/marklogic/hub/oneui/controllers/EnvironmentController.java
@@ -18,12 +18,15 @@ package com.marklogic.hub.oneui.controllers;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.node.ObjectNode;
+import com.marklogic.appdeployer.AppConfig;
 import com.marklogic.hub.HubProject;
 import com.marklogic.hub.deploy.util.HubDeployStatusListener;
+import com.marklogic.hub.oneui.exceptions.BadRequestException;
 import com.marklogic.hub.oneui.models.HubConfigSession;
 import com.marklogic.hub.oneui.models.StatusMessage;
 import com.marklogic.hub.oneui.services.DataHubService;
 import com.marklogic.hub.oneui.services.EnvironmentService;
+import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -37,9 +40,9 @@ import org.springframework.web.bind.annotation.ResponseBody;
 
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
-import javax.servlet.http.HttpSession;
 import java.io.IOException;
 import java.io.OutputStream;
+import java.nio.file.Paths;
 import java.util.HashMap;
 import java.util.Map;
 
@@ -85,9 +88,7 @@ public class EnvironmentController {
 
     @RequestMapping(value = "/install", method = RequestMethod.POST)
     @ResponseBody
-    public JsonNode install(@RequestBody ObjectNode payload, HttpSession session) throws Exception {
-        // get original project directory value, so we can revert if failure occurs
-        String originalDirectory = environmentService.getProjectDirectory();
+    public JsonNode install(@RequestBody ObjectNode payload) throws Exception {
         final Exception[] dataHubConfigurationException = {null};
         HubDeployStatusListener listener = new HubDeployStatusListener() {
             int lastPercentageComplete = 0;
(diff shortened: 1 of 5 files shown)


one-ui/src/main/java/com/marklogic/hub/oneui/controllers/EnvironmentController.java
```diff
@@ -106,23 +108,23 @@ public void onError(String commandName, Exception exception) {
                 template.convertAndSend("/topic/install-status", new StatusMessage(lastPercentageComplete, message));
                 logger.error(message, exception);
                 dataHubConfigurationException[0] = exception;
-                environmentService.setProjectDirectory(originalDirectory);
             }
         };
         // setting the project directory will resolve any relative paths
         try {
             environmentService.setProjectDirectory(payload.get("directory").asText(""));
```
hao1st: String directory = payload.get("directory");
If (StringUtils.isEmpty(directory)) { // input validation should be checked in UI side, and here just in case
    //maybe logger.error(...)
    return payload;
}
hubConfig.createProject(directory);
...
we had better put the line  "environmentService.setProjectDirectory(directory)" to the end when dhf has been succcessfully installed.

one-ui/src/main/java/com/marklogic/hub/oneui/models/HubConfigSession.java
```diff
@@ -1000,14 +1002,14 @@ public void afterPropertiesSet() throws Exception {
         if (hubConfigImpl.getAdminManager() == null) {
             hubConfigImpl.setAdminManager(new AdminManager());
         }
+        String projectDirectory = environmentService.getProjectDirectory();
```
hao1st: After checking the HubConfigImpl, I think we can simplify unnecessary calls here. We even do not need to check if the hubConfigImpl.getAppConfig() is null because it is null for sure.

HubConfigImpl.java 
------------------------------------
    protected void hydrateAppConfigs(Properties properties) {
        com.marklogic.mgmt.util.PropertySource propertySource = properties::getProperty;

        if (appConfig != null) {
            // Still need to call this since the setter also "updates" appConfig with DHF-specific values
            setAppConfig(appConfig);
        } else {
            DefaultAppConfigFactory appConfigFactory = new DefaultAppConfigFactory(propertySource);
            if (hubProject != null && !StringUtils.isEmpty(hubProject.getProjectDirString())) {
                appConfigFactory.setProjectDir(hubProject.getProjectDir().toFile());
            }
            setAppConfig(appConfigFactory.newAppConfig());
        }
  }
...

after changing the HubConfigImpl class, we can only have the following three lines' codes

hubConfigImpl.createProject(projectDirectory);
hubConfigImpl.refreshProject();
this.dataHub = new DataHubImpl(this);

rjrudin: @hao1st Are you saying to change hydrateAppConfigs? I know for certain that AppConfig is not always null in the context of that method.

one-ui/src/main/java/com/marklogic/hub/oneui/models/HubConfigSession.java
```diff
@@ -1000,14 +1002,14 @@ public void afterPropertiesSet() throws Exception {
...
         if (hubConfigImpl.getAppConfig() == null) {
-            hubConfigImpl.setAppConfig(new AppConfig(), true);
+            hubConfigImpl.setAppConfig(new AppConfig(Paths.get(projectDirectory).toFile()), true);
         }
-        hubConfigImpl.createProject(environmentService.getProjectDirectory());
+        hubConfigImpl.createProject(projectDirectory);
```
rjrudin: I think we really need unit tests for the changes in here. My experience over the past couple months with trying to make changes to HubConfigImpl is that it's very difficult to know what any change might break, due to the class doing too many things. You can likely pull some of these changes into protected methods so that it's easier to test what they're doing.

one-ui/src/main/java/com/marklogic/hub/oneui/models/HubConfigSession.java
hao1st: Yes, I think the root cause is how we glue those HubConfig/HubProject/AppConfig together. For one-ui installation case, we only need to deal with the situation when AppConfig is null.  And my gut feeling is that passing a valid project directory to AppConfig is always what we expect.  Plus I think the one-ui higher level api call should be simplier during createProject/refreshProject. For me, the following calls are confused and a hacker way to workaround.
```
public void afterPropertiesSet() in HubConfigSession class

        if (hubConfigImpl.getAppConfig() == null) {
            hubConfigImpl.setAppConfig(new AppConfig(), true);
            hubConfigImpl.setAppConfig(new AppConfig(Paths.get(projectDirectory).toFile()), true);
        }
        hubConfigImpl.createProject(environmentService.getProjectDirectory());
        hubConfigImpl.createProject(projectDirectory);
        hubConfigImpl.refreshProject();
        // resetting app config to trigger updates to values
        hubConfigImpl.setAppConfig(hubConfigImpl.getAppConfig(), false);
        hubConfigImpl.hydrateConfigs();
```

ryanjdew:
Cleaned up some the code as suggested by @hao1st and added a unit test. There is another ticket for entering a bad path that is tangent to this (https://project.marklogic.com/jira/browse/DHFPROD-4349), where I'll add some more error handling, etc.

PR #3558, opened 2020-02-06
DHFPROD-4319: Output URI Replacement" and "Target URI preview" not working as expected

xnikhil08:
Fixed the issue where errors were not being flagged in Target URI Preview, if user provides the invalid output URI Replacement

### DIFF
--- a/one-ui/ui/src/components/load-data/new-data-load-dialog/new-data-load-dialog.tsx
+++ b/one-ui/ui/src/components/load-data/new-data-load-dialog/new-data-load-dialog.tsx
@@ -268,6 +268,15 @@ const NewDataLoadDialog = (props) => {
 
     if (event.target.id === 'outputUriReplacement') {
       setOutUriReplacement(event.target.value)
+      let dataPayload = {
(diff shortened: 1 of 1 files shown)


one-ui/ui/src/components/load-data/new-data-load-dialog/new-data-load-dialog.tsx
```diff
@@ -266,27 +266,30 @@ const NewDataLoadDialog = (props) => {
       setDescription(event.target.value)
     }
 
+    if(srcFormat && tgtFormat && (event.target.id === 'name' && event.target.value.length > 0)) {
+      setIsValid(true);
+    }
+    if(event.target.id === 'name' && event.target.value.length == 0){
+      setIsValid(false);
+    }
```
hao1st:
```suggestion:-6+7
if (event.target.id === 'name') {
   if (event.target.value.length == 0) {
        setIsValid(false);
  } else if (srcFormat && tgtFormat) {
       setIsValid(true);
  }
}
```

PR #3559, opened 2020-02-06
DHFPROD-4349: Middle-tier work to reject bad path input

ryanjdew:


one-ui/src/main/java/com/marklogic/hub/oneui/controllers/EnvironmentController.java
```diff
@@ -112,8 +113,11 @@ public void onError(String commandName, Exception exception) {
         String directory = payload.get("directory").asText("");
         // setting the project directory will resolve any relative paths
         try {
+            Path directoryPath = Paths.get(directory);
             if (StringUtils.isEmpty(directory)) {
                 throw new BadRequestException("Missing required directory property");
+            } else if (!(directoryPath.isAbsolute() || directoryPath.toFile().exists())) {
+                throw new IOException("Invalid directory supplied");
```
rjrudin: I recommend including the directory path in here for more context - e.g. 

    throw new IllegalArgumentException("Invalid project directory: " + directory);

I also try to avoid IOException whenever possible, as it's a checked exception. For almost all IO errors, it's unlikely that a client can do anything meaningful with it other than rethrow it.

one-ui/src/main/java/com/marklogic/hub/oneui/exceptions/CustomExceptionHandler.java
```diff
@@ -97,10 +97,10 @@ private String httpStatusSuggestion(HttpStatus httpStatus) {
 
     private String exceptionSuggestion(Exception exception) {
         Throwable rootException = exception;
-        if (exception instanceof RuntimeException && exception.getCause() != null) {
+        if (exception.getCause() != null) {
             rootException = exception.getCause();
         }
-        if (rootException instanceof IOException) {
+        if (rootException instanceof IOException || exception instanceof IOException) {
```
rjrudin: Oh okay - IOException is being used as way of indicating that the project directory is unavailable. I think IOException is too broad for that - I can definitely see some other feature in the future throwing an IOException that has nothing to do with the project directory.

I recommend creating an e.g. "ProjectDirectoryException" and catching any possible IOException from the code that tries to access the project directory. This code can then catch that exception without any concern that something else may throw the exception.

ryanjdew:
@rjrudin Good recommendations on the new exception. changes have been made.

one-ui/src/main/java/com/marklogic/hub/oneui/controllers/EnvironmentController.java
```diff
@@ -112,8 +115,13 @@ public void onError(String commandName, Exception exception) {
         String directory = payload.get("directory").asText("");
         // setting the project directory will resolve any relative paths
         try {
+            Path directoryPath = Paths.get(directory);
             if (StringUtils.isEmpty(directory)) {
                 throw new BadRequestException("Missing required directory property");
+            } else if (!directoryPath.isAbsolute()) {
+                throw new ProjectDirectoryException("Directory supplied must be an absolute path");
```
rjrudin: I think it's worth including the directory here as well - e.g. 

    "Project directory must be an absolute path: " + directory

one-ui/src/main/java/com/marklogic/hub/oneui/controllers/EnvironmentController.java
```diff
@@ -112,8 +115,13 @@ public void onError(String commandName, Exception exception) {
...
+            } else if (!directoryPath.toFile().exists()) {
+                throw new ProjectDirectoryException("Directory '" + directory + "' does not exist or cannot be read");
```
rjrudin: I'd clarify this too as "Project directory".

one-ui/src/main/java/com/marklogic/hub/oneui/controllers/EnvironmentController.java
```diff
@@ -112,8 +115,13 @@ public void onError(String commandName, Exception exception) {
         String directory = payload.get("directory").asText("");
         // setting the project directory will resolve any relative paths
         try {
+            Path directoryPath = Paths.get(directory);
             if (StringUtils.isEmpty(directory)) {
                 throw new BadRequestException("Missing required directory property");
```
rjrudin: Same thing - I think it's good to qualify this as the "Project directory" - e.g.

    "Project directory not specified"

one-ui/src/main/java/com/marklogic/hub/oneui/controllers/EnvironmentController.java
bsrikan: @ryanjdew before this check we could provide a path and the missing directory would be created during install. There is no requirement around that. But this implementation now enforces a directory to exist. Is that something we want to do?

Also @wooldridge, these exceptions are thrown as a "message" and UI shows "suggestion" for all the above conditions -- "Verify that the user account running the service has permissions to read from or write to the project directory."

This install-form.tsx snippet might have to change if we want the message to show up in GUI:

```
let message = (error.response.status === 500) ? 'Internal Server Error' : error.response.data.message;
      setIsLoading(false);
      setErrorMessage({
        show: true, 
        message: 'Installation Failure', 
          description: <><p>{message}</p><p>{error.response.data.suggestion}</p></>
      });
```

PR #3564, opened 2020-02-07
DHFPROD-4351: Fix typo in setting HttpStatus for error page

ryanjdew:


ryanjdew:
- Fixes 404s that should have been 401s
- Will allow the user to hit URLs directly, like /install

PR #3565, opened 2020-02-08
DHFPROD-4301: Uploading large file fails with no error

xnikhil08:
1. Increased the limit of total file size to 100MB.
2. Added an error message when the upload fails upon reaching the 100MB limit.
3. Moved the help icons to right side of the form fields in New Data load Modal.

### DIFF
--- a/one-ui/src/main/resources/application.properties
+++ b/one-ui/src/main/resources/application.properties
@@ -12,3 +12,7 @@ server.servlet.session.timeout=5m
 server.servlet.session.cookie.max-age=-1
 
 server.servlet.session.cookie.name=DHOneUISession
+
+spring.servlet.multipart.max-file-size=100MB
+
+spring.servlet.multipart.max-request-size=100MB
--- a/one-ui/src/main/resources/application.properties
+++ b/one-ui/src/main/resources/application.properties
@@ -15,4 +15,4 @@ server.servlet.session.cookie.name=DHOneUISession
 
 spring.servlet.multipart.max-file-size=100MB
 
-spring.servlet.multipart.max-request-size=100MB
+spring.servlet.multipart.max-request-size=105MB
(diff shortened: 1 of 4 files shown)


one-ui/src/main/resources/application.properties
```diff
@@ -12,3 +12,7 @@ server.servlet.session.timeout=5m
 server.servlet.session.cookie.max-age=-1
 
 server.servlet.session.cookie.name=DHOneUISession
+
+spring.servlet.multipart.max-file-size=100MB
+
+spring.servlet.multipart.max-request-size=100MB
```
aebadirad: Might want to raise the max request slightly above 100MB to handle multiple files that add up to 100mb themselves, but fall a few bytes over the request limit.

Also, we may want to look at a better way than just to 'hang' the app while uploading a huge file. Most systems use a form of streaming service in the background, not rest, to gracefully handle a fileupload in the background.

xnikhil08: @aebadirad - I can raise it to 105 MB approx. That should avoid such scenarios where the overall size of multiple files is slightly above the 100 MB limit.

xnikhil08:
> Nice work. I tested with the following:
> 
> 1. Create a new data load, load 3 small JSON files. SUCCESS.
> 2. Try loading 1 large file (> 100MB). ERROR.
> 3. Try loading 3 small JSON files again. ERROR.
> 
> I'd expect #3 to be OK but I cannot seem to load files (large or small) after the large attempt. See here:
> 
> https://drive.google.com/file/d/12nqVIud9yCVNHCIxKB4WBBig0yPXEdYP/view
> 
> Also, the spinner progress icon seems to get stuck in place after the large attempt/error.

@wooldridge - Please review again.

bsrikan:
I tried in FF and for a file > 105MB, dont get the error message. The upload seems to be stuck @ 2.5%.

xnikhil08:
> Fails in FF

@bsrikan - The issue is fixed in firefox. Please review again.

PR #3567, opened 2020-02-10
DHFPROD-3865: Can now run a step via a Data Services endpoint

rjrudin:
Upgraded to Java Client 5.1.0 to use the Bulk IO API for testing. This required upgrading to ml-gradle 4.0, which supports Java Client 5.1 (3.x does not). 

Extracted reusable code from mlRunFlow.sjs into flow.sjs/findMatchingContent. This helps minimize the code in mlRunFlow.sjs (REST) and runSteps.sjs (DS).

### DIFF
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/impl/HubConfigImpl.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/impl/HubConfigImpl.java
@@ -1947,8 +1947,6 @@ public class HubConfigImpl implements HubConfig
 
         initializeModulePaths(config);
 
-        config.setSchemasPath(getUserSchemasDir().toString());
-
         addDhfPropertiesToCustomTokens(config);
 
         String version = getJarVersion();
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/flow.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/flow.sjs
@@ -96,16 +96,6 @@ class Flow {
     return docs;
   }
 
-  deleteFlow(flowName) {
-    let uris = cts.uris("", null ,cts.andQuery([cts.orQuery([cts.directoryQuery("/flows/"),cts.collectionQuery("http://marklogic.com/data-hub/flow")]),
-      cts.jsonPropertyValueQuery("name", flowName)]));
-    for (let doc of uris) {
-      if (fn.docAvailable(doc)){
-        this.hubUtils.deleteDocument(doc, this.config.STAGINGDATABASE);
-      }
-    }
-  }
-
   //note: we're using uriMatch here to avoid case sensitivity, but still strongly match on the actual flow name itself
   //TODO: make this a flat fn.doc call in the future and figure out how to normalize the uri so we don't need this loop at all
   getFlow(name) {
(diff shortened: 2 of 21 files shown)


marklogic-data-hub/src/main/java/com/marklogic/hub/impl/HubConfigImpl.java
```diff
@@ -1947,8 +1947,6 @@ private void updateAppConfig(AppConfig config) {
 
         initializeModulePaths(config);
 
-        config.setSchemasPath(getUserSchemasDir().toString());
```
rjrudin: setSchemasPath was deprecated in ml-app-deployer 3.x and removed in 4.0 - this line of code was essentially having no impact, as the "user schemas dir" is src/main/ml-schemas, which is the default for ml-app-deployer (though it's better to use ml-config/databases/(db name)/schemas).

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/flow.sjs
```diff
@@ -96,16 +96,6 @@ class Flow {
     return docs;
   }
 
-  deleteFlow(flowName) {
```
rjrudin: I deleted this because a) it wasn't being used anywhere, and b) it didn't work - "this.hubUtils" did not resolve.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/flow.sjs
```diff
@@ -140,6 +130,53 @@ class Flow {
...
+    } else {
+      let sourceQuery = combinedOptions.sourceQuery || flow.sourceQuery;
+      query = sourceQuery ? xdmp.eval(sourceQuery) : null;
+    }
+
+    if (stepDetails.name === 'default-merging' && stepDetails.type === 'merging' && uris) {
+      return uris.map((uri) => { return { uri }; });
+    } else {
+      let sourceDatabase = combinedOptions.sourceDatabase || this.globalContext.sourceDatabase;
+      if (filterQuery) {
```
hao1st: query can be null at this point? If query and filterQuery are null, what will it happen? I guess we do not want to continue calling queryToContentDescriptorArray. Do we need to add some logs for debugging purpose?

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/HubConfigImpl.java
aebadirad: We might want to make sure we note the backwards dependency, as folks who were using the user schemas directory under older versions, will suddenly find their configs will not work when they upgrade.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/flow.sjs
aebadirad: Should probably fix this or, comment it out/leave it and bug log it not remove it, as the new app is going to be atomicly changing things on the server instead of doing full reloads.

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/HubConfigImpl.java
rjrudin: Looking at HubProjectImpl, getUserSchemasDir is set to src/main/ml-modules, and a user can't override that. That's also the default in ml-app-deployer. So that's why I say that line of code was having no impact.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/flow.sjs
rjrudin: I believe this is being handled now by artifacts/core.sjs

PR #3573, opened 2020-02-11
DHFPROD-4310: Keep filesystem project in sync when making updates

ryanjdew:


one-ui/src/main/java/com/marklogic/hub/curation/controllers/AbstractArtifactController.java
```diff
@@ -1,31 +1,49 @@
...
+    protected ArrayNode allArtifactTypeInfo = null;
+
     protected ResponseEntity<JsonNode> getArtifacts() {
         return new ResponseEntity<>(getArtifactService().getList(this.getArtifactType()), HttpStatus.OK);
     }
 
-    protected ResponseEntity<JsonNode> updateArtifact(String artifactName, JsonNode loadDataJson) {
-        return new ResponseEntity<>(getArtifactService().setArtifact(this.getArtifactType(), artifactName, loadDataJson), HttpStatus.OK);
+    protected ResponseEntity<JsonNode> updateArtifact(String artifactName, JsonNode artifactJson) throws IOException {
+        ResponseEntity<JsonNode> resp = new ResponseEntity<>(getArtifactService().setArtifact(this.getArtifactType(), artifactName, artifactJson), HttpStatus.OK);
```
rjrudin: This strikes me as something that would be in an ArtifactManager class that depends on both HubProject and ArtifactService. The controller ideally only has one dependency and only calls a single method; its job is then just to collect inputs from the HTTP request and convert those into a single method call, and then convert whatever the output is into something in the HTTP response. 

That allows the ArtifactManager to be used in any context and keeps the web tier as thin as possible. Is that possible here?

one-ui/src/main/java/com/marklogic/hub/curation/controllers/AbstractArtifactController.java
```diff
@@ -46,4 +64,62 @@ protected ArtifactService getArtifactService() {
...
+        Path fileLocation = buildArtifactProjectLocation(artifactName, artifactVersion);
+        Files.deleteIfExists(fileLocation);
+    }
+
+    protected String getNameFromArtifact(ObjectNode artifact) {
+        ObjectNode artifactTypeInfo = getArtifactTypeInfo();
+        return artifact.get(artifactTypeInfo.get("nameProperty").asText()).asText();
+    }
+
+    protected String getVersionFromArtifact(ObjectNode artifact) {
```
rjrudin: Do we have any use case for versions yet? Or will the value just be "1" for everything until we have a use case for separate versions of an artifact?

ryanjdew: We don't yet. It is more in preparation for when we move mapping to this new pattern. Right now the version is always null and the artifact type info won't provide a version property, so it isn't taken into account.

one-ui/src/main/java/com/marklogic/hub/curation/controllers/AbstractArtifactController.java
ryanjdew: It is definitely possible to move it to an Artifact Manager. I thought we wanted to move away from the manager approach, but I guess by having the Artifact Manager we can eventually deprecate the other managers and reduce the overhead of the all the managers we maintain that way.

rjrudin: I think the pattern of a Manager having dependencies on both ML and the filesystem makes sense, while a Service only depends on ML (and the Service class is code-generated based on a data service). In some use cases, a Controller may only need to talk to a Service. In probably most use cases for now, a Controller will talk to a Manager who knows how to handle operations that involve both ML and the filesystem.

ryanjdew: updated to use an artifact manager

marklogic-data-hub/src/main/java/com/marklogic/hub/ArtifactManager.java
```diff
@@ -0,0 +1,102 @@
...
+ */
+public interface ArtifactManager {
+
+    /**
+     * Gets artifacts of a given type
+     *
+     * @param artifactType - type of artifact the operation is dealing with
+     * @return JsonNode of the artifacts matching a type
+     */
+    public JsonNode getArtifacts(String artifactType);
```
rjrudin: Can this be more strongly typed as an ArrayNode? Or would it be an ObjectNode if there's a single artifact? I'd lean towards an ArrayNode.

marklogic-data-hub/src/main/java/com/marklogic/hub/ArtifactManager.java
```diff
@@ -0,0 +1,102 @@
...
+
+    /**
+     * Updates an artifact of a given type and name.
+     *
+     * @param artifactType - type of artifact the operation is dealing with
+     * @param artifactName - identifier of the artifact
+     * @param artifactJson - JSON of artifact to persist
+     * @return JsonNode of the artifact matching a type
```
rjrudin: Just to confirm, is this returning a representation of the updated artifact (as opposed to what the artifact currently looks like)? I think that would be good to confirm in the comment.

marklogic-data-hub/src/main/java/com/marklogic/hub/ArtifactManager.java
```diff
@@ -0,0 +1,102 @@
...
+    public JsonNode updateArtifact(String artifactType, String artifactName, JsonNode artifactJson) throws IOException;
+
+    /**
+     * Retrieves an artifact of a given type and name.
+     *
+     * @param artifactType - type of artifact the operation is dealing with
+     * @param artifactName - identifier of the artifact
+     * @return JsonNode of the artifact matching a type and name
+     */
+    public JsonNode getArtifact(String artifactType, String artifactName);
```
rjrudin: Can this be strongly typed as an ObjectNode since it should be a single thing?

I mention this for a few other methods. I think the distinction would be useful for clients that aren't just going to be a passthrough. And it's more self-documenting to use ObjectNode or ArrayNode if we know it's going to be one or the other.

marklogic-data-hub/src/main/java/com/marklogic/hub/ArtifactManager.java
```diff
@@ -0,0 +1,102 @@
...
+
+    /**
+     * Deletes an artifact of a given type and name.
+     *
+     * @param artifactType - type of artifact the operation is dealing with
+     * @param artifactName - identifier of the artifact
+     * @return JsonNode
+     */
+    public JsonNode deleteArtifact(String artifactType, String artifactName) throws IOException;
```
rjrudin: Why would this throw an IOException, and what would the client be expected to be able to do with it? I think Java should have made IOException a runtime exception, as it's rare that a client can do anything useful with it.

marklogic-data-hub/src/main/java/com/marklogic/hub/ArtifactManager.java
```diff
@@ -0,0 +1,102 @@
...
+     * @return JsonNode of the artifact matching a type and name
+     */
+    public JsonNode getArtifact(String artifactType, String artifactName);
+
+    /**
+     * Deletes an artifact of a given type and name.
+     *
+     * @param artifactType - type of artifact the operation is dealing with
+     * @param artifactName - identifier of the artifact
+     * @return JsonNode
```
rjrudin: Is this returning the contents of the artifact before it's deleted?

marklogic-data-hub/src/main/java/com/marklogic/hub/ArtifactManager.java
```diff
@@ -0,0 +1,102 @@
...
+
+    /**
+     * Validates an artifact of a given type and name.
+     *
+     * @param artifactType - type of artifact the operation is dealing with
+     * @param artifactName - identifier of the artifact
+     * @param artifactJson - JSON of artifact to persist
+     * @return JsonNode with validation details
+     */
+    public JsonNode validateArtifact(String artifactType, String artifactName, JsonNode artifactJson);
```
rjrudin: Can this be an ObjectNode?

marklogic-data-hub/src/main/java/com/marklogic/hub/ArtifactManager.java
```diff
@@ -0,0 +1,102 @@
...
+
+    /**
+     * Retrieves settings of an artifact of a given type and name.
+     *
+     * @param artifactType - type of artifact the operation is dealing with
+     * @param artifactName - identifier of the artifact
+     * @return JsonNode of the artifact settings matching a type and name
+     */
+    public JsonNode getArtifactSettings(String artifactType, String artifactName);
```
rjrudin: Can this be an ObjectNode?

marklogic-data-hub/src/main/java/com/marklogic/hub/ArtifactManager.java
```diff
@@ -0,0 +1,102 @@
...
+
+    /**
+     * Provides path to artifact in project for an artifact of a given type, name, and (optionally) version.
+     *
+     * @param artifactType - type of artifact the operation is dealing with
+     * @param artifactName - identifier of the artifact
+     * @param artifactVersion - version of artifact (optional)
+     * @return Path to artifact in project
+     */
+    public Path buildArtifactProjectLocation(String artifactType, String artifactName, String artifactVersion);
```
rjrudin: It looks like the only client of this is an automated test. For that reason, I don't think it should be the interface yet. The fewer methods in an interface, the better. The method in the impl can be public if needed, and the test can just call it directly instead of depending on the interface.

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/ArtifactManagerImpl.java
```diff
@@ -0,0 +1,136 @@
...
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Iterator;
+
+@Component
+public class ArtifactManagerImpl implements ArtifactManager {
+    @Autowired
+    HubConfig hubConfig;
```
rjrudin: I'd really like to get away from anything depending on HubConfig to be a singleton. I took a look at the code, and I think it'd be better for ArtifactManagerImpl not to be a Spring component. Rather, just instantiate one when needed by passing in a HubProject and a DatabaseClient (which would be constructed based on a HubConfig, which the client presumably has access to). 

I looked at how allArtifactTypeInfo is used, because if this is no longer a singleton, there won't be a reason to store that as a class field. I think that's fine though. The data is needed during updateArtifact and deleteArtifact. Those should be very fast operations, such that I think it's fine to make a call to ML each time to get the needed artifact type info (particularly since that endpoint looks like it'll be really fast).

hao1st: I think we can set member allArtifactTypeInfo as static, and some methods can be public static methods. My initial thought is why we have to get ArtifactTypeInfo from ML server side if we already know those pre-defined supported artifact types, file extensions ? They can simply set as a final static Map without having to keep the allArtifactTypeInfo object instance with extra Json operations.

hao1st: If we create different objects for the Singleton (ArtifactManagerImpl), it can cause unnecessary memory footprint...

rjrudin: The memory footprint of ArtifactManagerImpl would be as minimal as possible though, as it would have no state in it. And they'd be eligible for garbage collection immediately after a controller instantiates one and uses it. We'd be better off worrying about creating instances of String. 

If we feel it's important for ArtifactManagerImpl to be a singleton, then I recommend moving HubConfig from being a class field to being a method argument instead. That's my main goal - that we shouldn't have any classes autowiring in a HubConfig object since that object is tied to one user, and we know that multiple users need to be able to use an object like ArtifactManagerImpl. 

Note that since we don't have a use case for multiple HubProject instances, I think it's fine for that to be a singleton that Spring manages. So having ArtifactManagerImpl autowire in HubProject is fine.

rjrudin: Also, I think we want the ML endpoint to be the one that knows about artifact type info. That's knowledge that's derived from code in ML. 

I just wouldn't worry about caching this stuff in ML yet, I think it's premature optimization. Artifacts aren't going to be updated/deleted that often - maybe dozens of times a day? Even hundreds a day would be negligible from a performance perspective. This just isn't something we need to optimize yet, or probably ever.

ryanjdew:
@rjrudin @hao1st Changes since last review:
- addressed the stronger typing of the JsonNodes
- removed the unnecessary internal API in ArtifactManager interface
- pass the HubConfig into the ArtifactManagerImpl constructor
- catch IOExceptions and rethrow them as RuntimeExceptions.

one-ui/src/main/java/com/marklogic/hub/curation/controllers/AbstractArtifactController.java
```diff
@@ -1,49 +1,55 @@
...
-    protected ResponseEntity<JsonNode> updateArtifactSettings(String artifactName, JsonNode settings) {
-        return new ResponseEntity<>(getArtifactService().setArtifactSettings(this.getArtifactType(), artifactName, settings), HttpStatus.OK);
+    protected ResponseEntity<ObjectNode> updateArtifactSettings(String artifactName, JsonNode settings) {
+        return new ResponseEntity<>(artifactManager.updateArtifactSettings(this.getArtifactType(), artifactName, settings), HttpStatus.OK);
     }
 
     protected abstract String getArtifactType();
+
+    public void afterPropertiesSet() {
+        this.artifactManager = new ArtifactManagerImpl(this.hubConfig);
```
rjrudin: Because subclasses of this class are Spring Controller beans, and thus managed as singletons, I think this still runs into a problem where there's a single ArtifactManager tied to... I guess the first user that hits this? That's my guess - that if a second user logs in, that user will end up using an ArtifactManager that is tied to the first user's HubConfig. I'm not certain though.

I think the safer approach is just to instantiate an ArtifactManagerImpl when you need it - i.e. have a "protected ArtifactManager newArtifactManager()" method that calls "new ArtifactManagerImpl(this.hubConfig)". That ensures you get an ArtifactManager tied to the current user session.

hao1st: @rjrudin I believe that the HubConfig has a separate instance for a different session.

ryanjdew: HubConfigSession is SessionScoped. If we couldn't trust `this.artifactManager` to not be shared between sessions, I'm not sure how we could trust `this.hubConfig` to not be shared across sessions.

rjrudin: I think it's worth testing, just to verify that two users don't end up using the same HubConfig. 

I'm mostly wondering if the scoping is "lost" somehow when it's used to construct another object.

PR #3578, opened 2020-02-12
DHFPROD-3922: Move in flow run logic

ryanjdew:


one-ui/src/main/java/com/marklogic/hub/curation/services/FlowManagerService.java
```diff
@@ -23,14 +25,21 @@
...
 import java.io.InputStream;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
 
 @Service
 public class FlowManagerService {
```
rjrudin: No action on this, but I'm realizing we have some naming inconsistency with Data Services, as the interface generated for DS is naturally going to be called a "Service". And then I was thinking that a Manager would be a layer on top that provides additional Java-based logic on top of one or more Services. But that conflicts with this usage of Service here, since it's the Service on top with Manager and other components (e.g. FlowRunner) below it. 

That said, I'm not sure how to resolve this. My initial thought is that perhaps the logic in here should move into FlowManager and FlowRunner. But again, no action needed for now, we can move this furniture around later.

one-ui/ui/api/swagger/mocks.yaml
```diff
@@ -2275,6 +2275,39 @@ definitions:
                 - '{}'
                 - '{}'
             version: 0
+    runFlowResponse:
```
rjrudin: This is interesting, as I was thinking we should define a JSON Schema for objects like this that are returned by key classes. Do you know if stepResponse is expected to be defined? That has some pretty key information in it. 

I think it's fine to leave this here, just makes me wonder what's a good way to define this stuff in one place.

ryanjdew: It can be derived by the code. It is a little different than other definitions in that the property names are dynamic based on what steps are run and if they provide a response. I'm not sure if and/or how properties like that are defined in swagger yet. Perhaps I just start by adding properties "1" and "2" while noting that they are dynamic based on the steps run?

PR #3579, opened 2020-02-13
DHFPROD-4337: Create API to install uploaded project

hao1st:
https://project.marklogic.com/jira/browse/DHFPROD-4337
https://project.marklogic.com/jira/browse/DHFPROD-4358


one-ui/src/main/java/com/marklogic/hub/oneui/auth/HttpHandshakeInterceptor.java
```diff
@@ -0,0 +1,35 @@
...
+
+import org.springframework.http.server.ServerHttpRequest;
+import org.springframework.http.server.ServerHttpResponse;
+import org.springframework.http.server.ServletServerHttpRequest;
+import org.springframework.http.server.ServletServerHttpResponse;
+import org.springframework.security.authentication.BadCredentialsException;
+import org.springframework.web.socket.WebSocketHandler;
+import org.springframework.web.socket.server.HandshakeInterceptor;
+
+public class HttpHandshakeInterceptor  implements HandshakeInterceptor {
```
rjrudin: How is this used? I don't see any references to it anywhere. I'm also wondering what problem it's solving, as I don't think DHFPROD-4337 spells it out? This may be needed, but if so, we should have a JIRA ticket that spells out the problem that this is solving.

one-ui/src/main/java/com/marklogic/hub/oneui/services/DataHubService.java
```diff
@@ -29,6 +29,19 @@ public boolean install(HubDeployStatusListener listener) {
         return false;
     }
 
+    public boolean unInstall(HubDeployStatusListener listener) {
```
rjrudin: Since the client for this method doesn't care about the return value, let's declare this as void. It appears the expectation is that if anything goes wrong, the listener will be told about it. So there's no value in returning true/false - the method is expected to have succeeded unless the listener is given an error.

one-ui/src/main/java/com/marklogic/hub/oneui/controllers/EnvironmentController.java
```diff
@@ -90,31 +96,109 @@ public JsonNode getProject() {
...
+        // extract the uploaded & zipped project file into the current project folder
+        extractZipProject(project, uploadedFile, listener);
+
+        HubDeployStatusListener uninstalllistener = getInstallListener(dataHubConfigurationException, true);
+        dataHubService.unInstall(uninstalllistener);
+
+        install(project.getProjectDirString(), listener, dataHubConfigurationException);
+    }
+
+    private HubDeployStatusListener getInstallListener(final Exception[] dataHubConfigurationException, boolean isUninstall) {
```
rjrudin: I am generally not big on anonymous classes like this. I find them more difficult to read and understand, and they can also complicate the dependencies on a class. For example, EnvironmentController has a dependency on SimpMessagingTemplate, but it doesn't need it - this anonymous impl needs it.

Let's move this into a real class - call it "UIDeployListener" or something like that. Move the SimpMessagingTemplate dependency to it, which can be passed in via the constructor, along with "boolean isUninstall".

one-ui/src/main/java/com/marklogic/hub/oneui/controllers/EnvironmentController.java
```diff
@@ -90,31 +96,109 @@ public JsonNode getProject() {
...
                 logger.error(message, exception);
                 dataHubConfigurationException[0] = exception;
             }
         };
-        String directory = payload.get("directory").asText("");
-        // setting the project directory will resolve any relative paths
+        return listener;
+    }
+
+    private void install(String directory, HubDeployStatusListener listener, final Exception[] dataHubConfigurationException) throws Exception {
```
rjrudin: I don't see any use of dataHubConfigurationException - what's its purpose? If it's not used here, can it be removed from the anonymous listener above?

one-ui/src/main/java/com/marklogic/hub/oneui/utils/DataHubProjectUtils.java
```diff
@@ -0,0 +1,130 @@
...
+import java.nio.file.Path;
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipInputStream;
+import org.apache.commons.io.FileUtils;
+import org.apache.commons.io.FilenameUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.web.multipart.MultipartFile;
+
+public class DataHubProjectUtils {
```
rjrudin: I'd much rather see a real class here as opposed to a collection of static methods. This looks more like a "ProjectManager" class with methods for backing up a project, cleaning a project, and extracting a project zip. 

That also makes it easy to write tests for these methods so that we know they work.

one-ui/src/main/java/com/marklogic/hub/oneui/controllers/EnvironmentController.java
```diff
@@ -90,31 +96,109 @@ public JsonNode getProject() {
...
+        HubProject project = hubConfig.getHubProject();
+
+        //backup first
+        backupExistingFSProject(project, listener);
+
+        // delete contents from the current project folder
+        cleanExistingFSProject(project, listener);
+
+        // extract the uploaded & zipped project file into the current project folder
+        extractZipProject(project, uploadedFile, listener);
```
rjrudin: Why require the controller to make 3 separate calls here? We want controllers to be as thin and as simple as possible. The controller doesn't care about all of these details - it just wants to say - Hey, I've got this uploadedFile, please make this the new project. A ProjectManager class - see below - can handle all the details of backing up an existing project and then extracting the new one. I think this class just wants to call a method named "replaceProject".

one-ui/src/main/java/com/marklogic/hub/oneui/auth/HttpHandshakeInterceptor.java
hao1st: I accidently checked in the file. It was created on my local for some testing. We do not need. it.

one-ui/src/main/java/com/marklogic/hub/oneui/controllers/EnvironmentController.java
hao1st: dataHubConfigurationException is based on previous install call logic, it is an excepection holder for the first thrown. We check it at the end, and rethrow it as ProjectDirectoryException. I was not sure why we did that and just did not want to break existing logic. @ryanjdew do you have any idea about
`            if (exception instanceof IOException || rootCause instanceof IOException) {
                exception = new ProjectDirectoryException("Your user account does not have write permissions to the installation directory.", "Log in as a user with write permissions to the directory you specify, or provide the absolute path to another directory.", exception);
            }`

one-ui/src/main/java/com/marklogic/hub/oneui/utils/DataHubProjectUtils.java
hao1st: I think that those methods can be used as utilty methods and do not feel we should create instances for that. I did not see any difference from creating "ProjectManager" class. We also can create test cases, right?

one-ui/src/main/java/com/marklogic/hub/oneui/controllers/EnvironmentController.java
hao1st: I can create one new method to combine 3 calls together.

one-ui/src/main/java/com/marklogic/hub/oneui/controllers/EnvironmentController.java
hao1st: I can create a UIDeployListener class.

one-ui/src/main/java/com/marklogic/hub/oneui/utils/DataHubProjectUtils.java
rjrudin: First, I'll take back my suggestion, as since this is bound to HubDeployStatusListener, there may not be much reuse to achieve here. 

Second - generally, I only like static functions/variables for things that are truly "static" - i.e. not likely to change. Statics also create tight binding between classes - i.e. there's no way to modify a client's usage of a static. For example, if we instead had a ProjectManager interface, the EnvironmentController (or DataHubService) could be bound to that interface. And it could have a constructor or setter that allows for any impl of that interface to be passed in. That results in highly composable, reusable software. Statics prevent that.

one-ui/src/main/java/com/marklogic/hub/oneui/services/DataHubProjectUtils.java
```diff
@@ -35,17 +35,34 @@
 import org.springframework.web.multipart.MultipartFile;
 
 public class DataHubProjectUtils {
```
hao1st: The reason to move the utility class to services package is that "no such method found" thrown if it is put into utils package during junit test case was run. I did not figure out the exact reason...

one-ui/src/main/java/com/marklogic/hub/oneui/controllers/EnvironmentController.java
```diff
@@ -171,4 +117,42 @@ public JsonNode reset() {
...
+
+        dataHubService.unInstall(new UIDeployListener(template,true));
+        install(project.getProjectDirString(), listener);
+    }
+
+    private void install(String directory, UIDeployListener listener) throws Exception {
+        try {
+            // setting the project directory will resolve any relative paths
+            Path directoryPath = Paths.get(directory);
+            if (StringUtils.isEmpty(directory)) {
```
rjrudin: Seems like this first "if" should be done before calling Paths.get, right?

one-ui/src/main/java/com/marklogic/hub/oneui/controllers/EnvironmentController.java
```diff
@@ -171,4 +117,43 @@ public JsonNode reset() {
...
+
+    private void install(String directory, UIDeployListener listener) throws Exception {
+        try {
+            if (StringUtils.isEmpty(directory)) {
+                throw new BadRequestException("Property 'directory', identifying project location, not specified");
+            }
+            // setting the project directory will resolve any relative paths
+            Path directoryPath = Paths.get(directory);
+            if (!directoryPath.isAbsolute()) {
+                throw new ProjectDirectoryException("The Project Directory field requires an absolute path. You entered: " + directory, "Enter the absolute path to an existing local directory.");
```
rjrudin: Is this ever going to happen, and what would the user be able to do about it? The user isn't able to enter an absolute path - the user is just uploading a file.

If "directory" is invalid, then the system is misconfigured, and I don't think any user can do anything about that, right? An IT person would need to fix the path for the HubProject and then restart the middle tier.

@ryanjdew @wooldridge Am I correct about this? My impression is that a user doesn't have any control over the project directory, except when the application is first initialized. 

I think that means that the value of "directory" can be assumed to be valid. If it weren't valid, the user would never have gotten to this point, and there's not anything a user can do about it either. I think that means we just need a single try/catch around the this code, which is already in place.

one-ui/src/main/java/com/marklogic/hub/oneui/listener/UIDeployListener.java
```diff
@@ -0,0 +1,71 @@
...
+
+    @Override
+    public void onError(String commandName, Exception exception) {
+        String message = "Error encountered running command: " + commandName;
+        String msg = parseMessage(message);
+        template.convertAndSend("/topic/install-status", new StatusMessage(lastPercentageComplete, msg));
+        logger.error(message, exception);
+        this.exception = exception;
+        if (exception instanceof IOException || exception.getCause() instanceof IOException) {
+            this.exception = new ProjectDirectoryException("Your user account does not have write permissions to the installation directory.",
```
rjrudin: This seems like another error that the user can't possibly resolve. It seems like a confusing error message too, as the user will think it's their MarkLogic user account. But it's the OS user account that matters here - i.e. which OS user was used to fire up the middle tier. 

I think a more appropriate error message is (same with above) - "Installation (or uninstallation) failed, please contact a system administrator; cause: " + e.getMessage(). That makes it clear that the expected user can't do anything to resolve the issue, and an IT admin will need to get involved.

one-ui/src/main/java/com/marklogic/hub/oneui/controllers/EnvironmentController.java
hao1st: @rjrudin i am not sure i understand your input here...please clarify.

one-ui/src/main/java/com/marklogic/hub/oneui/listener/UIDeployListener.java
hao1st: @rjrudin do we want to have the message for suggesion or message?

one-ui/src/main/java/com/marklogic/hub/oneui/controllers/EnvironmentController.java
rjrudin: My impression is that the project directory is not specified by a user, but rather when One UI is first initialized. Thus, if there's an error with the value of the directory, there's nothing that a user can do about it. The only thing to do would be to tell the user that the project directory is not valid and to contact a systems administrator.

one-ui/src/main/java/com/marklogic/hub/oneui/listener/UIDeployListener.java
rjrudin: I don't know how the "suggestion" is used, but I suppose the message could be "Installation failed; please contact a system administrator for assistance." and the suggestion would be "Installation error: " + e.getMessage().

one-ui/src/main/java/com/marklogic/hub/oneui/controllers/EnvironmentController.java
ryanjdew: @rjrudin The directory is specified when the install is called via the install page at /install, which the user is guided to do if a project hasn't previously been installed.

rjrudin: Okay - I'm thinking of the use case where someone uploads a project - could be Ernie or Wei Lin - and that user doesn't have the ability to specify the project directory.

rjrudin: I'll stand down on this for now though - I think it's something that'll come up in testing though - pinging @bsrikan  for awareness.

bsrikan: What @rjrudin says makes sense to me. At the time a project is being uploaded, user is logged in and DH is installed viz. project Directory is valid.
Say DH is installed and project dir is "/Users/userA/myProject".
Upload is basically an option to reinstall, so when a user clicks on Upload button, presumably there is a Finder window to select project zip to be uploaded. This is another place where path is required to the project zip being uploaded. 
@hao1st can you pls clarify that the project dir in question is not for the zip being uploaded. If yes, then project dir is already established and the if condition is never satisfied IMO.

bsrikan:
@hao1st this PR includes UI changes so ideally should have included @wooldridge for review.
Lets not create one PR for both back-end and front-end and if we do lets atleast add a reviewer who is a SME.

hao1st:
@bsrikan The upload feature is to replace the current project located in the one-ui server (remote) with a attached zip file from a local folder of a privileged login user. Before overriding the existing project folder, we back up it as a backup zip file in a temporary folder in the on-ui server.  The uploaded zip file folder is  irrelevant with the project folder.

PR #3584, opened 2020-02-15
DHFPROD-4310: Cleanup data sets when load data config is deleted

ryanjdew:


one-ui/src/main/java/com/marklogic/hub/curation/controllers/LoadDataController.java
```diff
@@ -139,4 +133,19 @@ private void enrichLoadData(ObjectNode loadDataConfig) {
...
         loadDataConfig.put("filesNeedReuploaded", fileCount == 0);
     }
+
+    private void deleteDataSetDirectory(String artifactName) {
+        Path dataSetDirectoryPath = dataSetDirectory(artifactName);
+        if (dataSetDirectoryPath.toFile().exists()) {
+            try {
+                FileUtils.deleteDirectory(dataSetDirectoryPath.toFile());
+            } catch (IOException e) {
+                throw new RuntimeException(e);
```
rjrudin: I'm wondering if this should really rethrow the error. If a dataset cannot be deleted, how much of a problem is that? And the user wouldn't really be able to do anything about it, right? I'm wondering if this should just be logged at the "warn" level.

hao1st: @ryanjdew I got the "{"code":500,"message":"java.io.IOException: Unable to delete directory /Users/hliu/dhf/data-sets/validArtifact.","suggestion":"Contact your server administrator."}" exception because the //data-sets does not exists or inpuFilePath does not exist. I agree with Rob and think that we do not need to throw a runtime exception here but logging it with the warning level.

PR #3591, opened 2020-02-19
DHFPROD-4308: Middle-Tier: Create extension to existing mapping artif…

hao1st:
https://project.marklogic.com/jira/browse/DHFPROD-4308


ryanjdew:
Looks good. In DH today, we're are using "targetEntity" instead of "entityName". 

@rjrudin I know you're big on good naming conventions. Thoughts? One thing to consider is that in the current mappings, while it will work with only specifying an entity model name, I specifically made it so you could specify the absolute IRI of an entity, so versioned entities could be used.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/artifacts/core.sjs
```diff
@@ -56,11 +58,42 @@ function getArtifacts(artifactType) {
...
+            return getArtifactsGroupbyEntity(artifactType, queriesForAnd)
         }
     }
+    return [];
+}
+
+function getArtifactsGroupbyEntity(artifactType, queriesForAnd) {
+    const artifacts = [];
+    const entityNames = getEntityNames();
+    for (const ename of entityNames) {
```
rjrudin: This is doing N queries, where N = entityNames.length. It's faster/better to do a single query - just pass entityNames into the cts.jsonPropertyValueQuery. Then iterate over the results with a map function. Can get it all done this simply (I often refer to https://www.w3schools.com/jsref/jsref_obj_array.asp for things like this):

```
function getArtifactsGroupByEntity(artifactType, queriesForAnd) {
  cts.search(cts.andQuery(queriesForAnd.concat(cts.jsonPropertyValueQuery("entityName", getEntityNames())).map(config => {
    const configByEntity = {
      name: config.path.to.entityName (I don't know the path)
    };
    switch(artifactType) {
    ....
    }
    return configByEntity;
  });
}
```

Note the use of "concat" to avoid modifying the queriesForAnd array that's passed in.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/artifacts/core.sjs
```diff
@@ -56,11 +58,42 @@ function getArtifacts(artifactType) {
...
+        let configByEntity = {};
+        configByEntity.name = ename;
+        queriesForAnd.push(cts.jsonPropertyValueQuery("entityName", ename));
+        let configs = cts.search(cts.andQuery(queriesForAnd));
+        // artifactType can be mapping, mastering, merging
+        switch (artifactType) {
+            case "mapping":
```
rjrudin: Ideally, this service wouldn't have much knowledge of the concrete artifact types. Is it necessary to store the "configs" under different key names? Couldn't it just be "config" or "settings", regardless of the artifact type?

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/artifacts/core.sjs
```diff
@@ -225,6 +258,16 @@ function deleteArtifactSettings(artifactType, artifactName, artifactVersion = 'l
     return { success: true };
 }
 
+function getEntityNames() {
+    let entities = cts.search(cts.collectionQuery("http://marklogic.com/entity-services/models"))
```
rjrudin: You can use "map" here to cut out a good chunk of code:

    cts.search(...).toArray().map(e => e.xpath("//info//title");

one-ui/src/main/java/com/marklogic/hub/curation/controllers/CollectionsController.java
```diff
@@ -0,0 +1,4 @@
+package com.marklogic.hub.curation.controllers;
+
+public class CollectionsController {
```
rjrudin: This seems like it can be deleted?

one-ui/src/main/java/com/marklogic/hub/curation/services/CollectionsService.java
```diff
@@ -0,0 +1,76 @@
...
+import com.marklogic.client.util.RequestParameters;
+import com.marklogic.hub.HubConfig;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Service;
+
+import java.util.ArrayList;
+import java.util.List;
+
+@Service
+public class CollectionsService {
```
rjrudin: This doesn't appear to be used anywhere, so it should be deleted.

If there is a need for it, I'd like to talk about it. There is nothing preventing a database from having millions of collections or more. And collections are implementation details - it's very rare that we ever want to expose them to users. So I'd want to think carefully about the requirement of "Get me all the collections except 'internal' ones".

one-ui/src/main/java/com/marklogic/hub/curation/controllers/CollectionsController.java
hao1st: my bad, something wrong during commit and push.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/artifacts/core.sjs
hao1st: agree. we can use config. @ryanjdew do you agree?

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/artifacts/core.sjs
hao1st: `let entities = cts.search(cts.collectionQuery("http://marklogic.com/entity-services/models"))
const entityNames = entities.toArray().map(e => e.xpath("//info//title"));
const res = cts.search(cts.andQuery(queriesForAnd.concat(cts.jsonPropertyValueQuery("entityName", entityNames)))).toArray();

let res2 = res
    .reduce((acc, mapping) => {
      acc[mapping.entityName] = acc[mapping.entityName] || {};
      acc[mapping.entityName].mapping = (acc[mapping.entityName].mappings || []);
      acc[mapping.entityName].mapping.push(mapping);
      return acc;
    }, {});`

I tried to use the above code, and did not work.

hao1st:
> Looks good. In DH today, we're are using "targetEntity" instead of "entityName".
> 
> @rjrudin I know you're big on good naming conventions. Thoughts? One thing to consider is that in the current mappings, while it will work with only specifying an entity model name, I specifically made it so you could specify the absolute IRI of an entity, so versioned entities could be used.
> 
> And apologies @hao1st, I should have caught that when reviewing the ticket before.

@ryanjdew NP, I changed it and modified the ticket

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/artifacts/mapping.sjs
```diff
@@ -0,0 +1,78 @@
...
+
+function validateArtifact(artifact) {
+    const missingProperties = requiredProperties.filter((propName) => !artifact[propName]);
+    if (missingProperties.length) {
+        return new Error(`Missing the following required properties: ${JSON.stringify(missingProperties)}`);
+    }
+    return artifact;
+}
+
+module.exports = {
```
bsrikan: @hao1st can we please add a test for validateArtifact?

one-ui/src/main/java/com/marklogic/hub/curation/services/CollectionsManager.java
```diff
@@ -0,0 +1,75 @@
...
+import com.marklogic.client.io.JacksonHandle;
+import com.marklogic.client.util.RequestParameters;
+import com.marklogic.hub.oneui.models.HubConfigSession;
+import java.util.ArrayList;
+import java.util.List;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Service;
+
+@Service
+public class CollectionsManager {
```
bsrikan: Does this go with the discussion we had today. Should this be @Service or @Component.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/artifacts/core.sjs
hao1st: Fixed.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/artifacts/mapping.sjs
hao1st: i will talk to @ryanjdew tomorrow, and we can simply the validateArtifact implemented in core.sjs. Actually we do not need to have this one in the artifact module.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/artifacts/core.sjs
```diff
@@ -26,7 +27,8 @@ const cachedArtifacts = {};
 const registeredArtifactTypes = {
     loadData: LoadData,
     flows: Flows,
-    stepDefinitions: StepDefs
+    stepDefinitions: StepDefs,
+    mapping: Mapping
```
rjrudin: For consistency, shouldn't this be "mappings" and "Mappings"?

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/artifacts/core.sjs
```diff
@@ -49,19 +51,31 @@ function getTypesInfo() {
...
+            return cts.search(queriesForAnd.length > 1 ? cts.andQuery(queriesForAnd) : queriesForAnd[0]).toArray();
+        } else {
+            return getArtifactsGroupByEntity(queriesForAnd)
         }
     }
-    return artifacts;
+    return [];
+}
+
+function getArtifactsGroupByEntity(queriesForAnd) {
```
rjrudin: Is this returning artifacts, or configs, or settings? Are "configs" the same as "settings"? I don't think this is returning artifacts, which are the actual flows/stepDefinitions/mappings/etc. The variables indicate that these are "configs", so I'm wondering if those are the same as "settings".

one-ui/src/main/java/com/marklogic/hub/curation/controllers/CollectionsController.java
rjrudin: This is still in the PR, it should be removed, right?

one-ui/src/main/java/com/marklogic/hub/curation/services/CollectionsManager.java
rjrudin: I don't think this class is supposed to be in here, right? It's only used by CollectionsCOntroller, which isn't used - nor seems relevant to DHFPROD-4308.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/artifacts/core.sjs
hao1st: `[
  {
    "name": "CustomerRecords", 
    "config": [
      {
        "name": "Map 1",
        "description": "",
        "sourceFormat": "json",
        "selectedSource":"collection", //(collection  | query)
        "sourceQuery": "",
        "collections": ['Provider'],
        "lastUpdated": "2020-02-15T11:25:35.765299-08:00"
      },
      {
        "name": "Customer Mapping - New",
        "description": "",
        "sourceFormat": "xml",
        "selectedSource":"collection", //(collection  | query)
        "sourceQuery": "cts.uris(null,null,cts.andQuery[cts.CollectionQuery('Customer'),cts.elementAttributeValueQuery()])",
        "collections": ['Customer'],
        "lastUpdated": "2020-02-11T11:25:35.765299-08:00"
      }
    ]
  },
  {
    "name": "StudentCollegeRecordsNew",
    "config": [
      {
        "name": "Map 4",
        "description": "",
        "sourceFormat": "json",
        "selectedSource":"collection", //(collection  | query)
        "sourceQuery": "",
        "collections": ['Provider'],
        "lastUpdated": "2020-02-15T11:25:35.765299-08:00"
      },
      {
        "name": "StudentRec Mapping - New",
        "description": "",
        "sourceFormat": "xml",
        "selectedSource":collection", //(collection  | query),
        "sourceQuery": "",
        "collections": ['Customer'],
        "lastUpdated": "2020-02-11T11:25:35.765299-08:00",
      }
    ]
  }`

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/artifacts/core.sjs
hao1st: @rjrudin @ryanjdew We used to have a folder like "mappings/{mappingstepname}" to be used to store [entity model <=> doc instance] mapping configuration with a filename like {mappingstepname}.{version}.mapping.json .
currently if i use "mapping" as artifact type, a new "mapping" folder willl be created to store the mapping artifacts with filename  like {mappingName}.mapping.json. If we use "mapings", it will cause name conflict. I am not sure it is a good name or a better folder stucture we should define for "mapping" operations.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/artifacts/core.sjs
hao1st: I guess we had better use "configs" rather than "config" in the json above.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/artifacts/mapping.sjs
hao1st: @ryanjdew In the core.sjs, 
```
function validateArtifact(artifactType, artifactName, artifact) {
    const artifactLibrary =  getArtifactTypeLibrary(artifactType);
    const validatedArtifact = artifactLibrary.validateArtifact(artifact, artifactName);
    if (validatedArtifact instanceof Error) {
        returnErrToClient(400, 'BAD REQUEST', validatedArtifact.message);
    }
    return validatedArtifact;
}

I think we can move the function used in loadData.sjs to core.sjs

```
function validateArtifact(artifact) {
    const missingProperties = requiredProperties.filter((propName) => !artifact[propName]);
    if (missingProperties.length) {
        return new Error(`Missing the following required properties: ${JSON.stringify(missingProperties)}`);
    }
    return artifact;
}
```
Different artifacts just need to expose "requiredProperties" so that we do not need to override/duplicate validateArtifact with the same implmentation.

File: marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/artifacts/core.sjs (hunk 8 above)
Comment (rjrudin) [reply marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/artifacts/core.sjs]: I'm confused - the directory for mappings in DHF today is "mappings" (and "flows" for flows and "step-definitions" for step definitions). Aren't these mapping artifacts still going into the "mappings" directory?

File: marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/artifacts/core.sjs (hunk 9 above)
Comment (rjrudin) [reply marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/artifacts/core.sjs]: So the use case is "Get me all the artifacts for each known entity type", right? 

If so, shouldn't "configs" be replaced with "artifacts"? You're then returning an array of objects, where each object has an entity type (I recommend replacing "name" with "entityType") and an array of artifacts that reference the entity type.

File: marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/artifacts/core.sjs (hunk 8 above)
Comment (hao1st) [reply marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/artifacts/core.sjs]: Yes, It can.  we have {mappingName} as subfolder. i will modify it to mappings

File: marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/artifacts/core.sjs (hunk 9 above)
Comment (hao1st) [reply marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/artifacts/core.sjs]: OK, it is fine.

File: marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/artifacts/core.sjs (hunk 8 above)
Comment (bsrikan) [reply marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/artifacts/core.sjs]: The dir structure today for mappings is mappings/<flowName-stepName>/<flowName-stepName-version>.mapping.json

version starts with 0 when a mapping step is created and bumped to 1 when mapping is configured after ingestion. Version remains at 1 with further changes to mapping.

File: marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/artifacts/mapping.sjs (hunk 6 above)
Comment (ryanjdew) [reply marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/artifacts/mapping.sjs]: @hao1st  We'll want to extend the validation to be able to encompass more than the required properties. 

e.g., for mapping, we'll want to enable the validation to eventually do the validation XPath expressions, etc. We already have the libraries for that from what we used in QS. We'll want to eventually call those from the mapping artifact library.

Comment (hao1st) [reply marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/artifacts/mapping.sjs]: @bsrikan OK, I added the validation test case for mapping test.

File: marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/artifacts/core.sjs [hunk 10]
```diff
@@ -49,19 +51,31 @@ function getTypesInfo() {
...
     const artifactLibrary =  getArtifactTypeLibrary(artifactType);
     for (const coll of artifactLibrary.getCollections()) {
         queriesForAnd.push(cts.collectionQuery(coll));
     }
     if (queriesForAnd.length) {
-        const results = cts.search(queriesForAnd.length > 1 ? cts.andQuery(queriesForAnd) : queriesForAnd[0]);
-        for (const result of results) {
-            artifacts.push(result.toObject());
+        if (artifactType == "loadData") {
+            return cts.search(queriesForAnd.length > 1 ? cts.andQuery(queriesForAnd) : queriesForAnd[0]).toArray();
```
rjrudin: One nit - I would just call this "queries" (the plural implying that it's an array). And then it doesn't matter if there's one or more - you can just do:

    cts.search(cts.andQuery(queries)).toArray()

hao1st: Why do you think CollectionsController/Manager should be removed? I noticed that we could not reuse CollectionsService in com.marklogic.hub.web.service package and also depends on HubConfig.

one-ui/src/main/java/com/marklogic/hub/curation/controllers/CollectionsController.java
rjrudin: Let's put this in a separate PR for a separate sub-task under 3991. It needs a test for it, and we should instead use a new DS endpoint instead of continuing use of REST endpoints (and the mlCollections REST extension is doing too many things too, it's not just used for retrieving collections).

hao1st:
@rjrudin make sense. I created a subtask for it.

PR #3593, opened 2020-02-20
DHFPROD-3854: can now run arbitrary value queries for steps

aebadirad:


aebadirad:
Note: I've got the history of commits back because I did a full rebase merge due to conflicts. So there's more to this PR as its correcting the history of the branch back to a single line when this branch was first created

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/flows/test-data/valueStep.sjs
```diff
@@ -0,0 +1,35 @@
...
+    } else {
+    concatString += valueNormalized;
+    }
+  }
+  if(concatString !=  expectedValues) {
+          datahub.debug.log({message: "Expected concat of values did not match.", type: 'error'});
+          throw Error("Expected concat of values did not match. Got: " +concatString + " but was expecting:" + expectedValues);
+  }
+
+  return null;
```
rjrudin: What do you think of returning a content object here that has a document in it containing the values? The test can then verify that the document was inserted with the expected values from the values query.

aebadirad: I initially did that, but the order of the values is important to be preserved (in the order they were submitted) as they are processed (same with an acceptsBatch sequence) due to the ordered nature of sequence (esp if it is sorted by order in the batching already).

Could probably do another test with a list of values that are integer based and pass them in as a batch, being able to cleanly loop over and extract the values from them and compare them with an expected to cover ourselves both ways.

SameeraPriyathamTadikonda:
Please rebase with feature/5.3.0 to avoid conflicts.

PR #3595, opened 2020-02-21
DHFPROD-4478: loadData artifact is now included in the downloaded zip

bsrikan:
Fix was really simple and added a test but didnt run it. I would like to know how to run them locally.
Moved loadDataConfig object to testHelper so I can use it in EnvironmentControllerTest. We should probably discuss on how to arrange these tests so as to have a common helper class.

### DIFF
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/impl/HubProjectImpl.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/impl/HubProjectImpl.java
@@ -643,7 +643,7 @@ public class HubProjectImpl implements HubProject {
 
     private void writeToStream(OutputStream out) {
         try (ZipOutputStream zout = new ZipOutputStream(out)){
-            Stream.of("entities", "flows", "src" + File.separator + "main", "mappings", "step-definitions", "gradle",
+            Stream.of("entities", "flows", "src" + File.separator + "main", "mappings", "step-definitions", "loadData", "gradle",
                 "gradlew", "gradlew.bat", "build.gradle", "gradle.properties").forEach(file ->{
                 File fileToZip = getProjectDir().resolve(file).toFile();
                 try {
--- a/one-ui/src/test/java/com/marklogic/hub/curation/controllers/EnvironmentControllerTest.java
+++ b/one-ui/src/test/java/com/marklogic/hub/curation/controllers/EnvironmentControllerTest.java
@@ -5,13 +5,16 @@ import ch.qos.logback.classic.LoggerContext;
 import ch.qos.logback.classic.spi.ILoggingEvent;
 import ch.qos.logback.core.LogbackException;
 import com.fasterxml.jackson.databind.ObjectMapper;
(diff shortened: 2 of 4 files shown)


marklogic-data-hub/src/main/java/com/marklogic/hub/impl/HubProjectImpl.java
```diff
@@ -643,7 +643,7 @@ public void exportProject(OutputStream outputStream) {
 
     private void writeToStream(OutputStream out) {
         try (ZipOutputStream zout = new ZipOutputStream(out)){
-            Stream.of("entities", "flows", "src" + File.separator + "main", "mappings", "step-definitions", "gradle",
+            Stream.of("entities", "flows", "src" + File.separator + "main", "mappings", "step-definitions", "loadData", "gradle",
```
rjrudin: @aebadirad @ryanjdew Do you know the history of why these folders aren't under "src/main" to begin with? If they were, we'd only need to grab "src/main" here.

rjrudin: Also, the name "loadData" hurts my head, since it's a "verb + noun" and everything else is "nouns". Guess that ship has sailed though.

one-ui/src/test/java/com/marklogic/hub/curation/controllers/EnvironmentControllerTest.java
```diff
@@ -49,10 +53,8 @@
 import org.springframework.test.context.junit.jupiter.SpringExtension;
 import org.springframework.web.multipart.MultipartFile;
 
-import static org.junit.jupiter.api.Assertions.assertFalse;
-import static org.junit.jupiter.api.Assertions.assertThrows;
-import static org.junit.jupiter.api.Assertions.assertTrue;
-import static org.junit.jupiter.api.Assertions.fail;
+import static org.junit.jupiter.api.Assertions.*;
```
hao1st: we had better use single class import. You can check your java code style configuration. Make sure the option "Use single class import" checked.

bsrikan: I have single class import checked already:

image

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/HubProjectImpl.java
aebadirad: artifacts != code, but that's why I designed it that way. Current trajectory seems to be more 'java' oriented, so it may make sense to change that in the future.

one-ui/src/test/java/com/marklogic/hub/curation/controllers/EnvironmentControllerTest.java
hao1st: Screen Shot 2020-02-21 at 9 31 04 AM

PR #3601, opened 2020-02-21
DHFPROD-4358: fix button layout issue during resizing

hao1st:
https://project.marklogic.com/jira/browse/DHFPROD-4358

1. Fixed button layout;
2. Enable/Disable upload button;


hao1st:
@bsrikan The PR is only for button alignment. I will have another PR for confirmation page and progression status bar issues.

bsrikan:
@hao1st if this is only for alignment, I would suggest taking the part for disabling Upload button out and including all the Upload code wrt role based access kept together in one PR.

PR #3602, opened 2020-02-22
DHFPROD-4308: return all the entity types even if no artifacts are as…

hao1st:
…sociated with


rjrudin:
I missed this before it was merged, but we ought to have a unit test for this scenario.

PR #3603, opened 2020-02-22
DHFPROD-3922: incorporate loadData options when running ingest

ryanjdew:


marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/LoadUserArtifactsCommand.java
```diff
@@ -188,44 +201,45 @@ public FileVisitResult preVisitDirectory(Path dir, BasicFileAttributes attrs) th
...
-                            flowResourceToURI,
-                            buildMetadata(hubConfig.getFlowPermissions(), "http://marklogic.com/data-hub/flow"),
-                            stagingFlowDocumentWriteSet,
-                            finalFlowDocumentWriteSet
-                        );
-                        return FileVisitResult.CONTINUE;
-                    }
-                });
+            ArtifactManager artifactManager = new ArtifactManagerImpl(hubConfig);
+            List<String> roles = new LinkedList<>();
```
hao1st:
```suggestion
            Set<String> roles = new HashSet<>();
```

marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/LoadUserArtifactsCommand.java
```diff
@@ -188,44 +201,45 @@ public FileVisitResult preVisitDirectory(Path dir, BasicFileAttributes attrs) th
...
+                (roleNode) -> roles.add(roleNode.asText())
+            );
+            ArtifactService artifactService = ArtifactService.on(stagingClient);
+            ObjectMapper mapper = new ObjectMapper();
+            // New way to deploy Artifacts via Data Services
+            for (ArtifactTypeInfo typeInfo: artifactManager.getArtifactTypeInfoList()) {
+                final String artifactType = typeInfo.getType();
+                final Path artifactPath = hubConfig.getHubProject().getProjectDir().toAbsolutePath().resolve(typeInfo.getDirectory().replaceAll("^/","")).toAbsolutePath();
+                logger.info("Deploying artifacts of type '" + artifactType + "' located at '"+ artifactPath.toString() +"'" );
+                if (roles.contains(typeInfo.getUpdateRole()) && artifactPath.toFile().exists()) {
```
hao1st: put the following right after for loop
`
if (roles.contains(typeInfo.getUpdateRole()) {
     continue;
 }
`

marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/LoadUserArtifactsCommand.java
```diff
@@ -188,44 +201,45 @@ public FileVisitResult preVisitDirectory(Path dir, BasicFileAttributes attrs) th
...
+            List<String> roles = new LinkedList<>();
+            RolesService.on(stagingClient).getRoles().elements().forEachRemaining(
+                (roleNode) -> roles.add(roleNode.asText())
+            );
+            ArtifactService artifactService = ArtifactService.on(stagingClient);
+            ObjectMapper mapper = new ObjectMapper();
+            // New way to deploy Artifacts via Data Services
+            for (ArtifactTypeInfo typeInfo: artifactManager.getArtifactTypeInfoList()) {
+                final String artifactType = typeInfo.getType();
+                final Path artifactPath = hubConfig.getHubProject().getProjectDir().toAbsolutePath().resolve(typeInfo.getDirectory().replaceAll("^/","")).toAbsolutePath();
```
rjrudin: I think this is a good method to add to HubProject. I think HubProject should be aware of an ArtifactTypeInfo, and thus have a method of:

    Path getArtifactTypePath(ArtifactTypeInfo)

marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/LoadUserArtifactsCommand.java
rjrudin: I was thinking - first check to see if the artifactPath doesn't exist, and if not, log an info message of "Artifact directory (directory) does not exist, so will not load any artifacts from it". Then check to see if the user has the update role, and if so, load the artifacts; if not, log an info message stating "User is not permitted to update artifacts of type (type), so will not load any artifacts of that type".

marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/LoadUserArtifactsCommand.java
```diff
@@ -188,44 +201,45 @@ public FileVisitResult preVisitDirectory(Path dir, BasicFileAttributes attrs) th
...
+                logger.info("Deploying artifacts of type '" + artifactType + "' located at '"+ artifactPath.toString() +"'" );
+                if (roles.contains(typeInfo.getUpdateRole()) && artifactPath.toFile().exists()) {
+                    final String fileExtension = "*" + typeInfo.getFileExtension();
+                    BaseModulesFinder modulesFinder = new BaseModulesFinder(){
+                        @Override
+                        protected Modules findModulesWithResolvedBaseDir(String resolvedBaseDir) {
+                            Modules modules = new Modules();
+                            logger.info("resolvedBaseDir: " + resolvedBaseDir);
```
rjrudin: This looks like it should be a debug message or removed.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/artifacts/core.sjs
```diff
@@ -43,7 +43,8 @@ function getTypesInfo() {
                 fileExtension: getArtifactFileExtension(artifactType),
                 directory: getArtifactDirectory(artifactType),
                 nameProperty: artifactLibrary.getNameProperty(),
-                versionProperty: artifactLibrary.getVersionProperty()
+                versionProperty: artifactLibrary.getVersionProperty(),
+                updateRole: artifactLibrary.getPermissions().filter((perm) => perm.capability === 'update').map((perm) => xdmp.roleName(perm.roleId))[0]
```
rjrudin: Similar to https://github.com/marklogic/marklogic-data-hub/pull/3605, I'm wondering if it'd be better for the "roles" service to instead include true/false properties for e.g. "canLoadMappings", "canLoadFlows", etc. That prevents the Java tier from having knowledge of any roles. 

More importantly, ML knows the answer to the question of whether a user can load a mapping, so it should be the one to determine a true/false value for that, not the Java code.

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/ArtifactManagerImpl.java
```diff
@@ -29,18 +30,20 @@
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.Iterator;
+import java.util.LinkedList;
+import java.util.List;
 
 @Component
 public class ArtifactManagerImpl implements ArtifactManager {
     HubConfig hubConfig;
+    protected List<ArtifactTypeInfo> allArtifactTypeInfo = null;
```
rjrudin: What's the benefit of saving this state in the class? Are there use cases where we may invoke multiple methods on an instance of this class for the same user?

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/ArtifactManagerImpl.java
```diff
@@ -29,18 +30,20 @@
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.Iterator;
+import java.util.LinkedList;
+import java.util.List;
 
 @Component
```
rjrudin: I think "Component" can be removed, right? I'd rather see the tests that are autowiring this in just instantiate an instance when it needs it (there are only 3 of them).

marklogic-data-hub/src/main/java/com/marklogic/hub/flow/impl/FlowRunnerImpl.java
```diff
@@ -592,4 +606,66 @@ public void setFlowManager(FlowManager flowManager) {
     public HubConfig getHubConfig() {
         return hubConfig;
     }
+
+    // Brings in artifact settings for running a step
+    private void applyArtifactSettings(Step step, Map<String, Object> stepConfig) {
```
rjrudin: How about making this a protected method? Then you can easily unit test the various branches in this method without having to connect to ML or setup a test flow.

marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/LoadUserArtifactsCommand.java
ryanjdew: Refactored to add a canUserUpdate property artifactTypeInfo class, so checking the roles is no longer necessary.

marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/LoadUserArtifactsCommand.java
ryanjdew: Changed to use continue calls and added more specific logging as to why an artifact type may not be loaded.

marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/LoadUserArtifactsCommand.java
ryanjdew: Moved the logic for the path into the HubProject

marklogic-data-hub/src/main/java/com/marklogic/hub/artifact/ArtifactTypeInfo.java
```diff
@@ -0,0 +1,101 @@
...
+
+    public String getUpdateRole() {
+        return updateRole;
+    }
+
+    public void setUpdateRole(String updateRole) {
+        this.updateRole = updateRole;
+    }
+
+    public static ArtifactTypeInfo deserialize(ObjectNode jsonObject) {
```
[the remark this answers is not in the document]
ryanjdew: Refactored to use an ObjectMapper

marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/LoadUserArtifactsCommand.java
ryanjdew: Moved the logic for the path into the HubProject

ryanjdew: Moved the logic for the path into the HubProject

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/ArtifactManagerImpl.java
ryanjdew: No longer keeping the state. With the change to deliberately state whether the current user can update an artifact type, we want to ensure that information is current anyway.

marklogic-data-hub/src/main/java/com/marklogic/hub/artifact/ArtifactTypeInfo.java
```diff
@@ -0,0 +1,77 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.marklogic.hub.artifact;
+
+/**
+ * Provides metadata about an artifact type
+ */
+public class ArtifactTypeInfo {
+    private String type = null;
```
hao1st: We do not need to initialize these member variable. They shoud be default.

ryanjdew: Removed unnecessary initializations and setters.

marklogic-data-hub/src/main/java/com/marklogic/hub/flow/impl/FlowRunnerImpl.java
```diff
@@ -599,4 +613,66 @@ public void setFlowManager(FlowManager flowManager) {
     public HubConfig getHubConfig() {
         return hubConfig;
     }
+
+    // Brings in artifact settings for running a step
+    protected void applyArtifactSettings(Step step, Map<String, Object> stepConfig) {
```
rjrudin: I think we should have unit tests for this, but I'd actually like to write them. I created a subtask for the work.

ryanjdew:
Updated it so that the RunFlowWithoutProjectTests would pass by also updating referenced loadData artifacts with an absolute inputFilePath

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/artifacts/core.sjs
```diff
@@ -36,14 +36,16 @@ function getTypesInfo() {
     for (const artifactType of Object.keys(registeredArtifactTypes)) {
         if (registeredArtifactTypes.hasOwnProperty(artifactType)) {
             const artifactLibrary = registeredArtifactTypes[artifactType];
+            const updateRoles = artifactLibrary.getPermissions().filter((perm) => perm.capability === 'update').map((perm) => perm.roleId);
+            const currentRoles = xdmp.getCurrentRoles().toArray();
```
hao1st:
```suggestion
            const updateRoles = artifactLibrary.getPermissions().filter((perm) => perm.capability === 'update').map((perm) => String(perm.roleId));
            const currentRoles = xdmp.getCurrentRoles().toArray().map(String);
```

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/artifacts/core.sjs
```diff
@@ -36,14 +36,16 @@ function getTypesInfo() {
...
             typesInfo.push({
                 type: artifactType,
-                storageDatabases: artifactLibrary.getStorageDatabases(),
-                collections: artifactLibrary.getCollections(),
                 fileExtension: getArtifactFileExtension(artifactType),
                 directory: getArtifactDirectory(artifactType),
                 nameProperty: artifactLibrary.getNameProperty(),
-                versionProperty: artifactLibrary.getVersionProperty()
+                versionProperty: artifactLibrary.getVersionProperty(),
+                userCanUpdate: userCanUpdate
```
hao1st: @ryanjdew I just tested it, userCanUpdate is always false. After I did the above changes as suggestion, it works.
Also can we add "userCanRead" as part of artifact type info?

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/artifacts/core.sjs
ryanjdew: Updated to map to strings so they're accurately compared.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/artifacts/core.sjs
ryanjdew: Added the logic for userCanRead

PR #3605, opened 2020-02-22
DHFPROD-4337: grant manage privileges to users allowing install/uploa…

hao1st:
…d operation


one-ui/src/main/java/com/marklogic/hub/oneui/auth/LoginLogoutHandler.java
```diff
@@ -43,19 +42,35 @@ public void onAuthenticationSuccess(HttpServletRequest request, HttpServletRespo
...
             try {
                 RolesService rolesService = RolesService.on(authenticationToken.getHubConfigSession().newStagingClient());
                 resp.putArray("roles").addAll((ArrayNode) rolesService.getRoles());
+                if (authenticationToken.hasManagePrivileges()) {
+                    ArrayNode rolesNodes = (ArrayNode) resp.get("roles");
+                    rolesNodes.forEach(e -> {
+                        if ("data-hub-environment-manager".equals(e.asText())) {
+                            managePrivilege[0] = true;
+                            return;
+                        } else if ("manager-admin".equals(e.asText())) {
```
bsrikan: This should be "manage-admin".
manage-admin, security and data-hub-environment-manager are roles and so managePrivilege should probably be called hasManageAdminAndSecurityRole? 
final variable should be all capital.
`rolesNodes.forEach(role ->` is more readable

one-ui/src/main/java/com/marklogic/hub/oneui/auth/LoginLogoutHandler.java
```diff
@@ -43,19 +42,35 @@ public void onAuthenticationSuccess(HttpServletRequest request, HttpServletRespo
...
                 if (e.getServerStatusCode() == HttpStatus.SC_NOT_FOUND) {
                     isInstalled = false;
                 }
             }
         }
+
         resp.put("isInstalled", isInstalled);
-        resp.put("hasManagePrivileges", authenticationToken.hasManagePrivileges());
+        resp.put("hasManagePrivileges", !authenticationToken.hasManagePrivileges() ? false :
+            (managePrivilege[0] || managePrivilege[1] && managePrivilege[2]) ? true : false);
```
bsrikan: We should consider renaming hasManagePrivileges to something else. It seems like we are comparing privilege to a role.
Can we simplify this, by having a variable for line 73. And arr[0]/[1]/[2] dont read too well.. maybe something like envManager || manageAdmin && security

hao1st: added test case. It is a temporary workaround and we will revisit it and once a better function exposed from RoleService. And right now one array defintion is better than 3 arrays for lamda.

bsrikan:
If the added tests had passed, they should probably be updated as well. They didnt catch the issue I ran into.

one-ui/src/test/java/com/marklogic/hub/oneui/AuthTest.java
```diff
@@ -34,6 +35,12 @@
     @Value("${test.mlHost:localhost}")
     public String mlHost;
 
+    @Value("${test.dataHubDeveloperUsername}")
+    public String mlDHDeveloper;
```
rjrudin: Let's not come up with new names for these - just use "dataHubDeveloperUsername" and "dataHubDeveloperPassword".

one-ui/src/main/java/com/marklogic/hub/oneui/auth/LoginLogoutHandler.java
```diff
@@ -43,19 +42,34 @@ public void onAuthenticationSuccess(HttpServletRequest request, HttpServletRespo
         ConnectionAuthenticationToken authenticationToken = (ConnectionAuthenticationToken)authentication;
         ObjectNode resp = mapper.createObjectNode();
         boolean isInstalled = authenticationToken.stagingIsAccessible();
+        final boolean[] managePrivilege = new boolean[3];
         if (authenticationToken.stagingIsAccessible()) {
             try {
                 RolesService rolesService = RolesService.on(authenticationToken.getHubConfigSession().newStagingClient());
                 resp.putArray("roles").addAll((ArrayNode) rolesService.getRoles());
+                if (authenticationToken.hasManagePrivileges()) {
```
rjrudin: Pinging @ryanjdew on this - let's move this logic into ML - I don't know if we're referencing ML role names in our Java code yet, but I'd rather see our Java code at a more abstract level. 

In this case, the Java code needs to know if the user can install/upload. So let's tweak the role service to return both the set of roles (though I'm wondering what those are used for? I don't think the UI should be tightly bound to the names of ML roles) and e.g. "canInstall:true/false". This logic will then go into the roles endpoint, where we can easily unit test it.

ryanjdew: @rjrudin I agree, we should move this logic to ML. The trouble is that we cover the scenario where the Data Hub isn't installed yet, so hitting a Data Service isn't an option at that point.

It is something to revisit, but I think it will take more time to come to the right approach and I don't feel it should hold up this request.

rjrudin: Ah, good point.

one-ui/src/main/java/com/marklogic/hub/oneui/auth/LoginLogoutHandler.java
```diff
@@ -43,19 +42,34 @@ public void onAuthenticationSuccess(HttpServletRequest request, HttpServletRespo
...
+                    ArrayNode rolesNodes = (ArrayNode) resp.get("roles");
```
rjrudin: I don't think this logic is really being exercised by the test, right? i.e. there's just a test for an admin and a test for a non-admin.

I recommend instead moving this logic into a protected method and then writing unit tests (i.e. don't extend HubTestBase) that test the logic in here. The protected method would be like this:

    protected boolean userCanInstall(ArrayNode roles)

The scenarios would then be:

1. User has admin
1. User has manage-admin and security, but not admin
1. User has manage-admin but not security nor admin
1. User has security but not manage-admin nor admin
1. User has none of the 3 roles 

And note the method returns a single boolean - that simplifies the code below, where it doesn't need to look at an array of 3 booleans any more.

hao1st: @rjrudin only 1 or 2 or both can install ? right?

bsrikan: Story says user with role manage-admin and security should be able to install. 
So for the test 1 and 2 should be able to install.

rjrudin: Right, 1 and 2 are true and 3/4/5 are false.

PR #3606, opened 2020-02-24
DHFPROD-4492: Renamed runFlowWithoutProject to runFlow

rjrudin:
And deprecated existing runFlow methods.

This is based on beta and tech summit feedback.


aebadirad:
Just as a notice, this fundamentally changes the default signature in the interface for runflow without deprecation, as such, it'll break every application without code changes.

rjrudin:
I'll hold off on this for now, going to discuss it again with a group.

rjrudin:
Reopening this, as feedback from some folks in the field has been positive about deprecating the methods that depend on Spring and the project files. I don't think there's ever a reason to use those, given that the calls to ML depend on the flow/stepDefinition docs in ML. That raises the possibility of the flow/stepDefinition files on disk differing from the ones in ML, but both sets will be used, which will lead to confusing results.

PR #3611, opened 2020-02-24
DHFPROD-4487: Add user creation to run unit tests

kkanthet:
create user and assign role to run unit tests with different users.

### DIFF
--- a/one-ui/src/test/java/com/marklogic/hub/oneui/TestHelper.java
+++ b/one-ui/src/test/java/com/marklogic/hub/oneui/TestHelper.java
@@ -16,6 +16,13 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.PropertySource;
+import com.marklogic.mgmt.ManageClient;
+import com.marklogic.mgmt.ManageConfig;
+import com.marklogic.mgmt.api.API;
+import com.marklogic.mgmt.api.security.User;
+import org.junit.jupiter.api.AfterAll;
+import java.util.stream.Collectors;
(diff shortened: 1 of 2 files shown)


one-ui/src/test/java/com/marklogic/hub/oneui/TestHelper.java
```diff
@@ -102,4 +122,20 @@ public void addStagingDoc(String uri, DocumentMetadataHandle meta, String resour
...
+        adminAPI = new API(client);
+
+        user = new User(adminAPI, username);
+        user.setUserName(username);
+        user.setPassword(password);
+        user.setRole(Stream.of(role).collect(Collectors.toList()));
+        user.save();
+    }
+
+    @AfterAll
```
rjrudin: Is this actually called? I believe this method must be static - see https://howtodoinjava.com/junit5/after-all-annotation-example/ . 

I also don't think there's any need to delete the user, it can be considered part of the test setup. Just make sure it has a name that makes it clear it's for running tests.

one-ui/src/test/java/com/marklogic/hub/oneui/TestHelper.java
```diff
@@ -102,4 +122,20 @@ public void addStagingDoc(String uri, DocumentMetadataHandle meta, String resour
...
     }
+    private void createUser(String username, String password, String role) {
+        client = new ManageClient();
+        client.setManageConfig(new ManageConfig(mlHost, 8002, adminUserName, adminPassword));
+        adminAPI = new API(client);
+
+        user = new User(adminAPI, username);
+        user.setUserName(username);
+        user.setPassword(password);
+        user.setRole(Stream.of(role).collect(Collectors.toList()));
```
rjrudin: Can just call "addRole" here

PR #3613, opened 2020-02-25
DHFPROD-3918: Bench/flows implementation

wooldridge:
- Add Tool Bench view for managing flows at /bench
- Allow for creating, editing, and deleting flows
- Implement Add Step functionality for Data Load artifacts at /data-load

Screencasts of functionality...

Full privileges (data-hub-flow-reader, data-hub-flow-writer):

https://drive.google.com/file/d/1LidLz5TUzGiZNOHOmdLb8nc9qZ-aoRU9/view

Operator privileges (data-hub-flow-reader):

https://drive.google.com/file/d/1_PN_u2zafHiRpyl-iWZPz0OGQjp7MK2y/view

Minimal privileges:

https://drive.google.com/file/d/1t-P-U8Pkoy08eAw7M8j3mL3OHknwxMAz/view

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/hub-utils/invoke-update-node-path.mjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/hub-utils/invoke-update-node-path.mjs
@@ -1,0 +1,25 @@
+/**
+ Copyright 2012-2020 MarkLogic Corporation
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+ http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+'use strict';
+
+const doc = cts.doc(external.docURI);
+
+if (fn.exists(doc)) {
+    const oldNode = fn.head(cts.doc(external.docURI).xpath(external.xpath));
+    if (fn.exists(oldNode)) {
+        xdmp.nodeReplace(oldNode, external.newNode);
+    }
+}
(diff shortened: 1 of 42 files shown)


one-ui/ui/src/components/flows/flows.tsx
```diff
@@ -0,0 +1,283 @@
...
+            )
+        })
+    }
+
+   return (
+    <div id="flows-container" className={styles.flowsContainer}>
+        {props.canReadFlows || props.canWriteFlows ?
+            <>
+                <div className={styles.createContainer}>
+                    <Button
```
bsrikan: @wooldridge should this be MLButton?
The edit feature should be similar to load data config. An operator should be able to click open the "Edit Flow" dialog to view.
Just for readability, I am not too particular about this. Can properties that belong to one element be in one line? Its ok when there is only 1 or 2 elements. When there are more like in this return block to me its more readable if they are in one line.

wooldridge: Added MlButton and committed. Let's move the discussion of general coding style out to the broader group (including the Explorer team).

wooldridge:
DHFPROD-4513 fixed with the latest commit. Screencast here:

https://drive.google.com/file/d/10D-ay4mMRFKJHTniX0SncF0V2YF74q7-/view

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/hub-utils/invoke-update-node-path.mjs
```diff
@@ -0,0 +1,25 @@
+/**
```
rjrudin: I think this file shouldn't be here, because we can't support .mjs files because we have to still support ML9?

one-ui/src/main/java/com/marklogic/hub/curation/services/FlowManagerService.java
```diff
@@ -38,6 +38,10 @@
     private Scaffolding scaffolding;
 
     public Flow updateFlow(String flowJson) {
+        return updateFlow(flowJson, false);
```
rjrudin: No action needed in this PR, but the next time we have to touch this class, let's fold it into FlowManager.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/hub-utils/invoke-update-node-path.mjs
ryanjdew: yes. it was missed between branches. I'll add a commit to address it.

SameeraPriyathamTadikonda:
Please rebase with feature/5.30

PR #3614, opened 2020-02-26
DHFPROD-4308: add/modify test cases for mapping artifact api 

SameeraPriyathamTadikonda:


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/artifacts/core.sjs
```diff
@@ -249,7 +249,7 @@ function deleteArtifactSettings(artifactType, artifactName, artifactVersion = 'l
     return { success: true };
 }
 
-function getEntityNames() {
+function getEntityTitles() {
```
rjrudin: This is fine here - just an FYI, we'll soon have a library for functions pertaining to entity models, and this will be refactored to use that library.

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/data-services/artifacts/suiteSetup.sjs
```diff
@@ -0,0 +1,35 @@
+const DataHubSingleton = require('/data-hub/5/datahub-singleton.sjs');
+const dataHub = DataHubSingleton.instance();
+
+const permissions = [xdmp.permission('data-hub-entity-model-writer', 'update'), xdmp.permission('data-hub-entity-model-reader', 'read')];
+
+dataHub.hubUtils.writeDocument('/entities/TestEntity-NoMappingConfig', {
```
rjrudin: You don't need to do this - just toss the content into a file under ./test-data/entities in the test file directory, and then call hub-test:load-entities($test:__CALLER_FILE__) (search for references to that for an example). 

Though I did notice you're writing to staging - any reason for that? We usually run the unit tests against the final app server.

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/data-services/artifacts/suiteTeardown.sjs
```diff
@@ -0,0 +1,5 @@
+const DataHubSingleton = require('/data-hub/5/datahub-singleton.sjs');
+const dataHub = DataHubSingleton.instance();
+
+dataHub.hubUtils.deleteDocument('/entities/TestEntity-NoMappingConfig', dataHub.config.STAGINGDATABASE);
```
rjrudin: Don't need to do this either, just call hub-test:delete-artifacts($test:__CALLER_FILE__)

PR #3615, opened 2020-02-26
WIP DHFPROD-4379: Prototyping new library for useful ES functions

rjrudin:
This is an attempt to make a reusable library based on terminology being defined on an internal Wiki page. Let me know what you think about the function signatures in particular.

This is not intended to be complete by any means. It's just a starting point for what is intended to be a widely used library that encapsulates all knowledge about entity models.


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/entity-lib.sjs
```diff
@@ -0,0 +1,133 @@
...
+  const parts = getEntityIRIParts(entityIRI);
+  const entityTitle = parts.entityTitle;
+  return modelDoc.toObject().definitions[entityTitle];
+}
+
+/**
+ * @param entityIRI sem.iri or string
+ * @return {{entityTitle: string, baseUri: string, modelTitle: string, version: string}}
+ */
+function getEntityIRIParts(entityIRI) {
```
ryanjdew: I don't think we can do much better than this approach with what we have available form entity services today, but I think raises additional needs we have the logical model. There is no validation from the ES libraries that prevent the Entity Model TItle or the Entity Type Title from having slashes in them, so we can technically have an IRI that looks like: `http://marklogic.com/example/models/ProductDetail/d-0.0.1/ProductDetail/d` I guess perhaps we could drop into the Entity Model Descriptor document itself and get the baseUri and then use the triples to get the rest of the information?

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/entity-lib.sjs
```diff
@@ -0,0 +1,133 @@
...
+  const map = {};
+  const modelIRI = getModelIRI(model);
+  for (var entityTitle of Object.keys(model.definitions)) {
+    map[modelIRI + "/" + entityTitle] = model.definitions[entityTitle];
+  }
+  return map;
+}
+
+function getModelIRI(model) {
+  const info = model.info;
```
ryanjdew: I think pulling this from the logical model described in the triples would be the ideal way to get this information. Unfortunately, that would require a node stored in the database. The `tde.nodeDataExtract` function looked promising for in-memory nodes, but unfortunately, the collection condition of the TDE is apparently enough to cause no triples to be found. That seems like a short-coming in `tde.nodeDataExtract`

PR #3617, opened 2020-02-27
DHFPROD-4379: Adding library for encapsulating ES logic

rjrudin:
Only modification to existing code is for the entity-services mapper to use this library. We can gradually refactor other areas of DHF to use the functions in here, and of course this library can be enhanced over time.


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mapping/entity-services/lib.sjs
```diff
@@ -155,7 +156,7 @@ function getRelatedMappings(mapping, related = [mapping]) {
 
 function getTargetEntity(targetEntityType) {
   if (!entitiesByTargetType[targetEntityType]) {
-    let entityModel = getModel(targetEntityType);
+    let entityModel = entityLib.findModelForEntityTypeId(targetEntityType);
```
rjrudin: This is the only change to existing code, as it was a trivial refactoring to perform.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/entity-lib.sjs
```diff
@@ -0,0 +1,140 @@
...
+  }
+
+  entityTypeId = entityTypeId.toString();
+
+  const tokens = entityTypeId.split("/");
+  if (tokens.length < 3) {
+    throw Error("Could not get EntityTypeId parts from invalid EntityTypeId: " + entityTypeId);
+  }
+
+  const infoTokens = tokens[tokens.length - 2].split("-");
```
rjrudin: This still has the issue of not working if the EntityTypeTitle has a slash in it - which seems very unlikely. There's a test to verify that it fails. 

@ryanjdew you had mentioned grabbing the model doc and figuring the baseUri from that. I'd like for this method to not avoid any DB calls though - in fact, I have a convention of "find" implying the need to query ML, and "get" not involving any queries (I've used that convention in the past, it's a handy way for a client to know how performant a function should be).

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/entity-lib.sjs
```diff
@@ -0,0 +1,140 @@
...
+ */
+function getEntityTypeIdParts(entityTypeId) {
+  if (entityTypeId == null) {
+    throw Error("Cannot get EntityTypeId parts from null EntityTypeId");
+  }
+
+  entityTypeId = entityTypeId.toString();
+
+  const tokens = entityTypeId.split("/");
+  if (tokens.length < 3) {
```
hao1st: Since we always have baseUrl, the length should be 5 after split with "/"

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/entity-lib.sjs
```diff
@@ -0,0 +1,140 @@
...
+  if (infoTokens.length < 2) {
+    throw Error("Could not get EntityTypeId parts; expected info part did not contain a hyphen; EntityTypeId: " + entityTypeId);
+  }
+
+  const title = tokens[tokens.length - 1];
+  const baseUri = tokens.slice(0, tokens.length - 2).join("/") + "/";
+  return {
+    baseUri: baseUri,
+    modelTitle: infoTokens[0],
+    version: infoTokens.slice(1).join("-"),
```
hao1st:
```suggestion
    version: infoTokens.slice(1)[0],
```

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/entity-lib.sjs
hao1st:
```suggestion
  if (tokens.length < 5) {
```

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/entity-lib.sjs
```diff
@@ -0,0 +1,140 @@
...
+function convertModelToEntityTypeMap(model) {
+  const map = {};
+  const modelId = getModelId(model);
+  for (var entityTypeTitle of Object.keys(model.definitions)) {
+    map[modelId + "/" + entityTypeTitle] = model.definitions[entityTypeTitle];
+  }
+  return map;
+}
+
+function getModelId(model) {
```
hao1st: for consistency, add java doc?

hao1st:
https://github.com/marklogic/marklogic-data-hub/blob/8273a3aed8e201455b8f834d4f043282c82459ab/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/entity-lib.sjs#L66-L72

function name does not match return value "entity definitions"

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/entity-lib.sjs
rjrudin: The problem is ES doesn't require a particular format for baseUri. ES will accept this too: "something/title-version/title". So the minimum length we can support is 3, which then assumes that the second token contains the model title and version.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/entity-lib.sjs
rjrudin: The "join" is needed to support a version like "1.0-SNAPSHOT" - see the test for that.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/entity-lib.sjs
rjrudin: I didn't yet because it's not an exported function. We also don't have any expectation of comments on methods - the vast majority of DH functions don't have comments. I only included them where I felt they added value in this library, and I didn't think they did for getModelId, which is simple enough to understand by looking at the code.

rjrudin:
Fixed a couple unit tests that were failing because they were asserting on the count of entity type IDs, and some entity types were lingering from previous tests, causing the assertion to fail.

PR #3622, opened 2020-02-28
DHFPROD-4285: Can now map multiple properties of the same entity type

rjrudin:
"m:entity" now refers to a unique property path instead of an entity type title.

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mapping/entity-services/lib.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mapping/entity-services/lib.sjs
@@ -27,69 +27,94 @@ const xsltPermissions = [
 
 const reservedNamespaces = ['m', 'map'];
 
-function buildMappingXML(mappingJSON) {
-  // Obtain all linked JSON mappings
-  const relatedMappings = getRelatedMappings(mappingJSON).map((mappingDoc) => mappingDoc.toObject());
+/**
+ * Build an XML mapping template in the http://marklogic.com/entity-services/mapping namespace, which can then be the
+ * input to the Entity Services mappingPut function that generates an XSLT template.
+ *
+ * @param mappingDoc expected to be a document-node and not an object asdfasdf
+ * @return {*}
+ */
+function buildMappingXML(mappingDoc) {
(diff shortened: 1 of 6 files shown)


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mapping/entity-services/lib.sjs
```diff
@@ -141,16 +167,34 @@ function buildMapProperties(mapping, entityModel) {
...
   if (dhMappingTraceIsEnabled) {
     xdmp.trace(dhMappingTrace, `Getting related mappings for '${xdmp.describe(mapping)}'`);
   }
-  let internalMappings = mapping.xpath('/properties//object-node()[exists(targetEntityType) and exists(properties)]');
-  for (let internalMapping of internalMappings) {
-    related.push(internalMapping);
-  }
-  return related;
+  Object.keys(mapping.properties).forEach(propertyTitle => {
+    const property = mapping.properties[propertyTitle];
```
bsrikan: We have a failing test. null check required here.

`com.marklogic.client.FailedRequestException: Local message: failed to apply resource at documents: Internal Server Error. Server Message: JS-JAVASCRIPT: Object.keys(mapping.properties).forEach(propertyTitle => { -- Error running JavaScript request: TypeError: Cannot convert undefined or null to object . See the MarkLogic server error log for further detail.`

rjrudin: I'll fix it, but it's really due to an invalid test file - that test is generating a mapping file with "null" as the value of "properties", which doesn't make any sense.

rjrudin:
Fixed the failing test, which had an invalid mapping file, by ensuring that the mapping passed into getObjectPropertyMappings has a non-null "properties" property.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mapping/entity-services/lib.sjs
bsrikan: True that about the mapping file in the test. When a mapping step is created a version 0 file has similar structure ie no properties.

PR #3623, opened 2020-02-28
DHFPROD-3922: Add Jobs APIs to One UI

ryanjdew:


marklogic-data-hub/src/main/resources/ml-modules/services/mlJobs.sjs
```diff
@@ -43,7 +43,11 @@ function get(context, params) {
     resp = datahub.jobs.getJobDocsByFlow(flow);
   }
   else if (fn.exists(latest)) {
-    resp = datahub.jobs.getLastestJobDocPerFlow();
+    flowNames = (fn.exists(flowNames)) ? datahub.hubUtils.normalizeToSequence(flowNames) : flowNames;
```
rjrudin: I think this REST extension is prime for refactoring. The "get" function is difficult to understand because it supports so many operations, and the error that's thrown - "Incorrect options" - isn't helpful. 

What do you think of starting a new "jobs" service with a "getLatestJobForFlows" endpoint? (I don't think "doc" should be part of the name, that's an implementation detail). I understand if it's too much as I know Mike needs this right away, just throwing it out for consideration, as I do think we want to shift to a set of well-defined endpoints for jobs as soon as possible.

one-ui/src/main/java/com/marklogic/hub/curation/controllers/JobsController.java
```diff
@@ -0,0 +1,110 @@
...
+@RequestMapping(value="/api/jobs")
+public class JobsController extends SearchableManager {
+    @Autowired
+    private HubConfigSession hubConfig;
+
+
+    @RequestMapping(method = RequestMethod.GET)
+    @ResponseBody
+    public ResponseEntity<?> getJobs(@RequestParam(value = "flowName", required = false) String flowName) {
+        JobDocManager jobDocManager = getJobDocManager();
```
rjrudin: Would this be appropriate in JobDocManager, or is this logic too specific to the definition of "getJobs" here?

one-ui/src/main/java/com/marklogic/hub/curation/controllers/JobsController.java
```diff
@@ -0,0 +1,110 @@
...
+    @RequestMapping(value = "latest", method = RequestMethod.GET, produces = {MediaType.APPLICATION_JSON_VALUE})
+    @ResponseBody
+    public JsonNode getLatestJobDocsPerFlow() {
+        return flattenJobsJson(getJobDocManager().getLatestJobDocumentForFlows(Collections.emptyList()));
+    }
+
+    private JsonNode getJobs(JobDocManager jobDocManager, String jobId, String flowName) {
+        JsonNode jobsJson = (ArrayNode) jobDocManager.getJobDocument(jobId, flowName);
+        if (jobsJson == null) {
+            throw new RuntimeException("Unable to get job document");
```
rjrudin: I recommend adding the jobId and the flowName to this message - e.g. "Unable to find document with ID ${jobId} for flow ${flowName}".

ryanjdew:
@rjrudin I'm going to agree with @hao1st at refactoring later if just for the reason that @wooldridge is waiting for these changes. 

@wooldridge is there anything that you can see that is missing from the APIs on your side?

PR #3626, opened 2020-02-28
DHFPROD-4523: Initial check-in of Azure DHF Flow sample

vgouraba:
Initial code for Azure Function to invoke DHF Flow.


vgouraba:
Will submit a new one based off of "develop"

PR #3627, opened 2020-03-01
DHFPROD-4054: Reworking Spring Security to use granted authorities

rjrudin:
Changes:

- Added SecurityService.getAuthorities, which has a stubbed out implementation
- Reworked Spring Security so that nearly everything is in the AuthenticationFilter class
- Enabled method-level security so that granted authorities can be used with the @Secured annotation on controller methods

I tested this manually - with DH installed and without it installed - and all seemed fine. But the method-level security test - see what's commented out in EnvironmentController - didn't seem to work yet. Not sure if the proper response is being sent to the browser yet.


one-ui/src/main/java/com/marklogic/hub/oneui/auth/AuthenticationFilter.java
```diff
@@ -0,0 +1,171 @@
...
+        final DatabaseClient stagingClient = hubConfig.newStagingClient(null);
+
+        boolean dataHubInstalled = false;
+        List<GrantedAuthority> authorities = new ArrayList<>();
+        ArrayNode roles = null;
+
+        try {
+            dataHubInstalled = stagingClient.checkConnection().isConnected();
+            if (dataHubInstalled) {
+                roles = (ArrayNode) RolesService.on(stagingClient).getRoles();
```
hao1st: very nice refactors... One question is that why we need to have two services (RolesService and SecurityServicce)? Would it better use a SecurityService which consolidate the roles and authorities along with one DS call?

hao1st: BTW, I still tried to figout out @Secured, @PreAuthorize, @PostAuthorize. Another question is that do we really need to return roles of the current users to front-end. I thought that we wanted to client-side has little knowledge about roles rather than authorities (canInstallDataHub, canRead{artifactType}, canWrite{artifact})  ?

rjrudin: I didn't want to get rid of RolesService and "roles" until the GUI is updated to depend on authorities instead of roles.

rjrudin: I think @Secured should be all we need. That will reference authorities, but will include a "ROLE_" prefix to make Spring Security happy. And yes, the GUI shouldn't care about roles anymore.

PR #3628, opened 2020-03-02
DHFPROD-4418: Fixed caching in getFlow

rjrudin:
Also renamed the two test flow files so they have the correct names as expected by getFlow.

Had accidentally closed this yesterday, not realizing the PR hadn't been merged due to rebase isuses.


rjrudin:
Thanks @akshaysonvane - those tests were still passing for me because I had still the previously-named flow files for that test in my modules database. After a clean deployment, they started failing for me too. Fixed now.

PR #3629, opened 2020-03-02
DHFPROD-4540: Added test for applyArtifactSettings

rjrudin:
Temporarily removed code for applying artifact settings in support of batch size, thread count, a custom hook, and collections/permissions. No tests were depending on this code, and I'm wondering if this feature is needed yet - these things can all still be specified in the flow.

### DIFF
--- a/marklogic-data-hub/src/test/java/com/marklogic/hub/flow/impl/FlowRunnerTest.java
+++ b/marklogic-data-hub/src/test/java/com/marklogic/hub/flow/impl/FlowRunnerTest.java
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package com.marklogic.hub.flow;
+package com.marklogic.hub.flow.impl;
 
 import com.marklogic.bootstrap.Installer;
 import com.marklogic.client.DatabaseClient;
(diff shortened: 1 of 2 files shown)


marklogic-data-hub/src/main/java/com/marklogic/hub/flow/impl/FlowRunnerImpl.java
```diff
@@ -641,34 +641,6 @@ protected void applyArtifactSettings(Step step, Map<String, Object> stepConfig)
                     stepOptions.put("outputFormat", artifactJson.get("targetFormat").asText());
                     stepConfig.put("fileLocations", fileLocations);
                 }
-                ObjectNode artifactSettingsJson = null;
```
rjrudin: I am wondering if this code is needed yet. Is it possible for a client to specify any of these inputs as part of artifact settings? Basically, I'm wondering what the schema for a "settings" is, as all of these things can already be specified in multiple other places.

ryanjdew: It is a combination of this [ticket](https://project.marklogic.com/jira/browse/DHFPROD-3921) for creating the settings and this [ticket](https://project.marklogic.com/jira/browse/DHFPROD-3922) for running an ingestion. We should update the Swagger Docs to contain the settings schema.

rjrudin:
Thanks, I just backed out the removal of that code in FlowRUnnerImpl, though it would be helpful to get tests in place for it as part of the current effort for consuming the batchSize/threadCount/etc stuff in the artifact settings.

PR #3631, opened 2020-03-02
DHFPROD-4169: Map JSON source data to entity type

xnikhil08:
Contains the changes for all the subtasks including DHFPROD-4311, DHFPROD-4312 and DHFPROD-4313


one-ui/ui/src/components/entities/entity-tiles.tsx
```diff
@@ -47,7 +41,25 @@ const EntityTiles: React.FC = () => {
...
-              //handleError(error);
+          }
+    }
+
+    const getMappingArtifactByMapName = async (entityName,mapName) => {
+        try {
+            let response = await axios.get(`/api/artifacts/mapping/${mapName}`);
+            
+            if (response.status === 200) {
+                let entArt = response.data;
```
rjrudin: This is a "mapping", right? "entArt" implies it's an entity model artifact, but the URL indicates it's a mapping.

one-ui/ui/src/components/entities/entity-tiles.tsx
```diff
@@ -63,28 +75,47 @@ const EntityTiles: React.FC = () => {
...
+              return true;
+            } else {
+                return false;
             }
           }
           catch (error) {
-            let message = error.response.data.message;
-            console.log('Error While creating the Load Data artifact!', message)
+            let message = error;
+            console.log('Error While creating the Mapping artifact!', message)
```
rjrudin: If these console.log messages are going to be retained, I recommend some consistency in them - there's "Mapping Artifacts", "MappingArtifact", "Mapping artifact", and "mapping artifact". I believe the term "artifact" is redundant here and doesn't need to be used - "mapping" should suffice, just like "step definition" and "flow".

one-ui/ui/src/components/entities/entity-tiles.tsx
```diff
@@ -63,28 +75,47 @@ const EntityTiles: React.FC = () => {
...
             setIsLoading(false);
-            //handleError(error);
+            return false;
+          }
+    }
+
+    const updateMappingArtifact = async (mapObj) => {
```
rjrudin: Can this be "mapping" instead of "mapObj"? "mapObj" makes me think it's different from a "mapping" in terms of what data it holds. If they're the same, let's only use one term, that's simpler.

one-ui/ui/src/components/entities/entity-tiles.tsx
```diff
@@ -95,10 +126,13 @@ const EntityTiles: React.FC = () => {
...
                     entityName={entMaps.entityType}
+                    getMappingArtifactByMapName={getMappingArtifactByMapName}
                     deleteMappingArtifact={deleteMappingArtifact}
                     createMappingArtifact={createMappingArtifact}
-                    canReadWrite={canReadWrite}
-                    canReadOnly={canReadOnly} />
+                    updateMappingArtifact={updateMappingArtifact}
+                    canReadWrite={props.canReadWrite}
+                    canReadOnly={props.canReadOnly}
+                    entitiesInfo={props.entitiesInfo[entMaps.entityType]} />
```
rjrudin: What does "entities info" refer to? I'm guessing it's the entityType? If so, then I think this should be "entityType={props.entityTypes[entMaps.entityType]}". 

But I think "entMaps" should be renamed as well to something that's more clear. Based on its usage - it has "entityType" and "artifacts" as properties - I'm guessing it's a collection of things pertaining to an entityType. Perhaps "entityCardData" would be better here, as it seems specific to the data needed to display an entity-specific card?

one-ui/ui/src/components/entities/mapping/mapping-card.tsx
```diff
@@ -1,30 +1,47 @@
...
+    const [mapName, setMapName] = useState({});
     const [dialogVisible, setDialogVisible] = useState(false);
     const [loadArtifactName, setLoadArtifactName] = useState('');
+    const [mappingVisible, setMappingVisible] = useState(false);
+    const [sourceData, setSourceData] = useState<any[]>([]);
+    const [sourceURI,setSourceURI] = useState('');
+    const [docNotFound, setDocNotFound] = useState(false);
+
+    //For Entity table
+    const [entityData, setEntityData] = useState<any[]>([]);
```
rjrudin: The user is really constructing a mapping artifact, not an entity. Can we instead say there's a "source table" and a "mapping table"?

one-ui/ui/src/components/entities/mapping/mapping-card.tsx
```diff
@@ -103,6 +110,142 @@ const MappingCard: React.FC<Props> = (props) => {
         <span style={{fontSize: '16px'}}>Are you sure you want to delete this?</span>
         </Modal>;
 
+
+    const getSourceData = async (index) => {
+
+        let database = props.data[index].sourceDatabase || 'data-hub-STAGING';
```
rjrudin: This seems like implementation details that the GUI shouldn't have to worry about. Could the GUI instead say - Hey middle tier, get me a source document for this step?

one-ui/ui/src/components/entities/mapping/mapping-card.tsx
```diff
@@ -103,6 +110,142 @@ const MappingCard: React.FC<Props> = (props) => {
...
+
+        return nestedDoc;
+        
+        
+        
+    }
+    
+    
+    const extractEntityInfoForTable = () => {
+        let entProps = props.entitiesInfo.definitions.definitions[props.entityName].properties;
```
rjrudin: I recommend "entityModel" instead of "entitiesInfo" (though why are there two levels of "definitions" here?) and then "entityTypeTitle" instead of "entityName".

one-ui/ui/src/components/entities/mapping/mapping-card.tsx
```diff
@@ -103,6 +110,142 @@ const MappingCard: React.FC<Props> = (props) => {
...
+        let entTableTempData: any = [];
+        entProps.map(prop => {
+            let propty = {
+                name : prop.name,
+                type : prop.datatype
+            }
+            entTableTempData.push(propty)
+        })
+        setEntityData([...entTableTempData]);
```
rjrudin: "EntityData" sounds like property values from entity instances. But this is really an array of entity type properties, right? So I'd call it "setEntityTypeProperties".

one-ui/ui/src/components/entities/mapping/source-entity-map/source-to-entity-map.module.scss
```diff
@@ -0,0 +1,214 @@
...
+        background-color: #fafafa;
+        line-height: 2px;
+      };
+    & tbody > tr > td {
+        line-height: 2px;
+        vertical-align: top;
+    }
+}
+
+.mapExpContainer {
```
rjrudin: Is "Exp" short for "Expression" here? If so, I think since "mapExpression" is used below, use "mapExpression" here too.

one-ui/ui/src/components/entities/mapping/source-entity-map/source-to-entity-map.module.scss
```diff
@@ -0,0 +1,214 @@
...
+.sourceTable {
+    & thead > tr > th {
+        background-color: #fafafa;
+        line-height: 2px;
+      };
+    & tbody > tr > td {
+        line-height: 2px;
+    }
+}
+.srcTableRows{
```
rjrudin: Since "source" is used above, use "source" instead of "src" for consistency here. I don't think there's ever going to be a reason to abbreviate "source" or "entity".

one-ui/ui/src/pages/Entities.tsx
```diff
@@ -0,0 +1,55 @@
...
+    //Role based access
+    const roleService = useContext(RolesContext);
+    const canReadOnly = roleService.canReadMappings();
+    const canReadWrite = roleService.canWriteMappings();
+
+    const getEntityInfo = async () => {
+        try {
+        let response = await axios.get(`/api/entities`);
+        if(response.status === 200) {
+            let entData:any = {};
```
rjrudin: I recommend renaming "/api/entities" to "/api/entityTypes" and then using "entityTypes" here instead of "entData" and "entitiesInfo".

one-ui/ui/src/util/search-service.tsx
```diff
@@ -0,0 +1,32 @@
+import axios from 'axios'
+
+const getResultsByQuery = async (database: string, ctsQuery: string, count: number, urisOnly: boolean) => {
```
rjrudin: Can this be reworked to accept a flow name and step number? I'd be happy to make the DS endpoint to implement this, as this seems like too much implementation detail in the GUI.

one-ui/ui/src/components/entities/mapping/mapping-card.tsx
xnikhil08: @rjrudin - I used the name 'EntityData' here because it seems consistent with the UI as we have been calling the right-hand side table in mapping UI as Entity table only.

one-ui/ui/src/components/entities/mapping/mapping-card.tsx
xnikhil08: @rjrudin - This is how it works in the current QS as well. We pass the source database name and Source Query to the middle-tier via the API '/search/sjsSearch' to get the list of URIs and then supply one of those URIs to the endpoint '/search/doc' to get the required Doc.
If we want to send only the step number then we will have to redesign this process but the hurdle here is that a mapping artifact will not have a step number unless its added t a flow. Either ways should we take that up in this story itself?

one-ui/ui/src/components/entities/mapping/mapping-card.tsx
xnikhil08: Done

one-ui/ui/src/components/entities/mapping/mapping-card.tsx
xnikhil08: Done.

one-ui/ui/src/components/entities/entity-tiles.tsx
xnikhil08: Done.

one-ui/ui/src/components/entities/mapping/source-entity-map/source-to-entity-map.module.scss
xnikhil08: Done.

one-ui/ui/src/components/entities/mapping/source-entity-map/source-to-entity-map.module.scss
xnikhil08: Done

one-ui/ui/src/pages/Entities.tsx
xnikhil08: i renamed entityInfo to entityModel and also changed the corresponding temporary variables to maintain consistency. And since we have been using the api '/api/entities' in current QS as well so we may want to use the same naming conventions here as well right?

one-ui/ui/src/util/search-service.tsx
xnikhil08: @rjrudin - We can definitely create an endpoint later which accepts only flow name and step number and returns the documents but in this case the mapping artifact is not part of any particular flow unless its added to it, as per the new one-ui design. So we might need to think about it in detail in another story.

one-ui/ui/src/components/entities/entity-tiles.tsx
```diff
@@ -47,7 +41,25 @@ const EntityTiles: React.FC = () => {
           } catch (error) {
               let message = error;
               console.log('Error while fetching mapping artifacts', message);
-              //handleError(error);
+          }
+    }
+
+    const getMappingArtifactByMapName = async (entityName,mapName) => {
```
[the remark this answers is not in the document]
xnikhil08: Done.

one-ui/ui/src/components/entities/entity-tiles.tsx
xnikhil08: Done.

one-ui/ui/src/components/entities/entity-tiles.tsx
xnikhil08: Done.

one-ui/ui/src/components/entities/mapping/mapping-card.tsx
rjrudin: Yes we can revisit later, I may open a ticket to rework this.

xnikhil08:
> I'm seeing the following when trying to build:
> 
> ```
> Creating an optimized production build...
> Failed to compile.
> 
> /Users/mwooldri/projects/marklogic-data-hub/one-ui/ui/src/App.tsx
> TypeScript error in /Users/mwooldri/projects/marklogic-data-hub/one-ui/ui/src/App.tsx(50,39):
> Property 'from' does not exist on type '{}'.  TS2339
> 
>     48 |       if (location.state && !user.redirect && user.error.type === '') {
>     49 |         if (location.state.hasOwnProperty('from')) {
>   > 50 |           history.push(location.state.from.pathname);
>        |                                       ^
>     51 |         }
>     52 |       }
>     53 |       if (user.redirect || location.pathname === '/') {
> ```
> 
> I don't get this error when building on the latest develop branch.
> I was able to avoid the above error after I rebased DHFPROD-4169 onto the current develop.

@wooldridge - Please review now.

wooldridge:
@xnikhil08, I made a PR to your branch that fixes a couple type-checking issues that were failing my local build (including the location.state.from issue above):
https://github.com/xnikhil08/marklogic-data-hub/pull/6

PR #3634, opened 2020-03-02
DHFPROD-4544: Make FlowRunner compatible with session based HubConfig

ryanjdew:


marklogic-data-hub/src/main/java/com/marklogic/hub/flow/impl/FlowRunnerImpl.java
```diff
@@ -545,7 +551,7 @@ public void afterExecute(Runnable r, Throwable t) {
                 //Run the next step
                 else {
                     if(!(threadPool != null && threadPool.isTerminating())) {
```
hao1st: It does not relate to PR. But the condition does not seem correct. 

```suggestion
                    if (!threadPool != null && !threadPool.isTerminating()) {
```

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/ScaffoldingImpl.java
```diff
@@ -43,7 +43,7 @@
 
 @Component
 public class ScaffoldingImpl implements Scaffolding {
-    
+
     @Autowired
```
hao1st: still @Autowired?

ryanjdew: I'd haven't verified that isn't autowired by other Data Hub code. It isn't autowired through any of the OneUI code paths.

marklogic-data-hub/src/main/java/com/marklogic/hub/flow/impl/FlowRunnerImpl.java
```diff
@@ -112,8 +113,9 @@ public FlowRunnerImpl(String host, String username, String password) {
      */
     public FlowRunnerImpl(HubConfig hubConfig) {
         this.hubConfig = hubConfig;
-        this.stepRunnerFactory = new StepRunnerFactory(this.hubConfig);
-        this.stepRunnerFactory.setStepDefinitionProvider(new MarkLogicStepDefinitionProvider(this.hubConfig.newStagingClient(null)));
+        this.flowManager = new FlowManagerImpl(hubConfig);
+        this.stepRunnerFactory = new StepRunnerFactory(hubConfig);
```
hao1st: My gut feeling is the StepRunnerFactory can be implemented as a singleton object.   StepRunnerFactory does not need to keep those states' reference when getStepRunner gets called. MarkLogicStepDefinitionProvider can be implemented as a singleton object. I am not sure why we have to pass stepRunnerFactory to the Runnable?

wooldridge:
@ryanjdew, I'm not seeing the error anymore after the force-push.

marklogic-data-hub/src/main/java/com/marklogic/hub/flow/impl/FlowRunnerImpl.java
rjrudin: The StepDefinitionProvider can't easily be a singleton because we have multiple implementations of it. As for FlowRunnerTask depending on StepRunnerFactory - it's likely that whole design can change, as I don't know that a FlowRunnerImpl is ever really used to run multiple flows at once. But I don't think it's worth changing anything about that right now.

marklogic-data-hub/src/main/java/com/marklogic/hub/flow/impl/FlowRunnerImpl.java
rjrudin: I think that's correct, because it avoids a null-pointer issue if the threadPool is null, since the next line tries to do something with the threadPool.

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/MappingManagerImpl.java
```diff
@@ -47,18 +46,20 @@
 @Component
```
rjrudin: If we're not autowiring this anymore, I think the Component annotation should be removed.

marklogic-data-hub/src/main/java/com/marklogic/hub/flow/impl/FlowRunnerImpl.java
```diff
@@ -118,8 +119,9 @@ public FlowRunnerImpl(String host, String username, String password) {
...
+        this.stepRunnerFactory.setStepDefinitionProvider(new MarkLogicStepDefinitionProvider(hubConfig.newStagingClient(null)));
```
rjrudin: You can get rid of this line of code now that StepRunnerFactory is doing this part itself in its constructor.

one-ui/src/main/java/com/marklogic/hub/curation/controllers/FlowController.java
```diff
@@ -163,4 +163,13 @@ protected ArtifactService getArtifactService() {
         return ArtifactService.on(dataServicesClient);
     }
 
+    private FlowManagerService getFlowManagerService() {
+        lastFlowManagerService = new FlowManagerService(this.hubConfig.getHubConfigImpl());
```
rjrudin: I think this introduces a race condition between two threads hitting this controller. Both threads could hit line 167 at the same time, and then lastFlowManagerService would be equal to whichever thread finished 167 last, and both threads would use the same one. 

Is "lastFlowManagerService" only needed for unit testing purposes? If so, the way to get a reference to it is to subclass FlowController in your test and override this method to do the same thing - i.e. hang onto a reference of the FlowManagerService that is constructed.

one-ui/src/main/java/com/marklogic/hub/curation/services/FlowManagerService.java
```diff
@@ -365,4 +370,13 @@ public Flow stop(String flowName) {
         return getFlow(flowName);
     }
 
+    protected FlowRunnerImpl getFlowRunner() {
+        lastFlowRunner = new FlowRunnerImpl(this.hubConfig);
+        return lastFlowRunner;
+    }
+
+    // TODO public so the FlowControllerTest can see it
+    public FlowRunnerImpl getLastFlowRunner() {
```
rjrudin: FlowManagerService isn't accomplishing anything in terms of reuse - FlowController is its only client (just like in QS). Its name doesn't fit with our naming conventions either. How about just shoving all of it into FlowController, and then you can just do the trick I mentioned above for getting a reference to lastFlowRunner?

marklogic-data-hub/src/main/java/com/marklogic/hub/flow/impl/FlowRunnerImpl.java
hao1st: @rjrudin if threadpool is null, then the current condition is definitly true. It will produce NullPointerExeption! It is a old bug! If the threadpool is not null and it is not terminating, we want to continue execute next task.

marklogic-data-hub/src/main/java/com/marklogic/hub/flow/impl/FlowRunnerImpl.java
hao1st: @rjudin I do not think I need to do now. But I think I can refractor for sure. I even think that we should be able to run multiple flows concurrently with a better threadpool (i/o operations) management in backend.

rjrudin: Currently we don't have any requirement for running multiple flows though the same FlowRunnerImpl instance, so let's not tackle this until such a requirement comes up. When it does, I'd rather see a FlowRunnerImpl with far less state in it, such that we can spin up a new instance easily to run a new flow.

hao1st: yes, I thought that maybe we did not have use cases to do that in the past. But I do not know the reason.

PR #3637, opened 2020-03-03
DHFPROD-4523: Initial checkin of Azure Flow Sample

vgouraba:
Azure Flow sample - to invoke DHF Flow from an Azure Function


azure/flow-runner-function/src/test/java/com/marklogic/dhf/azure/FunctionDhfFlowTest.java
```diff
@@ -0,0 +1,64 @@
...
+        // Running invalid flow throws Exception. The test fails without the try/catch block.
+        try {
+            final HttpResponseMessage ret = new FunctionDhfFlow().run(req, msg, context);
+        } catch (java.lang.RuntimeException ex) {
+            System.out.println (ex.getMessage());
+        }
+
+        //TO-DO: Fix this
+        // Verify - forcing OK to account for RuntimeException.
+        assertEquals(HttpStatus.OK, HttpStatus.OK);
```
hao1st: Should check ret.getStatus() == HttpStatus.OK) ? And move the line to try/catch block?

azure/flow-runner-function/src/test/java/com/marklogic/dhf/azure/HttpResponseMessageMock.java
```diff
@@ -0,0 +1,81 @@
...
+
+        @Override
+        public Builder status(HttpStatusType httpStatusType) {
+            this.httpStatusCode = httpStatusType.value();
+            this.httpStatus = httpStatusType;
+            return this;
+        }
+
+        @Override
+        public HttpResponseMessage.Builder header(String key, String value) {
```
hao1st: for consistency, just use Builder ?

vgouraba: HttpResponseMessageMock is a sample provided by Azure. Didn't want to make changes to it, unless its absolutely necessary. There might be other use cases (later on) that may require other variations.

azure/flow-runner-function/src/test/java/com/marklogic/dhf/azure/FunctionDhfFlowTest.java
vgouraba: For this Jira, the intent is to just invoke an invalid flow. As part of next set of Jiras, we would end up testing both positive and negative flows/scenarios. In this case, the expected behavior is to get java exception.

hao1st: I guess you can use assertThrows
https://howtodoinjava.com/junit5/expected-exception-example/

rjrudin: @hao1st For the scope of this story, we're not concerned much with the code, we're just focused on being able to package something up that includes DHF and then run it in Azure. Most of this code was auto-generated by an Azure archetype for Maven.

azure/flow-runner-function/src/test/java/com/marklogic/dhf/azure/HttpResponseMessageMock.java
rjrudin: Same thing here, this is mostly auto-generated by the Azure Maven archetype. We can change it later once we actually try to run a flow.

azure/flow-runner-function/pom.xml
```diff
@@ -0,0 +1,267 @@
...
+            </snapshots>
+        </repository>
+      <!-- Required for some DHF dependencies, such as ml-app-deployer -->
+      <repository>
+        <id>jcenter</id>
+        <url>https://jcenter.bintray.com/</url>
+      </repository>
+      <repository>
+        <id>marklogic-internal</id>
+        <url>http://distro.marklogic.com/nexus/repository/maven-releases/</url>
```
rjrudin: Just a note, we can remove this once 5.2.0 is officially out.

PR #3639, opened 2020-03-03
DHFPROD-4524: 5.3 Roles service doesn't provide granular enough feedb…

hao1st:
…ack to cover manage-admin/privileges

https://project.marklogic.com/jira/browse/DHFPROD-4524

We will remove roles and hasManagePrivileges once UI gets changed.


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/security/getAuthorities.sjs
```diff
@@ -15,14 +15,49 @@
...
+if (!hasManagerRole) {
+  hasManagerRole = manageRolesMustAllMatched.every((role) => currentRoles.indexOf(role) !== -1);
+}
+
+if (currentRoleNames.includes('admin')) {
+  currentRoleNames = xdmp.roles().toArray().map(roleIdToName);
+}
+const datahubRoles  = currentRoleNames.filter((roleName) => fn.startsWith(roleName, 'data-hub'));
+
+const privileges = {
```
rjrudin: Let's not use the term "privileges" here, as that already has a definition within ML. Just call it "response", since this is the response object.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/security/getAuthorities.sjs
```diff
@@ -15,14 +15,49 @@
...
+
+let hasManagerRole = defaultManageRoles.some((role) => currentRoles.includes(role));
+if (!hasManagerRole) {
+  hasManagerRole = manageRolesMustAllMatched.every((role) => currentRoles.indexOf(role) !== -1);
+}
+
+if (currentRoleNames.includes('admin')) {
+  currentRoleNames = xdmp.roles().toArray().map(roleIdToName);
+}
+const datahubRoles  = currentRoleNames.filter((roleName) => fn.startsWith(roleName, 'data-hub'));
```
rjrudin: You can do this and the line near the end in one simpler line of code:

    response.roles = currentRolesNames.filter(roleName => fn.startsWith(roleName, "data-hub");

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/data-services/security/getAuthoritiesTest.sjs
```diff
@@ -0,0 +1,24 @@
+const test = require("/test/test-helper.xqy");
```
rjrudin: There are a lot of other scenarios to test here, each as a different user. Switching users is very difficult in marklogic-unit-test though. For that reason, I think this test should be converted into a JUnit test - GetAuthoritiestTest - where each test method uses a different user. We have test-data-hub-developer and test-data-hub-operator already, and I think you can use "admin" too.

one-ui/src/main/java/com/marklogic/hub/oneui/auth/AuthenticationFilter.java
```diff
@@ -137,15 +139,24 @@ protected AuthenticationToken authenticateUser(String username, String password)
      * - see https://docs.spring.io/spring-security/site/docs/current/reference/htmlsingle/#appendix-faq-role-prefix.
      *
      * @param stagingClient
+     * @param roles
+     * @param canInstallDataHub
      * @return
      */
-    protected List<GrantedAuthority> getAuthoritiesForAuthenticatedUser(DatabaseClient stagingClient) {
+    protected List<GrantedAuthority> getAuthoritiesForAuthenticatedUser(DatabaseClient stagingClient, ArrayNode[] roles, boolean[] canInstallDataHub) {
```
rjrudin: When we need a method to return a tuple of different things, I'd much rather see us use the Pair class or a custom class instead of passing arrays as ways to capture values. 

So this can be simplified by returning a Pair<List\<GrantedAuthority\>, ArrayNode> . That makes it much more obvious what the method is doing.

For "canInstallDataHub" - can't LoginHandler just iterate over the list of GrantedAuthoritys to see if "canInstallDataHub" exists, and if so, write that to the JSON response? It shouldn't need to care about hasManagePrivileges anymore, right? That would get rid of this from this function, which shouldn't care about setting bits specific to the login response. It should just care about getting all the authorities (and for now, all the roles too, though that will soon go away).

hao1st: Now UI needs hasManagePrivileges for convenience. We can remove it once UI do not need it. Since we will remove roles, we can refactor the codes later on.

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/data-services/security/getAuthoritiesTest.sjs
hao1st: I thought about it. It is why I added more test cases in the AuthTest.

one-ui/src/main/java/com/marklogic/hub/oneui/auth/AuthenticationFilter.java
hao1st: I do not see any difference between use a Pair as return and pass roles as a parameter...

rjrudin: It's more clear what the inputs and outputs are. A signature of:

    List<GrantedAuthority> getGrantedAuthorities(DatabaseClient, ArrayNode); 

implies that the DatabaseClient and ArrayNode are inputs needed to process a single List as the output. The reader has to do extra work to realize that one of the inputs is actually an output, not an input. 

But this signature:

    Pair<List, ArrayNode> getGrantedAuthorities(DatabaseClient);

implies that a single input is needed to produce two outputs. 

But it's fine now not to change it - ideally, the roles will go away soon and we'll have a single output - the List of authorities.

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/data-services/security/getAuthoritiesTest.sjs
rjrudin: Consider though that we'll likely add a lot more logic to this endpoint. Will it be easier to write tests against SecurityService, where all we need to do is instantiate a SecurityService, or against the Spring MVC layer, where we need to deal with mock request and response objects and a lot of other logic that we don't really care about? 

It's good to have one test against Spring MVC to verify that all the parts are integrated together nicely. But for all the different scenarios that getAuthorities.sjs will handle, it'll be a lot cheaper in the long run to verify those scenarios directly against the endpoint.

hao1st: @rjrudin Can you give me an example of a java test class about how to invoke sjs ?

rjrudin: You'd just use the DS-generated one - SecurityService.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/security/getAuthorities.sjs
```diff
@@ -15,14 +15,49 @@
...
+const roleIdToName = (roleID) => xdmp.roleName(roleID);
+const currentRoles = xdmp.getCurrentRoles().toArray().map(String);
+let currentRoleNames = currentRoles.map(roleIdToName);
+
+const defaultManageRoles = ['admin', 'data-hub-environment-manager'].map((roleName) => String(xdmp.role(roleName)));
+const manageRolesMustAllMatched = ['manage-admin', 'security'].map((roleName) => String(xdmp.role(roleName)));
+
+let hasManagerRole = defaultManageRoles.some((role) => currentRoles.includes(role));
+if (!hasManagerRole) {
+  hasManagerRole = manageRolesMustAllMatched.every((role) => currentRoles.indexOf(role) !== -1);
```
bsrikan: This should be hasManageAdminRole.
With manageRolesMustAllMatched are we trying to check if a role has both manage-admin & security? If so, hasManageAdminAndSecurity sounds better to me.

one-ui/src/test/java/com/marklogic/hub/oneui/AuthTest.java
```diff
@@ -46,14 +48,67 @@ void loginWithInvalidCredentials() throws Exception {
...
+        mockMvc
+            .perform(post(LOGIN_URL).contentType(MediaType.APPLICATION_JSON).content(payload))
+            .andDo (
+                result -> {
+                    session[0] = (MockHttpSession) result.getRequest().getSession();
+                    String strResponse = result.getResponse().getContentAsString();
+                    ObjectMapper mapper = new ObjectMapper();
+                    JsonNode jsonResonse = mapper.readTree(strResponse);
+                    assertTrue(jsonResonse.get("roles").isArray());
+                    assertTrue(strResponse.contains("\"hasManagePrivileges\":true"));
```
bsrikan: @ryanjdew we are still at the early stages of 5.3 so maybe we can change this now. Is hasManagePrivileges telling us if the user has manage-admin role? If so, then shouldnt we name it that way. I am saying this because data-hub-operator has manage privilege, will this flag then be true.

bsrikan:
Does this PR have anything to do with session management. Is there anything else for us to think about, since we will be using session management the way explorer does for auto-logout on session timeout.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/security/getAuthorities.sjs
```diff
@@ -15,14 +15,49 @@
...
-    "canInstallDataHub",
-    "canReadLoadData",
-    "canWriteLoadData"
-  ]
+const Artifacts = require('/data-hub/5/artifacts/core.sjs');
+const roleIdToName = (roleID) => xdmp.roleName(roleID);
+const currentRoles = xdmp.getCurrentRoles().toArray().map(String);
+let currentRoleNames = currentRoles.map(roleIdToName);
+
+const defaultManageRoles = ['admin', 'data-hub-environment-manager'].map((roleName) => String(xdmp.role(roleName)));
```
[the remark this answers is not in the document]
rjrudin: This is a really good point - this needs a quick test, but if a user has role "apple", and "apple" inherits "manage-admin", then when you get the list of the current user's roles, you won't get "manage-admin" back, you'll only get "admin". 

The story needs to account for this scenario - e.g. a user that inherits manage-admin, but not directly. It may be better to check to see if the user has the "manage-admin" privilege, I just don't know off the top of my head if there's an easy way to get "all" roles and privileges, including inherited ones.

hao1st:
@bsrikan @rjrudin @ryanjdew Currently we just did special handling for "admin" by populating all the roles. But only "canInstallDataHub" appearing in authorities based on current getAuthorities. It seems not correct. I could not find any function to get the inherited roles for a user...
```
{
"authorities": [
"canInstallDataHub"
], 
"roles": [
"data-hub-environment-manager", 
"data-hub-mapping-reader", 
"data-hub-module-writer", 
"data-hub-admin-role", 
"data-hub-module-reader", 
"data-hub-entity-model-reader", 
"data-hub-explorer-architect", 
"data-hub-security-admin", 
"data-hub-flow-reader", 
"data-hub-step-definition-writer", 
"data-hub-operator", 
"data-hub-mapping-writer", 
"data-hub-admin", 
"data-hub-load-data-writer", 
"data-hub-job-reader", 
"data-hub-job-internal", 
"data-hub-monitor", 
"data-hub-entity-model-writer", 
"data-hub-developer", 
"data-hub-portal-security-admin", 
"data-hub-load-data-reader", 
"data-hub-step-definition-reader", 
"data-hub-flow-writer"
]
}
```

rjrudin:
I'll look into that sort of function tomorrow morning - the "Describe" tab in the Admin UI is able to collect all the inherited roles and privileges somehow, though it may be making multiple function calls to do so.

bsrikan:
There is xdmp.roleRoles("role", true) api that returns inherited roles as well

ryanjdew:
@hao1st I use `xdmp.getCurrentRoles()` for determining canRead/canWrite for the different artifacts. According to the documentation it "Returns all the current roles, **both assigned and inherited** by the current user and any received from amps."

one-ui/src/test/java/com/marklogic/hub/oneui/AuthTest.java
ryanjdew: It isn't explicitly depending on whether we have access to the Data Service endpoint it is really either don't get forbidden from the manage REST app server or if the Data Service endpoint is available it was checking the role, but now should check "canInstallDataHub".

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/security/getAuthorities.sjs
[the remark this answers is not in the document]
rjrudin: I take that back - xdmp.getCurrentRoles() (and xdmp.roleRoles()) do return all inherited roles. 

@hao1st You don't need to check for "data-hub-environment-manager" then.  If a user has that role, then xdmp.getCurrentRoles() will return both "manage-admin" and "security".

The only role that requires special handling is "admin", since for a user with only "admin", xdmp.getCurrentRoles() will only return "admin".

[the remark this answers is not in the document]
hao1st: @rjrudin if the user has only "data-hub-portal-security-admin" role,  should we give "canInstallDataHub" authority?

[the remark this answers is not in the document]
hao1st: @ryanjdew If the user has "admin" role, all the canRead{artifactType} and canWrite{artifactType} are false. Only "canInstallDataHub" is the authortity. It does not seem right. I mean that "cores.sjs" also needs to do special handling for "admin" role?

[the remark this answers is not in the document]
rjrudin: We don't care if the user has "data-hub-portal-security-admin" - that role is one of an infinite number of ways that a user may inherit manage-admin and security. All we care about is if the user has both manage-admin and security.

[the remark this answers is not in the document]
rjrudin: Re: the admin role - IMO, the design should be that if the user has the admin role, then the user gains all authorities. That seems like the expected and intuitive approach, along with being consistent with how the admin role works within ML.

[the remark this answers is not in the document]
hao1st: @rjrudin the following response for "admin" is correct? As we talked before, we will remove the roles once UI starts to respect authorities. 
```
{
"authorities": [
"canInstallDataHub"
], 
"roles": [
"data-hub-environment-manager", 
"data-hub-mapping-reader", 
"data-hub-module-writer", 
"data-hub-admin-role", 
"data-hub-module-reader", 
"data-hub-entity-model-reader", 
"data-hub-explorer-architect", 
"data-hub-security-admin", 
"data-hub-flow-reader", 
"data-hub-step-definition-writer", 
"data-hub-operator", 
"data-hub-mapping-writer", 
"data-hub-admin", 
"data-hub-load-data-writer", 
"data-hub-job-reader", 
"data-hub-job-internal", 
"data-hub-monitor", 
"data-hub-entity-model-writer", 
"data-hub-developer", 
"data-hub-portal-security-admin", 
"data-hub-load-data-reader", 
"data-hub-step-definition-reader", 
"data-hub-flow-writer"
]
}
```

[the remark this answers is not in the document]
rjrudin: If that's the complete set of authorities and DHF-related roles, then yes, it's correct. With the assumption that roles goes away very soon.

one-ui/src/main/java/com/marklogic/hub/oneui/auth/AuthenticationFilter.java
```diff
@@ -137,17 +141,38 @@ protected AuthenticationToken authenticateUser(String username, String password)
...
         JsonNode response = SecurityService.on(stagingClient).getAuthorities();
-        if (response != null && response.has("authorities")) {
-            response.get("authorities").iterator().forEachRemaining(node -> {
-                authorities.add(new SimpleGrantedAuthority("ROLE_" + node.asText()));
-            });
+        if (response != null) {
+            if (response.has("authorities")) {
+                response.get("authorities").iterator().forEachRemaining(node -> {
+                    String authority = node.asText();
+                    if ("canInstallDataHub".equals(authority))  {
```
rjrudin: Instead of this special case for this one capability, just add a method to AuthenticationToken like this:

- boolean hasAuthority(String authorityName);

LoginFilter can then just call token.hasAuthority("canInstallDataHub")

Then you don't need the boolean array of a single value.

one-ui/src/main/java/com/marklogic/hub/oneui/auth/LoginHandler.java
```diff
@@ -47,6 +47,9 @@ public void onAuthenticationSuccess(HttpServletRequest request, HttpServletRespo
         if (token.getRoles() != null) {
             jsonResponse.putArray("roles").addAll(token.getRoles());
         }
+        if (token.getAuthorites() != null) {
```
rjrudin: "Authorites" is misspelled, but if it were spelled correctly, you'd get a compilation error because "getAuthorities" already exists. 

We don't need to store the authorities in two separate fields in AuthenticationToken. Instead, just add a "getAuthorityNames" method that returns a List<String>. It will iterate over each GrantedAuthority and get its value and remove the "ROLE_", since the UI doesn't want to deal with that.

one-ui/src/main/java/com/marklogic/hub/oneui/auth/AuthenticationFilter.java
hao1st: @rjrudin I thought about using another method before,.But as we discussed, we do not need to have hasManagerPrivileges after UI change. And I will change the method to remove the canInstallDataHub array.

rjrudin: Okay, that's fair.

one-ui/src/main/java/com/marklogic/hub/oneui/auth/LoginHandler.java
hao1st: oops, i did not notice that. and will modify it.

PR #3644, opened 2020-03-04
Prototype of using springfox-swagger

rjrudin:
After authenticating, goto http://localhost:8080/swagger-ui.html to see the generated Swagger docs. Can also see the JSON at http://localhost:8080/v2/api-docs .


rjrudin:
Closing this, as the inability to provide examples using springfox annotations is a blocker for us.

PR #3646, opened 2020-03-04
DHFPROD-4557: Scaffold out new artifact directories as part of init

ryanjdew:


marklogic-data-hub/src/main/java/com/marklogic/hub/impl/HubProjectImpl.java
```diff
@@ -80,6 +80,8 @@
     private Path stepDefinitionsDir;
     private String userModulesDeployTimestampFile = USER_MODULES_DEPLOY_TIMESTAMPS_PROPERTIES;
 
+    private String[] artifactTypes = new String[]{"mappings", "entities", "loadData", "matching", "merging", "step-definitions"};
```
rjrudin: Should this be "load-data" to be consistent with "step-definitions"? And shouldn't "flows" be in this list?

rjrudin: I'm also wondering how aware HubProject should be of every artifact type. Could some other class make a call to ML to get a list of all the artifact types, and then pass that list of objects to HubProject, which would then initialize a directory for each one, using the name/type of the artifact as the directory name?

I'm thinking right now of all the places that need to be modified in order to define a new artifact type. Ideally, that's just one place - the ArtifactService in ML.

PR #3650, opened 2020-03-05
DHFPROD-4025: Unwrap array retrieving the lastest job for a single flow

ryanjdew:
Adjusting the Jobs API to match the documentation for the latest job for a single flow endpoint. That API was described in QuiskStart Documentation, but wasn't used due to timing of the 5.0 release.

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/services/mlJobs.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/services/mlJobs.sjs
@@ -1,18 +1,18 @@
 /**
-  Copyright 2012-2019 MarkLogic Corporation
+ Copyright 2012-2019 MarkLogic Corporation
 
-  Licensed under the Apache License, Version 2.0 (the "License");
-  you may not use this file except in compliance with the License.
-  You may obtain a copy of the License at
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
 
-     http://www.apache.org/licenses/LICENSE-2.0
+ http://www.apache.org/licenses/LICENSE-2.0
 
-  Unless required by applicable law or agreed to in writing, software
-  distributed under the License is distributed on an "AS IS" BASIS,
-  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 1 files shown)


marklogic-data-hub/src/main/resources/ml-modules/services/mlJobs.sjs
```diff
@@ -36,19 +36,22 @@ function get(context, params) {
...
   }
-  else if (fn.exists(flowNames)) {
-    resp = datahub.jobs.getJobDocsForFlows(flowNames);
-  }
-  else if (fn.exists(flow)) {
-    resp = datahub.jobs.getJobDocsByFlow(flow);
-  }
   else if (fn.exists(latest)) {
-    flowNames = (fn.exists(flowNames)) ? datahub.hubUtils.normalizeToSequence(flowNames) : flowNames;
+    flowNames = (fn.exists(flowNames)) ? datahub.hubUtils.normalizeToSequence(flowNames) : datahub.hubUtils.normalizeToSequence([flow]);
```
rjrudin: I'm thinking a test can be added easily here that would run against JobDocManager. On the other hand, there are currently zero tests that hit JobDocManager directly. 

And - I just searched for uses of "latest", and I found none. JobDocManager never passes this as a parameter. So this looks like dead code?? I'm confused.

ryanjdew: There are a few latest uses in the JobDocManager, like com/marklogic/hub/job/JobDocManager.java:106 `params.put("latest", "true");`. I'll add a test for it.

rjrudin: Are we looking at different things? I'm looking at https://github.com/marklogic/marklogic-data-hub/blob/develop/marklogic-data-hub/src/main/java/com/marklogic/hub/job/JobDocManager.java on the develop branch, and there's no mention of "latest" there.

rjrudin: Ah - I see it now, separate branch - https://github.com/marklogic/marklogic-data-hub/blob/feature/DHFPROD-3922/marklogic-data-hub/src/main/java/com/marklogic/hub/job/JobDocManager.java

PR #3651, opened 2020-03-05
DHFPROD-4562: Export project got exception if a folder does not exist…

hao1st:
… in the exported project folder

https://project.marklogic.com/jira/browse/DHFPROD-4562

### DIFF
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/impl/HubProjectImpl.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/impl/HubProjectImpl.java
@@ -647,6 +647,11 @@ public class HubProjectImpl implements HubProject {
             Stream.of("entities", "flows", "src" + File.separator + "main", "mappings", "step-definitions", "loadData", "gradle",
                 "gradlew", "gradlew.bat", "build.gradle", "gradle.properties").forEach(file ->{
                 File fileToZip = getProjectDir().resolve(file).toFile();
+                if (!fileToZip.exists()) {
+                    // should we add check if some missing files/folders are mandatory and give error info?
+                    logger.warn(String.format("%s does not exist during project export.", fileToZip.toString()));
+                    return;
+                }
                 try {
                     if (fileToZip.isDirectory()) {
                         zout.putNextEntry(new ZipEntry(file + File.separator));


marklogic-data-hub/src/main/java/com/marklogic/hub/impl/HubProjectImpl.java
```diff
@@ -647,6 +647,11 @@ private void writeToStream(OutputStream out) {
             Stream.of("entities", "flows", "src" + File.separator + "main", "mappings", "step-definitions", "loadData", "gradle",
                 "gradlew", "gradlew.bat", "build.gradle", "gradle.properties").forEach(file ->{
                 File fileToZip = getProjectDir().resolve(file).toFile();
+                if (!fileToZip.exists()) {
+                    // should we add check if some missing files/folders are mandatory and give error info?
```
ryanjdew: I think that logging that a folder doesn't exist is sufficient.

bsrikan: @hao1st what if there are gradle-<evn>.properties?

rjrudin: The original requirement was to exclude gradle-*.properties, as there may be sensitive data in those that shouldn't be shared.

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/HubProjectImpl.java
```diff
@@ -647,6 +647,11 @@ private void writeToStream(OutputStream out) {
             Stream.of("entities", "flows", "src" + File.separator + "main", "mappings", "step-definitions", "loadData", "gradle",
                 "gradlew", "gradlew.bat", "build.gradle", "gradle.properties").forEach(file ->{
                 File fileToZip = getProjectDir().resolve(file).toFile();
+                if (!fileToZip.exists()) {
```
rjrudin: @hao1st Since a bug is being fixed, there should be a test that reproduces the bug such that if I comment out this change, the test fails, and then the test succeeds with this change in place. 

The only exception to this is if we have some sort of intermittent bug with no idea of how to reproduce it, which would be indicated by the JIRA ticket lacking reproducible steps.

hao1st: @rjrudin I think that the change is trying to make the code more robust so that we can make sure the exported zip includes all the folder/files from the project folder. If any of them is missing against necessary folder structure, we can identify it for sure.

rjrudin: It looks like the change is trying to prevent the export from breaking when a directory doesn't exist. So an automated test would simulate that condition by e.g. deleting one of these directories or files before calling the export method.

hao1st: @bsrikan upload project test case was not commented out, but it was broken. The PR is a part of fix.

PR #3655, opened 2020-03-06
DHFPROD-4365: User should be able to browse between source documents

xnikhil08:


xnikhil08:
> I'm seeing the following error when I create a new mapping:
> 
> Screen Shot 2020-03-05 at 6 00 49 PM
> 
> Here's a screencast: https://drive.google.com/file/d/1Ctr6MJ9K3UAceTge_k5at9hATzlOt8OF/view

@wooldridge - I can't reproduce this error. Could be the cache or something. Let's try to reinstall the data hub and see if it comes again.

PR #3662, opened 2020-03-07
DHFPROD-4568: Adding node and npm versions to gradle

SameeraPriyathamTadikonda:


brucean52:
> @wooldridge @xnikhil08 @brucean52 12.16.1 is the latest LTS for Node.js. Is there any downside to use a latest version of Node.js. I am on 10.9 and recently I was running into compile issues with node-sass modules when I include 8.10.0 in one-ui/build.gradle

I don't think there is a downside to using the latest version of node. Sometimes I have to delete my node_modules folder when I upgrade my node version

wooldridge:
@bsrikan I'm fine moving to the LTS version of Node.js.

SameeraPriyathamTadikonda:
> @wooldridge @xnikhil08 @brucean52 12.16.1 is the latest LTS for Node.js. Is there any downside to use a latest version of Node.js. I am on 10.9 and recently I was running into compile issues with node-sass modules when I include 8.10.0 in one-ui/build.gradle

Updated the node version to 12.16.1

PR #3664, opened 2020-03-09
DHFPROD-4337: Middle-tier support upload with various zip types & add…

hao1st:
… test cases

Because a project zip file can be archived with a top-level folder in which DHF project foldes/files reside. It seems not trivial to deal with the situation. Previous implementation did not consider the situation and assumed all the DHF project folder/files are zipped as top-level.

https://project.marklogic.com/jira/browse/DHFPROD-4337


one-ui/src/main/java/com/marklogic/hub/oneui/services/DataHubProjectUtils.java
```diff
@@ -140,4 +167,62 @@ public static void extractZipProject(HubProject project, InputStream in, UIDeplo
         }
         listener.onStatusChange(0, String.format("Completed extracting the uploaded zip project into (%s)", currProjectDir.toFile().getAbsolutePath()));
     }
+
+    private static InputStream toByteArrayInputStream(InputStream in) {
+        if (in instanceof ByteArrayInputStream) {
+            return in;
+        }
+        byte[] buff = new byte[8000];
```
rjrudin: Just use FileCopyUtils here instead:

    ByteArraOutputStream out = new ByteArrayOutputStream();
    FileCopyUtils.copy(in, out);
    return new ByteArrayInputStream(out.toByteArray());

one-ui/src/test/java/com/marklogic/hub/oneui/TestHelper.java
```diff
@@ -79,6 +79,12 @@ public void authenticateSession() {
        hubConfig.setCredentials(environmentInfo, dataHubDeveloperUsername, dataHubDeveloperPassword);
     }
 
+    public void authenticateSessionAsAdmin() {
+        createUser("admin", "admin","admin");
```
rjrudin: It looks like TestHelper already has properties for test.adminUsername and test.adminPassword, so there's no need to do this - the line doesn't really make sense either, as we can safely assume there's an admin user with the admin role.

Also, to mirror similar methods in HubTestBase, call this "runAsAdminUser".

one-ui/src/main/java/com/marklogic/hub/oneui/services/DataHubProjectUtils.java
```diff
@@ -140,4 +167,62 @@ public static void extractZipProject(HubProject project, InputStream in, UIDeplo
...
+            }
+        } catch (IOException e) {
+            throw new RuntimeException(String.format("Failed to convert to ByteArrayInputStream due to (%s)", e.getMessage()));
+        }
+        byte[] data = bao.toByteArray();
+        ByteArrayInputStream bin = new ByteArrayInputStream(data);
+        return bin;
+    }
+
+    private static String getArchiveFolderOfZipFile(Path currProjectDir, InputStream in) {
```
rjrudin: I highly recommend unit tests to verify this functionality. There's a lot of code being added here, and the new test in EnvironmentControllerTest verifies it at an integration level. But if something goes wrong, it'll be difficult to debug because we don't have any unit tests verifying any of the new logic. 

Looking at what this method outputs, it's not clear to me what this method is doing. The name implies an "archive folder" is being returned, so I'm first wondering why a Path or File isn't returned. I also see "return null" at the bottom - what does that mean if that happens? Unit tests would both verify that this functionality works correctly and document the purposes of the method as well.

one-ui/src/test/java/com/marklogic/hub/oneui/TestHelper.java
rjrudin: @hao1st  Can't this line be removed? There shouldn't be any reason to try to create an admin user.

one-ui/src/test/java/com/marklogic/hub/oneui/services/DataHubProjectUtilsTest.java
```diff
@@ -0,0 +1,63 @@
...
+import static org.junit.jupiter.api.Assertions.fail;
+
+public class DataHubProjectUtilsTest {
+
+    @Test
+    public void testExtractZipFileWithoutArchiveFolder() throws Exception {
+        File file = new File(EnvironmentControllerTest.class.getClassLoader().getResource("dhfWithoutArchiveFolder.zip").getFile());
+        FileInputStream input = new FileInputStream(file);
+
+        Method method = DataHubProjectUtils.class.getDeclaredMethod("toByteArrayInputStream", InputStream.class);
```
rjrudin: Just make these methods protected - that's common for unit-testing purposes.

hao1st: @rjrudin it is arguable though i do not feel it is good to do that. But it is harmless for me.

rjrudin: Protected methods can still be changed between minor releases. You can also do package-level protection, which works as long as your test class is in the same package as your class.

one-ui/src/test/java/com/marklogic/hub/oneui/TestHelper.java
hao1st: @rjrudin I will fall back it to Environment Manager user instead of user though i did not get why we could not use "admin".

PR #3668, opened 2020-03-09
DHFPROD-4541: React-testing-library tests for search-bar component.

timur-isangulov:


timur-isangulov:
> I would like to remove any duplicate enzyme tests if its being covered by RTL

I removed the duplicate test. Thanks.

PR #3676, opened 2020-03-10
DHFPROD-3854: sourceQuery can now be a script

rjrudin:
Step option "sourceQueryIsScript:true" can be added so that the collector knows to evaluate the sourceQuery as a script instead of tossing it into a cts.uris call. 

Did some renaming of "uris" as a variable to "items". Unfortunately can't change e.g. "uris" in a Batch document, as that'd be a breaking change. 

FlowRunnerTest now depends on the indexes in the test final-database.json file, so I moved the setup of those indexes from RunMarkLogicUnitTestsTest into Installer.

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/endpoints/collector.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/endpoints/collector.sjs
@@ -69,19 +69,18 @@ if (!query) {
   fn.error(null, "RESTAPI-SRVEXERR", Sequence.from([404, "Not Found", "The collector query was empty"]));
 }
 
-query = new CollectorLib().prepareSourceQuery(combinedOptions, stepDefinition);
-
+const javascript = new CollectorLib().prepareSourceQuery(combinedOptions, stepDefinition);
 try {
-  let urisEval;
-  if (/^\s*cts\.(uris|values)\(.*\)\s*$/.test(query)) {
-    urisEval = query;
-  } else {
-    urisEval = "cts.uris(null, null, " + query + ")";
-  }
-  xdmp.eval(urisEval, {options: options}, {database: xdmp.database(database)});
+  /**
+   * DHF 5 has always used this eval, and it certainly is open for code injection. This is partially minimized
+   * by this collector code running in query mode, thus preventing updates. Additionally, while the DHF users
+   * almost all have the required privileges to evaluate code, other functions like xdmp.shutdown require their
+   * own privileges. Furthermore, a malicious user doesn't need to bother with exploiting this but rather can just
+   * hit /v1/eval directly. The broader issue is that DHF users should not have as many privileges as they do, and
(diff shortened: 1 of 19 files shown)


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/endpoints/collectorLib.sjs
```diff
@@ -19,13 +19,23 @@ const DataHub = require("/data-hub/5/datahub.sjs");
...
+   */
   prepareSourceQuery(combinedOptions, stepDefinition) {
     let sourceQuery = combinedOptions.sourceQuery;
 
     let isMergingStep = stepDefinition.name === 'default-merging' && stepDefinition.type === 'merging';
 
     if (isMergingStep) {
-      sourceQuery = fn.normalizeSpace(`cts.values(cts.pathReference('/matchSummary/URIsToProcess', ['type=string','collation=http://marklogic.com/collation/']), null, null, ${sourceQuery})`);
+      return fn.normalizeSpace(`cts.values(cts.pathReference('/matchSummary/URIsToProcess', ['type=string','collation=http://marklogic.com/collation/']), null, null, ${sourceQuery})`);
+    } else if (combinedOptions.sourceQueryIsScript) {
```
hao1st: if return from "if" code block, you do not need to use "else if", only "if" is enough.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/endpoints/collector.sjs
```diff
@@ -69,19 +69,18 @@ if (!query) {
   fn.error(null, "RESTAPI-SRVEXERR", Sequence.from([404, "Not Found", "The collector query was empty"]));
 }
 
-query = new CollectorLib().prepareSourceQuery(combinedOptions, stepDefinition);
-
+const javascript = new CollectorLib().prepareSourceQuery(combinedOptions, stepDefinition);
```
hao1st: can we have a better variable name for "javascript" here?

rjrudin: It's the same name as the argument for xdmp.eval - http://docs.marklogic.com/xdmp.eval .

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/endpoints/collectorLib.sjs
rjrudin: Good idea, I'll make the change.

rjrudin: Updated - definitely reads nicer now.

rjrudin:
Fixed 2 tests that were failing due to the test indexes not being applied by RunMarkLogicUnitTestsTest. That class now applies the test indexes as well.

PR #3677, opened 2020-03-10
DHFPROD-4173: Add map settings

ryanjdew:
This should now be ready to go.


wooldridge:
@ryanjdew, looks good!

PR #3684, opened 2020-03-11
DHFPROD-3922: Fix CSV run and logic for recording jobs

ryanjdew:


wooldridge:
Thanks, @ryanjdew. I tested running various Load Data artifacts in flows from the UI and the following source types worked:

- JSON
- XML
- CSV
- Binary
 
Loading a regular text file using the "Text" source format throws an error:
```
Local message: failed to apply resource at documents: Internal Server Error. Server Message: The input text document is not a valid json . (err:FOER0000): fn.error(null, flowResponse.errors[0].message, flowResponse.errors[0].stack); --  Error: The input text document is not a valid json .\n    at main (/data-hub/5/builtins/steps/ingestion/default/main.sjs:35:13)\n    at Flow.runMain (/data-hub/5/impl/flow.sjs:522:12)\n    at Flow.runStep (/data-hub/5/impl/flow.sjs:380:58)\n    at Flow.runFlow (/data-hub/5/impl/flow.sjs:248:12)\n    at transform (/marklogic.rest.transform/mlRunIngest/assets/transform.sjs:44:35)\n    at callExtension (/data-hub/5/rest-api/lib/extensions-util.sjs:94:30)\n    at applyList (/data-hub/5/rest-api/lib/extensions-util.sjs:129:7) . See the MarkLogic server error log for further detail.
```
But as we discussed that is expected given how Data Hub is currently implemented. The Text source format requires a text file with JSON or XML content.

ryanjdew:
@rjrudin The primary bug was that the separator value wasn't being populated from the loadData artifact. There were some secondary bugs with the Jobs code that caused the failure to not be recorded in the database if the error occurred in the Java layer. I've added some tests in my most recent commit to this PR.

PR #3695, opened 2020-03-13
DHFPROD-4519: Move job-info under models endpoint

akshaysonvane:
`/api/jobs` is now `/api/models/job-info`
`/api/jobs/models/{modelName}` is now `/api/models/{modelName}/job-info`

It makes more sense for job info to be part of models rather than `/jobs` as the job info is for the model.
This will also preemptively resolve conflicts when we merge the `feature/DHFPROD-4417` branch as oneui now has the jobs endpoint that deals with the job documents.


one-ui/src/main/java/com/marklogic/hub/oneui/controllers/ModelController.java
```diff
@@ -60,4 +62,18 @@ ModelManager modelManager() {
         JsonNode json = modelManager.getModel(modelName);
         return new ResponseEntity<>(json, HttpStatus.OK);
     }
+
+    @RequestMapping(value = "/job-info", method = RequestMethod.GET)
+    @ResponseBody
+    public ResponseEntity<?> getLatestJobInfoForAllModels() {
```
rjrudin: I agree with including this here, as this is job stuff that's specific to models, as opposed to job stuff that is global to a project.

one-ui/src/main/java/com/marklogic/hub/oneui/controllers/ModelController.java
```diff
@@ -60,4 +62,18 @@ ModelManager modelManager() {
...
+        List<JsonNode> modelList = modelManager.getLatestJobInfoForAllModels();
+        return new ResponseEntity<>(modelList, HttpStatus.OK);
```
rjrudin: One tweak I'd make because I like single-line methods in a controller - no need to change now, but I'd just do this:

    return new ResponseEntity<>(modelManager.getLatestJobInfoForAllModels(), HttpStatus.OK);

And since I'm really lazy, I'd have an AbstractController class with generic reusable methods, such as one that would let me do this here instead:

    return ok(modelManager.getLatestJobInfoForAllModels());

SameeraPriyathamTadikonda:
@akshaysonvane JIRA_ID is missing :)

one-ui/src/main/java/com/marklogic/hub/oneui/managers/ModelManager.java
```diff
@@ -154,4 +157,43 @@ public JsonNode getModel(String modelName) {
...
+            .collect(Collectors.toList());
+    }
+
+    JsonNode getJobInfoFromDB(DatabaseClient dbClient, String modelName) {
+        JsonNode node = null;
+        try {
+            node = JobInfo.on(dbClient).getLatestJobData(modelName);
+        }
+        catch (MarkLogicServerException e) {
+            /*
```
rjrudin: Why would the user not be authorized? I'm wondering more generally - why would we call this method if the user isn't authorized? This seems like something we'd want to handle via authorities - I think Spring Security has a method for checking this programmatically so an exception isn't thrown right away - e.g. https://stackoverflow.com/questions/3021200/how-to-check-hasrole-in-java-code-with-spring-security . 

I think this needs to be changed because if this exception occurs for some other reason, we don't necessarily want to just log it and move on.

one-ui/src/main/java/com/marklogic/hub/oneui/controllers/ModelController.java
akshaysonvane: Good idea.

one-ui/src/main/java/com/marklogic/hub/oneui/managers/ModelManager.java
akshaysonvane: Explore has two users, the architect user that can only access the entity models but not the entity instances and the user that can access both the entity models and instances.
The explorer architect user won't have access to this job info data service endpoint and in that case, the UI does not show any data in the last harmonized timestamp column. 

Now, that we have authorities we can rely on them to hide that column from the UI. 

I'll just create a task to check for the right authority for now as this would require additional work on the UI to hide the extra column for an explorer architect user. Making the change just on the backend will break the app as the frontend would not handle the 403 access denied when this call is made.

We can take up this task up once @brucean52 completes the task of merging the frontend explorer code with OneUI.

one-ui/src/main/java/com/marklogic/hub/oneui/controllers/ModelController.java
akshaysonvane: Added https://github.com/marklogic/marklogic-data-hub/pull/3695/commits/3029f5529211221da4ad83bd1e2a2403857c99fd

PR #3703, opened 2020-03-16
DHFPROD-4604:Fix failing one-ui tests

srinathgit:
Fix failing one-ui tests

### DIFF
--- a/one-ui/src/test/java/com/marklogic/hub/curation/controllers/EnvironmentControllerTest.java
+++ b/one-ui/src/test/java/com/marklogic/hub/curation/controllers/EnvironmentControllerTest.java
@@ -1,65 +1,32 @@
 package com.marklogic.hub.curation.controllers;
 
-import ch.qos.logback.classic.Logger;
-import ch.qos.logback.classic.LoggerContext;
-import ch.qos.logback.classic.spi.ILoggingEvent;
-import ch.qos.logback.core.LogbackException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.node.ArrayNode;
 import com.fasterxml.jackson.databind.node.ObjectNode;
-import com.marklogic.appdeployer.command.security.DeployPrivilegesCommand;
-import com.marklogic.appdeployer.command.security.DeployRolesCommand;
-import com.marklogic.client.DatabaseClient;
-import com.marklogic.client.ResourceNotFoundException;
-import com.marklogic.client.document.JSONDocumentManager;
-import com.marklogic.client.io.Format;
-import com.marklogic.client.io.StringHandle;
-import com.marklogic.hub.ApplicationConfig;
 import com.marklogic.hub.ArtifactManager;
 import com.marklogic.hub.artifact.ArtifactTypeInfo;
-import com.marklogic.hub.deploy.commands.DeployDatabaseFieldCommand;
-import com.marklogic.hub.deploy.commands.DeployHubOtherServersCommand;
-import com.marklogic.hub.deploy.commands.DeployHubTriggersCommand;
-import com.marklogic.hub.impl.HubConfigImpl;
-import com.marklogic.hub.oneui.Application;
 import com.marklogic.hub.oneui.TestHelper;
 import com.marklogic.hub.oneui.auth.AuthenticationFilter;
 import com.marklogic.hub.oneui.controllers.EnvironmentController;
 import com.marklogic.hub.oneui.exceptions.ProjectDirectoryException;
-import com.marklogic.hub.oneui.listener.UIDeployListener;
-import com.marklogic.hub.oneui.models.EnvironmentInfo;
 import com.marklogic.hub.oneui.models.HubConfigSession;
-import com.marklogic.hub.oneui.services.DataHubProjectUtils;
-import com.marklogic.hub.oneui.services.EnvironmentConfig;
 import com.marklogic.hub.oneui.services.EnvironmentService;
-import com.marklogic.hub.oneui.utils.TestLoggingAppender;
-import java.io.ByteArrayInputStream;
-import java.io.File;
-import java.io.FileInputStream;
-import java.io.IOException;
-import java.nio.file.Files;
-import java.nio.file.Path;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Set;
-import java.util.stream.Stream;
-import java.util.zip.ZipEntry;
-import java.util.zip.ZipInputStream;
-import org.apache.commons.lang3.StringUtils;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
-import org.junit.jupiter.api.extension.ExtendWith;
-import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.boot.test.context.SpringBootTest;
 import org.springframework.mock.web.MockHttpServletRequest;
 import org.springframework.mock.web.MockHttpServletResponse;
 import org.springframework.mock.web.MockMultipartFile;
-import org.springframework.test.context.junit.jupiter.SpringExtension;
-import org.springframework.web.multipart.MultipartFile;
+
+import java.io.ByteArrayInputStream;
+import java.io.File;
+import java.io.IOException;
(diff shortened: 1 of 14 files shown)


one-ui/src/test/java/com/marklogic/hub/curation/controllers/EnvironmentControllerTest.java
```diff
@@ -25,24 +25,9 @@
...
-import java.util.List;
-import java.util.Set;
-import java.util.stream.Stream;
-import java.util.zip.ZipEntry;
-import java.util.zip.ZipInputStream;
 import org.apache.commons.lang3.StringUtils;
-import org.junit.jupiter.api.AfterEach;
-import org.junit.jupiter.api.BeforeEach;
-import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.*;
```
hao1st: Could you change eclipse import setting and make sure that class count to use import with "*" is 30 or more.

Screen Shot 2020-03-15 at 9 41 43 PM

one-ui/src/test/java/com/marklogic/hub/curation/controllers/EnvironmentControllerTest.java
```diff
@@ -95,6 +88,7 @@ void after() {
     }
 
     @Test
+    @Order(1)
```
hao1st: it is nice to know we can use @order annatation!

one-ui/src/test/java/com/marklogic/hub/oneui/TestHelper.java
```diff
@@ -73,17 +74,26 @@ public TestHelper() throws IOException {
...
         hubConfig.setCredentials(environmentInfo, dataHubEnvironmentManagerUsername, dataHubEnvironmentManagerPassword);
     }
 
+    public void authenticateSessionAsAdmin() {
+        EnvironmentInfo environmentInfo = new EnvironmentInfo(mlHost, "DIGEST", 8000,"DIGEST", 8002,"DIGEST", 8010, "DIGEST", 8011);
+        hubConfig.setCredentials(environmentInfo, adminUserName, adminPassword);
+    }
+
+    @PostConstruct
+    public void createTestUsers() {
```
hao1st: it is a good idea to add it after running test cases!

srinathgit:
> installProjectWithNewDirectory (5) should be put in the front of testUploadProjectWithoutArchiveFolder (3) and testUploadProjectWithArchiveFolder (4).

I don't see any issue with that. The tests are running fine as is. The installProjectWithNewDirectory() installs vanilla datahub which should be good for the subsequent tests.

one-ui/src/test/java/com/marklogic/hub/oneui/TestHelper.java
srinathgit: @hao1st , I have called createTestUsers() wherever it's needed.

one-ui/src/test/java/com/marklogic/hub/oneui/TestHelper.java
```diff
@@ -73,17 +74,26 @@ public TestHelper() throws IOException {
...
+        createUser(dataHubDeveloperUsername,dataHubDeveloperPassword,"data-hub-developer");
```
rjrudin: The Installer program is already creating a data-hub-developer user with the name of "test-data-hub-developer". Why is this also needed?

one-ui/src/test/java/com/marklogic/hub/oneui/TestHelper.java
```diff
@@ -73,17 +74,26 @@ public TestHelper() throws IOException {
...
+        createUser(dataHubEnvironmentManagerUsername,dataHubEnvironmentManagerPassword,"data-hub-environment-manager");
```
rjrudin: I think that if we need another user, it should be created in the Installer program as well.

one-ui/src/test/java/com/marklogic/hub/curation/controllers/EnvironmentControllerTest.java
```diff
@@ -163,36 +208,8 @@ public void doAppend(ILoggingEvent event) throws LogbackException {
...
-    }
-
-    @Test
-    public void testUploadProjectWithArchiveFolder() throws Exception {
-        testUploadProject("dhfWithArchiveFolder.zip");
-    }
-
-    public void testUploadProject(String zipFileName) throws Exception {
-        testHelper.authenticateSessionAsEnvironmentManager();
+    private void testUploadProject(String zipFileName) throws Exception {
```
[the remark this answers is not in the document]
srinathgit: I have added comments and state is not changed after tests are run

one-ui/src/test/resources/application-test.properties
```diff
@@ -1,3 +1,9 @@
 test.mlHost=localhost
-test.dataHubDeveloperUsername=data-hub-developer-user
-test.dataHubDeveloperPassword=data-hub-developer-user
\ No newline at end of file
+test.dataHubDeveloperUsername=test-data-hub-developer
```
rjrudin: Are these still needed here? I doubt we need properties for them - I can't think of a reason why we'd change them in between test environments. 

If Installer is the class that is creating test users, I'd just define their username/passwords as static values on the Installer class.

one-ui/src/test/java/com/marklogic/hub/oneui/TestHelper.java
```diff
@@ -120,24 +127,48 @@ public void addStagingDoc(String uri, DocumentMetadataHandle meta, String resour
...
+            default:
+                loginInfo.username = "fake";
+                break;
+        }
+        return new ObjectMapper().writeValueAsString(loginInfo);
+    }
+
+    /*  EnvironmentControllerTest.testUpload* uninstalls/installs datahub, hence the roles have to be reassigned to users
+    after datahub is reinstalled. */
+    public void assignRoleToUsers() {
```
rjrudin: You'll also need to apply the database properties that Installer applies. 

I think the better approach is that after DHF is re-installed (and I assume it's being installed in the same fashion that "gradle bootstrap" installs it?), call a new static method on Installer that creates the additional users and applies the database properties. Call that method something like "afterInstall", and it'll have params of a DataHub and a HubConfig. That ensures that Installer does all of its additional post-install work.

srinathgit: I will be calling Installer.afterInstall() from tests in one-ui proj.  If I import Installer to the tests in one-ui,
1. HubtestBase.afterPropertiesSet() gets called which attempts to populate objects in HubTestBase. In order to do that, I will have to copy gradle.properties from marklogic-data-hub project to one-ui project (similar to what we have for web project). Is it ok to do that ?
2. I should be able to access test resources of marklogic-data-hub project  in order to update the database properties. Do you know of a way test resource from marklogic-data-hub project be used in one-ui rather than making an explicit copy ? I couldn't find the

one-ui/src/main/java/com/marklogic/hub/oneui/Application.java
```diff
@@ -29,9 +30,9 @@
 import java.util.Set;
 import java.util.stream.Stream;
 
-@ComponentScan(basePackages = "com.marklogic.hub", excludeFilters = {@ComponentScan.Filter(
-		type = FilterType.REGEX,
-		pattern = "com\\.marklogic\\.hub\\.impl\\.HubConfigImpl")
+@ComponentScan(basePackages = "com.marklogic.hub",
+    excludeFilters = {@ComponentScan.Filter(type = FilterType.REGEX, pattern = "com\\.marklogic\\.hub\\.impl\\.HubConfigImpl"),
+    @ComponentScan.Filter(type = FilterType.ASSIGNABLE_TYPE, classes = HubTestBase.class)
```
srinathgit: @rjrudin ,
The only way I can prevent spring container from instantiating HubTestBase and it's derived classes is to exclude it here. I don't feel right as one-ui code has dependency on test code in marklogic-data-hub. Is that ok or is there a better way to prevent its instantiation

one-ui/src/test/resources/application-test.properties
srinathgit: Since there is not going to be dependency on marklogic-data-hub:test, we will need this.

one-ui/src/test/java/com/marklogic/hub/oneui/TestHelper.java
srinathgit: The installation tests have been moved to a different test suite

PR #3704, opened 2020-03-16
DHFPROD-4603:Set absolute schema path in 'AppConfig'

srinathgit:
Set absolute schema path in 'AppConfig'. There is a test already for this running in regression, its failure resulted in this issue
https://jenkins.marklogic.com/blue/organizations/jenkins/Datahub_CI/detail/develop/759/tests

### DIFF
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/impl/HubConfigImpl.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/impl/HubConfigImpl.java
@@ -1947,6 +1947,8 @@ public class HubConfigImpl implements HubConfig
 
         initializeModulePaths(config);
 
+        config.setSchemaPaths(List.of(getUserSchemasDir().toString()));
+
         addDhfPropertiesToCustomTokens(config);
 
         String version = getJarVersion();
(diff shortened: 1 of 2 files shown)


marklogic-data-hub/src/main/java/com/marklogic/hub/impl/HubConfigImpl.java
```diff
@@ -1947,6 +1947,8 @@ private void updateAppConfig(AppConfig config) {
 
         initializeModulePaths(config);
 
+        config.setSchemaPaths(List.of(getUserSchemasDir().toString()));
```
rjrudin: I couldn't reproduce the bug, I'm adding a comment to the JIRA ticket. If it is a bug, we'll need a test to confirm that it's fixed.

srinathgit: As mentioned in the description, the issue was flagged by Sameera after the failure of test com.marklogic.bootstrap.DataHubInstallTest.testProjectScaffolding() in Jenkins. I was able to reproduce this failure in my local machine. The added fix resolved the failure. I will look into the comment in JIRA ticket

rjrudin: Hmm, it's odd that this is failing in that test but not when I test the code against a real project. 

Can you add a test or assertion to HubConfigImplTest to verify this and use the assertion message for why it's being done? It's good to have a targeted unit test for a change like this, as no one will ever know that it's being verified by a broad-ranging test like DataHubInstallTest.

srinathgit: Add an assertion to HubConfigImplTest

PR #3705, opened 2020-03-16
DHFPROD-4559: added check on view page if final database is empty

brucean52:


one-ui/ui/src/pages/View.tsx
```diff
@@ -60,7 +60,12 @@ const View: React.FC = () => {
         setIsLoading(false);
       }
     } catch (error) {
-      handleError(error);
+      if (error['response']['data']['message'] === "REST-INVALIDPARAM: (err:FOER0000) Invalid parameter: No configured options: exp-final-entity-options") {
```
rjrudin: This seems very brittle - the UI is coupling itself to a specific error format being returned by MarkLogic. This could change with any release of MarkLogic. 

At a minimum, I would only check for "No configured options: exp-final-entity-options". But even that is subject to change with any release of ML. 

Can the middle tier catch this instead and throw a better error?

one-ui/ui/src/pages/View.tsx
```diff
@@ -60,7 +60,12 @@ const View: React.FC = () => {
...
+        setEntites([]);
```
rjrudin: Can this be renamed to "setEntities" so it's spelled correctly?

one-ui/ui/src/pages/View.tsx
brucean52: I think this would be a better approach

akshaysonvane: Yeah, I think we should handle it in the middle-tier.

brucean52:
Closing this for now until we have a better solution

PR #3706, opened 2020-03-16
DHFPROD-4519: Inherit from "data-hub-module-reader" role 

akshaysonvane:
Removed try/catch now that data-hub-explorer-architect can access the job-info data service.

### DIFF
--- a/one-ui/src/main/java/com/marklogic/hub/oneui/managers/ModelManager.java
+++ b/one-ui/src/main/java/com/marklogic/hub/oneui/managers/ModelManager.java
@@ -182,18 +182,6 @@ public class ModelManager {
     }
 
     JsonNode getJobInfoFromDB(DatabaseClient dbClient, String modelName) {
-        JsonNode node = null;
-        try {
-            node = JobInfo.on(dbClient).getLatestJobData(modelName);
-        }
-        catch (MarkLogicServerException e) {
-            /*
-             * As MarkLogic server throws a FailedRequestException for both data-service module not found
-             * and if the user is unauthorized to access the module; we need to fail silently and log the
-             * issue.
-             */
-            logger.error(e.getMessage());
-        }
-        return node;
+        return JobInfo.on(dbClient).getLatestJobData(modelName);
     }
 }
--- a/one-ui/src/main/java/com/marklogic/hub/oneui/managers/ModelManager.java
+++ b/one-ui/src/main/java/com/marklogic/hub/oneui/managers/ModelManager.java
@@ -165,7 +165,7 @@ public class ModelManager {
      * @return - a JsonNode containing job info
      */
     public JsonNode getLatestJobInfo(String modelName) {
-        return getJobInfoFromDB(finalDataServiceClient, modelName);
+        return getLatestJobData(finalDataServiceClient, modelName);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


one-ui/src/main/java/com/marklogic/hub/oneui/managers/ModelManager.java
```diff
@@ -182,18 +182,6 @@ public JsonNode getLatestJobInfo(String modelName) {
...
-            /*
-             * As MarkLogic server throws a FailedRequestException for both data-service module not found
-             * and if the user is unauthorized to access the module; we need to fail silently and log the
-             * issue.
-             */
-            logger.error(e.getMessage());
-        }
-        return node;
+        return JobInfo.on(dbClient).getLatestJobData(modelName);
     }
```
bsrikan: Do we have test for this in place?

akshaysonvane: Yes, the data service is tested via ml-unit-test

akshaysonvane: This PR just removes the try/catch block surrounding the code.

one-ui/src/main/java/com/marklogic/hub/oneui/managers/ModelManager.java
```diff
@@ -182,18 +182,6 @@ public JsonNode getLatestJobInfo(String modelName) {
...
-        catch (MarkLogicServerException e) {
-            /*
-             * As MarkLogic server throws a FailedRequestException for both data-service module not found
-             * and if the user is unauthorized to access the module; we need to fail silently and log the
-             * issue.
-             */
-            logger.error(e.getMessage());
-        }
-        return node;
+        return JobInfo.on(dbClient).getLatestJobData(modelName);
```
rjrudin: I think the client should just call JobInfo then - there's no point in having this additional layer to go through. If we did keep it, I think we'd want the method names to be the same - i.e. the Manager method should be getLatestJobData.

akshaysonvane: The separate method is used for mocking the database call in the ModelManagerTest class
```
doReturn(expectedNode).when(modelManager).getJobInfoFromDB(dataServiceClient, modelName);
```

I'll change the method name to `getLatestJobData`

rjrudin: Can you mock ArtifactsService instead? That's supposed to be one of the benefits of data services - that there's an interface class that can be mocked for testing purposes.

rjrudin: But - if it's easier to just mock a single thing - i.e. ModelManager - that's fine too.

akshaysonvane: > Can you mock ArtifactsService instead? That's supposed to be one of the benefits of data services - that there's an interface class that can be mocked for testing purposes.

By ArtifactsService I am guessing you mean the JobInfo interface that was generated by the data service. 
JobInfo has a static `on` method which we call as 
```
JobInfo.on(finalDataServiceClient).getLatestJobData(s)
```
I believe mockito does not support mocking static stuff and would require us to use a testing framework like powermockito. 

Hence, in this case, it's simpler to just mock the `getLatestJobData`  method of ModelManager.

PR #3707, opened 2020-03-16
DHFPROD-4598: Update tooltips text

ryanjdew:


bsrikan:
Can you pls confirm with Julia if the tooltips should be on far right just like edit load modal. All the settings modal need to be fixed to have the tooltip moved over to the right. If possible, can we fix the one for mappings in this PR?

PR #3711, opened 2020-03-17
DHFPROD-4608: Settings for loadData/mapping/matching/merging should b…

hao1st:
…e persisted in filesystem

https://project.marklogic.com/jira/browse/DHFPROD-4608

### DIFF
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/impl/ArtifactManagerImpl.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/impl/ArtifactManagerImpl.java
@@ -91,7 +91,7 @@ public class ArtifactManagerImpl implements ArtifactManager {
         if (!isSetting) {
             fileLocation = buildArtifactProjectLocation(artifactType, getNameFromArtifact(artifactType, artifact), getVersionFromArtifact(artifactType, artifact), false);
         } else {
-            if (artifact.get("artifactName") == null ||  artifact.get("artifactName").asText().length() == 0) {
+            if (!artifact.has("artifactName") || artifact.get("artifactName").asText().length() == 0) {
                 logger.error("Failed to write to project folder due to invalid settings! artifactType: (%s), artifact settings: (%s).", artifactType, artifact.toString());
                 return;
             }
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/impl/ArtifactManagerImpl.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/impl/ArtifactManagerImpl.java
@@ -32,15 +32,18 @@ import java.nio.file.Paths;
 import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
+import org.apache.commons.io.FilenameUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class ArtifactManagerImpl implements ArtifactManager {
-    HubConfig hubConfig;
+    protected static final Logger logger = LoggerFactory.getLogger(ArtifactManagerImpl.class);
 
+    HubConfig hubConfig;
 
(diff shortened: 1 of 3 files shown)


marklogic-data-hub/src/main/java/com/marklogic/hub/impl/ArtifactManagerImpl.java
```diff
@@ -70,16 +76,28 @@ public ObjectNode getArtifactSettings(String artifactType, String artifactName)
...
     }
 
-    protected void writeArtifactInProject(String artifactType, ObjectNode artifact) {
-        Path fileLocation = buildArtifactProjectLocation(artifactType, getNameFromArtifact(artifactType, artifact), getVersionFromArtifact(artifactType, artifact));
+    protected void writeArtifactInProject(String artifactType, ObjectNode artifact, boolean isSetting) {
+        Path fileLocation;
+        if (!isSetting) {
+            fileLocation = buildArtifactProjectLocation(artifactType, getNameFromArtifact(artifactType, artifact), getVersionFromArtifact(artifactType, artifact), false);
+        } else {
+            if (artifact.get("artifactName") == null ||  artifact.get("artifactName").asText().length() == 0) {
```
rjrudin: Use !artifact.has("artifactName") . 

Also - would this really ever happen? And if it did, wouldn't we want to throw an error? Just logging it and moving on doesn't seem good, as something is really wrong if an artifact doesn't have a name. 

I think this code can assume that artifactName exists. If it doesn't, either the proceeding code will throw an error, or we can eagerly throw an exception. But it seems like the call to save the artifact in ML would have already failed too, right?

hao1st: yes, it is why i just log it and stop here without throwing an exception. I can change !artifact.has("artifactName")

rjrudin: I'm thinking we do want an exception thrown though. It's a big deal if an artifact can't be written to the filesystem, right? A remote user will never see that log message.

hao1st: @rjrudin it is a artifact setting info and should always have an artifactName associated. We double-check it (may be bad data imported) because we do not want to store settings without any artifact. I can throw an exception and I am not sure what exception we should throw.

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/ArtifactManagerImpl.java
```diff
@@ -70,16 +76,28 @@ public ObjectNode getArtifactSettings(String artifactType, String artifactName)
...
 
-    protected void writeArtifactInProject(String artifactType, ObjectNode artifact) {
-        Path fileLocation = buildArtifactProjectLocation(artifactType, getNameFromArtifact(artifactType, artifact), getVersionFromArtifact(artifactType, artifact));
+    protected void writeArtifactInProject(String artifactType, ObjectNode artifact, boolean isSetting) {
+        Path fileLocation;
+        if (!isSetting) {
+            fileLocation = buildArtifactProjectLocation(artifactType, getNameFromArtifact(artifactType, artifact), getVersionFromArtifact(artifactType, artifact), false);
+        } else {
+            if (!artifact.has("artifactName") || artifact.get("artifactName").asText().length() == 0) {
+                logger.error("Failed to write to project folder due to invalid settings! artifactType: (%s), artifact settings: (%s).", artifactType, artifact.toString());
```
rjrudin: Since we assume that artifactName is not null, just let an exception propagate when "asText()" is called. If we want something nicer, we can check to see if "artifactName" doesn't exist and then do something like this:

    throw new RuntimeException("Unable to write artifact to project folder; no artifactName found")

PR #3714, opened 2020-03-17
DHFPROD-4162: Initial design for managing entity types

rjrudin:
This is intended for a design review for 4162 before we start on stories.


specs/models/EntityModelDescriptor.v1.json
```diff
@@ -0,0 +1,109 @@
...
+              "type": "array",
+              "items": {
+                "type": "string"
+              }
+            },
+            "properties": {
+              "type": "object",
+              "additionalProperties": false,
+              "patternProperties": {
+                "^[a-z]+$": {
```
hao1st: PropertyName can only be alphabetic? We do not allow its name like Field1, Field2..

specs/models/EntityTypes.v1.json
```diff
@@ -0,0 +1,28 @@
+{
+  "title": "EntityTypes",
+  "type": "array",
+  "x-examples": {},
+  "items": {
+    "type": "object",
+    "properties": {
+      "entityTypeTitle": {
+        "type": "string"
```
hao1st: should be "datatype"?

specs/models/EntityModelDescriptor.v1.json
rjrudin: I'll check with Erik. This is an attempt to define a schema for ES - we won't be validating anything with it, so it doesn't yet need to be 100% correct. Ideally ML would provide it.

specs/models/EntityTypes.v1.json
rjrudin: "type" is what JSON schema uses, while Entity Services uses "datatype".

specs/models/EntityModelDescriptor.v1.json
rjrudin: I'll update this, and the other pattern, to a-zA-Z0-9.

specs/models/EntityModelDescriptor.v1.json
```diff
@@ -0,0 +1,109 @@
+{
+  "title": "EntityModelDescriptor",
+  "type": "object",
+  "description": "JSON schema representation of a MarkLogic Entity Services model",
+  "x-marklogic": {
```
ryanjdew: This is great! Being able to extend the APIs with ML specific opens a lot of possibilities for automated testing driven by these API documents.

rjrudin:
I'm going to rework this to assume that ArtifactService can be reused.

rjrudin:
My intent is that the models DS API can reuse artifacts/core.sjs. But the models DS API has a simpler interface so that a client isn't required to build the entire JSON for an entity model, which I'd like to avoid, as it may change in the near future.

specs/reference/data-services/models.v1.json
```diff
@@ -0,0 +1,141 @@
...
+    },
+    "/createModel.sjs": {
+      "post": {
+        "summary": "createModel",
+        "operationId": "post-createModelsjs",
+        "responses": {
+          "200": {
+            "description": "OK",
+            "content": {
+              "application/json": {
```
ryanjdew: Do we want to change this to actually be a multipart request and use encoding to reference the JSON schema as part of the "input" parameter? Like this: https://swagger.io/docs/specification/describing-request-body/multipart-requests/ That would more accurately describe the request being made and could be eventually be leveraged for testing data services. Of course we have a good pattern for ML unit tests and Data Services, so that may be less important.

rjrudin: Yeah I struggled with how to map to DS parameters. The fact that it's multi-part is an implementation detail that's hidden from developers too, buried away in the DS-generated code. So if we want to use OpenAPI to describe DS interfaces, I think we just use a combo of query params and request body as JSON.

PR #3720, opened 2020-03-18
DHFPROD-4582:Adding test for "source query as a script" feature

srinathgit:
Adding test for "source query as a script" feature


srinathgit:
> I'm not clear on what gap this is filling that isn't addressed already by the new test in FlowRunnerTest, which tests several different kinds of source queries.

Ah .. I didn't see the new test in FLowRunnerTest. I only saw the MarkLogicUnittests. I will close this out.

PR #3724, opened 2020-03-19
DHFPROD-4605: Updated CONTRIBUTING.md file

rjrudin:
Updated Gradle wrapper to latest 5.x release - 5.6.4. 

Also updated README for one-ui.

I removed some of the instructions for forking and rebasing - I think we expect people to know how to manage a fork and branches, and there's of course plenty of info on the Internet for that. Just didn't want to maintain that information here too.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -5,7 +5,7 @@ through the process.
 
  - [Issues and Bugs](#found-an-issue)
  - [Feature Requests](#want-a-feature)
- - [Building from Source](#building-marklogic-data-hub-from-source)
+ - [Developing Data Hub](#developing-data-hub)
  - [Submission Guidelines](#submission-guidelines)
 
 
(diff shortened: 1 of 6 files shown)


CONTRIBUTING.md
```diff
@@ -1,425 +1,365 @@
...
+- the Data Hub Gradle plugin in ./ml-data-hub-plugin
+- the One-UI web application in ./one-ui
+- the QuickStart web application in ./web
+
+### Prerequisites
+
+You need:
+
+- MarkLogic Server (see [Version Compatibility](https://docs.marklogic.com/datahub/refs/version-compatibility.html) for the correct version)
+- Java JDK 8 or later
```
SameeraPriyathamTadikonda: Should It be Java 8 or 11?

rjrudin: Good catch - technically, it should be Java 9, as that's what is in our build.gradle files. I'll update.

CONTRIBUTING.md
```diff
@@ -1,425 +1,365 @@
...
+tests, written in JUnit and also written using [marklogic-unit-test](https://github.com/marklogic-community/marklogic-unit-test). 
+While developing, you'll create/modify/run tests to verify the changes that you're making. 
+
+To run the tests, you first need to deploy a test instance of Data Hub to MarkLogic. The configuration for this instance
+is in the ./marklogic-data-hub/gradle.properties; you can override these values in gradle-local.properties. Verify
+that the values for mlSecurityUsername and mlSecurityPassword are correct, as that user will be used to deploy the test
+instance of Data Hub. Also, ensure that you do not have a Data Hub instance already deployed to the MarkLogic that you 
+will connect to. Then, run the following Gradle task from the root project:
+
+    ./gradlew -Pskipui= bootstrap
```
SameeraPriyathamTadikonda: should we add -Pskipui=true? missing "true"

rjrudin: Nope, as long as the property is defined, it'll achieve the desired effect.

CONTRIBUTING.md
SameeraPriyathamTadikonda: @rjrudin Can we build the project using java9? currently we are building using 11.

rjrudin: We can - our build.gradle files have source and target compatibilities of Java 9.

rjrudin: In practice though, people are likely using Java 11.

PR #3730, opened 2020-03-19
DHFPROD-3647: Save personal search query and its associated filters

rahulvudutala:
DHFPROD-3647: Save Query endpoint implementation
DHFPROD-3647: Adding reader and writer roles for saved query documents
DHFPROD-3647: Save Query dataservice modules

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/resources/hub-internal-config/security/roles/data-hub-operator.json
+++ b/marklogic-data-hub/src/main/resources/hub-internal-config/security/roles/data-hub-operator.json
@@ -12,7 +12,9 @@
     "data-hub-step-definition-reader",
     "data-hub-load-data-reader",
     "data-hub-match-merge-reader",
-    "tde-view"
+    "tde-view",
+    "data-hub-saved-query-reader",
+    "data-hub-saved-query-writer"
   ],
   "privilege": [
     {
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/savedQueries/saveSavedQuery.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/savedQueries/saveSavedQuery.sjs
@@ -1,0 +1,52 @@
+'use strict';
+declareUpdate();
+
+var saveQuery;
+var userCollections = ["http://marklogic.com/data-hub/saved-query"];
+var queryDocument = JSON.parse(saveQuery);
+
+if (queryDocument == null || queryDocument.savedQuery == null) {
+    throw Error("The request is empty or malformed");
+}
+
+if (queryDocument.savedQuery.name == null || !queryDocument.savedQuery.name) {
+    throw Error("Query name is missing");
+}
+
+if (queryDocument.savedQuery.query == null || Object.keys(queryDocument.savedQuery.query) == 0) {
+    throw Error("Query to be saved cannot be empty");
+}
+
+if (queryDocument.savedQuery.propertiesToDisplay == null || queryDocument.savedQuery.propertiesToDisplay.length == 0) {
+    throw Error("Entity type properties to be displayed cannot be empty");
+}
+
+let id = queryDocument.savedQuery.id;
+if (cts.doc("/saved-queries/" + id + ".json")) {
+    queryDocument.savedQuery.systemMetadata.lastUpdatedBy = xdmp.getCurrentUser();
+    queryDocument.savedQuery.systemMetadata.lastUpdatedDateTime = fn.currentDateTime();
+    xdmp.nodeReplace(cts.doc("/saved-queries/" + id + ".json"), queryDocument);
+} else {
+    queryDocument.savedQuery.id = sem.uuidString();
+    queryDocument.savedQuery.owner = xdmp.getCurrentUser();
+    queryDocument.savedQuery.systemMetadata = {
+        "createdBy": xdmp.getCurrentUser(),
+        "createdDateTime": fn.currentDateTime(),
+        "lastUpdatedBy": xdmp.getCurrentUser(),
+        "lastUpdatedDateTime": fn.currentDateTime()
+    };
+    insertDocument(queryDocument);
+}
+
+function insertDocument(queryDocument) {
+    let docUri = "/saved-queries/" + queryDocument.savedQuery.id + ".json";
+    let permissions = [xdmp.permission('data-hub-saved-query-reader', 'read'),
+        xdmp.permission('data-hub-saved-query-writer', 'update'),
+        xdmp.defaultPermissions()];
+    xdmp.documentInsert(docUri, queryDocument, {
+        permissions: permissions,
+        collections: userCollections
+    });
+}
+
+queryDocument;
--- a/marklogic-data-hub/src/test/java/com/marklogic/hub/dataservices/savedQueries/SavedQueriesServiceTest.java
+++ b/marklogic-data-hub/src/test/java/com/marklogic/hub/dataservices/savedQueries/SavedQueriesServiceTest.java
@@ -1,0 +1,175 @@
+package com.marklogic.hub.dataservices.savedQueries;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import com.marklogic.client.FailedRequestException;
+import com.marklogic.client.io.DocumentMetadataHandle;
+import com.marklogic.hub.ApplicationConfig;
+import com.marklogic.hub.HubTestBase;
+import com.marklogic.hub.dataservices.SavedQueriesService;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestInstance;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.springframework.test.context.ContextConfiguration;
+import org.springframework.test.context.junit.jupiter.SpringExtension;
+
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Set;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+@ExtendWith(SpringExtension.class)
+@ContextConfiguration(classes = ApplicationConfig.class)
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+public class SavedQueriesServiceTest extends HubTestBase {
+
+    private static JsonNode queryDoc;
+    private SavedQueriesService savedQueriesService;
+
+    @BeforeAll
+    void createSavedQueriesServiceInstance() {
+        this.savedQueriesService = SavedQueriesService.on(adminHubConfig.newFinalClient(null));
+    }
+
+    @BeforeEach
+    void before() throws IOException {
+        String jsonString = "{\n" +
+                "  \"savedQuery\": {\n" +
+                "    \"id\": \"\",\n" +
+                "    \"name\": \"some-query\",\n" +
+                "    \"description\": \"some-query-description\",\n" +
+                "    \"query\": {\n" +
+                "      \"searchText\": \"some-string\",\n" +
+                "      \"entityTypeIds\": [\n" +
+                "        \"Entity1\"\n" +
+                "      ],\n" +
+                "      \"selectedFacets\": {\n" +
+                "        \"Collection\": {\n" +
+                "          \"dataType\": \"string\",\n" +
+                "          \"stringValues\": [\n" +
+                "            \"Entity1\",\n" +
+                "            \"Collection1\"\n" +
+                "          ]\n" +
+                "        },\n" +
+                "        \"facet1\": {\n" +
+                "          \"dataType\": \"decimal\",\n" +
+                "          \"rangeValues\": {\n" +
+                "            \"lowerBound\": \"2.5\",\n" +
+                "            \"upperBound\": \"15\"\n" +
+                "          }\n" +
+                "        },\n" +
+                "        \"facet2\": {\n" +
+                "          \"dataType\": \"dateTime\",\n" +
+                "          \"rangeValues\": {\n" +
+                "            \"lowerBound\": \"2020-01-01T13:06:17\",\n" +
+                "            \"upperBound\": \"2020-01-22T13:06:17\"\n" +
+                "          }\n" +
+                "        }\n" +
+                "      }\n" +
+                "    },\n" +
+                "    \"propertiesToDisplay\": [\"facet1\", \"EntityTypeProperty1\"]\n" +
+                "  }\n" +
+                "}";
+        ObjectMapper mapper = new ObjectMapper();
+        queryDoc = mapper.readTree(jsonString);
+    }
+
+    @Test
+    void testSaveAndModifyQuery() {
+        JsonNode firstResponse = savedQueriesService.saveSavedQuery(queryDoc);
+        String id = firstResponse.get("savedQuery").get("id").asText();
+
+        assertNotNull(firstResponse);
+        assertNotNull(firstResponse.get("savedQuery"));
+        assertTrue(!id.isEmpty());
+        assertNotNull(firstResponse.get("savedQuery").get("systemMetadata"));
+        assertEquals("some-query", firstResponse.get("savedQuery").get("name").asText());
+        assertEquals(4, firstResponse.get("savedQuery").get("systemMetadata").size());
+        assertEquals("flow-developer", firstResponse.get("savedQuery").get("owner").asText());
+        assertEquals("flow-developer", firstResponse.get("savedQuery").get("systemMetadata").get("createdBy").asText());
+        assertPermissionsAndCollections("/saved-queries/" + id + ".json");
+
+        ObjectNode savedQueryNode = (ObjectNode) firstResponse.get("savedQuery");
+        savedQueryNode.put("name", "modified-name");
+        JsonNode modifiedResponse = savedQueriesService.saveSavedQuery(firstResponse);
+
+        assertNotNull(modifiedResponse);
+        assertEquals(id, modifiedResponse.get("savedQuery").get("id").asText());
+        assertEquals("flow-developer", modifiedResponse.get("savedQuery").get("owner").asText());
+        assertEquals("modified-name", modifiedResponse.get("savedQuery").get("name").asText());
+        assertPermissionsAndCollections("/saved-queries/" + id + ".json");
+    }
+
+    @Test
+    void testModifyQueryWithNonExistentId() {
+        JsonNode firstResponse = savedQueriesService.saveSavedQuery(queryDoc);
+        ObjectNode savedQueryNode = (ObjectNode) firstResponse.get("savedQuery");
+        savedQueryNode.put("id", "some-non-existent-id");
+        savedQueryNode.put("name", "modified-name");
+        JsonNode modifiedResponse = savedQueriesService.saveSavedQuery(firstResponse);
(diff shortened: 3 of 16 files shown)


marklogic-data-hub/src/main/resources/hub-internal-config/security/roles/data-hub-operator.json
```diff
@@ -12,7 +12,9 @@
     "data-hub-step-definition-reader",
     "data-hub-load-data-reader",
     "data-hub-match-merge-reader",
-    "tde-view"
+    "tde-view",
+    "data-hub-saved-query-reader",
+    "data-hub-saved-query-writer"
```
srinathgit: I believe "data-hub-saved-query-writer" should be assigned to "data-hub-developer"

rahulvudutala: Users able to run flows and see entity instances should be able to save personal queries. So added the write capability to data-hub-operator.
data-hub-developer inherits data-hub-operator role. So didn't any role for data-hub-developer

rjrudin: Yeah, normally the reader role would go to operator and writer to developer, but this is different since an operator has to be able to read and write saved queries.

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/data-services/security/getAuthoritiesTest.sjs
```diff
@@ -8,17 +8,18 @@ function invokeService() {
...
         "canWriteStepDefinitions,canReadStepDefinitions,canWriteMappings,canReadMappings,canWriteMatching,canReadMatching",
         response.authorities.toString()),
-    test.assertEqual(17, response.roles.length),
-    test.assertEqual("data-hub-operator,data-hub-entity-model-reader,data-hub-mapping-writer,data-hub-mapping-reader,"+
-    "data-hub-match-merge-writer,data-hub-load-data-writer,data-hub-module-writer,data-hub-load-data-reader,"+
-    "data-hub-job-reader,data-hub-flow-reader,data-hub-step-definition-reader,data-hub-match-merge-reader,data-hub-step-definition-writer,"+
-    "data-hub-flow-writer,data-hub-developer,data-hub-entity-model-writer,data-hub-module-reader",
-        response.roles.toString())
-]
+    test.assertEqual(19, response.roles.length),
```
rjrudin: I'd change that to be an assertTrue(response.roles.length > 17, "with some message here explaining what the minimum number of expected roles is"). This assertion is otherwise going to break whenever roles are added.

one-ui/src/main/java/com/marklogic/hub/oneui/controllers/SavedQueriesController.java
```diff
@@ -0,0 +1,47 @@
...
+    public ResponseEntity<JsonNode> saveQueryDocument(@RequestBody JsonNode queryDocument) {
+        JsonNode savedQuery = savedQueriesManager.saveOrUpdateQueryDocument(queryDocument);
+        return new ResponseEntity<>(savedQuery, HttpStatus.CREATED);
+    }
+
+    @RequestMapping(method = RequestMethod.PUT)
+    @ResponseBody
+    public ResponseEntity<JsonNode> updateQueryDocument(@RequestBody JsonNode queryDocument) {
+        JsonNode savedQuery = savedQueriesManager.saveOrUpdateQueryDocument(queryDocument);
+        return new ResponseEntity<>(savedQuery, HttpStatus.OK);
```
rjrudin: No need to change now, but there's a ResponseEntity.ok method you can use to get this all done in one line of code (I think by rule, a controller method should be one line of code, but there will of course be exceptions).

There's also ResponseEntity.created, but it requires a URI - which makes me think that perhaps the saveQueryDocument method should just return the queryId?

one-ui/src/main/java/com/marklogic/hub/oneui/controllers/SavedQueriesController.java
```diff
@@ -0,0 +1,47 @@
...
+@RequestMapping(value = "/api/savedQueries")
+public class SavedQueriesController {
+
+    @Autowired
+    private SavedQueriesManager savedQueriesManager;
+
+    @Autowired
+    private HubConfigSession hubConfig;
+
+    @Bean
```
rjrudin: I don't think Bean is buying you anything here, right? I think the method works the same without these two annotations on it. I recommend removing them to avoid any confusion.

one-ui/src/main/java/com/marklogic/hub/oneui/controllers/SavedQueriesController.java
```diff
@@ -0,0 +1,47 @@
...
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestMethod;
+import org.springframework.web.bind.annotation.ResponseBody;
+
+@Controller
+@RequestMapping(value = "/api/savedQueries")
+public class SavedQueriesController {
+
+    @Autowired
+    private SavedQueriesManager savedQueriesManager;
```
rjrudin: You don't need this if you're using savedQueriesManager(), right?

one-ui/src/main/java/com/marklogic/hub/oneui/managers/SavedQueriesManager.java
```diff
@@ -0,0 +1,43 @@
...
+import com.marklogic.hub.dataservices.SavedQueriesService;
+import com.marklogic.hub.oneui.exceptions.BadRequestException;
+import com.marklogic.hub.oneui.exceptions.DataHubException;
+
+public class SavedQueriesManager {
+
+    private DatabaseClient finalDataServiceClient;
+
+    public SavedQueriesManager(HubConfig hubConfig) {
+        finalDataServiceClient = hubConfig.newFinalClient();
```
rjrudin: Does this really work? I'm thinking this will break because newFinalClient() passes in a finalDbName, which will cause Data Services to throw an error. I think you want newFinalClient(null).

one-ui/src/main/java/com/marklogic/hub/oneui/managers/SavedQueriesManager.java
```diff
@@ -0,0 +1,43 @@
...
+
+        if (queryDocument.get("savedQuery").get("propertiesToDisplay") == null || queryDocument.get("savedQuery")
+                .get("propertiesToDisplay").isEmpty()) {
+            throw new BadRequestException("Entity type properties to be displayed cannot be empty");
+        }
+
+        try {
+            return SavedQueriesService.on(finalDataServiceClient).saveSavedQuery(queryDocument);
+        } catch (MarkLogicServerException mse) {
+            throw new DataHubException(mse.getServerMessage(), mse);
```
rjrudin: What's the purpose of wrapping this in a DataHubException?

one-ui/src/main/java/com/marklogic/hub/oneui/managers/SavedQueriesManager.java
rjrudin: Also, instead of hanging onto a reference to the DatabaseClient, why not instantiate the SavedQueriesService here?

one-ui/src/test/java/com/marklogic/hub/oneui/managers/SavedQueriesManagerTest.java
```diff
@@ -0,0 +1,136 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.test.context.junit.jupiter.SpringExtension;
+
+@ExtendWith(SpringExtension.class)
+@SpringBootTest(classes = {Application.class})
+public class SavedQueriesManagerTest extends TestHelper {
+
+    @Autowired
+    private SavedQueriesManager savedQueriesManager;
```
rjrudin: We don't want to autowire in Manager classes - just create one with the HubConfig in the test framework.

one-ui/src/main/java/com/marklogic/hub/oneui/managers/SavedQueriesManager.java
```diff
@@ -0,0 +1,43 @@
...
+public class SavedQueriesManager {
+
+    private DatabaseClient finalDataServiceClient;
+
+    public SavedQueriesManager(HubConfig hubConfig) {
+        finalDataServiceClient = hubConfig.newFinalClient();
+    }
+
+    public JsonNode saveOrUpdateQueryDocument(JsonNode queryDocument) {
+        if (queryDocument == null || queryDocument.get("savedQuery") == null) {
```
rjrudin: I think all of this logic should be in the DS endpoint - this is all validation logic, and similar to artifacts, we want to enforce this in the DS, not in Java code.

By doing that, you don't even need this class - the controller can just interact directly with the SavedQueryService. 

Also, @ryanjdew - this seems like a case in the future where we could use JSON schema to validate the incoming JSON so we don't have to write a lot of imperative code like this, right?

one-ui/ui/api/swagger/mocks.json
```diff
@@ -3768,6 +3768,128 @@
...
+        "produces":[
+          "application/json"
+        ],
+        "parameters":[
+          {
+            "in":"body",
+            "name":"body",
+            "required":true,
+            "schema":{
+              "$ref":"#/definitions/saveQueryRequest"
```
rjrudin: For consistency, this should be "savedQueryRequest" with a "d"

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/data-services/savedQueries/saveSavedQueryTest.sjs
```diff
@@ -0,0 +1,82 @@
...
+      "/data-hub/5/data-services/savedQueries/saveSavedQuery.sjs",
+      {
+        "saveQuery": saveQuery
+      }
+  ));
+}
+
+
+function testSaveNewQuery() {
+  const result = invokeService(saveQuery);
```
rjrudin: This test should also confirm the expected URI, permissions, and collections on the saved query.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/savedQueries/saveSavedQuery.sjs
```diff
@@ -0,0 +1,36 @@
...
+    queryDocument.savedQuery.systemMetadata = {
+      "createdBy": xdmp.getCurrentUser(),
+      "createdDateTime": fn.currentDateTime(),
+      "lastUpdatedBy": xdmp.getCurrentUser(),
+      "lastUpdatedDateTime": fn.currentDateTime()
+    }
+  }
+
+  let docUri = "/saved-queries/"+queryDocument.savedQuery.id;
+  let permissions = [xdmp.permission('data-hub-saved-query-reader', 'read'),
```
rjrudin: Also include "xdmp.defaultPermissions()"

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/savedQueries/saveSavedQuery.sjs
```diff
@@ -0,0 +1,36 @@
+'use strict';
+declareUpdate();
+
+var saveQuery;
+var userCollections = ["http://marklogic.com/data-hub/saved-query"];
+var queryDocument = JSON.parse(saveQuery);
+
+try {
+  if(queryDocument.savedQuery.id) {
```
rjrudin: In the case of an update, we actually don't want to do a documentInsert. That's because a user may have adjusted the permissions/collections on the document. By doing a documentInsert, we instead have to be careful not to lose any such additional permissions/collections. 

Instead, do this:

    xdmp.nodeReplace(cts.doc("/saved-queries/" + id + ".json"), queryDocument);

That way, you only update the contents of the document, which is all you want to update here.

ryanjdew: @rahulvudutala If need it, you can use this helper function in hubUtils: https://github.com/marklogic/marklogic-data-hub/blob/develop/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/hub-utils.sjs#L67

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/savedQueries/saveSavedQuery.sjs
rahulvudutala: Added it in the new commit

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/savedQueries/saveSavedQuery.sjs
rahulvudutala: Added xdmp.nodeReplace

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/data-services/savedQueries/saveSavedQueryTest.sjs
rahulvudutala: the complete js file is run as a single transaction. So cannot get the document until the transaction is committed. Added the tests in the java layer

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/savedQueries/saveSavedQuery.sjs
```diff
@@ -0,0 +1,56 @@
...
+
+let id = queryDocument.savedQuery.id;
+
+if(id) {
+    if(cts.doc("/saved-queries/"+id+".json")) {
+        queryDocument.savedQuery.systemMetadata.lastUpdatedBy = xdmp.getCurrentUser();
+        queryDocument.savedQuery.systemMetadata.lastUpdatedDateTime = fn.currentDateTime();
+        xdmp.nodeReplace(cts.doc("/saved-queries/" + id+".json"), queryDocument);
+    } else {
+        throw Error("The requested resource with " + id + " is not found")
```
rjrudin: I don't think this is necessary. I think you just need to say that if /saved-queries/id.json exists already, then nodeReplace it, else documentInsert it.

one-ui/src/test/java/com/marklogic/hub/oneui/controllers/SavedQueriesControllerTest.java
```diff
@@ -0,0 +1,215 @@
...
+        Assertions.assertEquals(200, modifiedResponse.getStatusCodeValue());
+        Assertions.assertNotNull(modifiedResponseBody);
+        Assertions.assertEquals(id, modifiedResponseBody.get("savedQuery").get("id").asText());
+        Assertions.assertEquals("test-data-hub-developer", modifiedResponseBody.get("savedQuery").get("owner").asText());
+        Assertions.assertEquals("modified-name", modifiedResponseBody.get("savedQuery").get("name").asText());
+    }
+
+    @Test
+    void testCollectionsAndPermissionOnSavedQuery() {
+        JsonNode response = savedQueriesController.saveQueryDocument(queryDoc).getBody();
```
rjrudin: There's no need for a separate test here. Just add these assertions to the first test for creating a saved query here.

one-ui/src/test/java/com/marklogic/hub/oneui/controllers/SavedQueriesControllerTest.java
```diff
@@ -0,0 +1,215 @@
...
+        Assertions.assertNotNull(permissions.get("data-hub-saved-query-writer"));
+        Assertions.assertNotNull(permissions.get("data-hub-saved-query-reader"));
+        Assertions.assertTrue(readQueryCap.equals(permissions.get("data-hub-saved-query-reader")));
+        Assertions.assertTrue(writeQueryCap.equals(permissions.get("data-hub-saved-query-writer")));
+        Assertions.assertTrue(expectedCollections.equals(collections));
+    }
+
+    @Test
+    void testCollectionsAndPermissionOnModifiedQuery() {
+        JsonNode firstResponse = savedQueriesController.saveQueryDocument(queryDoc).getBody();
```
rjrudin: Same thing, just add these assertions to the test for modifying a query. 

Really, you can cover both things - creating a query and modifying it - in the same test.

one-ui/src/test/java/com/marklogic/hub/oneui/controllers/SavedQueriesControllerTest.java
```diff
@@ -0,0 +1,215 @@
...
+import org.springframework.http.ResponseEntity;
+import org.springframework.test.context.junit.jupiter.SpringExtension;
+
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Set;
+
+@ExtendWith(SpringExtension.class)
+@SpringBootTest(classes = {Application.class})
+public class SavedQueriesControllerTest extends TestHelper {
```
rjrudin: I emailed Ryan and Srikanth about this - I'd rather see all these tests against the data service and thus the tests should be in the core project, not in one-ui. The core project is its own thing, and we want to verify its functionality within the core project. We don't want to depend on a UI project to verify it.

Additionally, I don't think testing the controller proves much of anything, since the controller only needs to construct the appropriate DatabaseClient. 

No action yet on this, waiting to hear back from Ryan and Srikanth.

marklogic-data-hub/src/test/java/com/marklogic/hub/dataservices/savedQueries/SavedQueriesServiceTest.java
```diff
@@ -0,0 +1,192 @@
...
+
+        Assertions.assertNotNull(modifiedResponse);
+        Assertions.assertEquals(id, modifiedResponse.get("savedQuery").get("id").asText());
+        Assertions.assertEquals("flow-developer", modifiedResponse.get("savedQuery").get("owner").asText());
+        Assertions.assertEquals("modified-name", modifiedResponse.get("savedQuery").get("name").asText());
+    }
+
+    @Test
+    void testCollectionsAndPermissionOnSavedAndModifiedQuery() {
+        JsonNode firstResponse = savedQueriesService.saveSavedQuery(queryDoc);
```
rjrudin: I really recommend combining these assertions with the first test. The setup of the test is identical - you're saving a single saved query. You then want to assert on the contents of the persisted document, along with its URI and collections/permissions. There's no reason to do that in two separate tests, it just makes our test suite take a little longer to finish.

marklogic-data-hub/src/test/java/com/marklogic/hub/dataservices/savedQueries/SavedQueriesServiceTest.java
```diff
@@ -0,0 +1,192 @@
...
+                "}";
+        ObjectMapper mapper = new ObjectMapper();
+        queryDoc = mapper.readTree(jsonString);
+    }
+
+    @Test
+    void testSaveAndModifyQuery() {
+        JsonNode firstResponse = savedQueriesService.saveSavedQuery(queryDoc);
+
+        Assertions.assertNotNull(firstResponse);
```
rjrudin: I recommend adding this so that this reads a bit easier; it's also consistent with most of our existing tests (unfortunately not all of them):

    import static org.junit.jupiter.api.Assertions.*;

PR #3732, opened 2020-03-19
DHFPROD-4615: Facet selections should be relocated under the search bar

timur-isangulov:
Also contains fix for DHFPROD-4648: Change api path for numeric facets.


timur-isangulov:
Screen Shot 2020-03-20 at 10 26 08 AM

PR #3736, opened 2020-03-20
DHFPROD-4375: added session timer and popup warning

brucean52:
- Implementation requires all endpoints to call resetSessionTimer in finally block.
- Popup Warning timer is configurable in application.config.js ( use 290 for testing )
- Auto logout time can be changed for testing ( line 105 in App.tsx ) set to 280


brucean52:
Will re-create and re-open this PR

PR #3737, opened 2020-03-23
DHFPROD-4504: Not able to persist mapping artifact with same name in another entity

xnikhil08:
This PR resolves the bug 4504.

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/one-ui/ui/src/components/entities/mapping/create-edit-mapping-dialog/create-edit-mapping-dialog.tsx
+++ b/one-ui/ui/src/components/entities/mapping/create-edit-mapping-dialog/create-edit-mapping-dialog.tsx
@@ -21,6 +21,8 @@ const CreateEditMappingDialog = (props) => {
   const [isSelectedSourceTouched, setSelectedSourceTouched] = useState(false);
 
   const [isValid, setIsValid] = useState(false);
+  const [isNameDuplicate,setIsNameDuplicate] = useState(false);
+  const [errorMessage,setErrorMessage] = useState('');
 
   const [deleteDialogVisible, setDeleteDialogVisible] = useState(false);
   const [tobeDisabled, setTobeDisabled] = useState(false);
@@ -44,10 +46,10 @@ const CreateEditMappingDialog = (props) => {
     } else {
       setMapName('');
       setMapNameTouched(false);
-      //setCollections([]);
       setCollections('');
       setDescription('');
       setSrcQuery('')
+      setIsNameDuplicate(false);
     }
 
     return (() => {
@@ -60,6 +62,7 @@ const CreateEditMappingDialog = (props) => {
       setSelectedSourceTouched(false);
       setCollectionsTouched(false);
       setTobeDisabled(false);
+      setIsNameDuplicate(false);
     })
 
   }, [props.mapData, props.title, props.newMap]);
@@ -67,11 +70,6 @@ const CreateEditMappingDialog = (props) => {
   const onCancel = () => {
 
     if (checkDeleteOpenEligibility()) {
-      console.log(isMapNameTouched
-        , isDescriptionTouched
-        , isSelectedSourceTouched
-        , isCollectionsTouched
-        , isSrcQueryTouched)
       setDeleteDialogVisible(true);
     } else {
       props.setNewMap(false);
@@ -149,9 +147,16 @@ const CreateEditMappingDialog = (props) => {
 
     //Call create Mapping artifact API function
 
-    props.createMappingArtifact(dataPayload);
+    let status = await props.createMappingArtifact(dataPayload);
 
-    props.setNewMap(false);
+    if (status.code === 200) {
+      props.setNewMap(false);
+    } else if (status.code === 400) {
+      
+      setErrorMessage(status.message)
+      setIsNameDuplicate(true);
+      setIsValid(false);
+    }
   }
 
   const handleChange = (event) => {
@@ -164,7 +169,8 @@ const CreateEditMappingDialog = (props) => {
         setMapName(event.target.value);
         if (event.target.value.length > 0) {
           if (JSON.stringify(collections) !== JSON.stringify([]) || srcQuery) {
-            setIsValid(true);
+              setIsValid(true);
+              setIsNameDuplicate(false);
           }
         } else {
           setIsValid(false);
@@ -215,9 +221,7 @@ const CreateEditMappingDialog = (props) => {
       else {
         setCollectionsTouched(true);
         setCollections(event.target.value);
-        console.log('event.target.value',event.target.value === props.mapData.collection, props.mapData && props.mapData.collection)
         if (props.mapData && props.mapData.collection) {
-          console.log('props.mapData.collection',props.mapData.collection,event.target.value)
           if (props.mapData.collection === event.target.value) {
             
             setCollectionsTouched(false);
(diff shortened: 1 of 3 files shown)


one-ui/ui/src/components/entities/mapping/create-edit-mapping-dialog/create-edit-mapping-dialog.tsx
```diff
@@ -320,8 +321,8 @@ const CreateEditMappingDialog = (props) => {
           Name:&nbsp;<span className={styles.asterisk}>*</span>
           &nbsp;
             </span>} labelAlign="left"
-          validateStatus={(mapName || !isMapNameTouched) ? '' : 'error'}
-          help={(mapName || !isMapNameTouched) ? '' : 'Name is required'}
+          validateStatus={(mapName || !isMapNameTouched) ? (props.title === 'Edit Mapping' ? '' : (!props.mapNames.hasOwnProperty(mapName) ? '' : 'error')) : 'error'}
+          help={(mapName || !isMapNameTouched) ? (props.title !== 'Edit Mapping'? (props.mapNames.hasOwnProperty(mapName) ? 'A mapping step with the same name already exists. Enter a unique name.':'') : '') : 'Name is required'}
```
rjrudin: Is this doing its own form of validation? The server is validating this already, and throwing an error with a message of "A mapping with the same name but for a different entity type already exists. Please choose a different name.". 

If this is doing the validation by comparing the mapping names to the known set of names already - I don't think that's 100% reliable, because a new mapping could have been created between the time that the UI fetches the set of names and the user then tries to create one. That's why the server has to do the validation. The UI should just catch the error from the server and then display it.

bsrikan: +1

one-ui/ui/src/components/entities/mapping/mapping-card.test.tsx
```diff
@@ -12,6 +12,7 @@ describe('Mapping cards view component', () => {
       updateMappingArtifact
       canReadOnly
       canReadWrite
-      entityModel/>);
+      entityModel
+      mapNames/>);
```
bsrikan: does this test pass when running locally? The props "data" is not right for the mapping card. Try using RTL to render like so and verify if you get back mapping card HTML or an empty div
```
import { render } from "@testing-library/react"; 

describe.... 

test('it renders mapping card div correctly' , () => {
const { debug } = render(<MappingCard data />);
debug();
};
```

one-ui/ui/src/components/entities/mapping/create-edit-mapping-dialog/create-edit-mapping-dialog.tsx
xnikhil08: @bsrikan - The way it works currently is that as soon as you create a mapping artifact, the react refreshes the entity page showing the new map artifact created. By this time the mapping names object is already refreshed along with the entire mapping list for that entity or else you will not even see the newly created artifact in UI.
This approach is really good for validating on the fly as user types the names. Otherwise use will have to click on Save button several times to hit the right name.

However, we can add another validation on top of this on Save button, just to take care of the case where the newly added artifact has not shown up yet in the UI and for some reason user again tries to add another artifact with same name.
Let me know your thoughts.

rjrudin: One problem is that multiple users could try to create the same mapping at the same time. The UI shouldn't do any validation here. It should depend on the server to ensure that two mappings with the same name can't be created. The server will throw an error when that happens, and the UI should then display the error message from the server.

wooldridge: Just handling the error based on the backend server validation (as Rob describes) seems like the cleaner way to do this. I don't think this error is common enough to require the on-the-fly checking. @xnikhil08 are you OK with that?

xnikhil08: @rjrudin - OK, but the current server validation throws error only when the same mapping name exists in another entity but doesn't throw any error when it's the same entity. I am assuming we need this validation for same entity as well right?

xnikhil08: @wooldridge - Yeah i am fine with it. I guess my only concern was the user experience, if there are a lot more mappings then user would have to keep clicking on the save button to get to the unique value. Thats why now i was thinking of having a sort of combination of both backend and on-the-fly validation.
But yeah we can remove the on-the-fly validation completely if thats not needed here.

wooldridge: @xnikhil08 if the backend can be fixed to recognize the case where there are same-name mappings for the same entity, then I think we should just let the backend handle the validation and keep the UI code cleaner.

one-ui/ui/src/components/entities/mapping/mapping-card.test.tsx
xnikhil08: @bsrikan - Yes, it already passes as is when running locally.

one-ui/ui/src/components/entities/mapping/create-edit-mapping-dialog/create-edit-mapping-dialog.tsx
xnikhil08: @rjrudin & @bsrikan & @wooldridge - Please review now. Updated the PR as discussed.

one-ui/ui/src/components/entities/mapping/create-edit-mapping-dialog/create-edit-mapping-dialog.tsx
```diff
@@ -148,9 +149,11 @@ const CreateEditMappingDialog = (props) => {
...
     let status = await props.createMappingArtifact(dataPayload);
 
-    if (status === '200') {
+    if (status.code === 200) {
       props.setNewMap(false);
-    } else if (status === '400') {
+    } else if (status.code === 400) {
+      
+      setErrorMessage(status.message)
       setIsNameDuplicate(true);
```
rjrudin: This is probably safe for now, but I wanted to note that the UI is making an assumption here - that a 400 status means that the name is a duplicate. That isn't guaranteed to be true, and it could certainly change in the future when some other validation rule is added.

wooldridge: @rjrudin I'm reading this as that the UI is seeing a 400 error status and displaying the associated error message sent from the backend, there's not an assumption/interpretation of the type of error (e.g., duplicate name) on the UI's part other than it's a 400 status.

SameeraPriyathamTadikonda:
@wooldridge / @rjrudin  please check the reviewers task list

xnikhil08:
since this is just a bug fix so it doesn't involve any changes in the tests.

PR #3739, opened 2020-03-23
DHFPROD-4615: Facet selections should be relocated under the search bar

timur-isangulov:


one-ui/ui/src/components/numeric-facet/numeric-facet.tsx
```diff
@@ -30,7 +30,7 @@ const NumericFacet: React.FC<Props> = (props) => {
   const getFacetRange = async () => {
     const response = await axios({
       method: 'POST',
-      url: `/api/search/facet-values/range`,
+      url: `/datahub/v2/search/facet-values/range`,
```
bsrikan: @akshaysonvane does this look right to you.

bsrikan: Looks like @timur-isangulov reverted the change. Files changed is 0.

bsrikan: To fix incorrect git commits in a PR, rather than having multiple commits like cleanup or revert, I always use `git rebase -i HEAD~10` which opens up an editor and I can reword commit messages. There are others like fixup, which is again something I use to squash commits that are merely corrections of an incorrect commit.

PR #3740, opened 2020-03-23
DHFPROD-4656: Added endpoint for getting primary entity types

rjrudin:
Added test plumbing for using the reference model for tests. 

Also added more reliable way of waiting for post-commit triggers to finish - waitForTasksToFinish.

### DIFF
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/impl/HubProjectImpl.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/impl/HubProjectImpl.java
@@ -476,7 +476,9 @@ public class HubProjectImpl implements HubProject {
 
     private void writeResourceFile(String srcFile, Path dstFile, boolean overwrite) {
         if (overwrite || !dstFile.toFile().exists()) {
-            logger.info("Getting file: " + srcFile);
+            if (logger.isDebugEnabled()) {
+                logger.debug("Getting file: " + srcFile);
+            }
             InputStream inputStream = null;
             try {
                 inputStream = HubProject.class.getClassLoader().getResourceAsStream(srcFile);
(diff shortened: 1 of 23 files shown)


marklogic-data-hub/src/main/java/com/marklogic/hub/impl/HubProjectImpl.java
```diff
@@ -476,7 +476,9 @@ private void writeResourceFile(String srcFile, Path dstFile) {
 
     private void writeResourceFile(String srcFile, Path dstFile, boolean overwrite) {
         if (overwrite || !dstFile.toFile().exists()) {
-            logger.info("Getting file: " + srcFile);
+            if (logger.isDebugEnabled()) {
```
rjrudin: I changed this to debug, as it was getting logged a ton when running a test, but not really adding any value.

marklogic-data-hub/src/test/java/com/marklogic/hub/HubTestBase.java
```diff
@@ -158,21 +162,6 @@
     @Autowired
     protected Versions versions;
 
-    @Autowired
```
rjrudin: Got rid of these, as it was hurting my head trying to figure out whether modifying one of these would affect future tests. There's no need for them to be Spring beans, it's easier to think about if we just instantiate a command when we need one.

marklogic-data-hub/src/test/java/com/marklogic/hub/mapping/MappingManagerTest.java
```diff
@@ -215,39 +210,4 @@ private void copyTestMap() {
     private void copySecondTestMap() {
         FileUtil.copy(getResourceStream("scaffolding-test/"+mappingName+"-2"+MappingManager.MAPPING_FILE_EXTENSION), getDataHubAdminConfig().getHubMappingsDir().resolve(mappingName+"/"+mappingName+"-2"+MappingManager.MAPPING_FILE_EXTENSION).toFile());
     }
-
```
rjrudin: This is all dead code.

rjrudin:
Going to add the other 3 endpoints first.

PR #3745, opened 2020-03-24
DHFPROD-4662: Adding checklist to PR Template

SameeraPriyathamTadikonda:
### Description
Adding a check list to the PR template.
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,20 @@
+### Description
+
+#### Checklist: 
+```diff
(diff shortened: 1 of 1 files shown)


akshaysonvane:
Does the checkbox for `Squashed all commits into a single commit` mean that we need to squash all commits and then open a PR?
Having separate commits helps with reviewing the PR as it provides more info about why the specific change was made.
Can we squash and merge instead of squashing all commits to start with?

PR #3746, opened 2020-03-24
DHFPROD-4559: Explorer returns empty results when there is no data

rahulvudutala:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests

- ##### Reviewer:

- [x] Reviewed Tests


one-ui/src/main/java/com/marklogic/hub/oneui/util/SearchHelper.java
```diff
@@ -96,6 +99,12 @@ public StringHandle search(SearchQuery searchQuery) {
 
         }
         catch (MarkLogicServerException e) {
+            // If there are no entityTypes to search, return empty result set
```
rjrudin: This seems like something we can check on before trying to do the search, right? 

Also, I'd say we should add a test - but it doesn't look like SearchHelper is actually being tested at all??? I see in SearchManagerTest that it's being mocked. 

Why is SearchHelper separate from SearchManager? SearchManager isn't doing anything other than delegating to SearchHelper.

I recommend trying to write a real test here. This one is easy, because you don't need any data - just pass in a SearchQuery with no entity names, and verify you get back an empty StringHandle.

rahulvudutala: The reason for adding this check in the catch block is we have to make a getModels call for every search request. This is handled by adding empty list to the collection query if the searchQuery.getQuery().getEntityNames() is empty which results in no results.
 If there are no entityTypes, then there are no search options and we get 400 error with a missing options file. So I added this check in case of exception.

rahulvudutala: I will add a comment as you suggested
"if there are no entity models, then we expect an error because no search options will exist"

rjrudin: What about a test for this? Can you submit a SearchQuery with no entity names and get this expected exception, and then verify that you get an empty StringHandle back?

rahulvudutala: Addressed in new commits

one-ui/src/test/java/com/marklogic/hub/oneui/TestHelper.java
```diff
@@ -188,11 +190,35 @@ private void assignRoleToUser(String username, String role) {
...
-    protected GenericDocumentManager getFinalGenericDocumentManager(DatabaseKind databaseKind) {
-        return getFinalClient(databaseKind).newDocumentManager();
+    protected EvalResultIterator runJsInDatabase(String query, String databaseName) {
+        try {
+            return getServerEval(databaseName).javascript(query).eval();
+        }
+        catch(FailedRequestException e) {
+            logger.error("Failed run code: " + query, e);
+            e.printStackTrace();
+            throw e;
```
rjrudin: This is a lot of redundant logging for an exception. If the point is to make it clear what query failed, just rethrow the exception:

    throw new RuntimeException("Failed to run query:  " + query,e);

one-ui/src/main/java/com/marklogic/hub/oneui/util/SearchOptionBuilder.java
```diff
@@ -51,7 +51,7 @@ public String buildSearchOptions(String query, SearchQuery searchQuery) throws I
         // Setting search string if provided by user
         if (StringUtils.isNotEmpty(searchQuery.getQuery().getSearchStr())) {
             sb.append("<qtext>")
-                .append(StringEscapeUtils.escapeXml(searchQuery.getQuery().getSearchStr()))
+                .append(StringEscapeUtils.escapeXml10(searchQuery.getQuery().getSearchStr()))
```
rjrudin: This is worth adding a unit test for - why escapeXml10 and not escapeXml? What scenario is this handling, such that it would break if we didn't call escapeXml10?

rahulvudutala:
I am making 2 commits here as 1 commit includes a refactoring of the code

PR #3747, opened 2020-03-24
DHFPROD-4656: Added models service and controller

rjrudin:
Much of this is updates to existing tests and test plumbing:

- Also added more reliable way of waiting for post-commit triggers to finish - waitForTasksToFinish. This required a new privilege to be added to data-hub-developer in the test app. 
- Tweaked test logging in core so it's not so verbose 
- The Load*Command objects are no longer autowired in HubTestBase - they're just manually constructed when needed now

```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/impl/HubProjectImpl.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/impl/HubProjectImpl.java
@@ -478,7 +478,9 @@ public class HubProjectImpl implements HubProject {
 
     private void writeResourceFile(String srcFile, Path dstFile, boolean overwrite) {
         if (overwrite || !dstFile.toFile().exists()) {
-            logger.info("Getting file: " + srcFile);
+            if (logger.isDebugEnabled()) {
+                logger.debug("Getting file: " + srcFile);
+            }
             InputStream inputStream = null;
             try {
                 inputStream = HubProject.class.getClassLoader().getResourceAsStream(srcFile);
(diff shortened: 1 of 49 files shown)


brucean52:
GET /primaryEntityTypes works 
primaryEntityTypes

{modelName}/info & {modelName}/entityType both are returning 405
model-entityTypes
model-info

rjrudin:
Those are both POST endpoints for updating the info and entityTypes. See the API under ./specs.

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/HubProjectImpl.java
```diff
@@ -478,7 +478,9 @@ private void writeResourceFile(String srcFile, Path dstFile) {
 
     private void writeResourceFile(String srcFile, Path dstFile, boolean overwrite) {
         if (overwrite || !dstFile.toFile().exists()) {
-            logger.info("Getting file: " + srcFile);
+            if (logger.isDebugEnabled()) {
+                logger.debug("Getting file: " + srcFile);
```
akshaysonvane: I'm guessing we are using the guard condition in this case since we are performing string concatenation.
No need to change right now but the following approach would allow us to get rid of the "if" block.
Since we are using slf4j we can get rid of the guard condition by using 
```suggestion
                logger.debug("Getting file: {}",  srcFile);
```
This gets rid of the superfluous string concatenation when the logger is disabled for DEBUG level.

rjrudin: There are a lot of little things like that to clean up - I'm just changing the logging from info to debug here, since the info logging was cluttering up logging in a test. And this doesn't seem like it warrants "info" level.

rjrudin:
Just re-requesting a review as I changed the branch to develop.

akshaysonvane:
This requires a rebase.

rjrudin:
Rebased to solve a few merge conflicts

rjrudin:
Had to push a fix to a mapping test used by MasterTest, example.org was somehow changed to example.com when I rebased. Don't know how.

PR #3750, opened 2020-03-24
DHFPROD-4632: Mock api endpoints using jest for RTL tests.

timur-isangulov:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/one-ui/ui/src/api/__mocks__/facets.tsx
+++ b/one-ui/ui/src/api/__mocks__/facets.tsx
@@ -1,0 +1,7 @@
+import { numericRange } from '../../assets/mock-data/facet'
+
+export const rangeFacet = async () => {
+    return await new Promise((resolve) => {
+        resolve(numericRange)
+    })
+}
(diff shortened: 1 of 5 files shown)


bsrikan:
@wooldridge @xnikhil08 @ryanjdew @srinathgit @kkanthet  FYI. 
Please note that this PR from Timur tries to test by mocking an api call. Looks like the API calls(POST/GET etc) should be refactored to be outside a component, so as to enable testing via mocks.

timur-isangulov:
> @wooldridge @xnikhil08 @ryanjdew @srinathgit @kkanthet FYI.
> Please note that this PR from Timur tries to test by mocking an api call. Looks like the API calls(POST/GET etc) should be refactored to be outside a component, so as to enable testing via mocks.

Thanks Srikanth, After experimenting for several hours using different mocking mechanisms (jest, nock, unmock, etc) I think the jest mock works great. According to instruction it requires to have api calls in a separate file but I think we may need to separate them anyway. It will look cleaner and easier to manage if we have them all in one place.

rjrudin:
I like that pattern of api calls in a separate file or set of files - that essentially creates your client library for talking to the server.

PR #3753, opened 2020-03-25
DHFPROD-4610:Remove "fileLocations" from flow declaration

srinathgit:
Remove "fileLocations" from flow declaration
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/one-ui/src/test/java/com/marklogic/hub/curation/controllers/FlowControllerTest.java
+++ b/one-ui/src/test/java/com/marklogic/hub/curation/controllers/FlowControllerTest.java
@@ -12,6 +12,7 @@ import com.marklogic.hub.flow.impl.FlowRunnerImpl;
 import com.marklogic.hub.oneui.TestHelper;
 import com.marklogic.hub.oneui.models.HubConfigSession;
 import com.marklogic.hub.oneui.models.StepModel;
+import com.marklogic.hub.util.json.JSONUtils;
 import org.apache.commons.io.FileUtils;
 import org.junit.jupiter.api.AfterEach;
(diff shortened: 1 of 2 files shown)


one-ui/src/test/java/com/marklogic/hub/curation/controllers/FlowControllerTest.java
```diff
@@ -124,8 +125,13 @@ void getFlow() throws IOException {
 
             //POST step
             StepModel stepModel = (StepModel)controller.createStep("testFlow", 1, stepString).getBody();
+            JsonNode flowJson = JSONUtils.convertArtifactToJson(controller.getFlow("testFlow").getBody());
 
             Assertions.assertNotNull(stepModel);
+            //"fileLocations" should not be present in flow declaration
+            Assertions.assertNotNull(flowJson.get("steps").get("1"));
```
rjrudin: What happens if an existing flow file has fileLocations in a step, and then saves the flow via QS? Won't the user lose the fileLocations data? This seems risky to me.

rjrudin: I just realized this is the one-ui class. I think it's worth adding a comment to the JsonIgnore annotation to define why this is being ignored. Should it even be in the class if it's being ignored??

srinathgit: Made changes so that "fileLocations", if present in the step is retained.

PR #3767, opened 2020-03-27
DHFPROD-3841: Can now reference $URI in mappings

rjrudin:
This opens the door to all sorts of variables in a mapping artifact. However, to achieve this, I had to make a copy of the ES functions map-to-canonical and node-map-to-canonical so that I can pass in a map of params to the underlying xdmp:xslt calls. I opened up bug 54632 so that the ES functions will allow this in a future version.

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/builtins/steps/mapping/entity-services/test-data/mappings/PersonMapping/PersonMapping-7.mapping.json
+++ b/marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/builtins/steps/mapping/entity-services/test-data/mappings/PersonMapping/PersonMapping-7.mapping.json
@@ -1,0 +1,24 @@
+{
+  "lang": "zxx",
+  "name": "PersonMapping",
+  "version": 7,
+  "sourceContext": "/",
+  "targetEntityType": "http://marklogic.com/data-hub/example/Person-1.0.0/Person",
+  "properties": {
+    "id": {
+      "sourcedFrom": "personId"
+    },
+    "nickname": {
+      "sourcedFrom": "$URI"
+    },
+    "name": {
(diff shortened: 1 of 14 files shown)


marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/builtins/steps/mapping/entity-services/test-data/mappings/PersonMapping/PersonMapping-7.mapping.json
```diff
@@ -0,0 +1,24 @@
...
+  "name": "PersonMapping",
+  "version": 7,
+  "sourceContext": "/",
+  "targetEntityType": "http://marklogic.com/data-hub/example/Person-1.0.0/Person",
+  "properties": {
+    "id": {
+      "sourcedFrom": "personId"
+    },
+    "nickname": {
+      "sourcedFrom": "$URI"
```
srinathgit: @rjrudin ,
Can "sourcedFrom" be set to something like concat($URI, 'test-uri') ? If so, adding a test for that would be good.

rjrudin: I don't think it would prove anything additional. The contents of sourcedFrom are evaluated by ML, and thus from a DHF testing perspective, there's no difference between "$URI" and an expression that incorporates multiple functions and predicates. Either "$URI" gets resolved, or it doesn't - the complexity of the expression doesn't affect that.

PR #3772, opened 2020-03-30
DHFPROD-4186: Map from XML source data

xnikhil08:
This PR enables XML data support in Mapping source data table.

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/one-ui/ui/src/components/entities/entity-tiles.tsx
+++ b/one-ui/ui/src/components/entities/entity-tiles.tsx
@@ -37,13 +37,12 @@ const EntityTiles = (props) => {
                 let mapArtifacts = response.data;
                 mapArtifacts.sort((a, b) => (a.entityType > b.entityType) ? 1 : -1)
                 setEntityArtifacts([...mapArtifacts]);
-              console.log('GET Mapping Artifacts API Called successfully!',response);
             }
           } catch (error) {
               let message = error;
-              console.log('Error while fetching the mappings!', message);
-          } finally {
-            resetSessionTime();
+              console.error('Error while fetching the mappings!', message);
+            } finally {
+              resetSessionTime();  
           }
     }
 
@@ -58,11 +57,10 @@ const EntityTiles = (props) => {
                 return mapArtifacts;
                }
 
-              console.log('GET Mapping Artifacts API Called successfully!',response);
             }
           } catch (error) {
               let message = error;
-              console.log('Error while fetching the mapping!', message);
+              console.error('Error while fetching the mapping!', message);
           } finally {
             resetSessionTime();
           }
@@ -80,7 +78,7 @@ const EntityTiles = (props) => {
             }
           } catch (error) {
               let message = error.response.data.message;
-              console.log('Error while deleting the mapping!', message);
+              console.error('Error while deleting the mapping!', message);
               setIsLoading(false);
           }
     }
@@ -91,7 +89,6 @@ const EntityTiles = (props) => {
 
             let response = await axios.post(`/api/artifacts/mapping/${mapping.name}`, mapping);
             if (response.status === 200) {
-              console.log('Create MappingArtifact API Called successfully!')
               setIsLoading(false);
               return {code: response.status};
             } else {
@@ -101,7 +98,7 @@ const EntityTiles = (props) => {
             let message = error;
             let code = error.response.data.code;
             let details = error.response.data.details
-            console.log('Error while creating the mapping!', message)
+            console.error('Error while creating the mapping!', message)
             setIsLoading(false);
             let err={code: code,
                     message: details}
@@ -116,14 +113,13 @@ const EntityTiles = (props) => {
 
             let response = await axios.post(`/api/artifacts/mapping/${mapping.name}`, mapping);
             if (response.status === 200) {
-              console.log('Update MappingArtifact API Called successfully!')
               return true;
             } else {
                 return false;
             }
           } catch (error) {
             let message = error;
-            console.log('Error while updating the mapping!', message)
+            console.error('Error while updating the mapping!', message)
             return false;
           } finally {
             resetSessionTime();
@@ -136,9 +132,8 @@ const EntityTiles = (props) => {
             if (response.status === 200) {
                 let entArt = response.data;
                 entArt.sort((a, b) => (a.entityType > b.entityType) ? 1 : -1)
-                console.log('entArt',entArt)
                 setMatchingArtifacts([...entArt]);
-              console.log('GET matching Artifacts API Called successfully!',response);
+              console.log('GET matching Artifacts API Called successfully!');
             }
           } catch (error) {
               let message = error;
--- a/one-ui/ui/src/components/entities/mapping/mapping-card.tsx
+++ b/one-ui/ui/src/components/entities/mapping/mapping-card.tsx
@@ -13,6 +13,7 @@ import { AdvMapTooltips } from '../../../config/tooltips.config';
 import {RolesContext} from "../../../util/roles";
 import { getSettingsArtifact } from '../../../util/manageArtifacts-service';
 import axios from 'axios';
+import { xmlParserForMapping } from '../../../util/xml-parser';
 
 interface Props {
     data: any;
@@ -56,6 +57,13 @@ const MappingCard: React.FC<Props> = (props) => {
 
     //For storing  mapping functions
     const [mapFunctions,setMapFunctions] = useState({});
+
+    //For storing namespaces
+    const [namespaces, setNamespaces] = useState({});
+    let nmspaces: any = {};
+    let mapIndexLocal: number = -1;
+    const [mapIndex, setMapIndex] = useState(-1);
+    let namespaceString = '';
 
     useEffect(() => {
         setSourceData([]);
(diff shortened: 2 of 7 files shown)


one-ui/ui/src/components/entities/entity-tiles.tsx
```diff
@@ -55,7 +55,7 @@ const EntityTiles = (props) => {
                 return mapArtifacts;
                }
 
-              console.log('GET Mapping Artifacts API Called successfully!',response);
+              console.log('GET Mapping Artifacts API Called successfully!');
             }
```
bsrikan: Can we get rid of these console.logs completely? Logging errors is still ok, and maybe we should try console.lerror for those.

one-ui/ui/src/config/data.config.js
```diff
@@ -52,10 +61,14 @@ const mapProps = {
...
   },
+  namespaces: {
+    nutFree: 'http://namespaces/nutfree',
+    withNuts: 'http://namespaces/withNuts'
+  },
   mapName: 'testMap',
   getMappingArtifactByMapName: jest.fn(),
   updateMappingArtifact: jest.fn(),
-  mappingVisible:  false,
+  mappingVisible:  true,
```
bsrikan: hope this doesnt break any other tests which requires mappingVisible to be false. We can always render with props overridden from this config file.. Like below so we can keep data.config generic:
render(<SourceToEntityMap {...data.mapProps} mappingVisible={true}/>)

xnikhil08: Updated the test. Please review now.

one-ui/ui/src/components/entities/entity-tiles.tsx
xnikhil08: @bsrikan - Yes, we can definitely remove these console logs completely and keep only the error logs. Updated the PR with the same. Please review now.

srinathgit:
I am seeing issues in xml documents with namespaces compared to 5.2 interface.
1. Every child node has the namespace (example.org) which is missing in current implementation
2. There is space after ':' in this string  " example.org: person"  which is causing issues with xpath. The space should be removed
3. If I hit the arrow in the source table next to "FirstName" , the entire table collapses whereas only that "FirstName" subtree should collapse.

xnikhil08:
> I am seeing issues in xml documents with namespaces compared to 5.2 interface.
> 
> 1. Every child node has the namespace (example.org) which is missing in current implementation
> 2. There is space after ':' in this string  " example.org: person"  which is causing issues with xpath. The space should be removed
> 3. If I hit the arrow in the source table next to "FirstName" , the entire table collapses whereas only that "FirstName" subtree should collapse.
>    The images from 5.2, 5.3 and source doc have been included
> 
> 
> 
> 
> 
> ```
> <?xml  version="1.0" encoding="UTF-8"?>
> <envelope xmlns="http://marklogic.com/entity-services">
> <headers>
> </headers>
> <triples>
> </triples>
> <instance>
> <person xmlns="http://example.org">
> <personId>111</personId>
> <name>
> <FirstName prefix="Mrs">Jane</FirstName>
> <middle>Joan</middle>
> <last>Smith</last>
> </name>
> </person>
> </instance>
> <attachments>
> </attachments>
> </envelope>
> ```

@srinathgit - All the issues have been fixed. Can you check again?

xnikhil08:
> Looks really good! I'm seeing a minor issue regarding the handling of nested elements. Looking at the following sample XMl file: SampleProtein.xml
> 
> 1. Open root element (entry). OK.
> 2. Open/close first nested level (protein, gene, organism). OK.
> 3. However second nested level (recommendedName, name) always appears open after first nested level is opened. And then trying to close the second nested level results in the entire hierarchy closing (up to root entry level).
> 
> See screencast: https://drive.google.com/file/d/12QYKof4Y2b4war5Qrfboa_wKyhRS1D-p/view

@wooldridge - Its fixed now. Please check now.

one-ui/ui/src/components/entities/mapping/mapping-card.tsx
```diff
@@ -57,6 +58,13 @@ const MappingCard: React.FC<Props> = (props) => {
     //For storing  mapping functions
     const [mapFunctions,setMapFunctions] = useState({});
 
+    //For storing namespaces
+    const [namespaces, setNamespaces] = useState({});
+    let nmspaces: any = {};
+    let mapIndexLocal: number = -1;
+    const [mapIndex, setMapIndex] = useState(-1);
+    let namespaceString = '';
```
bsrikan, line 66: Trying to understand what these variables mean and how they are different from each other-- namespaces, nmspaces and namespaceString. Can you pls explain.

With all of us trying to write unit test, I feel its important we use standard naming conventions so its easy to read and understand the code. `nmspc`, `nmspaces`, `namespaces`, `ind`, `sDta`, `mData`, `mapArt` and `propty` are some examples.

one-ui/ui/src/components/entities/mapping/mapping-card.tsx
```diff
@@ -177,83 +185,233 @@ const MappingCard: React.FC<Props> = (props) => {
...
+                if(parsedDoc['envelope']){
+                    if(parsedDoc['envelope'].hasOwnProperty('@xmlns')){
+                        
+                        let nmspcURI = parsedDoc['envelope']['@xmlns']
+                        let indCheck = nmspcURI.lastIndexOf('/');
+                        let ind = indCheck !== -1 ? indCheck + 1 : 0;
+                        let nmspcString = nmspcURI.slice(ind);
+                        namespaceString = nmspcString;
+                        nmspaces = { ...nmspaces, [namespaceString]: nmspcURI};
+                        setNamespaces({ ...namespaces, [namespaceString]: nmspcURI})
```
bsrikan, line 207: Could we have just used `namespaceString = nmspcURI.slice(ind);`?
Also looks like we are trying to get a namespace string in 4 to 5 different places similar to line 201 to 205. Can we create a function and reuse that to return the namespace string.

one-ui/ui/src/components/entities/mapping/source-entity-map/source-to-entity-map.tsx
```diff
@@ -72,7 +72,7 @@ const SourceToEntityMap = (props) => {
 
     const handleSubmitUri = (uri) => {
         props.getMappingArtifactByMapName();
-        props.fetchSrcDocFromUri(uri);
+        props.fetchSrcDocFromUri(uri,props.mapIndex);
```
bsrikan, line 75: Can you pls explain what this prop is for -- props.mapIndex. Should this be added to mapProps in data.config?

one-ui/ui/src/components/entities/mapping/mapping-card.tsx
xnikhil08: nmspaces - Local variable to store namespaces when iterating over  xml source doc. This will be used locally at other places as well.
namespaces - This is a state variable and will be passed on to the child component and have state management properties.
namespaceString - Just a string local variable to store the most recent namespace at a given time for any xml node. This is only a local temporary variable and has no significance for child component.

Any variable which is not part of the react state is a local variable for that component and has no relation with the child component.

one-ui/ui/src/components/entities/mapping/mapping-card.tsx
xnikhil08: This logic to set namespaceString is used only at this very place so seems an overkill to create a function for only one instance.

one-ui/ui/src/components/entities/mapping/source-entity-map/source-to-entity-map.tsx
xnikhil08: It wouldn't affect any of the existing tests. I ran them locally and made sure they are working fine. But yeah when we will write more tests using the function props.fetchSrcDocFromUri, then yes we would have to add that in the data.config.

one-ui/ui/src/components/entities/mapping/mapping-card.tsx
xnikhil08: nmspcString is extracted only at two places, one is inside the function updateParentNamespace() (already being called at multiple places) and other at line 204-205. The reason why that function is not called at these lines is that the concept of "parentNamespace" didn't made sense at this point earlier but only once you enter into the nested source logic. If still however we want to replace that code with a function call, then I will have to rename the function to something else which fits both the scenarios.

bsrikan: Discussed with @xnikhil08 . Will resolve this conversation and refactor the code later when writing tests or as a tech debt.

srinathgit:
> I still see the spaces that I had mentioned in the previous comment. Other issues have been resolved.

Spoke with @xnikhil08 , he will include that minor change in PR for DHFPROD-4172. Approving now

one-ui/ui/src/components/entities/entity-tiles.tsx
```diff
@@ -37,13 +37,10 @@ const EntityTiles = (props) => {
                 let mapArtifacts = response.data;
                 mapArtifacts.sort((a, b) => (a.entityType > b.entityType) ? 1 : -1)
                 setEntityArtifacts([...mapArtifacts]);
-              console.log('GET Mapping Artifacts API Called successfully!',response);
             }
           } catch (error) {
               let message = error;
-              console.log('Error while fetching the mappings!', message);
-          } finally {
-            resetSessionTime();
```
bsrikan, line 46: was deleting finally block intentional?

bsrikan: Fixed with recent push. Resolving conversation.

xnikhil08: No that got deleted by mistake it seems. Thanks for letting me know. Updated the code again. Please review now.

xnikhil08:
> A couple things...
> 
> Source values that are truncated (have an ellipsis) don't have an accompanying tooltip when I mouse over them. (But this isn't part of the explicit acceptance criteria):
> 
> dhfprod-4186_ellipsis
> 
> When I log in as a developer user (data-hub-developer role), I can't see the source data documents associated with a mapping that I could see when I logged in as admin:
> 
> Screen Shot 2020-04-07 at 5 41 52 PM

@wooldridge - The tooltip only works only when you hover over the ellipsis and the other issue seems like the permission issue of the user(admin vs developer).

PR #3791, opened 2020-04-05
DHFPROD-4399:See up to 10 mapping errors after step runs

srinathgit:
### Description
See up to 10 mapping errors after step runs

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/one-ui/ui/src/config/bench.config.ts
+++ b/one-ui/ui/src/config/bench.config.ts
@@ -1,0 +1,145 @@
+import {IRolesContextInterface} from "../util/roles";
+
+class MockRolesService implements IRolesContextInterface{
+  public roles: string[] = [];
+
+  public setRoles: (roles: string[]) => void = (roles: string[]) => {
+    this.roles = roles;
+  };
+  public canReadMappings:() => boolean = () => {
+    return false;
+  };
+  public canWriteMappings:() => boolean = () => {
+    return false;
+  };
+  public canReadMatchMerge:() => boolean = () => {
+    return false;
+  };
+  public canWriteMatchMerge:() => boolean = () => {
+    return false;
+  };
+  public canReadLoadData:() => boolean = () => {
+    return true;
+  };
+  public canWriteLoadData:() => boolean = () => {
+    return true;
+  };
+  public canReadEntityModels:() => boolean = () => {
+    return false;
+  };
+  public canWriteEntityModels:() => boolean = () => {
+    return false;
+  };
+  public canReadFlows:() => boolean = () => {
+    return true;
+  };
+  public canWriteFlows:() => boolean = () => {
+    return true;
+  };
+  public canReadStepDefinitions:() => boolean = () => {
+    return false;
+  };
+  public canWriteStepDefinitions:() => boolean = () => {
+    return false;
+  };
+  public hasOperatorRole:() => boolean = () => {
+    return true;
+  };
+}
+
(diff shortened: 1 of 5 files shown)


one-ui/ui/src/pages/Bench.test.tsx
```diff
@@ -1,9 +1,185 @@
...
\ No newline at end of file
+import { render, fireEvent, waitForElement, cleanup } from '@testing-library/react'
+import '@testing-library/jest-dom/extend-expect'
+import axiosMock from 'axios'
+import  Bench from '../pages/Bench';
+import {IRolesContextInterface, RolesContext} from '../util/roles';
+
+jest.mock('axios');
+
+class MockRolesService implements IRolesContextInterface{
```
bsrikan, line 10: MockRolesService is something that can be reused by other "pages" in test, so we should probably create a separate module. There is a __mocks__ folder under src/api to store mock functions

one-ui/ui/src/pages/Bench.test.tsx
```diff
@@ -1,9 +1,185 @@
...
+    };
+    public canWriteStepDefinitions:() => boolean = () => {
+        return false;
+    };
+    public hasOperatorRole:() => boolean = () => {
+        return true;
+    };
+}
+
+//Mock Responses
```
bsrikan, line 57: Mock responses below can also be reused. Can you pls move them into data.config.
Also pls change the flow/step name and uri

bsrikan:
UI unit test looks good and meets AC. However, there is an AC: "Then the expected Job/Batch documents will be created in the jobs database, "
We probably need a backend unit test for this. @srinathgit do you know if we have it in place already

one-ui/ui/src/config/bench.config.ts
```diff
@@ -0,0 +1,145 @@
...
+    "lastAttemptedStep": "1",
+    "lastCompletedStep": "1",
+    "timeStarted": "2020-04-04T01:17:44.918282-07:00",
+    "timeEnded": "2020-04-04T01:17:45.012137-07:00",
+    "stepResponses": {
+      "1": {
+        "flowName": "testFlow",
+        "stepName": "failedIngest",
+        "stepDefinitionName": "default-ingestion",
+        "stepDefinitionType": "ingestion",
```
bsrikan, line 65: Should the stepName and definitions be mapping since its a test for mapping story. We have to create a similar one for ingestion as well and have a separate story for that.

srinathgit:
> UI unit test looks good and meets AC. However, there is an AC: "Then the expected Job/Batch documents will be created in the jobs database, "
> We probably need a backend unit test for this. @srinathgit do you know if we have it in place already

The whole sentence reads like "When I run the mapping step and errors are returned
Then the expected Job/Batch documents will be created in the jobs database" . This story has nothing to do with running mapping step. Those test will be in place for "running the mapping". story. This story makes use of the job/batch documents that were created when mapping story is run

one-ui/ui/src/config/bench.config.ts
srinathgit: It can be either, this is just mock data. The UI code for rendering mapping is not in place yet, so used ingestion

bsrikan: Fair enough.

PR #3793, opened 2020-04-06
DHFPROD-4656: Now merging mocks.json with OpenAPI files

rjrudin:
Requires running the buildSwaggerFile task, which is now documented in the README.

@SameeraPriyathamTadikonda This will impact Jenkins in that the Gradle task will need to be run first. 

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [ ] Reviewed Tests


one-ui/ui/api/swagger/mocks.json
```diff
@@ -1,9 +1,8 @@
 {
   "swagger": "2.0",
   "info": {
-    "description": "This is a sample of a DataHub Swagger document defining endpoints to be mocked",
-    "version": "1.0.0",
-    "title": "Data Hub Curation API",
+    "version": "5.3.0-SNAPSHOT",
```
SameeraPriyathamTadikonda, line 4: We need to include this File in the task "updateVersion" in the main project which will update the versions whenever we release.
@rjrudin  Are we going to release these API's in future?

rjrudin: We don't have any requirements for the version yet - I just figured "5.3.0-SNAPSHOT" was more meaningful than "1.0.0".

rjrudin: I added mocks.json to the task in ./build.gradle, just in case.

SameeraPriyathamTadikonda:
I compared the old one with the changes you made and I see that the "error" tag under "models" is missing. Do we require this?
Screen Shot 2020-04-06 at 11 01 31 AM

rjrudin:
Thanks @SameeraPriyathamTadikonda - looks like json-refs isn't handling the "definitions" object correctly, I'll see what's up.

PR #3799, opened 2020-04-08
DHFPROD-4173: Apply additional settings in flow runner

ryanjdew:
### Description

Certain settings weren't being accounted for when running a data load. Reading in those settings now. I also added the missing target format on the mapping settings form.

https://drive.google.com/open?id=1ymeBD_m5KvOnUBtLo3d_DVFFbEJFErW4

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/flow/impl/FlowRunnerImpl.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/flow/impl/FlowRunnerImpl.java
@@ -657,17 +657,32 @@ public class FlowRunnerImpl implements FlowRunner{
                     if (artifactSettingsJson.hasNonNull("customHook")) {
                         step.setCustomHook(artifactSettingsJson.get("customHook"));
                     }
-                    List<String> stepOptionProperties = Arrays.asList("collections", "permissions");
                     ObjectMapper mapper = new ObjectMapper();
+                    List<String> collections = new ArrayList<>();
+                    // collections needs to be read in a special way so it is treated as a list
                     if (artifactSettingsJson.hasNonNull("collections")) {
                         try {
-                            stepOptions.put("collections", Arrays.asList(mapper.readValue(artifactSettingsJson.get("collections").toString(), String[].class)));
+                            collections.addAll(Arrays.asList(mapper.readValue(artifactSettingsJson.get("collections").toString(), String[].class)));
                         } catch (IOException e) {
                             logger.warn("Unable to parse collections from artifact settings", e);
                         }
                     }
-                    if (artifactSettingsJson.hasNonNull("permissions")) {
-                        stepOptions.put("permissions", artifactSettingsJson.get("permissions").asText());
+                    if (artifactSettingsJson.hasNonNull("additionalCollections")) {
+                        try {
+                            collections.addAll(Arrays.asList(mapper.readValue(artifactSettingsJson.get("additionalCollections").toString(), String[].class)));
+                        } catch (IOException e) {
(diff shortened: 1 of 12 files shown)


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/artifacts/mappings.sjs
```diff
@@ -91,5 +102,6 @@ module.exports = {
     getPermissions,
     getArtifactNode,
     validateArtifact,
-    getArtifactSettingNode
+    getArtifactSettingNode,
+    defaultArtifactSettings
```
rjrudin, line 106: Shouldn't there be a unit test against getArtifactSettings for a mapping to verify that the defaults are returned if there aren't settings yet?

marklogic-data-hub/src/main/java/com/marklogic/hub/flow/impl/FlowRunnerImpl.java
```diff
@@ -657,17 +657,32 @@ protected void applyArtifactSettings(Step step, Map<String, Object> stepConfig)
                     if (artifactSettingsJson.hasNonNull("customHook")) {
                         step.setCustomHook(artifactSettingsJson.get("customHook"));
                     }
-                    List<String> stepOptionProperties = Arrays.asList("collections", "permissions");
                     ObjectMapper mapper = new ObjectMapper();
+                    List<String> collections = new ArrayList<>();
```
rjrudin, line 661: I think verifyThatLoadDataSettingsAreAppliedToStepConfig in FlowRunnerTest should be expanded to cover these fixes.

ryanjdew: done

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/artifacts/mappings.sjs
ryanjdew: Added to the ML unit test to test the retrieval of default settings.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/artifacts/mappings.sjs
```diff
@@ -83,6 +83,17 @@ function getArtifactSettingNode(collectionName, artifactName, artifactVersion) {
   return fn.head(results);
 }
 
+function defaultArtifactSettings(artifactName) {
+    return {
+        artifactName,
+        collections: ['default-mapping'],
```
bsrikan, line 89: @ryanjdew wondering if this should be default-mapping or entity-services-mapping?

ryanjdew: default-mapping is the collection set by entity-services-mapping step definition.

ryanjdew:
Updated the controller tests to expect default settings instead of empty settings.

xnikhil08:
@ryanjdew - Just a suggestion. Not related to this PR, but the getArtifactsSettings axios call in the function getArtifactsSettings() in file activity-settings-dialog.tsx seems to be a duplicate of the similar API call present in the file utils/manageArtifacts-service.tsx(function getSettingsArtifact).
I know we have this discussion already in place to move all the API calls to the api folder in future but not sure when will that happen so until then we can use the one present in the utils folder as that was created to be used commonly across all components.

PR #3803, opened 2020-04-09
DHFPROD-4440: Improving how resetProject works in HubTestBase

rjrudin:
Submitting this as part of DHFPROD-4440, which involves step processors. I started writing a JUnit test for this and realized that our tests can be sped up by not having to load hub artifacts all the time (because clearDatabases no longer needs them) and by not having to generate and deploy entity-based search options, which are never used. 

Main change is that clearDatabases in HubTestBase no longer deletes the hub-core-artifact collection. That keeps all DH core artifacts present. Had to improve some legacy tests that were assuming that the database was empty after calling clear (making assertions on the number of docs in a database leads to brittle tests).

Also made a tweak to LoadUserModulesCommand that both speeds up tests and fixes a long-standing issue when running "mlWatch", which is - when you run mlWatch, DHF generates and deploys entity-based search options over and over and over. That also adds some time to every test, and none of our tests depend on LoadUserModulesCommand deploying entity-based search options.

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/test/java/com/marklogic/hub/AbstractHubTest.java
+++ b/marklogic-data-hub/src/test/java/com/marklogic/hub/AbstractHubTest.java
@@ -1,16 +1,20 @@
 package com.marklogic.hub;
 
-import com.fasterxml.jackson.databind.ObjectMapper;
+import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.extension.ExtendWith;
 import org.springframework.test.context.ContextConfiguration;
 import org.springframework.test.context.junit.jupiter.SpringExtension;
 
 /**
- * Adding this so that a class can subclass HubTestBase without having to define the same two Spring annotations
+ * Adding this so that a class can subclass HubTestBase without hing to define the same two Spring annotations
  * over and over. Plan is to move a lot of classes to under this so they don't duplicate the annotations.
  */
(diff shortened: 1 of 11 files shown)


marklogic-data-hub/src/test/java/com/marklogic/hub/AbstractHubTest.java
```diff
@@ -1,16 +1,20 @@
...
 
-import com.fasterxml.jackson.databind.ObjectMapper;
+import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.extension.ExtendWith;
 import org.springframework.test.context.ContextConfiguration;
 import org.springframework.test.context.junit.jupiter.SpringExtension;
 
 /**
- * Adding this so that a class can subclass HubTestBase without having to define the same two Spring annotations
+ * Adding this so that a class can subclass HubTestBase without hing to define the same two Spring annotations
```
srinathgit, line 9: A small typo here. Other than that everything looks good.

rjrudin: Thanks @srinathgit  - I'll have another PR for 4440 with the real impl in it; I'll fix the typo in that one.

PR #3808, opened 2020-04-10
DHFPROD-4751: Reworked HC to default to connecting to DHS

rjrudin:
The new HubCentral class is now responsible for collecting all configuration and helping configure a new HubConfigSession when a user logs in. This makes it easy to determine whether to use the new defaults for DHS or not. 

In the process, I removed a lot of the stuff pertaining to features that we know are either being removed (installing DHF) or being redesigned (upload/download/reset). 

Also bumped up the bugfix version numbers of our build.gradle file and simplified the set of dependencies. 

The README has been updated to describe how to run HC locally now. 

Finally, did a little more cleanup from the PR for renaming one-ui to HubCentral - I renamed AbstractOneUiTest, and I moved all the curation.controllers under central.controllers.

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/controllers/EnvironmentController.java
```diff
@@ -18,145 +18,94 @@
...
+    Versions versions;
 
     @Autowired
-    private SimpMessagingTemplate template;
+    Environment environment;
 
     private ObjectMapper mapper = new ObjectMapper();
 
-    @RequestMapping(value = "/initialized", method = RequestMethod.GET)
+    @RequestMapping(method = RequestMethod.GET)
```
akshaysonvane, line 54: We still need the request path to be
```suggestion
    @RequestMapping(value = "/api/info", method = RequestMethod.GET)
```

rjrudin: Thanks, forgot that when copying the method over.

rjrudin:
Updated the README for HC to include pointers to overriding properties when running the executable war file.

akshaysonvane:
The PR needs a rebase.

rjrudin:
Rebased

rjrudin:
Added a fix so that mlAppServicesPort defaults to 8010, and 8000 otherwise, as DHS blocks 8000.

Also had to set mlAppServicesUsername and mlAppServicesPassword to the same value as mlUsername and mlPassword.

PR #3812, opened 2020-04-14
DHFPROD-4625: Hub Central Security functional spec

ryanjdew:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/resources/hub-internal-config/security/roles/data-hub-common-writer.json
+++ b/marklogic-data-hub/src/main/resources/hub-internal-config/security/roles/data-hub-common-writer.json
@@ -1,0 +1,34 @@
+{
+  "role-name": "data-hub-common-writer",
+  "description": "Contains a common set of privileges/roles needed for Data Hub 5+ write operations",
+  "role": [
+    "data-hub-common"
+  ],
+  "privilege": [
+    {
+      "privilege-name": "xdmp:document-load",
+      "action": "http://marklogic.com/xdmp/privileges/xdmp-document-load",
+      "kind": "execute"
+    },
+    {
+      "privilege-name": "unprotected-collections",
+      "action": "http://marklogic.com/xdmp/privileges/unprotected-collections",
+      "kind": "execute"
+    },
+    {
+      "privilege-name": "unprotected-uri",
+      "action": "http://marklogic.com/xdmp/privileges/unprotected-uri",
+      "kind": "execute"
+    },
+    {
+      "privilege-name": "ps-user",
+      "action": "http://marklogic.com/xdmp/privileges/ps-user",
+      "kind": "execute"
+    },
+    {
+      "privilege-name": "rest-writer",
+      "action": "http://marklogic.com/xdmp/privileges/rest-writer",
+      "kind": "execute"
+    }
+  ]
+}
--- a/specs/docs/security/DataHubSecurity.md
+++ b/specs/docs/security/DataHubSecurity.md
@@ -1,0 +1,270 @@
+---
+tags: [security]
+---
+
+# Data Hub Security
+
+This document briefly covers best practices to follow in MarkLogic security with links to more indepth documentation. In addition, this document covers the privileges given to each role and the rationale behind the assigned privilege.
+
+## MarkLogic Security Practices
+
+### Checking Execute Privileges
+
+Execute privileges are used to protect functions from being called by users without the proper level of authorization. Functions can be guarded by calling the [`xdmp.securityAssert`](https://docs.marklogic.com/xdmp.securityAssert) function at the beginning of its execution to ensure the user has the proper execution privileges. See [Protecting Your XQuery and JavaScript Code with Execute Privileges](https://docs.marklogic.com/guide/security/execute#id_69547).
+
+#### Example xdmp.securityAssert call
+```javascript
+xdmp.securityAssert("http://marklogic.com/data-hub/function-access", "execute");
+```
+
+### Granting Privileges To Function Via Amps
+
+Execute privileges for a given role should never be broader than needed for the operations a role needs to execute. Amps are a way to grant temporary roles inside a designated function. The function can then institute guardrails to ensure the roles are only used for the narrow intended purpose. See [Temporarily Increasing Privileges with Amps](https://docs.marklogic.com/guide/security/execute#id_54880).
+
+#### Example amp configuration
+```javascript
+{
+  "local-name" : "updateJob",
+  "document-uri" : "/data-hub/5/impl/jobs.sjs",
+  "modules-database" : "%%mlModulesDbName%%",
+  "role" : [ "data-hub-job-internal" ]
+}
+```
+
+In JavaScript, in addition to the amp configuration file, you need to make use of the `module.amp` function to wrap your function on export. See [Amps and the module.amp Function](https://docs.marklogic.com/guide/jsref/functions#id_13020) 
+
+#### Example JavaScript amp using module.amp
+```javascript
+module.exports.updateJob = module.amp(
+  function updateJob(datahub, jobId, status, flow, step, lastCompleted, stepResponse) {
+  //...
+  }
+);
+```
+### When To Grant Privileges To User Roles And When To Use Amps
+
+User roles should be assigned privileges when the privilege is sufficiently restrictive and does not allow arbitrary code execution.
+
+If the privilege would allow arbitrary code execution, that is a sure indication that an amp should be used. For other privileges, consider if the privilege can be used beyond the intended purpose. Where possible use granular privileges (See [Granular Privileges](https://docs.marklogic.com/guide/security/granular)). If the user role can't use granular privileges and the required privilege would be too broad, (1) a new narrow execute privilege should be created, (2) the function that requires the privilege should check for the narrow privilege, and (3) the function should be amped with the broader privilege.
+
+Roles to be used with amps should be suffixed with `-internal` to indicate that the role is used for amps and NOT intended to be assigned to users.
+
+## Data Hub Roles, Privileges, and Authorities
+
+### data-hub-common
+
+Data Hub role containing the common privileges across Data Hub roles.
+
+#### Inherited Roles
+ - data-hub-module-reader
+ - rest-reader
+ - rest-extension-user
+
+#### Privileges
+
+Privilege | Notes
+---------|----------
+ xdmp-eval-in | Required to specify database for REST Client. (Note: This privilege alone doesn't allow eval to be called. The REST API code does the neccesary amping for eval execution)
+ xdbc-eval-in | Same as xdmp-eval-in
+
+### data-hub-common-writer
+
+Data Hub role containing the common privileges across Data Hub writer roles.
+
+#### Inherited Roles
+ - data-hub-common
+
+#### Privileges
+
+Privilege | Notes
+---------|----------
+ ps-user | Needed to record provenance data
+ xdmp-document-load | Needed to load data directly in the database
+ any-uri | Required to write documents to any URI without explicit privileges granted
+ any-collection | Required to write documents to any collection without explicit privileges granted
+ rest-writer | Required to perform write operations via the REST API
+
+### hub-central-user
+
+All Data Hub roles intended for Hub Central users inherit this role.
+
+#### Associate Authorities
+ - canLoginToHubCentral
+
+#### Inherited Roles
+ - data-hub-common
+
+### hub-central-downloader
+
+Role to determine if a user can download the Hub Central project.
+
+#### Associate Authorities
+ - canDownloadUserArtifacts
+
+#### Inherited Roles
+ - data-hub-common
+
+### hub-central-clear-user-data
+
+Role to determine if a user can clear their data through Hub Central.
+
+#### Associate Authorities
+ - canClearUserData
+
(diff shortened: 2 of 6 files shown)


specs/docs/security/DataHubSecurity.md
```diff
@@ -0,0 +1,266 @@
...
+ - canRunStep
+
+#### Inherited Roles
+ - hub-central-user
+ - data-hub-common-writer
+ - data-hub-flow-reader
+ - data-hub-load-data-reader
+ - data-hub-mapping-reader
+ - data-hub-match-merge-reader
+ - data-hub-step-definition-reader
```
bsrikan, line 131: IIRC, a hub-central-step-runner should only be able to see the steps in a flow(Run tile) but not the step configurations(Entities tile). So should not inherit load-data/mapping/match-merge/step-definition reader.

specs/docs/security/DataHubSecurity.md
```diff
@@ -0,0 +1,266 @@
+---
+tags: [security]
+---
+
+# Data Hub Security
```
rjrudin, line 5: This is great information, but I'm wondering if it should go into github yet. I don't believe we have any docs like this in github; currently, I think this information should go into a design spec on our internal Wiki. That's primarily for consistency. I do think there's long-term value in exposing this value to the public, I'm just not at the moment how best to do that - i.e. what's the right mix of info in github vs in our public docs? So for now, I think it's best to put this into a "Hub Central Security" design spec page on the Wiki.

specs/reference/data-services/security.v1.json
```diff
@@ -0,0 +1,56 @@
+{
+  "openapi": "3.0.0",
+  "tags":[{
+    "name": "security"
```
rjrudin, line 4: I know we have a "SecurityService" already (I may have been the one to suggest that name, can't remember), but I'm wondering if we should rename it to "UserService" because "getAuthorities" is specific to a user. I'm wondering what other functions we would add to this... and I'm not sure. So I don't really know a good name, but I lean towards UserService since this is currently specific to the user calling it. Thoughts?

specs/docs/security/DataHubSecurity.md
```diff
@@ -0,0 +1,266 @@
...
+
+Can read and write load data artifacts through Hub Central
+
+#### Associate Authorities
+ - canWriteLoadData
+
+#### Inherited Roles
+ - hub-central-load-reader
+ - data-hub-common-writer
+ - data-hub-load-data-reader
```
bsrikan, line 165: hub-central-load-reader inherits data-hub-load-data-reader already.

specs/docs/security/DataHubSecurity.md
```diff
@@ -0,0 +1,266 @@
...
+
+Can read and write mapping artifacts through Hub Central
+
+#### Associate Authorities
+ - canWriteMapping
+
+#### Inherited Roles
+ - hub-central-mapping-reader
+ - data-hub-common-writer
+ - data-hub-mapping-writer
```
bsrikan, line 188: hub-central-mapping-reader inherits data-hub-mapping-writer

specs/docs/security/DataHubSecurity.md
```diff
@@ -0,0 +1,266 @@
...
+Can read and write match and merge artifacts through Hub Central
+
+#### Associate Authorities
+ - canWriteMatching
+ - canWriteMerge
+
+#### Inherited Roles
+ - hub-central-match-merge-reader
+ - data-hub-common-writer
+ - data-hub-match-merge-writer
```
bsrikan, line 213: hub-central-match-merge-reader inherits data-hub-match-merge-writer

specs/docs/security/DataHubSecurity.md
```diff
@@ -0,0 +1,266 @@
...
+Can read and write custom artifacts through Hub Central
+
+#### Associate Authorities
+ - canWriteCustom
+
+#### Inherited Roles
+ - hub-central-mapping-reader
+ - data-hub-common-writer
+ - data-hub-step-definition-writer
+ - data-hub-modules-writer
```
bsrikan, line 237: Should hub-central-custom-writer inherit hub-central-custom-reader, so we can do away with inheriting data-hub-step-definition-writer. 
Typo in modules-writer, should be data-hub-module-writer. Should data-hub-common-writer inherit data-hub-module-writer?

specs/docs/security/DataHubSecurity.md
```diff
@@ -0,0 +1,266 @@
...
+#### Inherited Roles
+ - hub-central-user
+ - hub-central-step-writer
+ - hub-central-load-writer
+ - hub-central-mapping-writer
+ - hub-central-match-merge-writer
+ - hub-central-custom-writer
+ - hub-central-downloader
+
+### hub-central-custom-writer
```
bsrikan, line 252: Typo, should be hub-central-operator

specs/docs/security/DataHubSecurity.md
```diff
@@ -0,0 +1,266 @@
...
+ - hub-central-mapping-reader
+ - data-hub-common-writer
+ - data-hub-step-definition-writer
+ - data-hub-modules-writer
+
+### hub-central-developer
+
+Can create any type of step and run steps through Hub Central
+
+#### Inherited Roles
```
bsrikan, line 243: Should a HC developer inherit HC operator like its data-hub-* counterpart?

specs/docs/security/DataHubSecurity.md
ryanjdew: That is a difference between the `hub-central-`prefixed roles and the `data-hub-` prefixed roles. The `hub-central` roles dictate what can be done through the UI/middle-tier and the `data-hub` roles are primarily document-level permissions. In order for a `hub-central-step-runner` to be able to run a step they need the document-level read permissions to the various artifacts.

specs/docs/security/DataHubSecurity.md
ryanjdew: good catch. that should be `data-hub-load-data-writer`

specs/docs/security/DataHubSecurity.md
ryanjdew: yes

specs/reference/data-services/security.v1.json
ryanjdew: Makes sense. Leaves the door open for Hub Central to do some limited security management, etc. in the future.

specs/docs/security/DataHubSecurity.md
ryanjdew: Why would hub-central-mapping-reader inherit data-hub-mapping-writer?

specs/docs/security/DataHubSecurity.md
```diff
@@ -0,0 +1,266 @@
...
+
+#### Associate Authorities
+ - canDownloadProject
+
+#### Inherited Roles
+ - data-hub-common
+
+### hub-central-clear-data
+
+Role to determine if a user can clear data through Hub Central.
```
bsrikan, line 108: clear data from Staging, Final and Jobs only? Will this role be allowed to delete provenance?
Are these the artifacts this role is allowed to delete: /entities,/flows(user and default?), /step-definitions(user and default?), /loadData, /mappings, /matching, activity-settings, /jobs, /batch, all the ingested/mapped/mastered documents including triples/graph.

specs/docs/security/DataHubSecurity.md
bsrikan: My bad. I read data-hub-mapping-writer as reader

marklogic-data-hub/src/main/resources/hub-internal-config/security/roles/data-hub-common-writer.json
```diff
@@ -0,0 +1,30 @@
+{
+  "role-name": "data-hub-common-writer",
+  "description": "Contains a common set of privileges/roles needed for Data Hub 5+ write operations",
+  "role": [
+    "data-hub-common",
+    "rest-writer"
```
bsrikan, line 6: do we need to inherit data-hub-module-writer here?

ryanjdew: no. that isn't a common write permission that we want to grant to any user what has write privileges

specs/docs/security/DataHubSecurity.md
```diff
@@ -0,0 +1,267 @@
...
+
+## MarkLogic Security Practices
+
+### Checking Execute Privileges
+
+Execute privileges are used to protect functions from being called by users without the proper level of authorization. Functions can be guarded by calling the [`xdmp.securityAssert`](https://docs.marklogic.com/xdmp.securityAssert) function at the beginning of its execution to ensure the user has the proper execution privileges. See [Protecting Your XQuery and JavaScript Code with Execute Privileges](https://docs.marklogic.com/guide/security/execute#id_69547).
+
+#### Example xdmp.securityAssert call
+```javascript
+xdmp.securityAssert("http://marklogic.com/data-hub/function-access", "execute");
```
rjrudin, line 17: I'm all for guarding every DS endpoint with this. On one project, we wrote a little tool to parse the xdmp.securityAssert call on the first line that it was found in an endpoint module and then constructed an HTML table output of every endpoint + the privilege that it required. Very handy, may be worth doing in the future.

specs/docs/security/DataHubSecurity.md
```diff
@@ -0,0 +1,267 @@
...
+Roles to be used with amps should be suffixed with `-internal` to indicate that the role is used for amps and NOT intended to be assigned to users.
+
+## Data Hub Roles and Privileges
+
+### data-hub-common
+
+Data Hub role containing the common privileges across Data Hub roles.
+
+#### Inherited Roles
+ - data-hub-module-reader
```
rjrudin, line 59: I think we need rest-extension-user here as well. I ran into some surprising issues that were new to me, where if any module imported by a REST extension doesn't have at least one common read permission with every other module, an error will be thrown. So we guarantee that rest-extension-user is always present, and thus I think data-hub-common should have it.

specs/docs/security/DataHubSecurity.md
```diff
@@ -0,0 +1,267 @@
...
+ xdmp-eval-in | Required to specify database for REST Client. (Note: This privilege alone doesn't allow eval to be called. The REST API code does the neccesary amping for eval execution)
+ xdbc-eval-in | Same as xdmp-eval-in
+
+### data-hub-common-writer
+
+Data Hub role containing the common privileges across Data Hub writer roles.
+
+#### Inherited Roles
+ - data-hub-common
+ - rest-writer
```
rjrudin, line 75: We don't want to inherit this role, because it includes default permissions that work against document-level security. Need to inherit the rest-writer privilege instead.

specs/docs/security/DataHubSecurity.md
```diff
@@ -0,0 +1,267 @@
...
+
+#### Inherited Roles
+ - data-hub-common
+
+### hub-central-downloader
+
+Role to determine if a user can download the Hub Central project.
+
+#### Associate Authorities
+ - canDownloadProject
```
rjrudin, line 101: I think "canDownloadUserArtifacts" may be better here, since the entire project cannot be downloaded, just what we internally call "user artifacts".

rjrudin: Also, I don't know yet if this role should have all the read-access roles for user artifacts, or if we'd used an amp in the DS endpoint to grab all of them. The latter is certainly possible - I did a little prototype of using xdmp.zip. So a DS endpoint that requires a certain privilege and then uses an amp to construct the zip is possible. But I don't know if that's worth calling out here - it would seem a little early to do that.

specs/docs/security/DataHubSecurity.md
rjrudin: So I said "clear data" yesterday, and now I'm regretting it. :) 

The problem with "clear data" is that I think the expectation from clearing a database is that it's empty. But that won't be the case - the DH and user artifacts will still be present. We expect to do a clear first, and then add those artifacts back, but the net result is not a "clear". 

I think "clear user data" would work. I'd prefer if we had a noun here for consistency with other roles, but I don't have a good suggestion. Think we'll have to accept some verbs here, which is consistent with ML roles - many are nouns, some are verbs.

rjrudin: Also, I think "clear user data" to me implies that it applies to staging/final/jobs.

specs/docs/security/DataHubSecurity.md
ryanjdew: Changed to use the privilege.

specs/docs/security/DataHubSecurity.md
ryanjdew: added the rest-extension-user role here

ryanjdew:
@rjrudin Addressed the issues you've mentioned and I've created the following internal functional spec wiki: https://wiki.marklogic.com/display/ENGINEERING/Hub+Central+Security+Design+Specification. It naturally does quite a bit of referencing to other places we've documented things. 

I think it is important to have information about our security approach as part of the repo, rather than an internal wiki so there is a reference for potential contributors.

PR #3813, opened 2020-04-14
DHFPROD-3992:Backend work for running the mapping

srinathgit:
### Description
1. Backend work for running the mapping
2. Renamed artifact names (mappings,stepDefintions,flows -> mapping,stepDefinition,flow)
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/LoadUserArtifactsCommand.java
```diff
@@ -142,15 +142,23 @@ protected Modules findModulesWithResolvedBaseDir(String resolvedBaseDir) {
                         return modules;
                     }
                 };
-                Files.walkFileTree(artifactPath, new SimpleFileVisitor<Path>() {
-                    @Override
-                    public FileVisitResult preVisitDirectory(Path dir, BasicFileAttributes attrs) throws IOException {
-                        if (dir.toFile().isDirectory()) {
-                            executeWalkWithDataService(dir, modulesFinder, artifactService, typeInfo);
+                if("stepDefinition".equals(artifactType)){
```
rjrudin, line 145: Why does stepDefinition needs it own code here?

marklogic-data-hub/src/main/resources/hub-internal-config/triggers/ml-dh-json-new-mapping-create.json
```diff
@@ -0,0 +1,39 @@
+{
+  "name": "ml-dh-json-new-mapping-create",
+  "description": "MarkLogic Data Hub JSON mapping creation trigger",
+  "event": {
+    "data-event": {
+      "collection-scope": {
+        "uri": "http://marklogic.com/data-hub/mapping-artifact"
```
rjrudin, line 7: Why aren't we still using http://marklogic.com/data-hub/mappings ?

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mapping/default/lib.sjs
```diff
@@ -15,7 +15,12 @@ function getModel(targetEntity, version = '0.0.1') {
 }
 
 function getMapping(mappingName) {
-  return fn.head(cts.search(cts.andQuery([cts.collectionQuery('http://marklogic.com/data-hub/mappings'), cts.jsonPropertyValueQuery('name', mappingName, ['case-insensitive'])]), ["unfiltered", cts.indexOrder(cts.uriReference(), "descending")]));
+  //Look in new collection, if artifact is not present fallback to the old collection
+  let mapping =  fn.head(cts.search(cts.andQuery([cts.collectionQuery('http://marklogic.com/data-hub/mapping-artifact'), cts.jsonPropertyValueQuery('name', mappingName, ['case-insensitive'])]), ["unfiltered", cts.indexOrder(cts.uriReference(), "descending")]));
```
rjrudin, line 19: Assuming we do need a new collection (that's a separate question), should at least avoid the duplication of the query construction and make a function that accepts a collection name and returns the cts.andQuery.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/endpoints/collectorLib.sjs
```diff
@@ -39,11 +43,25 @@ class CollectorLib {
       return fn.normalizeSpace(`${sourceQuery}`);
     }
 
+    //Get sourceQuery from mapping artifact if present
+    if(combinedOptions.mapping) {
```
rjrudin, line 47: Whoa - why are we supporting another location for the sourceQuery? Why would it be on the mapping?

marklogic-data-hub-central/src/test/java/com/marklogic/hub/curation/controllers/FlowControllerTest.java
```diff
@@ -177,6 +178,39 @@ void runFlow() {
         JsonNode job = jobsController.getJob(resp.getJobId());
         String jobStatus = job.get("jobStatus").asText();
         assertEquals("finished", jobStatus, "Job status should be 'finished' once threads complete; job doc: " + job);
+
+
+        addStagingDoc("input/mapInput.json", "/input/mapInput.json", "default-ingestion");
+
+        //Equivalent to adding a step from GUI
+        controller.createStep("testFlow" , 2, "{\n" +
```
rjrudin, line 186: If the purpose of this test is to run a flow and not to test the creation of a step, I think it's better just to add the step to the test flow file.

marklogic-data-hub-central/src/test/java/com/marklogic/hub/curation/controllers/FlowControllerTest.java
```diff
@@ -166,7 +167,7 @@ void getFlow() throws IOException {
     }
 
     @Test
-    void runFlow() {
+    void runFlow() throws IOException {
```
rjrudin, line 170: Can you switch this to use the reference project instead of "run-flow-test"? I'd like for as many tests as possible to use the reference project. Adding a flow there with an entity-services-mapping would be a very useful addition.

rjrudin:
One other comment - I'm hoping to get https://github.com/marklogic/marklogic-data-hub/pull/3808 merged in soon today, which includes repackaging all of the controllers under central.controllers. I think it'd be good to wait for that and rebase on it.

marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/LoadUserArtifactsCommand.java
srinathgit: All artifacts are present in base dir except stepdef, hence calling File.walk in case of stepdef,
also we don't want to reload the previously loaded mappings(using REST) using DS.

marklogic-data-hub-central/src/test/java/com/marklogic/hub/curation/controllers/FlowControllerTest.java
srinathgit: Used reference-project for the tests

marklogic-data-hub/src/main/java/com/marklogic/hub/step/impl/MappingStepDefinitionImpl.java
```diff
@@ -42,14 +40,7 @@ public MappingStepDefinitionImpl(String name) {
...
         options.put("sourceDatabase", HubConfig.DEFAULT_STAGING_NAME);
         options.put("targetDatabase", HubConfig.DEFAULT_FINAL_NAME);
-        Versions versions = ApplicationContextReference.getBean(Versions.class);
-        if (versions == null || versions.isVersionCompatibleWithES()){
-            setModulePath("/data-hub/5/builtins/steps/mapping/entity-services/main.sjs");
-        }
-        else {
-            setModulePath("/data-hub/5/builtins/steps/mapping/default/main.sjs");
-        }
-
```
srinathgit, line 52: Spoke with @sbayatpur , she said DHF 5.3.0 will be supported on ML servers  >=9.0-11 for 9.x and >=10.0-2.1 for 10.x. Hence removed the check.

marklogic-data-hub/src/main/resources/hub-internal-config/triggers/ml-dh-json-new-mapping-create.json
srinathgit: Reverted to  http://marklogic.com/data-hub/mappings

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/data-services/artifacts/suite-setup.xqy
```diff
@@ -5,4 +5,18 @@ xquery version "1.0-ml";
 import module namespace hub-test = "http://marklogic.com/data-hub/test" at "/test/data-hub-test-helper.xqy";
 import module namespace test = "http://marklogic.com/test" at "/test/test-helper.xqy";
 
+xdmp:invoke-function(function() {
```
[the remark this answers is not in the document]
srinathgit: It loads it to final db, I want entities to be load to staging db . I used hub-test:load-artifacts for loading it to final.

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/data-services/artifacts/suite-teardown.xqy
```diff
@@ -3,4 +3,11 @@ xquery version "1.0-ml";
 import module namespace hub-test = "http://marklogic.com/data-hub/test" at "/test/data-hub-test-helper.xqy";
 import module namespace test = "http://marklogic.com/test" at "/test/test-helper.xqy";
 
+xdmp:invoke-function(function() {
```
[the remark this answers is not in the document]
srinathgit: I have used hub-test:delete-artifacts which deletes from final db, this code is to delete from staging db

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/data-services/artifacts/mappingTest.sjs
```diff
@@ -1,11 +1,64 @@
...
+        }
+      },
+      "stepDefinitionName": "entity-services-mapping",
+      "stepDefinitionType": "MAPPING"
+    },
+    "3": {
+      "name": "mapjson",
+      "description": "",
+      "options": {
+        "additionalCollections": [],
```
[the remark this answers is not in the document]
srinathgit: 1. flow.addSettingsToFlow() method adds settings to the step options (for all artifacts except 'loadData') only if it looks like this
    ```  "options": {
        "<artifact_type>": {
          "name": "<artifact_name>"
        }
      } ```
Else, it should just return the step as is (old versions of step). This is an old version of step. So, the test was to ensure it is returned as is.
2.  I agree , I will add test to ensure that 'additionalCollections' is concatenated to 'collections' array

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/flow.sjs
```diff
@@ -102,22 +103,53 @@ class Flow {
...
         cachedFlows[name] = null;
+        return null;
       }
+      flowObject = this.addSettingsToFlow(name, flowObject);
+      cachedFlows[name] = flowObject;
     }
     return cachedFlows[name];
   }
 
+  addSettingsToFlow(flowName, flowNode){
```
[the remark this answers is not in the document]
srinathgit: 1. WriteStepRunner requires the "fully populated" document, I believe steps using QueryStepRunner don't
2. I initially changed core.sjs/getArtifact() to return "fully populated" flow doc  (FlowManager.getFlow() and all clients use it).  It worked fine but whenever a flow is edited (add/remove step) , the flows that gets written is the fully populated document (since creating a new step, gets a flow (fully populated doc) and adds steps to it and persists it. So, I moved this method from core.sjs 's getArtifact to impl/flows.sjs 's getFlow() method

marklogic-data-hub-central/src/test/resources/test-projects/reference-project/entities/Item.entity.json
```diff
@@ -0,0 +1,30 @@
+{
```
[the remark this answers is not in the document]
srinathgit: I will rework the test to use Customer or Order

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/controllers/FlowController.java
```diff
@@ -379,100 +370,13 @@ public StepModel createStep(String flowName, Integer stepOrder, String stepId, S
...
         if (existingStep != null && existingStep.isEqual(step)) {
-            return transformStepToWebModel(existingStep);
+            return existingStep;
         }
 
         flowManager.saveFlow(flow);
-        return transformStepToWebModel(step);
-    }
-
-    private List<StepModel> getStepsToWebModel(String flowName) {
```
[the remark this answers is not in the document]
srinathgit: If the step is going to look something like the one below, I don't think it's needed. We allowed creating custom steps in QS, hence StepModel contained "modulePath"  (which was not in Step) which was used to create a step definition and a step in the backend. Since we won't do that, I believe it wouldn't be necessary. We can include it if there is a need
```
    "1" : {
      "name" : "mapItem",
      "options" : {
        "mapping" : {
          "name" : "mapItem"
        }
      },
      "stepDefinitionName" : "entity-services-mapping",
      "stepDefinitionType" : "MAPPING"
    }
```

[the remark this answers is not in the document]
rjrudin: Sounds good!

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/data-services/artifacts/suite-setup.xqy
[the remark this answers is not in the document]
rjrudin: Okay - I think it'd be better to add a new function to data-hub-test-helper - e.g. "load-artifacts-into-staging" - but that can be done later, don't need to do that in this PR. 

I'm missing why the staging db matters though for the test - are there assertions on making sure something exists in both the staging and final db?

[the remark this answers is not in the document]
srinathgit: So, the artifact service writes mappings to both staging and final db. In order to write the mappings in both the db, the entities should be present(this is necessitated by the fact that we now write mappings to same collection (http://marklogic.com/data-hub/mappings) as before and it has associate triggers which will fail in generating xslt if entity is not found in staging db.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/flow.sjs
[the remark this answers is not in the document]
srinathgit: Added a new endpoint that gets fullFlow() to be used by FlowRunner

srinathgit:
@rjrudin ,
I have added an endpoint getFullFlow() that returns the full flow document for the FlowRunnerImpl to use

marklogic-data-hub/build.gradle
```diff
@@ -200,6 +200,10 @@ task generateEntitySearchInterface(type: com.marklogic.client.tools.gradle.Endpo
     serviceDeclarationFile = dataServicesPath + "/entitySearch/service.json"
 }
 
+task generateFlowInterface(type: com.marklogic.client.tools.gradle.EndpointProxiesGenTask, group: dataServicesGroup) {
```
rjrudin, line 203: You don't need this, the task is generated dynamically based on your service.json file. In fact, you can delete the two task definitions above it.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/artifacts/core.sjs
```diff
@@ -341,6 +353,82 @@ function getEntityTitles() {
     return cts.search(cts.collectionQuery("http://marklogic.com/entity-services/models")).toArray().map(e => e.xpath("//info//title"));
 }
 
+function getFullFlow(flowName, artifactVersion = 'latest') {
```
rjrudin, line 356: This looks good, but shouldn't this go into artifacts/flows.sjs since it's flow-specific? 

Also @ryanjdew - do we really get any benefit from caching artifacts? That's only good for the duration of the transaction, right? I doubt we need any caching in this function.

srinathgit: getFullFlow() has to fetch 'loadData', 'mapping', 'matching' artifacts and settings to build full flow. core.sjs provides a generic interface regardless of artifact type to access them(getArtifact, getArtifactSettings), otherwise if this is in flow.sjs, I will have to include each and every module('loadData'.sjs', 'mapping.sjs' etc) and have multiple 'if' conditions for each type of artifact. It seems better to add it here

marklogic-data-hub/build.gradle
srinathgit: Removed all 3 tasks

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/artifacts/core.sjs
srinathgit: Removed caching for fullFlow()

srinathgit:
@rjrudin ,
I have added a 'TODO' to add 'processors' to Step in getFullFlow() once we know where it will go into (artifact or settings)

srinathgit:
Fixed couple of failing tests with latest commiy

srinathgit:
Will merge to the feature branch and then will create PR against develop

srinathgit:
Rebased with latest develop, merging to feature branch now

PR #3830, opened 2020-04-16
DHFPROD-3650: Delete saved personal queries feature

rahulvudutala:
### Description
Delete saved personal queries feature
Refactoring Search API to match with SavedQueries feature

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/controllers/EntitySearchControllerTest.java
+++ b/marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/controllers/EntitySearchControllerTest.java
@@ -18,7 +18,7 @@ public class EntitySearchControllerTest extends AbstractMvcTest {
     private ObjectNode savedQueryResponse;
 
     @Test
-    void createThenUpdateSavedQuery() throws Exception {
+    void testCRUDOnSavedQuery() throws Exception {
         String json = "{\n" +
             "  \"savedQuery\": {\n" +
             "    \"name\": \"some-query\",\n" +
@@ -33,6 +33,7 @@ public class EntitySearchControllerTest extends AbstractMvcTest {
             "  }\n" +
             "}";
 
+        // testing save query document
         postJson(SAVED_QUERIES_PATH, json)
             .andExpect(status().isCreated())
             .andDo(result -> {
@@ -41,6 +42,7 @@ public class EntitySearchControllerTest extends AbstractMvcTest {
                     "can e.g. display a link to the saved query");
             });
 
+        // testing update savedQuery document
         ObjectNode query = (ObjectNode) savedQueryResponse.get("savedQuery");
         query.put("description", "Updated description");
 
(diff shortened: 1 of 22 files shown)


marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/controllers/EntitySearchControllerTest.java
```diff
@@ -110,4 +110,38 @@ void getSavedQuery() throws Exception {
                     assertEquals(savedQueryResponse.get("savedQuery").get("id").asText(), response.get("savedQuery").get("id").asText());
                 });
     }
+
+    @Test
+    void deleteSavedQuery() throws Exception {
```
rjrudin, line 115: I'm of the opinion that it's perfectly fine just to toss the test of deletion onto an existing test method, like the existing createThenUpdate test. Fewer test methods means less time setting up tests, which means our test suite is faster.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/entitySearch/deleteSavedQuery.sjs
```diff
@@ -0,0 +1,12 @@
+'use strict';
+declareUpdate();
+
+var id;
+
+let savedQuery = cts.search(cts.andQuery([cts.collectionQuery("http://marklogic.com/data-hub/saved-query"),
```
rjrudin, line 6: For deleting a document, since we know what the URI should be, just do:

    if (cts.exists(cts.documentQuery(uri)) then delete the uri

We only want to use cts.search when we want the document back.

rahulvudutala: Since only the user who saved the query should be able to delete the query, I am adding the following query:
cts.exists(cts.andQuery([cts.documentQuery(docUri), cts.jsonPropertyValueQuery("owner", xdmp.getCurrentUser())]))

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/entitySearch/deleteSavedQuery.sjs
```diff
@@ -0,0 +1,10 @@
+'use strict';
```
rjrudin, line 1: Pinging @ryanjdew on this - I am wondering what is needed for security here. Just occurred to me that we presumably at least need a check on the owner of the saved query being the same as the MarkLogic user that's connecting, right? Otherwise one user could delete anyone's saved query. 

@rahulvudutala Is there a role required for creating/deleting saved queries? If so, I believe we want an xdmp.securityAssert check here. I defer to @ryanjdew on whether we want to start requiring this now or wait a couple more stories, as @ryanjdew is getting a policy into place where we secure all of our DS endpoints with an xdmp.securityAssert check.

rahulvudutala: @rjrudin 
1) I am checking in on the user here, 
let canDelete = cts.exists(cts.andQuery([cts.documentQuery(documentUri), cts.jsonPropertyValueQuery("owner", xdmp.getCurrentUser())]));
2) Yes. We have data-hub-saved-query-writer role with update permissions on the savedQuery document
Yes.

rjrudin: Oops sorry, I missed that part of the query. 

@ryanjdew I defer to you if you want to get an xdmp.securityAssert in place here, or table it for now. We'll of course need to add those to a lot of existing endpoints at some point.

PR #3832, opened 2020-04-17
DHFPROD-4440: Improvements to EntityServicesAlignmentTest

rjrudin:
This wasn't failing for me locally, but made a few improvements to it in the hopes it'll pass on Jenkins.

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


SameeraPriyathamTadikonda:
@rjrudin This is failing only on 10.0-3,10.0-4 and 10.0-nightly 3node cluster environments.

This test passed on 10.0-4 single node environment and 9.0-12, 9.0-nightly 3node cluster environments.
Just in case if any of the above info helps.

rjrudin:
I'm wondering if waitForTasksToFinish is stopping too soon... likely just need an artificial wait time in there. I'll add that, or I'll see if I can make that method verify that the task server is empty on every host.

PR #3837, opened 2020-04-19
DHFPROD-4720: Core project now publishes a test fixture for reusability

rjrudin:
Intent is to remove duplication of test plumbing between core and central. Summary of changes:

- Added new AbstractHubTest as a base class for HubTestBase and AbstractHubCentralTest. It has logic for resetting the project before a test runs, and for load project files. This is accomplished via the new "testFixturesApi" - "java-test-fixture" is a new feature in Gradle 5.6 for this exact use case, where test fixtures in core can be reused in the other DH projects. 

- LoadHubArtifactsCommand no longer has a "force" property to it; artifacts are always loaded. Modified it to ensure it never loads two documents with the same URI at the same time, which happens in some scenarios in our tests. 

- Modified LoadHubModuledCommand to default to a batch size of 10. In some scenarios when bootstrapping the test app, the config.sjs/config.xqy modules are being loaded twice, which causes a conflicting updates error. This prevents that. 

- Replaced temporal-document-write.sjs with WriteDocumentsTest. 

- Moved jaeger-config dependency from core into QuickStart, as it's only used there. 

- Removed spring-integration dependency from core; it's not needed anywhere.

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [ ] Added Tests
  

- ##### Reviewer:

- [ ] Reviewed Tests


rjrudin:
@SameeraPriyathamTadikonda I still have Jenkinsfile modified in this one, but I first want to see if all the tests pass. If so, I'll update it to ensure that logging works. I want the Gradle/web tests to run last because if any of those tests are messing around with the test environment, I don't want them breaking any of the tests in hub-central that I know are not messing with the test environment.

PR #3838, opened 2020-04-19
DHFPROD-4720: Core project now publishes a test fixture for reusability

rjrudin:
Intent is to remove duplication of test plumbing between core and central. Summary of changes:

- Added new AbstractHubTest as a base class for HubTestBase and AbstractHubCentralTest. It has logic for resetting the project before a test runs, and for load project files. This is accomplished via the new "testFixturesApi" - "java-test-fixture" is a new feature in Gradle 5.6 for this exact use case, where test fixtures in core can be reused in the other DH projects. 

- LoadHubArtifactsCommand no longer has a "force" property to it; artifacts are always loaded. Modified it to ensure it never loads two documents with the same URI at the same time, which happens in some scenarios in our tests. 

- Modified LoadHubModuledCommand to default to a batch size of 10. In some scenarios when bootstrapping the test app, the config.sjs/config.xqy modules are being loaded twice, which causes a conflicting updates error. This prevents that. 

- The Installer program always install the test application. I changed this primarily for local convenience, as I often need to install the test app after e.g. adding a role, and the fact that the test app Installer doesn't run if it thinks DH is already installed is not helpful. 

- Replaced temporal-document-write.sjs with WriteDocumentsTest. 

- Moved jaeger-config dependency from core into QuickStart, as it's only used there. 

- Removed spring-integration dependency from core; it's not needed anywhere. 
8ed615

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


rjrudin:
Tests passed on this. I believe it's also possible for the contents of src/test/resources in core to be reused in central so we don't need to e.g. duplicate the reference model project. I'm holding off on that for now though.

Another thing I'm holding off on is standardizing all projects on JUnit 5 so we're no longer carrying JUnit 4 as a dependency. Different projects are also using different versions of JUnit 5. I'll tackle that in a future PR.

marklogic-data-hub/src/test/java/com/marklogic/hub/deploy/commands/BuildAssetFileLoaderTest.java
```diff
@@ -19,12 +19,10 @@ public void withCustomBatchSize() {
...
 
         TestAssetFileLoader loader = new TestAssetFileLoader(null);
-        command.prepareAssetFileLoader(loader, context);
-        assertNull(loader.batchSize, "The batch size should still be null because it is null on the AppConfig object");
 
-        appConfig.setModulesLoaderBatchSize(10);
+        appConfig.setModulesLoaderBatchSize(7);
         command.prepareAssetFileLoader(loader, context);
-        assertEquals(10, (int) loader.batchSize, "The batch size should now be 10; this gives the user a way to load hub " +
+        assertEquals(7, (int) loader.batchSize, "The batch size should now be 10; this gives the user a way to load hub " +
```
akshaysonvane, line 25: Not required now but we should remove the count from the message.

PR #3845, opened 2020-04-20
DHFPROD-4624: functional spec APIs for mastering in Hub Central

ryanjdew:
### Description

Internal functional spec can be found here: https://wiki.marklogic.com/display/ENGINEERING/Hub+Central+Mastering+Design+Specification

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/specs/models/Flow.schema.json
+++ b/specs/models/Flow.schema.json
@@ -40,178 +40,191 @@
       "additionalProperties": false,
       "patternProperties": {
         "^[0-9]+$": {
-          "type": "object",
-          "properties": {
-            "name": {
-              "type": "string",
-              "description": "Name of step, though the 'step number' is how a step is referred to when running a flow"
-            },
-            "description": {
-              "type": "string",
-              "description": "Optional description fo the step",
-            },
-            "stepDefinitionName": {
-              "type": "string"
-            },
-            "stepDefinitionType": {
-              "type": "string"
-            },
-            "batchSize": {
-              "type": "number",
-              "description": "If set, overrides the batchSize defined at the flow level and in the step definition"
-            },
-            "threadCount": {
-              "type": "number",
-              "description": "If set, overrides the threadCount defined at the flow level and in the step definition"
-            },
-            "retryLimit": {
-              "type": "number",
-              "description": "Unused, has no effect if set"
-            },
-            "customHook": {
-              "type": "object",
+          "oneOf": [
+            {
               "properties": {
-                "module": {
+                "name": {
+                  "type": "string",
+                  "description": "Name of step, though the 'step number' is how a step is referred to when running a flow"
+                },
+                "description": {
+                  "type": "string",
+                  "description": "Optional description of the step"
+                },
+                "stepDefinitionName": {
                   "type": "string"
                 },
-                "parameters": {
-                  "type": "object"
-                },
-                "user": {
+                "stepDefinitionType": {
                   "type": "string"
                 },
-                "runBefore": {
-                  "type": "boolean"
-                }
-              }
-            },
-            "processors": {
-              "type": "array",
-              "items": {
-                "type": "object",
-                "properties": {
-                  "path": {
-                    "type": "string",
-                    "description": "Path to a module in the modules database that will be invoked via xdmp.invoke"
-                  },
-                  "when": {
-                    "type": "string",
-                    "description": "When the processor should be invoked. Only 'beforeContentPersisted' is supported."
-                  },
-                  "vars": {
+                "batchSize": {
+                  "type": "number",
+                  "description": "If set, overrides the batchSize defined at the flow level and in the step definition"
+                },
+                "threadCount": {
+                  "type": "number",
+                  "description": "If set, overrides the threadCount defined at the flow level and in the step definition"
+                },
+                "customHook": {
+                  "type": "object",
+                  "properties": {
+                    "module": {
+                      "type": "string"
+                    },
+                    "parameters": {
+                      "type": "object"
+                    },
+                    "user": {
+                      "type": "string"
+                    },
+                    "runBefore": {
+                      "type": "boolean"
+                    }
+                  }
+                },
+                "processors": {
+                  "type": "array",
+                  "items": {
                     "type": "object",
-                    "description": "Any properties defined in this object are passed to the invoked module"
+                    "properties": {
+                      "path": {
+                        "type": "string",
+                        "description": "Path to a module in the modules database that will be invoked via xdmp.invoke"
+                      },
+                      "when": {
+                        "type": "string",
+                        "description": "When the processor should be invoked. Only 'beforeContentPersisted' is supported."
+                      },
+                      "vars": {
+                        "type": "object",
+                        "description": "Any properties defined in this object are passed to the invoked module"
+                      }
+                    }
+                  }
+                },
+                "options": {
+                  "type": "object",
+                  "properties": {
+                    "sourceQuery": {
+                      "type": "string",
+                      "description": "Defines the items to be processed by the step; must be a cts.query or cts.uris statement if sourceQueryIsScript is false"
+                    },
+                    "sourceQueryIsScript": {
+                      "type": "boolean",
+                      "description": "Added in 5.3.0; if true, then sourceQuery can be any JavaScript statement that can be passed into xdmp.eval"
+                    },
+                    "constrainSourceQueryToJob": {
+                      "type": "boolean",
+                      "description": "If true, the query is applied to the documents that were created or modified in the same job that executes the step"
+                    },
+                    "provenanceGranularityLevel": {
+                      "type": "string",
+                      "description": "The granularity of the provenance tracking information: coarse (default) to store document-level provenance information only, fine to store document-level and property-level provenance information, or off to disable provenance tracking in future job runs. Applies only to mapping, matching, merging, mastering, and custom steps."
+                    },
+                    "stepUpdate": {
+                      "type": "boolean",
+                      "description": "If true, custom modules can make changes directly to records in the database"
+                    },
+                    "acceptsBatch": {
+                      "type": "boolean",
+                      "description": "If true, the step module is invoked once with all records in the batch passed to it"
+                    },
+                    "collections": {
+                      "type": "array",
+                      "items": {
+                        "type": "string"
+                      }
+                    },
+                    "additionalCollections": {
+                      "type": "array",
+                      "items": {
+                        "type": "string"
+                      }
+                    },
+                    "permissions": {
+                      "type": "string",
+                      "description": "Comma-delimited string of role,capability,role,capability,etc"
+                    },
+                    "outputFormat": {
+                      "type": "string"
+                    },
+                    "sourceDatabase": {
+                      "type": "string"
+                    },
+                    "targetDatabase": {
+                      "type": "string"
+                    },
+                    "targetEntity": {
+                      "type": "string",
+                      "description": "Name of the entity type associated with the output of the step"
+                    },
+                    "validateEntity": {
+                      "type": "boolean",
+                      "description": "Applicable to mapping steps only"
+                    },
+                    "headers": {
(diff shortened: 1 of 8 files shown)


specs/models/ArtifactSettings.v1.json
```diff
@@ -0,0 +1,157 @@
...
+      ]
+    },
+    "targetDatabase": {
+      "type": "string"
+    },
+    "sourceDatabase": {
+      "type": "string"
+    },
+    "additionalCollections": {
+      "type": "array",
```
srinathgit, line 48: I feel like we don't need "additionalCollections" and can just have "collections". We save default artifact settings with "collections" set when someone creates an artifact. When users update the settings from UI, "collections" gets reset (as collections is not part of the payload or UI property) .

specs/models/Flow.schema.json
```diff
@@ -48,7 +51,7 @@
             },
             "description": {
               "type": "string",
-              "description": "Optional description fo the step",
+              "description": "Optional description fo the step"
```
rjrudin, line 54: Can you fix my terrible spelling of "of" while you're at it? :)

specs/models/MasterCollections.v1.json
```diff
@@ -0,0 +1,52 @@
+{
+  "title": "MasteringCollections.v1",
```
rjrudin, line 2: I think the file and title should have the same name - "MasteringCollections" seems appropriate.

specs/models/ArtifactSettings.v1.json
```diff
@@ -0,0 +1,157 @@
+{
+  "title": "ArtifactSettings.v1",
+  "type": "object",
+  "description": "Settings for an artifact",
+  "properties": {
+    "customHook": {
```
rjrudin, line 6: I think these properties make it clear that an "Artifact", as defined here, is really a "Step". 

Our existing (e.g. LoadHubArtifactsCommand and LoadUserArtifactsCommand) code also refers to flows and entity models as artifacts, but those concepts do not support properties like customHook and targetDatabase/sourceDatabase. 

Of course, if this is just capturing the as-is, that's fine, and this is correct. But I think it's critical that we rework the concept of the "artifact service" into a "step service", because we're really describing steps. 

"Artifact" will then continue to have its current meaning of "Config files created by users", which will consist of flows, entity models, steps, step definitions, and mappings (with mappings being separate from mapping steps because of the existing code that expects things to be that way).

specs/models/MasterCollections.v1.json
wooldridge: The UX designs are using "Match" and "Merge" in the UI, so I recommended leaving off the "ing" for related files in the code. Should we do the same for Mastering, so "Master"? I see "Load" and "Map" and "Curate" in the UI as well. Shorter is better.

rjrudin: Existing DH core uses mapping/matching/merging/mastering.

specs/models/ArtifactSettings.v1.json
ryanjdew: We need a way to differentiate between default collections that can't be removed in the UI and user-defined collections that can. Perhaps just have "collections" and "defaultCollections" can be added in on reading the step out?

specs/models/Flow.schema.json
```diff
@@ -40,178 +43,195 @@
...
-                  "type": "string",
-                  "description": "Name of the entity type associated with the output of the step"
+                "threadCount": {
+                  "type": "number",
+                  "description": "If set, overrides the threadCount defined at the flow level and in the step definition"
                 },
-                "validateEntity": {
-                  "type": "boolean",
-                  "description": "Applicable to mapping steps only"
+                "retryLimit": {
```
rjrudin, line 71: I included this at first to document it, but now I'm thinking we should remove it.

specs/models/Step.v1.json
```diff
@@ -0,0 +1,181 @@
...
+        "provenanceGranularityLevel": {
+          "type": "string",
+          "description": "The granularity of the provenance tracking information: coarse (default) to store document-level provenance information only, fine to store document-level and property-level provenance information, or off to disable provenance tracking in future job runs. Applies only to mapping, matching, merging, mastering, and custom steps.",
+          "enum": [
+            "off",
+            "coarse",
+            "fine"
+          ]
+        },
+        "stepUpdate": {
```
rjrudin, line 29: If the intent is to define a new Step schema that uses "options" strictly for things that make sense to override at runtime... you know, I'm wondering if "options" should be purely a runtime thing in the first place? i.e. put sourceQuery directly under Step, and a Step has no understanding of "options" in the first place. Then, some chunk of code can look at the options defined at runtime when a flow is run, and it'll use e.g. Object.assign to overwrite what's in the step definition. So e.g.:

```
options = {}; // provided by user at runtime
flow = getFlow();
stepId = flow.steps["1"].stepId;
step = getStep(stepId); // loads a step document from the database
step = applyOptionsToStep(step);
flow.steps["1"] = step;
```

That makes the data model much easier to design and understand, because now we don't need to think about "Which of these things would a user want to override at runtime?" If needed, the "applyOptionsToStep" function could prohibit certain things from being overridden.

Thoughts on dumping "options" completely from a Step?

specs/models/Step.v1.json
```diff
@@ -0,0 +1,181 @@
...
+          "type": "boolean",
+          "description": "If true, custom modules can make changes directly to records in the database"
+        },
+        "acceptsBatch": {
+          "type": "boolean",
+          "description": "If true, the step module is invoked once with all records in the batch passed to it"
+        },
+        "collections": {
```
rjrudin, line 37: Taking a page out of mlcp's book here - how about "outputCollections" to make it clear that these are applied to documents outputted by the step? Likewise for "outputPermissions", and I'd do "outputAdditionalCollections" as well. I remember being confused at first about what the scope of "collections" is - does that affect the query for documents to process, or are they applied to documents that will be written, or is it something else? "outputCollections" works well for mlcp, I think we should copy it here.

(Note that the function that adapts the new Step schema to what the existing Flow schema expects would then rename this to "collections").

specs/models/Step.v1.json
```diff
@@ -0,0 +1,181 @@
...
+            "xml"
+          ]
+        },
+        "sourceDatabase": {
+          "type": "string"
+        },
+        "targetDatabase": {
+          "type": "string"
+        },
+        "targetEntity": {
```
rjrudin, line 66: I think we should at a minimum say "targetEntityTypeId" here, as "entityTypeId" is our agreed-upon name for a unique identifier for an entity type. We can have code that accepts a name in here as well. 

Also, I'm proposing "inputDatabase" and "outputDatabase" to be consistent not just with mlcp, but also with the evolving concepts of a Step having an Input and an Output. Gorbet pointed out how the word "source" is a bit loaded because it has been typically used to refer to a source of data outside of ML, like some external system. So we want to move away from "sourceQuery" - though I wouldn't rename that yet. I don't know, if this is just a draft, I think we can stay with sourceDatabase/targetDatabase, but I think we'll need to get Gorbet and likely PAT involved because we're really defining critical concepts here.

specs/models/Step.v1.json
bsrikan: "options" for a Step(as they appear in a flow) really are "settings" based on current implementation right?

```
"steps": {
  "1": {
    "name": "dhsLoadData", 
    "description": "", 
    "options": {
        "sourceQuery": null, 
        "collections": ["default-ingestion"], 
        "loadData": {
            "name": "dhsLoadData"
        }, 
        "outputFormat": "json", 
        "targetDatabase": "data-hub-STAGING"
    }, 
```

rjrudin: "Settings" is an arbitrary concept that appears in the UI but could change at any point. We don't want to model our data based how form dialogs are constructed in the UI.

ryanjdew: @rjrudin  The only scenario where I'm not certain about dropping options all together is the scenario of custom steps. Do we just pass in the entire step into the step function instead of options in that case? If not, I'm not certain how we'd determine what qualifies as "options" for the custom step.

bsrikan: Each Step might need a different set of options. Command line options would apply it commonly across all steps right?

rjrudin: The intent of options is to allow for some parts of a step to be overridden at runtime. But "some" is not clearly defined, and there are clear instances where what's in and what's not in "options" doesn't make sense:

- batchSize and threadCount are not in options, but those clearly would be overridden at runtime
- fileLocations/input-file-path is not in options, but is likely to be overridden at runtime
- Examples of properties in options that are extremely unlikely to be overridden at runtime: outputFormat, acceptsBatch, stepUpdate, mapping

My point is that "options" as defined in a step are meaningless today. We should discard it - it's an unnecessary level of nesting in a step that provides no meaning, as demonstrated above. 

Runtime options make sense though. This should be an object that can overwrite any part of a step, though the function that applies options to a step can always prohibit certain things from being overridden as needed. For example, we likely don't want to allow for name/stepDefinitionName/stepDefinitionType to be overridden.

specs/models/Flow.schema.json
ryanjdew: removed

specs/models/Step.v1.json
ryanjdew: removed "options" from the "Step.v1" model. Left "options" as is in the Flow schema.

specs/models/Step.v1.json
ryanjdew: I've updated this to use "outputCollections" and "defaultOutputCollections".

specs/models/Step.v1.json
ryanjdew: changed to "targetEntityTypeId" and made the changes to input/output database

specs/models/Step.v1.json
```diff
@@ -0,0 +1,142 @@
...
+      "type": "string"
+    },
+    "outputFormat": {
+      "type": "string",
+      "enum": [
+        "json",
+        "xml"
+      ]
+    },
+    "permissions": {
```
rjrudin, line 85: Little tweak - I think "outputPermissions" here, for consistency with mlcp and to make it explicit how these will be applied.

specs/models/StepSettings.v1.json
```diff
@@ -0,0 +1,157 @@
...
+    },
+    "provenanceGranularityLevel": {
+      "type": "string",
+      "enum": [
+        "off",
+        "coarse",
+        "fine"
+      ]
+    },
+    "permissions": {
```
rjrudin, line 31: I think since the UI is under development, we should use the same terms here. It's fine that we have to translate the new terms to the existing Flow schema, but for the UI, it'll never be easier than now to change it to use these new terms.

specs/models/Step.v1.json
ryanjdew: updated to use outputPermissions

specs/models/Step.v1.json
```diff
@@ -0,0 +1,142 @@
...
+    },
+    "sourceQueryIsScript": {
+      "type": "boolean",
+      "description": "Added in 5.3.0; if true, then sourceQuery can be any JavaScript statement that can be passed into xdmp.eval"
+    },
+    "sourceQuery": {
+      "type": "string",
+      "description": "Defines the items to be processed by the step; must be a cts.query or cts.uris statement if sourceQueryIsScript is false"
+    },
+    "targetEntityTypeId": {
```
rjrudin, line 134: I'm realizing now that we really have a lot of interchangeability between input/source and output/target/destination in DH today. I think this should likely be "outputEntityTypeId", and "sourceQuery" should be "inputQuery", but I think we can hold off on those changes for now because I'm expecting there to be new abstractions we add soon. So I think these are fine for now, even though we still have consistency with input vs source and output vs target/destination.

PR #3847, opened 2020-04-20
DHFPROD-4662: Adding new item to checklist

SameeraPriyathamTadikonda:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [ ] Reviewed Tests

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -17,4 +17,4 @@
 - ##### Reviewer:
 
 - [ ] Reviewed Tests
-
+- [ ] Has good error handling and messaging


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -17,4 +17,4 @@
 - ##### Reviewer:
 
 - [ ] Reviewed Tests
-
+- [ ] Has good error handling and messaging
```
rjrudin, line 20: I don't think we can add this yet, particularly if "messaging" means "logging". 

The user story itself should call out how error scenarios should be handled. Thus, if the tests are sufficient for a story, then that means it is addressing the error scenarios. But if those scenarios are not defined in the story, then it's very subjective as to what "good error handling" means.

SameeraPriyathamTadikonda:
Closing this now. Will re-open once we define what good error handling means.

PR #3850, opened 2020-04-21
DHFPROD-4802: Adding authorities to saved-queries artifact

rahulvudutala:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/entitySearch/saveSavedQuery.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/entitySearch/saveSavedQuery.sjs
@@ -18,10 +18,17 @@
 declareUpdate();
 
 const ds = require("/data-hub/5/data-services/ds-utils.sjs");
+const Security = require("/data-hub/5/impl/security.sjs");
 
 var saveQuery;
 var userCollections = ["http://marklogic.com/data-hub/saved-query"];
 var queryDocument = JSON.parse(saveQuery);
+
+let authorized = new Security().getRolesAndAuthorities().authorities.includes("canManageSavedQuery");
+
+if(!authorized) {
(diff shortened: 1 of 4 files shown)


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/entitySearch/saveSavedQuery.sjs
```diff
@@ -18,11 +18,18 @@
 declareUpdate();
 
 const ds = require("/data-hub/5/data-services/ds-utils.sjs");
+const Security = require("/data-hub/5/impl/security.sjs");
 
 var saveQuery;
 var userCollections = ["http://marklogic.com/data-hub/saved-query"];
 var queryDocument = JSON.parse(saveQuery);
 
+let authorized = new Security().getRolesAndAuthorities().authorities.includes("canManageSavedQuery");
```
rjrudin, line 27: Pinging @ryanjdew - this is doing what xdmp.securityAssert does, but that needs a privilege. I'll get a PR together that has an example of this.

ryanjdew:
@rjrudin I think we're okay not having the securityAssert at the moment. With the new security library @rahulvudutala added, we'll be able to add a function in the future to handle securityAssert if the current user doesn't have the given authority (or privilege associated with that authority).

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/entitySearch/saveSavedQuery.sjs
rjrudin, line 27: One small note, and no need to change (I'm going to approve) - for variables that aren't going to change, prefer "const" over "let". It makes for more readable code too, as I can see this and know it's never going to change.

PR #3856, opened 2020-04-21
DHFPROD-4720: Can now download user configuration files

rjrudin:
Added an "all-artifacts" project in core to support this. 

Added our first use of "Secured" in HC, along with a pattern for how to verify that a user can or cannot access a Secured controller method. 

Also removed a number of unused classes that aren't needed now that we're not checking if the user can install DH - ForbiddenException(LoginFailureHandler can assume that any failure is due to the user being unauthorized); ProjectDirectoryException (no more project); UIDeployListener (no more install); DataHubProjectUtils (no more upload). Also removed BadRequestException - CustomExceptionHandler wasn't doing anything with it, and it was being improperly thrown by MapSearchManager instead of the generic DataHubException (it shouldn't have been thrown there, since it was extending AuthenticationException).

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


rjrudin:
@rahulvudutala Take a look at the privilege I've added and then how xdmp.securityAssert is used. Based on a conversation with @ryanjdew , I think this is the direction we want to go. While it does require adding a privilege per thing-user-can-do, it allows for very granular security and for reuse of xdmp.securityAssert. Ultimately, I think we'll want an xdmp.securityAssert as the first line in every DS endpoint.

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/managers/MapSearchManager.java
```diff
@@ -117,7 +117,7 @@ public StringHandle search(MapSearchQuery mapSearchQuery) {
 
     public JsonNode sjsSearch(SJSSearchQuery SJSSearchQuery) {
         if (SJSSearchQuery.sourceQuery == null && SJSSearchQuery.database == null) {
-            throw new BadRequestException("Query requires database or sourceQuery");
+            throw new DataHubException("Query requires database or sourceQuery");
```
ryanjdew, line 120: How are we differentiating from a 400 Bad Request and a 500 internal server error in this case?

rjrudin: The BadRequestException wasn't actually doing anything - it wasn't capturing a 400, and it wasn't being handled by CustomExceptionHandler. It was also extending Spring Security's AuthenticationException, but it's not an authentication exception. So the class effectively wasn't doing anything, so I removed it and switched this over to the generic DataHubException.  That exception isn't really do anything either - CustomExceptionHandler handles it in its generic exception handler - but it's better than an exception that wasn't doing what its name implies.

PR #3862, opened 2020-04-22
DHFPROD-4226: Export rows as csv

akshaysonvane:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/WebMvcConfig.java
+++ b/marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/WebMvcConfig.java
@@ -19,6 +19,8 @@ package com.marklogic.hub.central;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.core.io.ClassPathResource;
 import org.springframework.core.io.Resource;
+import org.springframework.scheduling.concurrent.ThreadPoolTaskExecutor;
+import org.springframework.web.servlet.config.annotation.AsyncSupportConfigurer;
 import org.springframework.web.servlet.config.annotation.ResourceHandlerRegistry;
 import org.springframework.web.servlet.config.annotation.WebMvcConfigurer;
 import org.springframework.web.servlet.resource.PathResourceResolver;
@@ -43,5 +45,29 @@ public class WebMvcConfig implements WebMvcConfigurer {
                         : new ClassPathResource("static/index.html");
                 }
             });
+    }
+
+    /**
+     * We need this since we are using a StreamingResponseBody for asynchronous request processing in the csv
+     * export feature. StreamingResponseBody by default uses a SimpleAsyncTaskExecutor which does not reuse threads,
+     * as a result, its advisable to provide our own TaskExecutor.
+     * Also, this allows us to configure the Default Timeout which at this point is 10 minutes.
+     * @param configurer
+     */
+    @Override
+    public void configureAsyncSupport(AsyncSupportConfigurer configurer) {
+        final ThreadPoolTaskExecutor taskExecutor = new ThreadPoolTaskExecutor();
+
+        // Setting core pool size to 10 since we aren't expecting more than 5~10 concurrent users.
+        // If all threads are busy and a request comes in then that request is queued.
+        taskExecutor.setCorePoolSize(10);
+
+        // Setting the max pool size, also to 10 makes this a fixed sized thread pool.
+        taskExecutor.setMaxPoolSize(10);
+        taskExecutor.initialize();
+
+        // Setting the default timeout value to 10 minutes since that aligns with the request timeout value of MarkLogic server
+        configurer.setDefaultTimeout(600000)
+            .setTaskExecutor(taskExecutor);
     }
 }
--- a/marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/managers/EntitySearchManager.java
+++ b/marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/managers/EntitySearchManager.java
@@ -16,6 +16,9 @@
  */
 package com.marklogic.hub.central.managers;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import com.marklogic.client.DatabaseClient;
 import com.marklogic.client.ForbiddenUserException;
 import com.marklogic.client.MarkLogicServerException;
@@ -23,6 +26,7 @@ import com.marklogic.client.ResourceNotFoundException;
 import com.marklogic.client.document.GenericDocumentManager;
 import com.marklogic.client.io.DocumentMetadataHandle;
 import com.marklogic.client.io.Format;
+import com.marklogic.client.io.ReaderHandle;
 import com.marklogic.client.io.StringHandle;
 import com.marklogic.client.io.marker.StructureWriteHandle;
 import com.marklogic.client.query.QueryManager;
@@ -30,21 +34,29 @@ import com.marklogic.client.query.RawCombinedQueryDefinition;
 import com.marklogic.client.query.StructuredQueryBuilder;
 import com.marklogic.client.query.StructuredQueryBuilder.Operator;
 import com.marklogic.client.query.StructuredQueryDefinition;
+import com.marklogic.client.row.RowManager;
 import com.marklogic.hub.HubClient;
 import com.marklogic.hub.central.exceptions.DataHubException;
 import com.marklogic.hub.central.models.DocSearchQueryInfo;
 import com.marklogic.hub.central.models.Document;
 import com.marklogic.hub.central.models.SearchQuery;
+import com.marklogic.hub.dataservices.EntitySearchService;
+import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.text.StringEscapeUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.util.CollectionUtils;
 
+import javax.servlet.http.HttpServletResponse;
+import java.io.IOException;
+import java.io.OutputStream;
 import java.time.LocalDate;
 import java.time.ZoneId;
+import java.time.ZonedDateTime;
 import java.time.format.DateTimeFormatter;
 import java.util.*;
+import java.util.function.Function;
 
 public class EntitySearchManager {
 
@@ -63,12 +75,16 @@ public class EntitySearchManager {
     private static final String SEARCH_HEAD = "<search xmlns=\"http://marklogic.com/appservices/search\">\n";
     private static final String SEARCH_TAIL = "</search>";
     private static final Set<String> METADATA_FIELD_NAME = new HashSet<>(Arrays.asList("datahubCreatedOn"));
+    private static final String CSV_CONTENT_TYPE = "text/csv";
+    private static final String CSV_FILE_EXTENSION = ".csv";
+    private static final String SEPARATOR = "_";
+    private static final String FILE_PREFIX = "DH_Export_";
 
     private static final Logger logger = LoggerFactory.getLogger(EntitySearchManager.class);
 
 
-
     public static String QUERY_OPTIONS = "exp-final-entity-options";
+    private static Function<String, String> replaceHyphenWithUnderscore = str -> str.replaceAll("-", "_");
     private DatabaseClient finalDatabaseClient;
     private ModelManager modelManager;
 
@@ -93,7 +109,7 @@ public class EntitySearchManager {
         }
         catch (MarkLogicServerException e) {
             // If there are no entityModels to search, then we expect an error because no search options will exist
-            if(searchQuery.getQuery().getEntityTypeIds().isEmpty() || modelManager.getModels().size() == 0) {
+            if (searchQuery.getQuery().getEntityTypeIds().isEmpty() || modelManager.getModels().size() == 0) {
                 logger.warn("No entityTypes present to perform search");
                 return new StringHandle("");
             }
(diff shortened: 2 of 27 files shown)


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/entitySearch/opticPlanExportService.api
```diff
@@ -0,0 +1,37 @@
+{
+    "functionName": "getOpticPlan",
```
rjrudin, line 2: For consistency, the functionName should be the same  as the filename, minus ".api".

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/entitySearch/opticPlanExportService.sjs
```diff
@@ -0,0 +1,85 @@
...
+  ' <xsl:template match="*:options">\n' +
+  '  <xsl:copy>\n' +
+  '   <xsl:apply-templates select="node()|@*" />\n' +
+  returnFlags +
+  '  </xsl:copy>\n' +
+  ' </xsl:template>\n' +
+  '\n' +
+  '</xsl:stylesheet>'));
+
+const entityOptionsFileName = '/Default/data-hub-FINAL/rest-api/options/exp-final-entity-options.xml';
```
rjrudin, line 54: This is not likely to work in DHS, as we don't know the group name. Today, in a typical DHS setup, there are 4 other groups, and Default is not used at all. 

Instead of depending on an options document to exist, what if we generate this based on the entity name? hub-entities.xqy provides a nice function for generating this.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/entitySearch/opticPlanExportService.sjs
```diff
@@ -0,0 +1,85 @@
...
+  See the License for the specific language governing permissions and
+  limitations under the License.
+*/
+'use strict';
+
+const op = require('/MarkLogic/optic');
+const search = require('/MarkLogic/appservices/search/search');
+const options = require("/data-hub/5/impl/hub-utils/invoke-search-options.sjs");
+
+const returnFlags = '<return-aggregates xmlns="http://marklogic.com/appservices/search">false</return-aggregates>\n' +
```
rjrudin, line 22: Pinging @ryanjdew - should this be in XQuery instead of SJS? I ask that because we're dealing with a lot of XML here, and that's of course more difficult to do in SJS vs XQuery.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/entitySearch/opticPlanExportService.sjs
```diff
@@ -0,0 +1,85 @@
...
+var schemaName;
+var limit;
+var structQuery;
+var searchQText;
+var columns = xdmp.getRequestField("columns");
+
+structQuery = fn.head(xdmp.unquote(structQuery)).root;
+searchQText = searchQText || "";
+
+const oldOptions = options.getSearchOptions(entityOptionsFileName);
```
rjrudin, line 66: Given that we need an amp plus a separate module, which then has to invoke a function in the modules database for a document whose URI will be dependent on the groups - I think we should just generate these on the fly. Check out hub-entities.xqy.

marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/controllers/EntitySearchControllerTest.java
```diff
@@ -57,30 +77,122 @@ void testCRUDOnSavedQuery() throws Exception {
...
+        String projectFolder = "test-projects/export-project";
+        installProjectInFolder(projectFolder);
+        entityManager.deployQueryOptions();
+
+        DocumentMetadataHandle metadata = new DocumentMetadataHandle()
+            .withCollections("Order")
+            .withPermission("data-hub-operator", DocumentMetadataHandle.Capability.READ, DocumentMetadataHandle.Capability.UPDATE);
+        GenericDocumentManager finalDocMgr = getHubConfig().newFinalClient().newDocumentManager();
+        DocumentWriteSet writeSet = finalDocMgr.newWriteSet();
+        File inputDir = new ClassPathResource(projectFolder + "/input").getFile();
```
rjrudin, line 190: I've worked on projects before where we create a lot of test files that are stored in version control, and it can quickly become difficult to manage. I think it's far better in the long term to write a simple program to generate documents in memory and write them to ML. It's fewer files in version control, and it's easier to change. 

Right now, it doesn't look like the test cares what's in the documents, though I think it should - shouldn't it be verifying that the correct value is present in each column in each row? 

For that reason, I would only insert 2 test documents (generally, there's zero, there's 1, and then there's 2 or more). And I'd only request maybe 3 columns. Now you only need to verify 6 values. And since you have 2 test documents, you can have a separate test that passes in a limit of 1, and you can verify you only get 1 row back. 

With that in mind, I'd design a bare minimum test document - no headers, no triples, no attachments. And if you're only requesting 3 properties, I'd populate 4 total properties - that allows you to verify that when you ask for 3 of them, you don't get the 4th. Then, make a simple method in here that allows you to insert a document with 3 known values. 

Also, let's use the entity-reference-model project. The Customer entity has three atomic properties so far - customerId, name, and customerSince. Feel free to add a fourth atomic property for your test. That avoids having to create a new example project, and the more tests we have using the same sample project, the easier it is to understand the tests.

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/WebMvcConfig.java
```diff
@@ -44,4 +46,14 @@ protected Resource getResource(String resourcePath, Resource location)
                 }
             });
     }
+
+    @Override
+    public void configureAsyncSupport(AsyncSupportConfigurer configurer) {
```
rjrudin, line 51: Is there something under src/main/java that needs this? The only reference I saw to "async" is in EntitySearchControllerTest.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/entitySearch/opticPlanExportService.sjs
ryanjdew: @rjrudin I'm fine with having this in JavaScript. I recommended to @akshaysonvane that we call XSLT from JavaScript. @akshaysonvane, for readability, can we adjust this to use template literals for multiline string? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/entitySearch/opticPlanExportService.sjs
ryanjdew: Another option would be to use http://docs.marklogic.com/xdmp.groupName which doesn't require any special permissions.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/entitySearch/opticPlanExportService.api
akshaysonvane: Got it.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/entitySearch/opticPlanExportService.sjs
akshaysonvane: Sure, @ryanjdew.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/entitySearch/opticPlanExportService.sjs
akshaysonvane: My understanding was that a user is free to edit the search options file.
Do we still allow a user to edit the options file for HC? 

Another alternative here could be to perform the search operation from the java middle-tier. Though it would increase a request, it would make sure that we use the correct search options file.

marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/controllers/EntitySearchControllerTest.java
akshaysonvane: Makes sense. I'll update the tests.

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/WebMvcConfig.java
akshaysonvane: We need this since we are using a StreamingResponseBody for asynchronous request processing.
StreamingResponseBody by default uses a SimpleAsyncTaskExecutor which does not reuse threads, as a result, its advisable to provide our own TaskExecutor. 
Also, this allows us to configure the Default Timeout which at this point is 10 minutes.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/entitySearch/opticPlanExportService.sjs
akshaysonvane: Right, I completely forgot about the groups. We need to rework this.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/entitySearch/opticPlanExportService.sjs
rjrudin: If the user does edit the options file, it'll get overwritten when an entity model is modified. 

I like the idea of having the middle tier fetch the options and then pass them to the DS endpoint. That way, the endpoint doesn't have to worry about what group to use when finding the options.

akshaysonvane:
Passing the query options file from the middle tier so I removed the amp.
Updated tests to generate docs in memory and the tests now check the content as well.
Also updated the test infrastructure to provide an option to load search query options.

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/WebMvcConfig.java
rjrudin: I think this method needs to make that clear. Spring is great at making it easy to setup useful, reusable stuff, but the few lines of code required to do so often beg the question of - "Why is this being configured?" 

The method name is usually a good way to answer that question - e.g. "configureAsyncProcessingForStreamingResponses". Each setting needs a comment as well - why 10 threads? Why 10 minutes? Why did we choose those numbers? 3 months from now, someone will look at this code and wonder why these settings were chosen.

I also think it's worth explicitly identifying the feature this is for - i.e. exporting entities. That may expand in the future - i.e. when we support exporting other results.

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/managers/EntitySearchManager.java
```diff
@@ -100,7 +112,7 @@ public StringHandle search(SearchQuery searchQuery) {
         }
         catch (MarkLogicServerException e) {
             // If there are no entityModels to search, then we expect an error because no search options will exist
-            if(searchQuery.getQuery().getEntityTypeIds().isEmpty() || modelManager.getModels().size() == 0) {
+            if (searchQuery.getQuery().getEntityTypeIds().isEmpty() || modelManager.getModels().size() == 0) {
```
rjrudin, line 115: Looking at this, it does seem like there's a chance for a null pointer. How about "searchQuery.hasEntityTypeIds()"? That results in less work for the controller, and that method can be coded to ensure an NPE can't occur.

marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/controllers/EntitySearchControllerTest.java
```diff
@@ -57,30 +71,198 @@ void testCRUDOnSavedQuery() throws Exception {
...
+    @Test
+    void testRowExport() throws Exception {
+        installReferenceProject(true);
+
+        Customer customer1 = new Customer(1, "Jane", 123456789, "2012-05-16");
+        Customer customer2 = new Customer(2, "John", 987654321, "2013-06-16");
+        createCustomer(customer1);
+        createCustomer(customer2);
+
+        waitForTasksToFinish();
```
rjrudin, line 111: Why is this here? I don't think any post-commit triggers fire when entity instances are saved?

marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/controllers/EntitySearchControllerTest.java
```diff
@@ -57,30 +71,198 @@ void testCRUDOnSavedQuery() throws Exception {
...
+            "    }\n" +
+            "}";
+
+        int limit = 2;
+        int totalColumns = 3;
+
+        // Export using query document
+        postWithParams(EXPORT_PATH, getRequestParams(limit, json))
+            .andExpect(request().asyncStarted())
+            .andDo(MvcResult::getAsyncResult)
```
rjrudin, line 148: Nice - is this verifying that the async support is configured correctly then?

marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/controllers/EntitySearchControllerTest.java
```diff
@@ -57,30 +71,198 @@ void testCRUDOnSavedQuery() throws Exception {
...
+
+    private void assertRowsAndColumns(int limit, int totalColumns, String response) {
+        int headerRow = 1;
+        int totalRows = limit + headerRow;
+
+        assertEquals(totalRows, response.chars().filter(i -> i == '\n').count() + 1);
+        assertEquals(totalColumns, response.substring(0, response.indexOf("\n")).chars().filter(i -> i == ',').count() + 1);
+    }
+
+    private void createCustomer(Customer customer) {
```
rjrudin, line 198: This is reusable stuff that is worth adding to the ReferenceModelProject class. 

The current "createCustomer" method in there is for "raw" customers. I think it's worth renaming that to "createRawCustomer", and then this can become "createCustomerInstance", which makes it clear we're creating entity instances here.

marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/controllers/EntitySearchControllerTest.java
```diff
@@ -57,30 +71,198 @@ void testCRUDOnSavedQuery() throws Exception {
...
+    private MultiValueMap<String, String> getRequestParams(int limit, String json) {
+        MultiValueMap<String, String> params = new LinkedMultiValueMap<>();
+        params.add("fileType", "csv");
+        params.add("limit", String.valueOf(limit));
+        params.add("queryDocument", json);
+
+        return params;
+    }
+
+    static class Customer {
```
rjrudin, line 250: This can move to the testFixture source set in core DH too. Make it a regular public class though, with private fields and getters/setters, as we can use Jackson to generate JSON for it automatically, which will save some effort. We'll likely add more to this class so to make writing tests easier as we go.

marklogic-data-hub-central/src/test/resources/test-projects/reference-project/entities/Customer.entity.json
```diff
@@ -10,6 +10,9 @@
         "name"
       ],
       "properties": {
+        "customerId": {
```
rjrudin, line 13: Just an FYI, we have the reference project in both core and central - I'll get that resolved soon, as central can definitely reuse the one in core.

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/managers/EntitySearchManager.java
```diff
@@ -278,6 +290,126 @@ protected String buildSearchOptions(String query, SearchQuery searchQuery) {
...
+    }
+
+    public void exportRows(JsonNode queryDocument, Long limit, OutputStream out) {
+        QueryManager queryMgr = finalDatabaseClient.newQueryManager();
+        SearchQuery searchQuery = transformToSearchQuery(queryDocument);
+        StructuredQueryDefinition structuredQueryDefinition = buildQuery(queryMgr, searchQuery);
+
+        String structQuery = structuredQueryDefinition.serialize();
+        String searchQText = searchQuery.getQuery().getSearchText();
+        String queryOptions = finalDatabaseClient.newServerConfigManager().newQueryOptionsManager().readOptionsAs(QUERY_OPTIONS, Format.XML, String.class);
```
rjrudin, line 315: I think we need to catch an exception here in case the options exist. They should exist - but if they don't, the user is going to get an error they likely won't understand and I'll probably get pinged with a message saying that Gradle is broken. 

So use a try/catch here to catch an exception if the options don't exist, and rethrow the error with a message of e.g. "Could not find search options: (name); system is not configured correctly, please contact a system administrator".

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/entitySearch/getOpticPlan.sjs
```diff
@@ -0,0 +1,79 @@
...
+   </xsl:template>
+</xsl:stylesheet>`));
+
+
+var viewName;
+var schemaName;
+var limit;
+var structQuery;
+var searchQText;
+var queryOptions;
```
rjrudin, line 54: I think you can define the DS API with queryOptions as type = jsonDocument, and then pass (hopefully) a JacksonHandle into it in the Java code. Same goes for the structQuery - if these are JSON objects, let's have a richer interface than using "String".

Also, "searchText" is widely used as the parameter name for a user's search text, so let's use that instead of "searchQText". 

One other nit - generally, I'm opposed to abbreviating words unless there's a really, really good reason. So let's say "structuredQuery" instead of "structQuery".

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/WebMvcConfig.java
akshaysonvane: Makes sense. I'll add the code comments.

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/managers/EntitySearchManager.java
akshaysonvane: Since the models have been designed to initialize these objects in the constructor it won't throw an NPE.

marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/controllers/EntitySearchControllerTest.java
akshaysonvane: Right, this needs to go. I missed that while refactoring the code.

marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/controllers/EntitySearchControllerTest.java
akshaysonvane: Yes, `request().asyncStarted()` checks for the async support. 
`getAsyncResult()` is needed to wait for the async result to return.

marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/controllers/EntitySearchControllerTest.java
akshaysonvane: To the ReferenceModelProject this goes then.

marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/controllers/EntitySearchControllerTest.java
akshaysonvane: Will do.

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/managers/EntitySearchManager.java
akshaysonvane: Good catch. I'll catch the exception.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/entitySearch/getOpticPlan.sjs
akshaysonvane: Both of them are XML docs which I get as a String from the relevant methods in the middle tier.
Converting them to an Object would mean a lot of unnecessary serialization/deserialization at both ends of the wire.

marklogic-data-hub-central/src/test/resources/test-projects/reference-project/entities/Customer.entity.json
akshaysonvane: Updated the hub central code and tests to use the reference model project from the core.

I haven't deleted the project since there's a test that does use the flows and other configs from the HC's reference project and I didn't want to copy all of that baggage over to the reference model project in the core.

akshaysonvane:
Added check for the authority in PR https://github.com/marklogic/marklogic-data-hub/pull/3870 to the middle-tier endpoints.

akshaysonvane:
Resolved conflicts.

PR #3864, opened 2020-04-23
DHFPROD-4757: Removing HC dependency on the filesystem

rjrudin:
Added a BaseController that is the one and only one bean that accesses HubConfigSession. Every controller extends that and is now abstracted from how a HubConfig is obtained.

Removed every controller method that both did not have a test and that I could not find any way to access it via the UI code. Doing this to bump up code coverage.

Renamed methods in subclasses of AbstractArtifactController and changed the tests to call those methods; again, to improve code coverage. 

Changes to core are minimal - removed file-writing from ArtifactManagerImpl; added a null-check to HubProjectImpl so it doesn't assume a token exists (this doesn't impact HC app code, but HC test code - which does need to initialize project - will otherwise break without this check); and made a couple methods in EntityManagerImpl and FlowManagerImpl visible to HC (the methods have useful logic that don't depend on a filesystem).

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/HubCentral.java
```diff
@@ -42,23 +37,19 @@
...
-
-        HubConfigImpl tempHubConfig = new HubConfigImpl(hubProject, environment);
-        tempHubConfig.initHubProject();
-        logger.info("Initialized project at: " + getProjectDirectory());
-
         if (useLocalDefaults) {
             logger.info("Local defaults of digest authentication and no SSL will be used when connecting to MarkLogic");
         }
 
+        temporaryLoadDataDirectory = new File("build/data-sets");
```
ryanjdew, line 44: This is called "temporaryLoadDataDirectory". Do we have any intention of cleaning up the directory or are we saying that it is temporary, as in it will go away when we rework the "Load Data" feature?

rjrudin: Right, it will go away once we tackle that set of stories.

rjrudin:
Removing a couple deleted endpoints from the mocks files too (if we support stopping a flow, it needs to be reworked since the current impl depend on FlowRUnnerImpl being a singleton).

rjrudin:
Added a "pari" user to the reference project for easy testing.

PR #3868, opened 2020-04-24
DHFPROD-4654: Added 500 error support to hub central

brucean52:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/src/components/modal-status/modal-status.test.tsx
+++ b/marklogic-data-hub-central/ui/src/components/modal-status/modal-status.test.tsx
@@ -1,12 +1,122 @@
 import React from 'react';
-import { render } from '@testing-library/react';
+import { BrowserRouter as Router } from 'react-router-dom';
+import { render, wait } from '@testing-library/react';
+import axiosMock from 'axios'
+import userEvent from "@testing-library/user-event";
+
 import ModalStatus from './modal-status';
+import NoMatchRedirect from '../../pages/noMatchRedirect';
+import { UserContext } from '../../util/user-context';
+import { 
+    userSessionWarning,
+    userModalError,
+    userNoErrorNoSessionWarning,
+    userHasModalErrorHasSessionWarning
+  } from '../../assets/mock-data/user-context-mock';
+
+jest.mock('axios');
 
 describe('Modal Status Component', () => {
-  test('Modal session status renders', () => {
-    const { getByTestId } = render( <ModalStatus visible={true} />);
-      expect(getByTestId('inactivity')).toBeInTheDocument();
-  })
+  afterEach(() => {
+    jest.clearAllMocks();
+  });
+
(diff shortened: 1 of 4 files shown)


marklogic-data-hub-central/ui/src/components/modal-status/modal-status.test.tsx
```diff
@@ -1,12 +1,211 @@
...
+      </Router>);
+
+      expect(getByText('500 Internal Server Error')).toBeInTheDocument();
+      expect(getByText('java.net.ConnectException: Failed to connect to localhost/0:0:0:0:0:0:0:1:8011')).toBeInTheDocument();
+
+      await wait(() => {
+        userEvent.click(getByText('Cancel'));
+      });
+
+      debug()
```
bsrikan, line 151: can we get rid of this debug from the test.

marklogic-data-hub-central/ui/src/components/modal-status/modal-status.test.tsx
```diff
@@ -1,12 +1,211 @@
...
-      expect(getByTestId('inactivity')).toBeInTheDocument();
-  })
+  afterEach(() => {
+    jest.clearAllMocks();
+  });
+
+  test('Modal session status renders and click continue session', async () => {
+    axiosMock.get.mockImplementation(() => Promise.resolve({ status: 200 }));
+
+    const context = {
```
bsrikan, line 21: Can you pls create a test config file to keep the context and export that to the test. This is being used in several tests, so it will be great if we can extract that out. Similar thing for the error context.. 
For mapping tests we have created a data.config.js under config directory. Probably not the right place to mix test data config with application configs, and I plan to move them to a better place later.

marklogic-data-hub-central/ui/src/components/modal-status/modal-status.test.tsx
brucean52: Removed the debug

marklogic-data-hub-central/ui/src/components/modal-status/modal-status.test.tsx
brucean52: I created a user-context-mock under '/assets/mock-data' since it's mocking Context data and not configuration for the app.

brucean52:
> @brucean52 , now that we have so many modals in the project, Is it better to rename the modal to be more specific so that it is more readable ? like showSessionModal something like that.

It's currently named ModalStatus since it's a modal for both the session status and the 500 error code. Eventually, DHFPROD-4739 will refactor this into 2 components. We can name the components more specific then. 

https://project.marklogic.com/jira/browse/DHFPROD-4739

timur-isangulov:
Reviewed and tested, looks good.

PR #3869, opened 2020-04-24
DHFPROD-4775: Introducing HubClient as a simplified interface

rjrudin:
Most of the changes in here are switching HC code over from HubConfig to HubClient. That's an easy change, since HC doesn't need any of the HubProject stuff in HubConfig. 

Note that AbstractHubCentralTest still has a HubProject, since tests are like Pari - they need to be able to deploy apps for testing. 

Then changed the following HC core classes:

- HubConfig now has newHubClient, as HubConfig has all the config needed to know how to create a HubClient
- Saw an unused method in HubProjectImpl, removed it
- CollectorImpl now gets a RestTemplate from HubClient. Removed 4 methods from the Collector interface that were not invoked anywhere too. 
- QueryStepRunner now depends on a HubClient and a sourceDatabase. The latter is used to control which database is sent to the collector endpoint. The RestTemplate still always talks to the staging app server when hitting the collector, which is now more clear in the code. 
- WriteStepRunner now uses getFinalClient() when an ingestion wants to write to the final database. Wasn't sure if this was tested anywhere, so added IngestInFinalTest to verify it works. 
- Removed "withSourceClient" from StepRunner, as that concept no longer exists. It only applied to QueryStepRunner, and QueryStepRunner will always use getStagingClient() to talk to the collector and mlRunFlow endpoints.

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


marklogic-data-hub/src/main/java/com/marklogic/hub/HubClient.java
```diff
@@ -0,0 +1,42 @@
...
+    DatabaseClient getJobsClient();
+
+    String getDbName(DatabaseKind kind);
+
+    /**
+     * This is needed by the CollectorImpl class, which is not yet able to use a DatabaseClient.
+     *
+     * @return
+     */
+    RestTemplate getStagingRestTemplate();
```
srinathgit, line 41: Is this going to be temporary and will go away when OkHttp is used to make requests for Collector ?

rjrudin: Correct.

PR #3874, opened 2020-04-25
DHFPROD-4864: updates to reference-entity-model project with flows

bsrikan:
### Description
Updates to reference entity model project to include loadData, flows, mappings and Zip and Address entities so mapping will not throw any error.

Sameera will use this project to spin up HC instance on an engrlab machine in a nightly basis.

It also already has hub-developer, hub-operator users and should be used to work with HC instead of admin. Demo should also be run as one of these users and not admin.
When you use this project to create a flow/mapping etc, you would have to download the project just like an HC user would and create a PR to this example project, which then will be available on successive nightlies.

Echoing whatever Rob has been suggesting so far. 

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


examples/reference-entity-model/entities/Address.entity.json
```diff
@@ -0,0 +1,46 @@
+{
```
rjrudin, line 1: We don't need nor want separate entity files. The approach for 5.3.0 is that each model file has one "primary" entity whose name is the same as info.title. Each other entity definition is considered a structured type.

Is there a particular feature that is requiring Address and Zip to be duplicated in separate files?

examples/reference-entity-model/gradle-dhs.properties
```diff
@@ -0,0 +1,19 @@
+mlHost=
```
rjrudin, line 1: I don't think it's worth checking this in, because everyone will either need to modify mlHost in this file or remember to add -PmlHost= to the command line. I think for now, it's better for this file to be gitignored. 

What we may want to add as a testing convenience is an e.g. "hubUseDhsDefaults=true" property so that people don't need to define the dozen or so properties below. That would make life easier in a number of ways.

examples/reference-entity-model/input/json/Cust1.json
```diff
@@ -0,0 +1,33 @@
+{
+    "CustomerID": 101,
```
rjrudin, line 2: I like have sample files, but let's not use abbreviations - just call it "customer1.json".

examples/reference-entity-model/input/xml/CustXMLDoc1.xml
```diff
@@ -0,0 +1,32 @@
+  <CustOrders xmlns="custOrderInfo">
```
rjrudin, line 1: Same thing, let's do customer1.xml

examples/reference-entity-model/loadData/CustomerLoadJSON.loadData.json
```diff
@@ -0,0 +1 @@
+{"name":"CustomerLoadJSON","description":"","sourceFormat":"json","targetFormat":"json","outputURIReplacement":".*CustomerLoadJSON,'/customer'","inputFilePath":"data-sets/CustomerLoadJSON"}
```
rjrudin, line 1: I'd like to encourage a naming practice for steps - which are processes - that are verb-oriented instead of a set of nouns. So let's do "loadCustomerJSON" and "mapCustomerJSON". I think customers should follow the same practice, so we should demonstrate that in our examples. 

Even better would be to have names for the datasets other than "JSON" and "XML", as in the real world, users are likely to have e.g. "loadAbcCustomers" and "loadXyzCustomers". I think we can go with JSON and XML for now, but ideally we can have real data source names here and not just use data formats.

examples/reference-entity-model/entities/Address.entity.json
bsrikan: Mapping doesnt seem to work without them and complains about missing target entitiy.

bsrikan: We can get them updated later to include merge matches. This is just so the nightly instance will have something for the team to build their demo and can be enhanced later.

examples/reference-entity-model/gradle-dhs.properties
bsrikan: We would use this project to spin up a DHS instance and make this project available in HC running against a DHS. Hence the dhs environment property file.

examples/reference-entity-model/loadData/CustomerLoadJSON.loadData.json
bsrikan: I agree with all the suggestions Anyone from the team can later enhance this to be standardized. For now this is just aimed at getting a project together for upcoming demo that can be run in an instance that is not localhost as we had discussed in Friday's standup.

examples/reference-entity-model/entities/Address.entity.json
rjrudin: Hmm - that's a concern, I'll try it out locally.

rjrudin: I verified that the mapping tool in HC is using "Customer" instead of the full entityTypeId. The backend does not yet support mappings that use just entityName and then try to refer to other entity definitions. I'll open a bug for this, as we do not want to go down the path of duplicating entity types again in HC.

PR #3880, opened 2020-04-27
DHFPROD-4865: primaryEntityTypes endpoint now includes entityTypeId

rjrudin:
Removed Address/Zip entity models from the reference project, as we don't want to support that approach anymore.

With this change, mappings no longer appear in the UI because the UI is expecting targetEntityType to be "Customer". But the backend does not support that; entityTypeId must be used instead. So the UI now needs to:

1. Create mappings with targetEntityType = entityTypeId
2. Find mappings for an entity type based on entityTypeId, not entityName

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [ ] Added Tests
  

- ##### Reviewer:

- [ ] Reviewed Tests

### DIFF
--- a/specs/models/Flow.schema.json
+++ b/specs/models/Flow.schema.json
@@ -40,178 +40,191 @@
       "additionalProperties": false,
       "patternProperties": {
         "^[0-9]+$": {
-          "type": "object",
-          "properties": {
-            "name": {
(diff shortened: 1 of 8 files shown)


srinathgit:
This requires a rebase

PR #3883, opened 2020-04-28
DHFPROD-4775: New endpoint for clearing user data

rjrudin:
The interesting new stuff is all in DataHubImpl.

Tossed a bunch of the download-testing stuff into AllArtifactsProject so that it can be reused easily by ClearUserDataTest.

Not updating mocks.json/mocks.yaml yet - going to look into springfox again to see if we can get this stuff generated dynamically from the middle tier endpoints.

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/security.sjs
```diff
@@ -64,7 +64,10 @@ class Security {
     }
 
     if (currentRoleNames.includes("hub-central-downloader")) {
-      response.authorities.push("canDownloadConfigurationFiles");
+      response.authorities.push("downloadConfigurationFiles");
```
bsrikan, line 67: Is this just a name refactor to the download story already merged?

rjrudin: Right - I had pointed out to @ryanjdew that it's redundant to include the word "can". 

I'm also hoping for a day where we can dynamically build authorities based on patterns in role and privilege names, and we're not likely to have the word "can" in either of those names.

PR #3889, opened 2020-04-29
DHFPROD-4880: Using the Java Client for collecting URIs

rjrudin:
This ticket will likely go into Sprint 30. It was a simple change to make (as someone else did the real work), so I want to see how the tests run with this. 

This is the result of DHFPROD-4533, where performance testing confirmed that this approach of accessing the underlying OkHttpClient results in good performance and the same functional behavior. I ran all the tests locally, and all of them passed.

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/collector/impl/CollectorImpl.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/collector/impl/CollectorImpl.java
@@ -17,42 +17,16 @@ package com.marklogic.hub.collector.impl;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.marklogic.client.DatabaseClient;
-import com.marklogic.client.DatabaseClientFactory;
-import com.marklogic.client.DatabaseClientFactory.SSLHostnameVerifier;
-import com.marklogic.client.MarkLogicIOException;
 import com.marklogic.hub.HubClient;
 import com.marklogic.hub.collector.Collector;
 import com.marklogic.hub.collector.DiskQueue;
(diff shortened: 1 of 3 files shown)


marklogic-data-hub/src/main/java/com/marklogic/hub/collector/impl/CollectorImpl.java
```diff
@@ -96,147 +64,29 @@ public CollectorImpl(HubClient hubClient, String sourceDatabase) {
...
-                    for (List<?> alt : alts) {
-                        if (alt != null && alt.size() == 2 && alt.get(1) instanceof String) {
-                            Integer type = (Integer) alt.get(0);
-                            if (type == type_dnsName || type == type_ipAddress) {
-                                subjectAltArray.add((String) alt.get(1));
-                            }
-                        }
-                    }
+            } finally {
+                if (bufferedReader != null) {
```
akshaysonvane, line 83: We don't need the check since `IOUtils.closeQuietly` will check and close.

ryanjdew: Antoher option would be to use a try-with-resource statement, since the bufferedReader isn't required outside of the try block. https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html

srinathgit: This PR also requires a rebase

PR #3891, opened 2020-04-29
DHFPROD-4894: Stop printing requests/responses on MockMvc test failure

akshaysonvane:
### Description
There is a bug in MockMvc that causes ConcurrentModificationException
when it tries to read the response headers for async requests in
order to print the response.
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/AbstractMvcTest.java
+++ b/marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/AbstractMvcTest.java
@@ -5,6 +5,7 @@ import com.marklogic.hub.central.auth.LoginInfo;
 import org.junit.jupiter.api.BeforeEach;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.test.autoconfigure.web.servlet.AutoConfigureMockMvc;
+import org.springframework.boot.test.autoconfigure.web.servlet.MockMvcPrint;
 import org.springframework.http.MediaType;
 import org.springframework.mock.web.MockHttpSession;
 import org.springframework.test.web.servlet.MockMvc;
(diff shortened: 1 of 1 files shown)


marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/AbstractMvcTest.java
```diff
@@ -25,7 +26,7 @@
  * Similar to AbstractOneUiTest, please be judicious about adding helper methods to this class. Any methods added
  * should be applicable to a wide swath of tests, not just a handful.
  */
-@AutoConfigureMockMvc
+@AutoConfigureMockMvc(print = MockMvcPrint.NONE)
```
rjrudin, line 29: Can you add your comment in this PR to the code here? This is a good example of how convenient annotations can be, but they often beg the question of "Why is this set?" Your comment will be a very useful reminder six months from now when a new developer sees this.

akshaysonvane: The PR already contains this
```
DHFPROD-4894: Stop printing requests/responses on MockMvc test failure
There is a bug in MockMvc that causes ConcurrentModificationException
when it tries to read the response headers for async requests in
order to print the response.
```
Do you want me to add anything more?

rjrudin: I think the comment is better placed on the code, not in the PR. PR comments are useful for explaining the scope of a PR, but they're not a good place for describing how a particular line of code works. A reader of this code will see the comment immediately, but may not even think to look for a commit comment.

akshaysonvane: Right, I'll add it to the code.

PR #3892, opened 2020-04-29
DHFPROD-4625: Adjust authorities to work with prefixed privileges

ryanjdew:
### Description

Making some adjustments to the security model based on feedback from the product architecture team. 

I've also altered the currently secured endpoints to follow the new conventions.

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/auth/AuthenticationFilter.java
```diff
@@ -106,6 +106,6 @@ protected AuthenticationToken authenticateUser(String username, String password)
             authorities.add(new SimpleGrantedAuthority("ROLE_" + authority));
         });
 
-        return new AuthenticationToken(username, password, hubCentral.getProjectName(), (ArrayNode) response.get("roles"), authorities);
+        return new AuthenticationToken(username, password, hubCentral.getProjectName(), authorities);
```
rjrudin, line 109: Nice - no more roles!

marklogic-data-hub-central/ui/src/util/authorities.tsx
```diff
@@ -0,0 +1,76 @@
...
+ * @description This service provides information about what capabilities the user has.
+ */
+export class AuthoritiesService implements IAuthoritiesContextInterface {
+    public authorities: string[] = [];
+
+    public setAuthorities: (authorities: string[]) => void = (authorities: string[]) => {
+        this.authorities = authorities;
+    };
+
+    public canReadMapping:() => boolean = () => {
```
rjrudin, line 33: Could the UI just use "readMapping" as well? It seems simpler and less effort to not have to map e.g. "readMapping" to "canReadMapping".

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/entitySearch/saveSavedQuery.sjs
```diff
@@ -24,11 +24,9 @@ var saveQuery;
 var userCollections = ["http://marklogic.com/data-hub/saved-query"];
 var queryDocument = JSON.parse(saveQuery);
 
-let authorized = new Security().getRolesAndAuthorities().authorities.includes("canManageSavedQuery");
-
-if(!authorized) {
-    ds.throwForbidden(xdmp.getCurrentUser() + " user doesn't have authority to save or update query");
-}
+const sec = new Security();
```
rjrudin, line 27: I agree with not using "Security" as the variable name here, but I think we need something besides "sec" because a reader could easily think that's an ML function that starts with "sec.". How about "dhsec", or "hubsec"? (hubsec is kind of cool, like infosec).

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/entitySearch/saveSavedQuery.sjs
```diff
@@ -24,11 +24,9 @@ var saveQuery;
...
+let authorized = new Security().getRolesAndAuthorities().authorities.includes("manageSavedQuery");
```
rjrudin, line 28: Do you still need this line of code? I would think the sec.dataHubAuthorityAssert function would suffice.

marklogic-data-hub-central/ui/src/util/authorities.tsx
bsrikan: @rjrudin I was wondering if we need 2 authorities, 1 each for read and write. Wouldnt canWrite*=false would mean its a read only authority. Please clarify.

rjrudin: I think having one for read and one for write avoids confusion. Either you have the ability to do something - you have the authority - or you don't.

bsrikan: Thanks!

ryanjdew: I'm not too strongly opposed to "readMapping". I just thought I'd leave "canReadMapping" as it makes clear it is a method that returns a boolean, rather than performing an action like "reading a mapping", whatever that action would entail.

@wooldridge @xnikhil08  Any preference here?

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/entitySearch/saveSavedQuery.sjs
ryanjdew: good point. will address that.

marklogic-data-hub-central/ui/src/util/authorities.tsx
rjrudin: Yeah, I'll withdraw my comment here - I think it's fine to do this in the UI, particularly if that makes the code read more nicely.

srinathgit:
The PR needs a rebase

ryanjdew:
rebased

PR #3893, opened 2020-04-29
DHFPROD-4830: Check for false positives on search for multiple merge …

ryanjdew:
…actions on document

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/builtins/steps/mastering/default/merging/falsePositivesForOtherMerges.sjs
+++ b/marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/builtins/steps/mastering/default/merging/falsePositivesForOtherMerges.sjs
@@ -1,0 +1,12 @@
+const merging = require("/data-hub/5/builtins/steps/mastering/default/merging.sjs");
+const test = require("/test/test-helper.xqy");
+
+const assertions = [];
+let caughtException = null;
+try {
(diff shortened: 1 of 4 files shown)


marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/builtins/steps/mastering/default/merging/falsePositivesForOtherMerges.sjs
```diff
@@ -0,0 +1,12 @@
+const merging = require("/data-hub/5/builtins/steps/mastering/default/merging.sjs");
+const test = require("/test/test-helper.xqy");
+
+const assertions = [];
+let caughtException = null;
+try {
+    merging.main({uri: '/merge-with-doc1.json'}, {mergeOptions: {}});
```
rjrudin, line 7: I think there should be an assertion here on the results of this, because shouldn't those be different now? I Otherwise, the reader has to analyze the test documents to figure out the scenario and why things are working now, which I wasn't able to do.

ryanjdew: No, the results won't differ, really. The only difference that is patch is making is that in one corner case it will no longer throw an exception.

It was hitting a null pointer error due to a false positive when looking for another match summary document that has other merge actions for the same document.

PR #3895, opened 2020-04-30
DHFPROD-4887: Generating Swagger docs dynamically via springfox

rjrudin:
Added jsonSchema2Pojo Gradle task to generate Java classes from ./specs/models. Where possible, these are being used for documentation purposes in controllers. I renamed some of these to not have ".v1" in their name, so e.g. "ModelDescriptorV1" seemed like a funky class name. I imagine we'll get rid of ".v1" in all the filenames soon, and replace it either with ".schema" or with nothing.

The README is updated to point to the new endpoint for Swagger docs. It also points to the Application class for more information on how to configure annotations on endpoints. Existing endpoints are a good source of info as well.

I tried to document every endpoint possible; some are using anyObject in mocks.json today, so I didn't make Java classes for some of those. We can gradually fill in those gaps.

If all looks good, I'll submit a separate PR to remove the existing mocks support in the ./ui folder.

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


marklogic-data-hub-central/build.gradle
```diff
@@ -85,6 +91,26 @@ dependencies {
...
+}
+
+task deleteSchemaPojos(type: Delete) {
+    delete fileTree("src/main/java/com/marklogic/hub/central/schemas") {
+        include "*.java"
+    }
+}
+generateJsonSchema2Pojo.mustRunAfter deleteSchemaPojos
+
+task generateSchemaPojos {
```
srinathgit, line 109: So, all the classes in com.marklogic.hub.central.schemas were generated by running the jsonSchema2Pojo gradle task ?

rjrudin: Correct. Unfortunately, the generateJsonSchema2Pojo library doesn't add comments to the classes indicating that they were auto-generated, like the ones generated by the ML DS plugin. That's why I added a README.md file to that schemas package.

PR #3897, opened 2020-04-30
DHFPROD-4861: Query needs to be selected by default after saving a query

timur-isangulov:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/e2e/cypress/integration/scenarios.queries.tsx
+++ b/marklogic-data-hub-central/ui/e2e/cypress/integration/scenarios.queries.tsx
@@ -46,6 +46,7 @@ describe('save/manage queries scenarios', () => {
         browsePage.getSaveQueryDescription().type('new-query description');
         browsePage.getSaveQueryButton().click();
         cy.wait(500);
+        browsePage.getSelectedQuery().should('contain', 'new-query');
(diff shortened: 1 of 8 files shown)


Sanjeevani19:
> Query name persists after clearing search
> https://drive.google.com/file/d/1Z2S53gfkOAjMwkh_KryedSUPYPWar98q/view

@brucean52 , I fixed this bug in my previous PR but marilyn gave me a feedback that clearall should only clear the selected facets not the selected query, because in that state the user can still save the existing query without facets.
- FYI, We have a separate state called reset to no query button that's when we can remove the selected value from dropdown
https://project.marklogic.com/jira/browse/DHFPROD-4879

timur-isangulov:
> Edited query name shows old name
> https://drive.google.com/file/d/1WzyOUGbYCE7doEYzZXtMsv8Vic9qfzyw/view
> 
> Query name persists after clearing search
> https://drive.google.com/file/d/1Z2S53gfkOAjMwkh_KryedSUPYPWar98q/view

Can you please review again.

timur-isangulov:
I noticed many bugs in the query implementation as well. I think we need to create a separate jira issues and fix them accordingly.

PR #3899, opened 2020-04-30
DHFPROD-4756:Fix Column alignment in entity table

srinathgit:
### Description
Fix Column alignment in entity table. This is due to an issue with antd where table headers are misaligned if scroll 'y' is set. It is not needed here anyway, so removing it. I don't think this requires a test
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


srinathgit:
image
When you scroll to the right after the fix

PR #3905, opened 2020-05-01
4865-mapping-bug: Updated the mapping UI components to use targetEntityId instead of entityName to create mapping artifact.

xnikhil08:
### Description
This contains the UI changes to use entityTypeId instead of entityName as targetEntityType for creating mapping artifacts.
Also removed few unnecessary console logs from entityType component.

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/src/components/entities/entity-tiles.tsx
+++ b/marklogic-data-hub-central/ui/src/components/entities/entity-tiles.tsx
@@ -46,14 +46,14 @@ const EntityTiles = (props) => {
           }
     }
 
-    const getMappingArtifactByMapName = async (entityTypeTitle,mapName) => {
+    const getMappingArtifactByMapName = async (entityTypeId,mapName) => {
         try {
             let response = await axios.get(`/api/artifacts/mapping/${mapName}`);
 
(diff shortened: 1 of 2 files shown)


marklogic-data-hub-central/ui/src/components/entities/entity-tiles.tsx
```diff
@@ -46,14 +46,14 @@ const EntityTiles = (props) => {
           }
     }
 
-    const getMappingArtifactByMapName = async (entityTypeTitle,mapName) => {
+    const getMappingArtifactByMapName = async (entityTypeId,mapName) => {
```
rjrudin, line 49: I'll try this out locally, but I was thinking that the UI needs to populate targetEntityType with entityTypeId instead of entityType too?

rjrudin: I just tested this locally - I created a new mapping, and the targetEntityType was correctly set using the entityTypeId.

PR #3910, opened 2020-05-04
DHFPROD-4887: Enabling Spring profiles

rjrudin:
Tossing SwaggerConfig into dev profile so we have an easy way of omitting it

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


bsrikan:
@rjrudin this is so that swagger docs are generated only in dev?

rjrudin:
@bsrikan  Eventually, yes. We're not selecting a Spring profile per environment yet. But this is the first step towards adopting Spring profiles to control what's enabled in an environment. The Swagger config is a good way to force us to take that first step, as our assumption is we won't want it in a production environment.

PR #3911, opened 2020-05-04
DHFPROD-4728: Grant role to allow user to export data

timur-isangulov:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/e2e/cypress/fixtures/users/developer.json
+++ b/marklogic-data-hub-central/ui/e2e/cypress/fixtures/users/developer.json
@@ -2,5 +2,5 @@
     "user-name": "dh-dev",
     "description": "A data hub dev user",
     "password": "dh-dev",
-    "role": ["data-hub-developer"]
+    "role": ["data-hub-developer", "hub-central-entity-exporter"]
 }
--- a/marklogic-data-hub-central/ui/e2e/cypress/integration/scenarios.queries.tsx
+++ b/marklogic-data-hub-central/ui/e2e/cypress/integration/scenarios.queries.tsx
@@ -15,7 +15,7 @@ const homePage = new HomePage();
 const queryComponent = new QueryComponent()
 
 
-describe('save/manage queries scenarios', () => {
+describe('save/manage queries scenarios, developer role', () => {
 
     beforeEach(() => {
         cy.visit('/');
@@ -73,7 +73,5 @@ describe('save/manage queries scenarios', () => {
         queryComponent.getDeleteQuery().click();
         queryComponent.getDeleteQueryYesButton().click({force: true})
     });
-
-
 
 });
--- a/marklogic-data-hub-central/ui/src/components/queries/managing/manage-query-modal/manage-query.test.tsx
+++ b/marklogic-data-hub-central/ui/src/components/queries/managing/manage-query-modal/manage-query.test.tsx
@@ -5,18 +5,28 @@ import ManageQuery from './manage-query';
 
 describe('Query Modal Component', () => {
 
-    test('Modal is not visible', () => {
-      const { queryByTestId } =  render(<ManageQuery />);
+  let query = [{ "savedQuery": { "id": "00529904-ed6e-4539-b7d8-eeddaf15aaa4", "name": "Lee", "description": "", "query": { "searchText": "", "entityTypeIds": ["Customer"], "selectedFacets": { "firstname": { "dataType": "xs:string", "stringValues": ["Lee"] } } }, "propertiesToDisplay": ["id"], "owner": "admin", "systemMetadata": { "createdBy": "admin", "createdDateTime": "2020-05-05T14:48:53.206721-07:00", "lastUpdatedBy": "admin", "lastUpdatedDateTime": "2020-05-05T14:48:53.206721-07:00" } } }]
 
-      expect(queryByTestId('manage-queries-modal')).toBeNull();
-    });
+  test('Verify modal is not visible', () => {
+    const { queryByTestId } = render(<ManageQuery />);
+    expect(queryByTestId('manage-queries-modal')).toBeNull();
+  });
 
-    test('Modal is visible', () => {
-        const { getByTestId } =  render(<ManageQuery />);    
+  test('Verify export button is visible', () => {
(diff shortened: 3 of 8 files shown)


marklogic-data-hub-central/ui/e2e/cypress/fixtures/users/developer.json
```diff
@@ -2,5 +2,5 @@
     "user-name": "dh-dev",
     "description": "A data hub dev user",
     "password": "dh-dev",
-    "role": ["data-hub-developer"]
+    "role": ["data-hub-developer", "hub-central-entity-exporter"]
```
rjrudin, line 5: Do the Cypress tests actually hit ML? I thought UI tests were just operating against mocks. 

If they do hit ML, I recommend having a generic "test user" and then enhancing the test framework so that it's easy to change roles on that user. That's something we have in the HC JUnit tests.

timur-isangulov: Yes, Cypress hits ML. Since we deploy data-hub project with all the test users and data before running tests it may be difficult to achieve that at test running stage. I will sync with Srikanth about that.

marklogic-data-hub-central/ui/e2e/cypress/integration/scenarios.queries.tsx
```diff
@@ -59,6 +59,13 @@ describe('save/manage queries scenarios', () => {
         queryComponent.getSubmitButton().click();
     });
 
+    it('check export icon displays', () => {
```
bsrikan, line 62: This test can be verified via RTL and not e2e worthy.

marklogic-data-hub-central/ui/e2e/cypress/integration/scenarios.queries.tsx
```diff
@@ -74,6 +81,51 @@ describe('save/manage queries scenarios', () => {
...
+        browsePage.getSaveQueryDescription().should('be.visible');
+        browsePage.getSaveQueryDescription().type('new-query description');
+        browsePage.getSaveQueryButton().click();
+        cy.wait(500);
+        browsePage.getSelectedQuery().should('contain', 'new-query');
+        browsePage.getSaveQueryButton().should('not.be.visible');
+        browsePage.getSaveQueriesDropdown().should('be.visible');
+    });
+
+    it('check export icon does not display', () => {
```
bsrikan, line 122: not e2e worthy. Should be covered using RTL.

marklogic-data-hub-central/ui/e2e/cypress/integration/scenarios.queries.tsx
```diff
@@ -74,6 +81,51 @@ describe('save/manage queries scenarios', () => {
...
+        cy.contains('MarkLogic Data Hub');
+        cy.loginAsOperator();
+        cy.wait(500);
+        homePage.getBrowseEntities().click();
+        cy.wait(1000);
+        browsePage.getFacetView();
+        browsePage.selectEntity('All Entities');
+    });
+
+    it('apply facet search,open save modal and save queries', () => {
```
bsrikan, line 99: I would recommend trying to achieve this using RTL as well. The story is for export data and this test seems to be for search/save queries. Exporting data could be an e2e scenario, although I would recommend trying to use RTL first.

bsrikan:
Most of the scenarios I see in e2e for Explorer seems to be possible to write using RTL. Although that could be a tech debt item to refactor/clean up later. 
New stories being implemented must have RTL tests as much as possible. e2e should involve a scenario that for ex verifies if a valid user can search, save a query and export data. Each step in this e2e -- search, save and export is a unit/functional test that we should try to cover using RTL.
We have to list in the test plan what does an e2e constitute for Explorer. Then add Cypress tests.

Meanwhile I will create a story under tech debt epic to be able to add roles to a generic user for the use case being tested as suggested by @rjrudin .

marklogic-data-hub-central/ui/src/components/queries/managing/manage-query-modal/manage-query.test.tsx
```diff
@@ -5,18 +5,35 @@ import ManageQuery from './manage-query';
...
 
-    test('Modal is not visible', () => {
-      const { queryByTestId } =  render(<ManageQuery />);
+  let query = [{ "savedQuery": { "id": "00529904-ed6e-4539-b7d8-eeddaf15aaa4", "name": "Lee", "description": "", "query": { "searchText": "", "entityTypeIds": ["Customer"], "selectedFacets": { "firstname": { "dataType": "xs:string", "stringValues": ["Lee"] } } }, "propertiesToDisplay": ["id"], "owner": "admin", "systemMetadata": { "createdBy": "admin", "createdDateTime": "2020-05-05T14:48:53.206721-07:00", "lastUpdatedBy": "admin", "lastUpdatedDateTime": "2020-05-05T14:48:53.206721-07:00" } } }]
 
-      expect(queryByTestId('manage-queries-modal')).toBeNull();
-    });
+  test('Modal is not visible', () => {
+    const { queryByTestId } = render(<ManageQuery />);
+    expect(queryByTestId('manage-queries-modal')).toBeNull();
```
bsrikan, line 12: I will approve this now. Just a thought for future tests we add, that the 2 modal visibility tests can be combined with a description "Verify modal visibility" and the assertion for not visible until you click on the modal icon can be moved into that test.

marklogic-data-hub-central/ui/src/components/queries/managing/manage-query-modal/manage-query.test.tsx
```diff
@@ -5,18 +5,35 @@ import ManageQuery from './manage-query';
...
+  });
 
-    test('Modal is visible', () => {
-        const { getByTestId } =  render(<ManageQuery />);    
+  test('Modal is visible', () => {
+    const { getByTestId } = render(<ManageQuery />);
+    expect(getByTestId('manage-queries-modal-icon')).toBeInTheDocument();
```
bsrikan, line 17: click event will fail if the testid is not in the document. I would like to avoid assertions that dont add value to a test.
Moreover I see a patter here, which we should avoid as well.
All 3 test steps in 'Modal is visible' is duplicated in subsequent steps. We have to click on the icon, there is no way around it, but in doing so 'Modal is visible' is getting tested automatically and so we dont need a separate test doing just that.

PR #3912, opened 2020-05-04
DHFPROD-4768:Fix route changes

srinathgit:
### Description
Add back the snippet used for page redirect which shouldn't have been removed in the previous PR
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/src/App.tsx
+++ b/marklogic-data-hub-central/ui/src/App.tsx
@@ -77,6 +77,16 @@ const App: React.FC<Props> = ({history, location}) => {
 
   }, [user]);
 
+  useEffect(() => {
+    // On route change...
+    axios.get('/api/environment/project-info')
+        .then(res => {})
+        // Timeouts throw 401s and are caught here
+        .catch(err => {
+            handleError(err);
+        })
+  }, [location.pathname]);
+
   const path = location['pathname'];
   const pageTheme = (themeMap[path]) ? themes[themeMap[path]] : themes['default'];
   document.body.classList.add(pageTheme['bodyBg']);
--- a/marklogic-data-hub-central/ui/src/util/environment.tsx
+++ b/marklogic-data-hub-central/ui/src/util/environment.tsx
@@ -3,10 +3,8 @@ import axios from 'axios';
 const defaultEnv = {
     projectName: '',
     dataHubVersion: '',
-    markLogicVersion: '',
-    host:'',
-    stagingPort:''
-  }
+    markLogicVersion: ''
+}
 
 export function setEnvironment()  {
     axios.get('/api/environment/project-info')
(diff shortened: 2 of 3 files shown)


marklogic-data-hub-central/ui/src/App.tsx
```diff
@@ -77,6 +77,16 @@ const App: React.FC<Props> = ({history, location}) => {
 
   }, [user]);
 
+  useEffect(() => {
+    // On route change...
+    axios.get('/api/environment/project-info')
```
bsrikan, line 82: This was '/api/environment/project' earlier. Should this be project-info now?

bsrikan:
@brucean52 I have added you as a reviewer, to verify if there is any impact to Explorer.

marklogic-data-hub-central/ui/src/App.tsx
srinathgit: yeah I believe this snippet is meant for route change (which I had removed in my earlier PR).  It should just call a valid endpoint. If user is logged out due to session expiry, this will enable the page to be redirected to login page. @wooldridge can you please confirm if that is the case

srinathgit:
> I am redirected to login page. But there used to be some POST requests firing every few seconds just before the timeout pop up appears. I dont see that anymore. Not sure if it is intentional, so requesting changes until we know for sure that it is and wont break other parts of the application..
Can you provide more info on the POST requests firing that you are referring to ? What endpoint is called and does that endpoint have anything to do with logout pop up ?

bsrikan:
Its a GET request

image

brucean52:
> @brucean52 I have added you as a reviewer, to verify if there is any impact to Explorer.

Explorer is a read only app so it should not be any impact. Tested out the PR on explorer pages and it worked without issues.

srinathgit:
> Its a GET request
> 
> image

As far as I can tell, there shouldn't be any impact. The /info endpoint should be called after user opts to continue the session in the popup, it is used to reset the timer. If you can find any issues, do let me know

marklogic-data-hub-central/ui/src/util/environment.tsx
```diff
@@ -32,4 +30,8 @@ export function getEnvironment():any {
     }
 }
 
-export default { setEnvironment, getEnvironment };
+export function resetEnvironment() {
```
rjrudin, line 33: When would this be called? I don't understand the concept of "reset environment" in HC on DHS, as the user has no control over the environment.

srinathgit: The function resetEnvironment() ensures the project name header link (currently hardcoded to "Data Hub Project") in the UI is not displayed when user logs out.

rjrudin: That name will soon be the DHS service name, which is intended to be a meaningful identifier to the user. I think it makes sense to have that displayed in the login page so the user knows which DHS instance they're logging into. Seems like a requirements question, check with @sbayatpur .

srinathgit: Yeah, that's a different story that will probably taken up next sprint. In this PR, I am just adding back the snippet I had removed inadvertently in my previous PR for removing unwanted code from HC

PR #3914, opened 2020-05-04
DHFPROD-4668: Add hub-central-user role

ryanjdew:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/auth/AuthenticationFilter.java
+++ b/marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/auth/AuthenticationFilter.java
@@ -17,13 +17,13 @@ package com.marklogic.hub.central.auth;
 
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
-import com.fasterxml.jackson.databind.node.ArrayNode;
 import com.marklogic.client.FailedRequestException;
 import com.marklogic.hub.central.HttpSessionHubClientProvider;
 import com.marklogic.hub.central.HubCentral;
 import com.marklogic.hub.dataservices.SecurityService;
 import org.springframework.security.authentication.AuthenticationServiceException;
 import org.springframework.security.authentication.BadCredentialsException;
+import org.springframework.security.authentication.InsufficientAuthenticationException;
 import org.springframework.security.core.Authentication;
 import org.springframework.security.core.AuthenticationException;
 import org.springframework.security.core.GrantedAuthority;
@@ -99,13 +99,16 @@ public class AuthenticationFilter extends AbstractAuthenticationProcessingFilter
             }
             throw e;
         }
-
         List<GrantedAuthority> authorities = new ArrayList<>();
+        final boolean[] hasLoginAuthority = {false};
         response.get("authorities").iterator().forEachRemaining(node -> {
             String authority = node.asText();
+            hasLoginAuthority[0] = ("loginToHubCentral".equals(authority)) || hasLoginAuthority[0];
             authorities.add(new SimpleGrantedAuthority("ROLE_" + authority));
         });
-
+        if (!hasLoginAuthority[0]) {
+            throw new InsufficientAuthenticationException("User doesn't have necessary privileges to access Hub Central");
+        }
         return new AuthenticationToken(username, password, hubCentral.getProjectName(), authorities);
     }
 }
--- a/marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/auth/LoginFailureHandler.java
+++ b/marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/auth/LoginFailureHandler.java
@@ -17,6 +17,7 @@ package com.marklogic.hub.central.auth;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.node.ObjectNode;
+import org.springframework.security.authentication.InsufficientAuthenticationException;
 import org.springframework.security.core.AuthenticationException;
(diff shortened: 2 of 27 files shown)


marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/auth/AuthenticationFilter.java
```diff
@@ -99,13 +99,16 @@ protected AuthenticationToken authenticateUser(String username, String password)
...
         List<GrantedAuthority> authorities = new ArrayList<>();
+        final boolean[] hasLoginAuthority = {false};
         response.get("authorities").iterator().forEachRemaining(node -> {
             String authority = node.asText();
+            hasLoginAuthority[0] = ("loginToHubCentral".equals(authority)) || hasLoginAuthority[0];
             authorities.add(new SimpleGrantedAuthority("ROLE_" + authority));
         });
-
+        if (!hasLoginAuthority[0]) {
+            throw new InsufficientAuthenticationException("User doesn't has necessary privileges to access Hub Central");
```
rjrudin, line 110: Little typo - have, not has

marklogic-data-hub/src/main/resources/hub-internal-config/security/roles/hub-central-clear-user-data.json
```diff
@@ -2,13 +2,20 @@
...
+    "hub-central-user",
     "data-hub-developer",
     "data-hub-entity-model-reader",
     "data-hub-flow-reader",
     "data-hub-load-data-reader",
     "data-hub-mapping-reader",
     "data-hub-match-merge-reader",
-    "data-hub-step-definition-reader"
+    "data-hub-step-definition-reader",
+    "data-hub-entity-model-writer",
```
rjrudin, line 13: I don't think the writer roles are actually needed, as they don't confer any privileges. And technically, we don't want this role to be able to overwrite any of these things - the role only needs to be able to clear the database and then do an initial insert of them.

bsrikan:
I verified that 

- a developer user can log in and has access to all the loadData/mappings/entities
- an operator user is not able to login and gets 403
- an admin user is not able to login and gets 403
- a user with only hub-central-user role is able to log in and see empty configurations for load/entities/run tile. View and Browse show 403 message in the screen. Allowed to click open Add button to create a Model and silently fails with 403

@ryanjdew  does everything for hub-central-user only use case look right?

ryanjdew:
@bsrikan That looks correct to me. For the test-* I've temporarily added the hub-central-user role to those until we have the hub-central equivalent roles in place. 

Once we address the other roles, those should address the hiding/disabling of the buttons as mentioned in the last bullet.

marklogic-data-hub/src/test/ml-config/security/roles/data-hub-test-internal.json
```diff
@@ -0,0 +1,56 @@
...
+      "action": "http://marklogic.com/xdmp/privileges/create-role",
+      "kind": "execute"
+    },
+    {
+      "privilege-name": "remove-role",
+      "action": "http://marklogic.com/xdmp/privileges/remove-role",
+      "kind": "execute"
+    },
+    {
+      "privilege-name": "get-user",
```
bsrikan, line 36: will this privilege impact running of tests in DHS?

marklogic-data-hub/src/test/ml-config/security/roles/data-hub-test-internal.json
```diff
@@ -0,0 +1,56 @@
...
+      "action": "http://marklogic.com/xdmp/privileges/get-user",
+      "kind": "execute"
+    },
+    {
+      "privilege-name": "create-user",
```
bsrikan, line 41: will this and other *-user privileges impact running of tests in DHS?

ryanjdew: It is only used as part of an amp on a testing library function. I'm not aware of any negative impact of using this in DHS.

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/auth/LoginFailureHandler.java
```diff
@@ -37,7 +38,11 @@ public void onAuthenticationFailure(HttpServletRequest request, HttpServletRespo
         String json = node.toString();
         response.setContentType("application/json");
         response.setCharacterEncoding("UTF-8");
-        response.setStatus(HttpServletResponse.SC_UNAUTHORIZED);
+        if (exception instanceof InsufficientAuthenticationException) {
```
bsrikan, line 41: was this exception causing the clearUserData test to fail?

rjrudin: I don't think so - that test was failing on 9.0-12 because it was depending on granular privileges that are in 10.0-3.

bsrikan: I was referring to the unit test failure for the PR which I think runs on ML 10.0-4:
https://jenkins.marklogic.com/blue/organizations/jenkins/Datahub_CI/detail/PR-3914/16/tests

ryanjdew: It turned out that I was wiping out the granular privileges when attempting to deploy the amp/role for testing in ml-unit-tests here: https://github.com/marklogic/marklogic-data-hub/pull/3914/files#diff-1ec8825c599ca5cc7abf57c14482d7e5R79 

Adding the GranularPrivilegeCommand here addressed the problem.

PR #3915, opened 2020-05-05
DHFPROD-4804: Added hubVersion Gradle task

rjrudin:
### Description

One note - there's some cleanup I'm aiming to do when we do the story for getting rid of mlDHFVersion - the existing Versions class can be simplified a bit at that point. But I didn't want to touch it now because it has to account for DH not being installed, which of course is not a use case for HC. VersionInfo is intended to be a clean, simple way to get version data when you know you can connect to DH. 

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/impl/VersionInfo.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/impl/VersionInfo.java
@@ -1,0 +1,55 @@
+package com.marklogic.hub.impl;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.marklogic.hub.HubClient;
+import com.marklogic.hub.dataservices.SystemService;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.Properties;
+
+/**
+ * This is intended to replace some of the Versions class with a HubClient-friendly and non-Spring-dependent way of
+ * accessing version data.
+ */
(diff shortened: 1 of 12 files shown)


marklogic-data-hub/src/main/java/com/marklogic/hub/impl/VersionInfo.java
```diff
@@ -0,0 +1,55 @@
...
+    public static String getBuildVersion() {
+        Properties properties = new Properties();
+        try (InputStream inputStream = VersionInfo.class.getClassLoader().getResourceAsStream("version.properties")) {
+            properties.load(inputStream);
+        } catch (IOException e) {
+            throw new RuntimeException("Unable to get library version from version.properties, cause: " + e.getMessage(), e);
+        }
+
+        String version = properties.getProperty("version");
+        return "${project.version}".equals(version) ? "5.3-SNAPSHOT" : version;
```
akshaysonvane, line 53: Can we add this file to the `updateVersion` task that's in root build.gradle?
The `updateVersion` task keeps track of all the hardcoded version locations.

rjrudin: Good idea, done

PR #3920, opened 2020-05-06
DHFPROD-4888: Add entity type title as collection to mapped documents

ryanjdew:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/controllers/MappingControllerTest.java
+++ b/marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/controllers/MappingControllerTest.java
@@ -165,8 +165,9 @@ public class MappingControllerTest extends AbstractHubCentralTest {
         JsonNode result = controller.getMappingSettings("TestCustomerMapping").getBody();
         // Check for defaults
         assertEquals("TestCustomerMapping", result.get("artifactName").asText());
-        assertEquals(1, result.get("collections").size());
-        assertEquals("default-mapping", result.get("collections").get(0).asText());
+        assertEquals(2, result.get("collections").size());
+        assertEquals("TestCustomerMapping", result.get("collections").get(0).asText());
+        assertEquals("Customer", result.get("collections").get(1).asText());
 
         ObjectNode settings = readJsonObject(MAPPING_SETTINGS);
 
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/hub-es.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/hub-es.sjs
@@ -16,6 +16,7 @@
 
 'use strict';
 const sem = require("/MarkLogic/semantics.xqy");
+const entityLib = require("/data-hub/5/impl/entity-lib.sjs");
 
 function getPropertyRangePath(entityIRI, propertyPath) {
   let properties = propertyPath.split("/");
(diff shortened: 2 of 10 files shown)


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/hub-es.sjs
```diff
@@ -121,8 +121,23 @@ function getRefEntityIdentifiers(entityIRI, entityInfo, prop) {
   return refEntityId;
 }
 
+function getEntityServiceTitle(iri) {
+  if (!(iri instanceof sem.iri)) {
+    iri = sem.iri(iri);
+  }
+  const triple = fn.head(cts.triples(iri,
```
rjrudin, line 128: Thoughts on parsing this out instead of doing a query? I'm just thinking that for mapping 1 million docs, this means 1 million additional triple lookups. 

I've also tried to start a naming convention in entity-lib.sjs where "find" indicates that a function will do a query, and "get" indicates that a function will not do a query. I've found this to be helpful on previous projects, though of course this is just one of dozens of DH libraries. But I do think it's a helpful way so readers can quickly detect where queries occur when analyzing code for performance.

ryanjdew: @rjrudin I'm trying to recall, did we add validation to the entity title to ensure we don't have spaces and other things that potentially require knowing the right decoding rules to get the value from the URI? If we have that, I don't have an issue with parsing it from the IRI. 

I can also make some adjustments to limit the triples query to the models and cache the value for the duration of a batch/transaction.

rjrudin: We do have validation in the new model endpoints in 5.3, but of course that won't impact existing entity models. Though since we'll have docs telling people how to upgrade their entity models for HC and 5.3, we could tell them to ensure their titles conform to the validation rules in 5.3.

entity-lib.sjs has a function too for doing the parsing - getEntityTypeIdParts .

marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/controllers/MappingControllerTest.java
```diff
@@ -166,7 +166,7 @@ public void testMappingSettings() {
         // Check for defaults
         assertEquals("TestCustomerMapping", result.get("artifactName").asText());
         assertEquals(1, result.get("collections").size());
-        assertEquals("default-mapping", result.get("collections").get(0).asText());
+        assertEquals("Customer", result.get("collections").get(0).asText());
```
bsrikan, line 169: Shouldnt the default collection be both entityType title and stepName to be consistent with pre 5.3?

rjrudin: "default-mapping" is technically the step definition name, though I don't think it has any meaning other than to be a sensible default for a collection for mapped documents. If the entityType is used as the sensible default, then I don't see any reason nor meaning in the step definition name being used as a default collection too.

bsrikan: I was looking for "TestCustomerMapping" and agree that "default-mapping" shouldnt be used as a default collection.

ryanjdew: @bsrikan added the step name to the default collections.

bsrikan:
@ryanjdew not sure if this is a result of this PR(I will check and confirm on latest develop), but when I try to update the collection via Edit(pencil icon) the mapping gets overwritten and I loose my expressions.

ryanjdew:
@bsrikan That is unrelated to these changes.

bsrikan:
Right just verified that its an issue in develop as well. Will file a bug. This PR is looking good.

rjrudin:
> Right just verified that its an issue in develop as well. Will file a bug. This PR is looking good.

@bsrikan The next version of the endpoints will address this.

PR #3921, opened 2020-05-06
DHFPROD-4634:Update an existing query after applying new changes

Sanjeevani19:
### Description

- This PR also includes DHFPROD-4878(Edit query details), DHFPROD-4862(default selection for unapplied facets) and Save a copy of current query functionality.

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


timur-isangulov:
Saving a new query replaces previously saved. https://drive.google.com/file/d/1-xNDYiexJHasgw7vjhbORgmXayg9dRhE/view

Sanjeevani19:
> Saving unapplied facets does not save the unapplied facets
> https://drive.google.com/file/d/1XRyAqy1eg-W1JkN6WU-iRJlnl8FJzukl/view
@brucean52 We are not saving unapplied facets, when i checked with yash about this he mentioned only applied facets should be saved when saving a query, that's the reason when user clicks on second radio option, there is a apply button which comes up so that the user can save it if he wants to.

> 1. Switching Entity in dropdown and saving another query can cause the last saved query to be lost
> 2. Saved Queries do not set correct Entity in dropdown
>    https://drive.google.com/file/d/1o_u7E73Y_kUQdox0f2pZ_OBuExnbC6VS/view
@brucean52 
I will be handling these two scenarios as a part of this story.
https://project.marklogic.com/jira/browse/DHFPROD-4964

PR #3928, opened 2020-05-08
DHFPROD-4961:Fix the logic for fetching entities,setting targetEntityType and sourceContext

srinathgit:
### Description
Fix the logic for fetching entities, setting targetEntityType and sourceContext. The existing tests should run fine. Additionally, this will require e2e which @bsrikan will add.
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/src/components/entities/mapping/mapping-card.tsx
+++ b/marklogic-data-hub-central/ui/src/components/entities/mapping/mapping-card.tsx
@@ -224,7 +224,7 @@ const MappingCard: React.FC<Props> = (props) => {
                 }
                 if(parsedDoc['envelope']){
                     if(parsedDoc['envelope'].hasOwnProperty('@xmlns')){
-                        
+
                         let nmspcURI = parsedDoc['envelope']['@xmlns']
                         let indCheck = nmspcURI.lastIndexOf('/');
                         let ind = indCheck !== -1 ? indCheck + 1 : 0;
@@ -260,7 +260,7 @@ const MappingCard: React.FC<Props> = (props) => {
         let {lastUpdated, ...dataPayload} = mapDataLocal;
         dataPayload['namespaces'] = nmspaces;
         await props.updateMappingArtifact(dataPayload);
-        
+
         let mapArt = await props.getMappingArtifactByMapName(dataPayload.targetEntityType,dataPayload.name);
 
         if(mapArt) {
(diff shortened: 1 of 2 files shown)


marklogic-data-hub-central/ui/src/components/entities/mapping/mapping-card.tsx
```diff
@@ -500,27 +498,18 @@ const MappingCard: React.FC<Props> = (props) => {
                 }
                 nestedEntityData.push(propty);
                 extractNestedEntityData(val.subProperties, propty.children, parentKey);
-
-                if(parentKey !== tempKey){
-                    parentKey = tempKey;
-                }
+                parentKey = (parentKey.indexOf("/")!=-1)?parentKey.substring(0,parentKey.lastIndexOf('/')):''
```
rjrudin, line 501: Is this based on the assumption that the entity title is the part of an entityTypeId that comes after the last slash? Would it be possible for the backend to provide this information instead? 

I'm leery of the UI having any knowledge of how to parse bits out of the entityTypeId. I'd like to consolidate all of that logic into entity-lib.sjs in the backend so that if any assumptions prove incorrect, we only have to change it in a single place in all of DHF.

srinathgit: No, this doesn't have anything to do with entityTypeId. This method converts the nested entity properties to array that is needed for entity table in the ui . The keys of the table are unique and it is built by concatenating the property names (name/firstName/prefix etc)..

rjrudin:
@wooldridge What's the error? For some changes, like to roles and privileges, you have to uninstall DH because ML won't allow for a role/privilege to be changed. Best bet is to spin up a new Docker VM and deploy the DH app to that.

wooldridge:
@rjrudin there's been some change recently that makes the old project folder that I've been using for development no longer work. This happened when I tried to test this PR above and it happened earlier today when I cherry-picked my tile UI work on to the latest develop. Now I can't log into the UI, my POST to http://localhost:3000/api/login ({username: "admin", password: "admin"}) results in:
```
{"message":"User doesn't have necessary privileges to access Hub Central"}
```
I will try to build a new project from scratch using the files described by @bsrikan or try the reference example project.

rjrudin:
@wooldridge Check with @ryanjdew , I believe this is due to the new requirement that all users must have the hub-central-user role.

bsrikan:
Mapped document looks like it did earlier which is good. Test button shows values from all the documents. Will file a separate bug for that. 
There should be a backend test that runs a flow with an entityType having same property name in structured types as well.

image

PR #3931, opened 2020-05-11
DHFPROD-4398: Fix the 400 error in mapping POST request for an XML doc.

xnikhil08:
### Description
This is just another PR for 4398, to fix the 400 error in mapping POST request for XML source doc.
Doesn't need any test.

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/src/components/entities/mapping/mapping-card.tsx
+++ b/marklogic-data-hub-central/ui/src/components/entities/mapping/mapping-card.tsx
@@ -315,7 +315,7 @@ const MappingCard: React.FC<Props> = (props) => {
         let ind = indCheck !== -1 ? indCheck + 1 : 0;
         let obj: any = {};
         let nmspace = val[el].slice(ind);
-        if (!nmspaces.hasOwnProperty(nmspace)) {
+        if (nmspace && !nmspaces.hasOwnProperty(nmspace)) {
             obj[nmspace] = val[el];
         }
         let colonIndex = el.indexOf(':');


marklogic-data-hub-central/ui/src/components/entities/mapping/mapping-card.tsx
```diff
@@ -258,6 +258,9 @@ const MappingCard: React.FC<Props> = (props) => {
 
     const updateMappingWithNamespaces = async (mapDataLocal) => {
         let {lastUpdated, ...dataPayload} = mapDataLocal;
+        if(nmspaces.hasOwnProperty("")){
```
bsrikan, line 261: rather than deleting an invalid namespace property, is it possible to not add/create such namespace in the first place.

rjrudin:
@bsrikan That XML is valid though; we're not saying that it's incorrect, are we? The xmlns="" must be there to indicate that those elements are not in the "es:" namespace, which is declared as the default namespace in the root element of the document. 

And XML data can be added to ML via any means, not just an ingestion step. We have to support xmlns="", it's a valid XML attribute that is often required to be present.

marklogic-data-hub-central/ui/src/components/entities/mapping/mapping-card.tsx
xnikhil08: Done.

PR #3935, opened 2020-05-12
Fixed Jest tests

phoenixeliot:
This definitely doesn't fit the PR style guidelines; this is just for reference for @brucean52 

Ignore the many changed files in each of the `components` and `pages` folders; those are just updating the name of MlButton to MLButton.


phoenixeliot:
The Jest things have been fixed on Bruce's end, so this PR is done.

PR #3936, opened 2020-05-12
DHFPROD-4945: New endpoints for flows and ingestion/mapping steps

rjrudin:
For reviewing/testing this locally - best thing to do is examine the new tests, both for HC and for the new DS endpoints. Running the HC tests verifies that we have nearly 100% coverage on the new controllers (the only thing not covered is a catch block for a JSON processing exception). 

Lots of changes in here - these are the categories of changes:

- New api/sjs files for the new flow/step DS endpoints
- 3 new HC controllers for flows, ingestion steps, and mapping steps
- Updates to our schemas and the classes generated from them
- Lots of tests, including several new improvements to the test frameworks

The one thing I purposefully did not write a test for was running a referenced ingestion step, as we know that functionality is going to change soon so that it requires files to be uploaded when the step is run.

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


bsrikan:
Dear reviewers, there is a lot to go through here so I am unchecking "Reviewed Tests" so I can get time to look through all the changes and test before this gets merged.

rjrudin:
@bsrikan  I'll mark it as a Draft for now so it doesn't get merged.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/step/saveStep.sjs
```diff
@@ -0,0 +1,60 @@
...
+  // For now, can assume the stepDefinitionName based on the type. Can add stepDefinitionType as a parameter once we need
+  // more flexibility.
+  const stepDefinitionName = "mapping" === stepDefinitionType.toLowerCase() ? "entity-services-mapping" : "default-ingestion";
+
+  stepProperties.stepDefinitionName = stepDefinitionName;
+  stepProperties.stepDefinitionType = stepDefinitionType;
+  stepProperties.stepId = stepName + "-" + stepDefinitionType;
+
+  const stepDef = new Step().getStepByNameAndType(stepDefinitionName, stepDefinitionType);
+  const stepDefOptions = stepDef.options;
```
srinathgit, line 52: The default collections for ingestion step should be the step name (current behavior) and for mapping step, it's the step name and the entity name(explorer requires that ).

rjrudin: That logic still exists in the loadData/mapping artifact libraries. 

Should we not copy the default collections from the step definition? In which case - shouldn't we remove the default collections from the step definition? Do they serve any other purpose?

Pinging @ryanjdew  and @bsrikan  for insight.

rjrudin: I think I see what to do - the default collections are currently applied in defaultArtifactSettings, which isn't called for step creation. I'll copy/paste that logic into defaultArtifact for mapping.sjs, and then I'll modify this to not pull the default collections from the step definition.

rjrudin: Thanks @srinathgit - I updated the tests (and added manageIngestionStep.sjs) so the assertions are on the step name and also the entity type for mapping tests.

PR #3941, opened 2020-05-12
DHFPROD-4704: Deploy entity model configs after saving an entity

akshaysonvane:
### Description
- As per @rjrudin's suggestion added the deployment code to the controller since EntityManagerImpl is tightly-coupled to a project filesystem right now.
- Deploying the model config only when we update entity model types since that's when the model will have the entity properties.
- Moved `applyDatabasePropertiesForTests` method to the testFixture `AbstractHubTest` to make it accessible to both core and HC.
- Also moved and renamed `marklogic-data-hub/src/test/ml-config` to `marklogic-data-hub/src/test/resources/test-config` since test resouces are shared with both core and HC projects. 
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/controllers/ModelController.java
```diff
@@ -76,11 +84,85 @@
     @RequestMapping(value = "/{modelName}/entityTypes", method = RequestMethod.PUT)
     @ApiImplicitParam(required = true, paramType = "body", dataType = "ModelDefinitions")
     public ResponseEntity<Void> updateModelEntityTypes(@ApiParam(hidden = true) @RequestBody JsonNode entityTypes, @PathVariable String modelName) {
+        // update the model
         newService().updateModelEntityTypes(modelName, entityTypes);
+
+        //deploy updated configs
+        deployModelConfigs();
```
rjrudin, line 91: Yeah I think this is fine here for now. Once we know it's all working correctly, we can consider how to refactor it into DH core and possibly replace parts of EntityManagerImpl with it.

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/controllers/ModelController.java
```diff
@@ -76,11 +84,85 @@
...
 
+    void deployModelConfigs() {
+        ManageClient manageClient = hubClientProvider.getHubClient().getManageClient();
+        DatabaseClient finalDatabaseClient = hubClientProvider.getHubClient().getFinalClient();
+        DatabaseClient stagingDatabaseClient = hubClientProvider.getHubClient().getStagingClient();
+
+        JsonNode modelConfigNode = newService().generateModelConfig();
+
+        // Deploy search options
+        String explorerOptions = modelConfigNode.get("search-options").get("explorer").asText();
```
rjrudin, line 105: Let's put each of these deploy steps in a separate private method and then put each call to ML in a try-catch iof the form:

```
try {
     make a single call to ML - e.g. to deploy explorer search options to staging
} catch (Exception ex) {
     throw new RuntimeException("Unable to deploy explorer search options to staging after updating entity models: cause: " + ex.getMessage(), ex);
}
```

The reason for this is that the typical errors we get back from the Manage API and REST API don't give a ton of context and often lead to the conclusion of "Gradle is broken, we must replace it". Gradle isn't being used here, so hopefully it won't be blamed, but providing some context around the source of the error will be helpful.

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/controllers/ModelController.java
```diff
@@ -76,11 +84,85 @@
...
+            queryOptionsManager.writeOptionsAs("exp-" + databaseKind + "-entity-options", Format.XML, explorerOptions);
+            queryOptionsManager.writeOptionsAs(databaseKind + "-entity-options", Format.XML, defaultOptions);
+        });
+
+        // Deploy protected-paths
+        modelConfigNode.get("protected-paths").forEach(jsonNode -> manageClient.postJson("/manage/v2/protected-paths", jsonNode.toString()));
+
+        // Deploy query-rolesets
+        try {
+            modelConfigNode.get("query-rolesets").forEach(jsonNode -> manageClient.postJson("/manage/v2/query-rolesets", jsonNode.toString()));
```
rjrudin, line 121: You need to do a try/catch for each individual postJson call. While we only expect one query roleset for now (pii-reader), there could be more in the future. And if the first one fails because it already exists, we still want to try to deploy the others.

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/controllers/ModelController.java
```diff
@@ -76,11 +84,85 @@
...
+
+        // Deploy index configuration
+        modelConfigNode.get("index").get("database-names").forEach(dbNameNode -> {
+            // Get updated entity index configs
+            JsonNode indexNode = modelConfigNode.get("index").get("index-config");
+
+            try {
+                // Get current db configuration
+                String dbConfig = manageClient.getJson("/manage/v2/databases/" + dbNameNode.asText() + "/properties");
+                JsonNode dbNode = new ObjectMapper().readTree(dbConfig);
```
rjrudin, line 135: Starting with this line, put the rest into a protected method so we can unit-test it easily without depending on ML at all. This is relatively complex logic here, and we want to make sure it behaves correctly. It's easiest to do that if we can pass in our own inputs for the existing dbConfig and the generated indexNode and then verify that they are merged together correctly. Those tests will be much easier to write and will run very quickly - I recommend putting them into a separate test class that doesn't extend anything, can call it "MergeDatabaseConfigTest".

marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/controllers/model/ModelControllerTest.java
```diff
@@ -60,26 +69,114 @@ private void updateModelInfo() {
...
+            assertTrue(indexConfig.contains("someOtherProperty"));
+        });
+    }
+
+    private void loadUnrelatedIndexes() {
+        String indexConfig = "{\n" +
+            "   \"lang\":\"zxx\",\n" +
+            "   \"path-namespace\":[\n" +
+            "      {\n" +
+            "         \"prefix\":\"es\",\n" +
```
rjrudin, line 154: Shouldn't this be a different namespace - isn't this the one that's generated too?

marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/controllers/model/ModelControllerTest.java
```diff
@@ -60,26 +69,114 @@ private void updateModelInfo() {
...
+        catch (JsonProcessingException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    private void assertIndexDeployment() {
+        ManageClient manageClient = getHubClient().getManageClient();
+        Stream.of(getHubConfig().getDbName(DatabaseKind.STAGING), getHubConfig().getDbName(DatabaseKind.FINAL)).forEach(databaseKind -> {
+            String indexConfig = manageClient.getJson("/manage/v2/databases/" + databaseKind + "/properties");
+            assertTrue(indexConfig.contains("testRangeIndexForDHFPROD4704"));
```
rjrudin, line 142: A testing technique - a problem with assertTrue/assertFalse is that if it fails, it's not known why. We really want to see the value of indexConfig to find out what's going on. 

To improve on that, I'd do this:

```
["someProperty", "someOtherProperty", etc].forEach(prop -> {
    assertTrue(indexConfig.contains(property), "Expected " + property + " to be in indexConfig: " + indexConfig);
});
```

That makes debugging a test failure much easier.

marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/controllers/model/ModelControllerTest.java
```diff
@@ -60,26 +69,114 @@ private void updateModelInfo() {
     }
 
     private void updateModelEntityTypes() {
+        // Loading unrelated indexes so that we can check for them after updating entity model
```
rjrudin, line 72: I believe you need to call Installer.applyDatabasePropertiesForTests after each test here because you're modifying the config of the staging/final databases, which could break other tests.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/models/generateModelConfig.sjs
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  // Add search options
+  options["default"] = hent.dumpSearchOptions(entityModels, false);
+  options["explorer"] = hent.dumpSearchOptions(entityModels, true);
+
+  // Add indexes
+  const dbProps = hent.dumpIndexes(entityModels);
+  index["index-config"] = dbProps.toObject();
+  index["database-names"].push(config.STAGINGDATABASE);
```
rjrudin, line 53: I don't think you should add the database-names here - it's assumed that the client knows what those are, which is the case here. If you were doing it here, you arguably should do it for search-options as well. But I don't think it's necessary - the middle tier has a HubClient, so it's good to go.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/models/generateModelConfig.sjs
```diff
@@ -0,0 +1,62 @@
...
+*/
+'use strict';
+
+const consts = require("/data-hub/5/impl/consts.sjs");
+const config = require("/com.marklogic.hub/config.sjs");
+const hent = require("/data-hub/5/impl/hub-entities.xqy");
+
+
+const result = {};
+const protectedPaths = [];
```
rjrudin, line 24: There's no need to define these ahead of time. Just assign properties directly to "result", makes for less code - i.e. declare "result" as:

```
result = {
  protectedPaths: [],
  queryRolesets: [],
  index-config: {},
  options: {}
};
```

That also has a nice benefit of immediately telling the reader what the response object looks like.

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/controllers/ModelController.java
akshaysonvane: Right, updated in here https://github.com/marklogic/marklogic-data-hub/pull/3941/commits/b6777afc22dfc3a5bd4d452ff1d0df426f91e0cc

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/controllers/ModelController.java
akshaysonvane: Good catch. Updated in here b6777af

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/controllers/ModelController.java
akshaysonvane: Added the test.

marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/controllers/model/ModelControllerTest.java
akshaysonvane: Yep, forgot to change it while copying.
Updated in here b6777af

marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/controllers/model/ModelControllerTest.java
akshaysonvane: Makes sense. Updated in here b6777af

marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/controllers/model/ModelControllerTest.java
akshaysonvane: Updated in here b6777af

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/models/generateModelConfig.sjs
akshaysonvane: Updated in here b6777af to use the middle tier for that database names.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/models/generateModelConfig.sjs
akshaysonvane: Makes sense. Updated in here b6777af

marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/controllers/MergeDatabaseConfigTest.java
```diff
@@ -0,0 +1,89 @@
...
+    @Test
+    public void testMergeIndexConfigs() throws JsonProcessingException {
+        String entityServicesRangeIndex = "testEntityServicesRangeIndexForDHFPROD4704";
+        String entityServicesPathIndex = "testEntityServicesPathIndexForDHFPROD4704";
+        String databaseRangeIndex = "testRangeIndexForDHFPROD4704";
+        String databasePathIndex = "testPathIndexForDHFPROD4704";
+        String databaseField1 = "language";
+        String databaseField2 = "forest";
+
+        String indexConfig = "{\n" +
```
rjrudin, line 22: For future reference - ml-app-deployer has a bunch of POJOs to build this stuff, which is a lot easier to write and read than doing inline JSON in Java. See UpgradeProjectTest for examples of using ElementIndex and Database.

marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/controllers/MergeDatabaseConfigTest.java
```diff
@@ -0,0 +1,89 @@
...
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+public class MergeDatabaseConfigTest {
+
+    @Test
+    public void testMergeIndexConfigs() throws JsonProcessingException {
```
rjrudin, line 14: I think you also need a scenario for where the entity indexes already exist. So keep this scenario, and add another one that has the same setup here, and the existing indexes includes the entity indexes.

marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/controllers/model/ModelControllerTest.java
```diff
@@ -60,26 +79,115 @@ private void updateModelInfo() {
...
```
rjrudin, line 82: Food for thought for the future - no change needed now:

"NounTest" - e.g. "ModelControllerTest" - is usually a decent approach for the first couple tests. But that doesn't scale well as the "Noun" becomes able to do more "Verbs". At that points, it's worth refactoring into "Verb1Test", "Verb2Test", etc, and putting them into a package specific to the Noun. 

I think ModelControllerTest isn't too big yet, but when we add more tests, it's likely going to be worth moving to a VerbTest approach.

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/controllers/ModelController.java
```diff
@@ -76,11 +86,132 @@
...
+                }
+                catch (JsonProcessingException e) {
+                    throw new RuntimeException(e);
+                }
+
+                manageClient.putJson("/manage/v2/databases/" + databaseName + "/properties", indexNode.toString());
+            });
+        }
+        catch (Exception e) {
+            throw new RuntimeException("Unable to deploy index configuration after updating entity models.", e);
```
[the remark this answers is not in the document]
akshaysonvane: Makes sense, updated in here https://github.com/marklogic/marklogic-data-hub/commit/da196c85281e73752f8f10bfd994508de7644daa

marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/controllers/MergeDatabaseConfigTest.java
akshaysonvane: Awesome, updated the test to use the POJOS.

marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/controllers/MergeDatabaseConfigTest.java
akshaysonvane: Added the scenario in here da196c8

rjrudin:
@akshaysonvane  I'm going to submit a small commit into your PR to add more assertions in the test; it will include depending on ml-app-deployer 4.0.1 instead of 4.0.0 to take advantage of a little fix I'm making in the Database class in ml-app-deployer, which is helpful for assertions.

rjrudin:
See https://github.com/akshaysonvane/marklogic-data-hub/pull/7

akshaysonvane:
Updated the `MergeDatabaseConfigTest` to use the new PathIndex fix in the Database class.

akshaysonvane:
Updated the PR to fix the `getFlowNames() – com.marklogic.hub.flow.FlowManagerTest` test which is failing on the develop branch.

PR #3946, opened 2020-05-13
DHFPROD-4951: UI is now using new step and flow endpoints

rjrudin:
This intentionally is not trying to make the uploading of files work for ingestion steps, nor are they runnable, as there are separate stories for changing how those features work. 

The tests are failing in Bench.test.tsx, can't figure out why - manual testing of the Bench seems to confirm that everything is working properly. 

Otherwise, the UI changes were fairly simple to make - the only interesting thing was modifying flows.tsx to take advantage of getting all the necessary data from /api/flows instead of having to get all the ingestion and mapping steps too. 

Removed the functions in artifacts/core for settings, along with related tests, as we no longer need a concept of a separate settings document. Will remove the rest later, once we know everything here is working.

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


wooldridge:
@rjrudin, I'm trying the Bench tests and they pass for me, are there still issues? 
Screen Shot 2020-05-15 at 1 57 55 AM

rjrudin:
@wooldridge  Nope, @srinathgit  was able to figure out the fixes for the tests.

rjrudin:
I modified LoadDataController to return an artifact type of "ingestion" instead of "loadData", and uploading files works again (as does running ingestion steps)

PR #3958, opened 2020-05-15
DHFPROD-4597: View Entity Types

brucean52:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


marklogic-data-hub-central/ui/src/assets/mock-data/modeling.js
```diff
@@ -459,4 +459,215 @@ export const editEntityPropertyRequest = { "AnotherModel": {
     }
   }
 }
-}
\ No newline at end of file
+}
+
+export const propertyTableEntities = [
+  {
+      "entityName": "Concept",
```
rjrudin, line 466: This is fine, but if possible, I'd like for us to use Customer/Order across the board as defined in reference-entity-model project. We can absolutely enhance those to account for scenarios that they don't account for yet. It keeps us all on the same sheet of music, which makes communicating/testing/debugging easier.

marklogic-data-hub-central/ui/src/components/modeling/entity-type-table/entity-type-table.test.tsx
```diff
@@ -28,11 +28,11 @@ describe('EntityTypeModal Component', () => {
 
       expect(getByText(/Customer/i)).toBeInTheDocument();
       expect(getByText(/1,000/i)).toBeInTheDocument();
-      expect(getByText(/2020-04-09 14:27/i)).toBeInTheDocument();
+      expect(getByText(/09-Apr-20 14:27/i)).toBeInTheDocument();
```
rjrudin, line 31: @wooldridge  @brucean52 Is DD-MMM-YY the standard for date formatting in the UI? I'm a big fan of YYYY-MM-DD because it's naturally sortable, but many users I've worked with prefer something else, which always disappoints me. Do we have an agreed-upon standard yet for HC?

marklogic-data-hub-central/ui/src/components/modeling/property-table/property-table.module.scss
```diff
@@ -0,0 +1,29 @@
...
+}
+
+.addIcon {
+  cursor: pointer;
+  font-size: 21px;
+  margin-left: 6px;
+  color: #5B69AF;
+}
+
+.addBtnContainer {
```
rjrudin, line 18: Unless we don't control the term "addBtn", I highly, highly, highly recommend avoiding abbreviations like this. They don't serve any purpose other than to make code a little harder to read. Plus, more verbose names are easier to search for when looking for references.

marklogic-data-hub-central/ui/src/config/tooltips.config.js
```diff
@@ -92,13 +92,22 @@ const AdvCustomOtherTooltips = {}
 
 const ModelingTooltips = {
   entityTypeName: 'Click to view or edit settings for this entity type.',
+  addNewEntity: 'Click here to add a new entity type',
```
rjrudin, line 95: This does not end with a period - which is what I'd vote for, but the other messages all end with a period, so I'm thinking this should end with a period too for consistency.

marklogic-data-hub-central/ui/src/components/modeling/entity-type-table/entity-type-table.test.tsx
brucean52: Let me get confirmation on this. The modeling design wiki shows YYYY-MMM-DD. The figma design shows the changes I made here

brucean52: Last processed now shows time relative time in the table, which is sortable. Hover text over time shows date in DD-MMM-YY

PR #3963, opened 2020-05-15
DHFPROD-4956:Migrate flows and mappings to flows and steps

srinathgit:
### Description
Migrate flows and mappings to flows and steps
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/flow/impl/FlowMigrator.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/flow/impl/FlowMigrator.java
@@ -1,0 +1,214 @@
+/*
+ * Copyright (c) 2020 MarkLogic Corporation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.marklogic.hub.flow.impl;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.ObjectWriter;
+import com.fasterxml.jackson.databind.node.JsonNodeFactory;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import com.marklogic.hub.FlowManager;
+import com.marklogic.hub.HubConfig;
+import com.marklogic.hub.HubProject;
+import com.marklogic.hub.MappingManager;
+import com.marklogic.hub.error.DataHubProjectException;
+import com.marklogic.hub.flow.Flow;
+import com.marklogic.hub.impl.FlowManagerImpl;
+import com.marklogic.hub.impl.MappingManagerImpl;
+import com.marklogic.hub.mapping.Mapping;
+import com.marklogic.hub.step.StepDefinition;
+import com.marklogic.hub.step.impl.Step;
+import org.apache.commons.io.FileUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Map;
+import java.util.Set;
+
+
+/**
+ * Class for migrating pre-5.3.0 flows to  5.3.0 and above versions
+ */
+
+public class FlowMigrator {
+
+    private HubProject hubProject;
+    private MappingManager mappingManager;
+    private FlowManager flowManager;
+    protected final Logger logger = LoggerFactory.getLogger(this.getClass());
+    ObjectMapper mapper = new ObjectMapper();
+
+    public FlowMigrator(HubConfig hubConfig){
+        hubProject = hubConfig.getHubProject();
+        mappingManager = new MappingManagerImpl(hubConfig);
+        flowManager = new FlowManagerImpl(hubConfig, mappingManager);
+    }
+
+    public void migrateFlows(){
+        //Backup flows and mappings
+        Path migratedFlows = hubProject.getProjectDir().resolve("migrated-flows");
+        try {
+            migratedFlows.toFile().mkdirs();
+            FileUtils.copyDirectory(hubProject.getFlowsDir().toFile(), migratedFlows.resolve("flows").toFile());
+            FileUtils.copyDirectory(hubProject.getHubMappingsDir().toFile(), migratedFlows.resolve("mappings").toFile());
+            logger.info("The original flows and mappings are backed up in migrated-flows/flows and migrated-flows/mappings"+
+                " directory respectively.");
+        } catch (Exception e) {
+            throw new RuntimeException("Couldn't migrate flows as backing up flows failed : " + e.getMessage());
+        }
+
+        Path stepsDir = hubProject.getProjectDir().resolve("steps");
+        Path ingestionDir = stepsDir.resolve(StepDefinition.StepDefinitionType.INGESTION.toString());
+        Path mappingDir = stepsDir.resolve(StepDefinition.StepDefinitionType.MAPPING.toString());
+
+        try {
+            ingestionDir.toFile().mkdirs();
+            mappingDir.toFile().mkdirs();
+        } catch (Exception e) {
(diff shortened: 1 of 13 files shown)


marklogic-data-hub/src/test/java/com/marklogic/hub/flow/impl/FlowMigratorTest.java
```diff
@@ -0,0 +1,63 @@
...
+        Assertions.assertTrue(migratedFlows.toFile().exists());
+        Assertions.assertTrue(migratedFlows.resolve("flows").toFile().listFiles().length > 0);
+        Assertions.assertTrue(migratedFlows.resolve("mappings").toFile().listFiles().length > 0);
+
+        verifyFiles(getResourceFile("flow-migration-test/keys/flows"), hubProject.getFlowsDir().toFile());
+        verifyFiles(getResourceFile("flow-migration-test/keys/steps/ingestion"), hubProject.getProjectDir().resolve("steps").resolve("ingestion").toFile());
+        verifyFiles(getResourceFile("flow-migration-test/keys/steps/mapping"), hubProject.getProjectDir().resolve("steps").resolve("mapping").toFile());
+    }
+
+    private void verifyFiles(File expectedDir, File actualDir) {
```
rjrudin, line 51: I ran the test locally (it passed), but I think two changes should be made to the test:

1. Narrow down the test files to the bare minimum of steps needed to verify that everything is working
1. Add assertions to verify that the step references are correct in every flow, and that each step document was created correctly

For the first item above - the ingestion_only-flow.flow.json file is testing 7 identical scenarios, because every one of the 7 steps is the same from a QA perspective. So we only need one non-custom ingestion step that has a fileLocations. We also need a non-custom ingestion step that doesn't have inputFileType in it so we can verify that the migrator adds sourceFormat=json. 

As for the logic of - "Should this step be migrated?" - the most effective way to do that is to make "stepRequiresMigration" a protected method and that pass different kinds of steps into it, verifying that the result is true or false. 

For the second item - I think it's best to make an assertion on every stepId in the migrated flow document, and also on every property of every migrated step. We specifically want to verify that everything under "options" was migrated to top-level properties; that "fileLocations" was handled correctly; and that "options/mapping" was *not* migrated (it currently is being migrated, but should not be). 

I also noticed that a "namespaces: {}" was added, even though no namespaces exist in the mappings. If that's something we need to account for, then we need a mapping artifact that has namespaces populated so we can verify that the special handling for it worked. 

Note that we really do not want any issues during this migration process; the last thing we want for new users on 5.3 is to migrate their flows and then run into a problem that requires involving Support and makes them think that 5.3 doesn't work. So we need to be certain to cover every scenario and ensure every line of the migrated flows/steps is correct.

marklogic-data-hub/src/test/java/com/marklogic/hub/flow/impl/FlowMigratorTest.java
```diff
@@ -0,0 +1,63 @@
...
+        flowMigrator.migrateFlows();
+        Path migratedFlows = hubProject.getProjectDir().resolve("migrated-flows");
+
+        Assertions.assertTrue(migratedFlows.toFile().exists());
+        Assertions.assertTrue(migratedFlows.resolve("flows").toFile().listFiles().length > 0);
+        Assertions.assertTrue(migratedFlows.resolve("mappings").toFile().listFiles().length > 0);
+
+        verifyFiles(getResourceFile("flow-migration-test/keys/flows"), hubProject.getFlowsDir().toFile());
+        verifyFiles(getResourceFile("flow-migration-test/keys/steps/ingestion"), hubProject.getProjectDir().resolve("steps").resolve("ingestion").toFile());
+        verifyFiles(getResourceFile("flow-migration-test/keys/steps/mapping"), hubProject.getProjectDir().resolve("steps").resolve("mapping").toFile());
```
rjrudin, line 48: I realized the story isn't clear here - the mapping artifacts should be moved, not copied. That is assuming that we don't need them anymore, which I think is true. I'll double check on 5.0.x "default" mappings.

marklogic-data-hub/src/main/java/com/marklogic/hub/flow/impl/FlowMigrator.java
```diff
@@ -0,0 +1,191 @@
...
+                }
+            }
+        });
+    }
+
+    // Only create step artifacts for ootb ingestion and mapping steps. Other steps(including custom ingestion and mapping)
+    // will be inline
+    private boolean stepRequiresMigration(Step step) {
+        return (StepDefinition.StepDefinitionType.MAPPING.equals(step.getStepDefinitionType()) && "entity-services-mapping".equalsIgnoreCase(step.getStepDefinitionName())) ||
+            (StepDefinition.StepDefinitionType.INGESTION.equals(step.getStepDefinitionType()) &&  "default-ingestion".equalsIgnoreCase(step.getStepDefinitionName()));
```
bsrikan, line 151: ingestion step seems to be show up as "Unknown" in the the flow, since its expecting a loadData config

marklogic-data-hub/src/main/java/com/marklogic/hub/flow/impl/FlowMigrator.java
```diff
@@ -0,0 +1,191 @@
...
+                    logger.error("Flow artifact " + flowFile.toString() + " creation failed: " + e.getMessage());
+                }
+            }
+        });
+    }
+
+    // Only create step artifacts for ootb ingestion and mapping steps. Other steps(including custom ingestion and mapping)
+    // will be inline
+    private boolean stepRequiresMigration(Step step) {
+        return (StepDefinition.StepDefinitionType.MAPPING.equals(step.getStepDefinitionType()) && "entity-services-mapping".equalsIgnoreCase(step.getStepDefinitionName())) ||
```
bsrikan, line 150: endpoint http://localhost:8080/api/artifacts/mapping seems to return all the versions of a mapping. Shouldnt it return only the latest version. UI ends up showing all the versions which is not desired.
image

srinathgit: It will return whatever version is specified in the flow. Also, "mapping" object will be removed from the mapping step

marklogic-data-hub/src/main/java/com/marklogic/hub/flow/impl/FlowMigrator.java
```diff
@@ -0,0 +1,191 @@
...
+                    else {
+                        logger.info("Step " + step.getName() + " is not an out of the box ingestion or mapping step. It will remain inline inside the flow artifact");
+                        newSteps.set(entry.getKey(), mapper.valueToTree(step));
+                    }
+                }
+                newFlow.set("steps",newSteps);
+                File flowFile = Paths.get(hubProject.getFlowsDir().toString(), flow.getName() + FlowManager.FLOW_FILE_EXTENSION).toFile();
+                try{
+                    writer.writeValue(flowFile, newFlow);
+                    logger.info("Flow " + flowFile.toString() + " successfully migrated.");
```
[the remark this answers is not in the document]
srinathgit: The flows should be run after DHFPROD-4951 and DHFPROD-4952 are merged into develop though implementation of DHFPROD-4956 doesn't have dependency on them

marklogic-data-hub/src/main/java/com/marklogic/hub/flow/impl/FlowMigrator.java
srinathgit: This has to be verified after DHFPROD-4951 and DHFPROD-4952 are merged

marklogic-data-hub/src/test/java/com/marklogic/hub/flow/impl/FlowMigratorTest.java
srinathgit: Since there will be no default-mapping steps and all mappings are migrated, ./mappings dir is removed.

bsrikan: @srinathgit when I tested this PR, ./mappings dir was not removed.

rjrudin:
@srinathgit Once 4951 is merged into develop, I'll rebase this PR off develop locally and do some manual testing in the UI of the migrated results.

marklogic-data-hub/src/test/java/com/marklogic/hub/flow/impl/FlowMigratorTest.java
srinathgit: It is removed in the latest commit

bsrikan:
I dont see migrated mapping and ingestion under steps folder deployed to ML if I run hubDeployAsDeveloper after running hubMigrateFlows.

srinathgit:
> I dont see migrated mapping and ingestion under steps folder deployed to ML if I run hubDeployAsDeveloper after running hubMigrateFlows.

Deployment story is DHFPROD-4952. Once a PR for that is merged, you will see the ingestion and mapping steps deployed as well.

PR #3964, opened 2020-05-18
DHFPROD-4916: Making saved query name to be unique

rahulvudutala:
### Description

- Making saved query name to be unique backend changes
- Making saved query name to be unique front end changes
- Unit and integration tests for FE
- New mlUnittest and moved existing EntitySearchServiceTest JUnit to mlUnittest.
- Temporary changes to integration tests

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


marklogic-data-hub-central/ui/e2e/cypress/integration/authorization.spec.tsx
```diff
@@ -42,7 +42,8 @@ describe('login', () => {
     cy.loginAsDeveloper()
     .wait(500)
     .url()
-    .should('include', '/home');
+     .should('include', '/tile')
+    // .should('include', '/home');
```
rjrudin, line 46: Is there any benefit in keeping the commented-out code? I think it's almost always best to delete commented-out code - in the rare case something might be needed later, it can always be retrieved from version control.

marklogic-data-hub-central/ui/src/components/queries/queries.tsx
```diff
@@ -59,17 +59,21 @@ const Query = (props) => {
...
-            handleError(error);
-        } finally {
-            props.setIsLoading(false);
-            resetSessionTime();
-        }
+        props.setIsLoading(true);
+        await creatNewQuery(query);
+        setOpenSaveModal(false);
+        getSaveQueries();
+        // try {
```
rjrudin, line 66: Same thing here - why not just delete this code? What's the benefit of keeping it around? It's in version history if it's ever needed again.

marklogic-data-hub-central/ui/src/components/queries/saving/edit-save-query/save-changes-modal.tsx
```diff
@@ -88,18 +94,28 @@ const SaveChangesModal: React.FC<Props> = (props) => {
...
+                    setAllSearchFacets(facets);
+                    props.setSaveChangesModalVisibility();
+                    applySaveQuery(searchOptions.query, searchOptions.entityTypeIds, facets, queryName);
+                    props.setCurrentQueryDescription(queryDescription);
                 }
             } catch (error) {
-                handleError(error);
+                if (error.response.status === 400) {
+                    if (error.response.data.hasOwnProperty('message')) {
+                        setErrorMessage(error['response']['data']['message']);
```
rjrudin, line 105: This seems to be repeated - it's in 3 different places. 

@wooldridge @xnikhil08  It seems there should be a way to reduce this duplication so that each "catch" block doesn't have to do so much work? The problem with repeating this work in multiple places is that if we ever want to change how we handle e.g. 400 errors, we have N places to modify as opposed to just one.

marklogic-data-hub-central/ui/src/components/queries/saving/save-query-modal/save-query-modal.tsx
```diff
@@ -57,16 +67,31 @@ const SaveQueryModal: React.FC<Props> = (props) => {
...
-
+            try {
+                await props.saveNewQuery(queryName, queryDescription, facets);
+                props.setSaveNewIconVisibility(false);
+                props.setSaveModalVisibility();
+                applySaveQuery(searchOptions.query, searchOptions.entityTypeIds, facets, queryName);
+                props.setCurrentQueryName(queryName);
+                props.setCurrentQueryDescription(queryDescription);
+            } catch (error) {
+                if (error.response.status === 400) {
```
rjrudin, line 78: This is the 4th occurrence of very similar logic for handling a 400 - hopefully there's a way to reduce this duplication??

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/entitySearch/saveSavedQuery.sjs
```diff
@@ -43,7 +43,15 @@ if (queryDocument.savedQuery.propertiesToDisplay == null || queryDocument.savedQ
     ds.throwBadRequest("Entity type properties to be displayed cannot be empty");
 }
 
-let id = queryDocument.savedQuery.id;
+const id = queryDocument.savedQuery.id;
+const positiveQuery = cts.andQuery([cts.collectionQuery("http://marklogic.com/data-hub/saved-query"), cts.jsonPropertyValueQuery("name", queryDocument.savedQuery.name),
```
rjrudin, line 47: A general technique for making queries easier to read is to put each query on a separate line - e.g.

```
cts.andQuery([
  cts.collectionQuery(".."),
  cts.jsonPropertyValueQuery(...),
  cts.jsonPropertyValueQuery(...)
])
```

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/entitySearch/saveSavedQuery.sjs
```diff
@@ -43,7 +43,15 @@ if (queryDocument.savedQuery.propertiesToDisplay == null || queryDocument.savedQ
...
+    cts.jsonPropertyValueQuery("owner", xdmp.getCurrentUser())]);
+const negativeQuery = cts.documentQuery("/saved-queries/" + id + ".json");
+const queryNameExists = cts.exists(cts.andNotQuery(positiveQuery, negativeQuery));
+if(queryNameExists) {
+    ds.throwBadRequest(`A query already exists with a name of ${queryDocument.savedQuery.name}`);
```
rjrudin, line 52: I think this message needs to capture the fact that the current user already has the query; otherwise, the user could interpret this message as implying that no two users can use the same named queries. 

I recommend: "You already have a saved query with a name of ${the name}".

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/data-services/entitySearch/saveSavedQueryTest.sjs
```diff
@@ -38,45 +40,150 @@ var saveQuery = JSON.stringify({
...
 });
 
-function invokeService(saveQuery) {
-  return fn.head(xdmp.invoke(
-      "/data-hub/5/data-services/entitySearch/saveSavedQuery.sjs",
-      {
-        "saveQuery": saveQuery
-      }
-  ));
+function beforeEach() {
```
rjrudin, line 43: I think this is an indicator that multiple test modules should be created. That's because marklogic-unit-test has built-in support for "do something before every test" - just make a setup.xqy module.

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/data-services/entitySearch/saveSavedQueryTest.sjs
```diff
@@ -38,45 +40,150 @@ var saveQuery = JSON.stringify({
...
+  declareUpdate();
+  xdmp.collectionDelete("http://marklogic.com/data-hub/saved-query");
```
rjrudin, line 45: Just call "reset-hub()" in data-hub-test-helper.xqy instead. That ensures that everything but the "hub-core-artifact" documents are retained.

marklogic-data-hub-central/ui/src/components/queries/saving/edit-save-query/save-changes-modal.tsx
rahulvudutala: @rjrudin the values that are set in the catch are specific to the component. Apart from setErrorMessage state, there are other states that are set in different catch blocks. This error handling is specific to the child components of the Query component. There is also a handleError generic error handler, which handles 400 error.

brucean52: Yeah, I'm not really sure of better alternatives. async await in a try catch block is already a pretty lean. One refactor I would recommend is combining the edit and save modals into one since they share inputs.

PR #3969, opened 2020-05-18
DHFPROD-4674: Add hub-central-writer role

ryanjdew:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/examples/reference-entity-model/src/main/ml-config/security/users/hub-developer.json
+++ b/examples/reference-entity-model/src/main/ml-config/security/users/hub-developer.json
@@ -4,6 +4,7 @@
   "password": "%%testPassword%%",
   "role": [
     "data-hub-developer",
-    "hub-central-user"
+    "hub-central-user",
+    "hub-central-mapping-writer"
   ]
 }
--- a/marklogic-data-hub-central/ui/src/components/tiles/toolbar.tsx
+++ b/marklogic-data-hub-central/ui/src/components/tiles/toolbar.tsx
@@ -1,6 +1,8 @@
-import React from 'react';
+import React, {useContext} from 'react';
 import { Tooltip } from 'antd';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
+import { AuthoritiesContext } from "../../util/authorities";
+
 import styles from './toolbar.module.scss';
 import './toolbar.scss';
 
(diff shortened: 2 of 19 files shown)


marklogic-data-hub/src/main/resources/hub-internal-config/security/privileges/data-hub-write-mapping.json
```diff
@@ -0,0 +1,5 @@
+{
```
rjrudin, line 1: Thoughts on using xdmp.securityAssert in the step endpoints? Would require conditionals on stepDefinitionType

marklogic-data-hub-central/ui/src/components/tiles/toolbar.tsx
```diff
@@ -12,23 +14,39 @@ interface Props {
 const Toolbar: React.FC<Props>  = (props) => {
 
     const tiles = props.tiles; // config/tiles.config.ts
+    const authorityService = useContext(AuthoritiesContext);
+    const idToAuthoritiesNeeded = {
+        load: () => authorityService.canReadLoadData(),
+        model: () => authorityService.canReadEntityModel(),
+        curate: () => authorityService.canReadMapping(),
+        run: () => authorityService.canReadFlow(),
+        explore: () => true
```
bsrikan, line 23: So even a hub-central-user can access Explore tile?

bsrikan:
A mapping-reader and writer user dont see tooltip:
image

A hub-central-user doesnt see any tooltip.

examples/reference-entity-model/src/main/ml-config/security/users/hub-developer.json
```diff
@@ -4,6 +4,7 @@
   "password": "%%testPassword%%",
   "role": [
     "data-hub-developer",
-    "hub-central-user"
+    "hub-central-user",
```
rjrudin, line 7: I think data-hub-developer should gain the data-hub:writeMapping privilege, which then means we don't have to modify these users or the cypress users that already have data-hub-developer. That privilege can then be used in xdmp.securityAssert calls in the DS endpoints.

marklogic-data-hub/src/main/resources/hub-internal-config/security/roles/data-hub-mapping-writer.json
```diff
@@ -1,4 +1,5 @@
 {
   "role-name": "data-hub-mapping-writer",
-  "description": "Permits updating mapping documents"
+  "description": "Permits updating mapping documents",
+  "privilege": []
```
rjrudin, line 4: I am wondering if this role should gain the data-hub:writeMapping privilege. Then we don't even need to modify data-hub-developer. It would seem surprising if this role doesn't inherit that privilege - thoughts?

marklogic-data-hub-central/ui/src/components/tiles/toolbar.tsx
rjrudin: I think that makes sense for now, until we have requirements stating otherwise.

rjrudin: @ryanjdew  @xnikhil08 Does this basically take care of the UI portion of https://project.marklogic.com/jira/browse/DHFPROD-4820 - i.e. don't show the Model tile unless canReadEntityModel returns true?

ryanjdew: @rjrudin It shows the tile still without the authority, but the tooltip states they need to contact their security admin to get access, and selecting the tile does nothing in that case.

marklogic-data-hub/src/main/resources/hub-internal-config/security/roles/data-hub-mapping-writer.json
ryanjdew: The only issue I have with that is that if we added the privileges to the lower-level data-hub-* roles then that means that for example when the hub-central-step-runner gets the reader roles necessary the privileges will come over with them. With the current roles defined, we could potentially have the writer privileges on the lower level roles and not the read. 

I'm thinking doing any assert on reading documents would be redundant anyway given the document permissions we're setting on the documents.

ryanjdew:
Note: Needed to adjust data-hub-mapping-writer role to have the data-hub-step-definition-reader role with the recent changes using the step data services endpoints.

PR #3970, opened 2020-05-18
DHFPROD-4456: Ensure all tools handle middle tier failure

wooldridge:
### Description

Update HC tool views such that all views handle cases when middle tier has gone down (a 504 error results and an error dialog appears in the UI). This includes when clicking the Download button under Manage Project. Here is a screencast:

https://drive.google.com/file/d/1LB2WTXT75m4O1Z-E0KBkUIrXIxcJRPMt/view

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/src/pages/Curate.tsx
+++ b/marklogic-data-hub-central/ui/src/pages/Curate.tsx
@@ -39,8 +39,8 @@ const Curate: React.FC = () => {
               setEntityModels({...models});
           }
         } catch (error) {
-            let message = error;
-            console.error('Error while fetching entities Info', message);
+            console.error('Error fetching entities', error);
+            handleError(error);
         } finally {
           resetSessionTime();
         }
--- a/marklogic-data-hub-central/ui/src/pages/Run.tsx
+++ b/marklogic-data-hub-central/ui/src/pages/Run.tsx
@@ -82,7 +82,8 @@ const Run = (props) => {
                 setFlows(response.data);
             }
         } catch (error) {
-            console.error('********* ERROR', error);
+            console.error('Error getting flows', error);
+            handleError(error)
         } finally {
           resetSessionTime();
         }
--- a/marklogic-data-hub-central/ui/src/util/environment.tsx
+++ b/marklogic-data-hub-central/ui/src/util/environment.tsx
@@ -20,7 +20,7 @@ export function setEnvironment()  {
 export function getEnvironment():any {
     let env: any;
     env = localStorage.getItem('environment');
-    if (env) {
+    if(env) {
         return JSON.parse(env);
     }
     else{


brucean52:
> On 3000 it seems to work as expected. The issue really is on 8080.

If you shut down middle tier, you also shut down the UI if you are using 8080. Not sure how to test this, or if it's necessary.

bsrikan:
> > On 3000 it seems to work as expected. The issue really is on 8080.
> 
> If you shut down middle tier, you also shut down the UI if you are using 8080. Not sure how to test this, or if it's necessary.

UI is expected to shutdown, but it doesnt seem so. I can click about different tiles, open modal, save something etc. The issue here is since middle tier is down, application should show the 504 error we get on port 3000. We should see the same behavior on 3000 & 8080 right.

wooldridge:
The following Jira ticket needs to be rewritten or replaced to describe the current problem and a proposed solution (FYI @sbayatpur): https://project.marklogic.com/jira/browse/DHFPROD-4456

If this PR solves a problem, and I think it does--it adds consistent async error handling across the page views--it should be merged if it passes tests. 

@bsrikan @brucean52 thoughts?

brucean52:
I think that is due to the web app being cached by the browser. If you shutdown 3000, 

> > > On 3000 it seems to work as expected. The issue really is on 8080.
> > 
> > 
> > If you shut down middle tier, you also shut down the UI if you are using 8080. Not sure how to test this, or if it's necessary.
> 
> UI is expected to shutdown, but it doesnt seem so. I can click about different tiles, open modal, save something etc. The issue here is since middle tier is down, application should show the 504 error we get on port 3000. We should see the same behavior on 3000 & 8080 right.

I see the same behavior if I shutdown port 3000. The app is still cached in the web browser, but if you navigate to a different page or try to hit an endpoint it'll be blank.

bsrikan:
Sure @wooldridge . The ticket was created when Install was still a thing in HC. The steps for currnet HC are updated in the comments. 
https://project.marklogic.com/jira/browse/DHFPROD-4456?focusedCommentId=173078&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-173078

Latest screencast: https://www.screencast.com/t/CvFg5eUr4
 Pasting the steps here. Issue was originally filed for 8080 and still is an issue on 8080. Somehow works on 3000, but the app is going to be available on 8080 and so should work there as expected:

    Run mlDeploy to install DH
    Build and launch HC  by running bootRun or the war file
    Login
    Kill the bootRun or the process running war file

Session timeout appears for the first time as usual after 4mins or so and never goes away. The 30 sec timeout keeps resetting to 30 after reaching 0 in an infinite loop. Hitting Logout also resets the timeout to 30sec. Attached screenshot showing network request and a screen recording.

inder:
@wooldridge @bsrikan how do we move this ticket along? it has been stuck here for 2 weeks.

bsrikan:
It is still a bug for me. The issue is for whatever reason, if the process that is running HC, gets killed, until it is brought back up, the application seems to work as usual because of caching, but no request is sent to backend and so such requests silently fail ie network request log will show error but UI wont, logout doesnt work either. If user refreshes the browser then page doesnt load, which doesnt really convey what is going on.

Expecting/suggesting some sort of redirect to an error page, so we can gracefully handle this situation, if possible. 
Thiss PR, as is, handles the error on dev server ie port 3000, but not the production port 8080. I am ok with whatever we decide to do with this and so deferring to @sbayatpur

PR #3971, opened 2020-05-19
DHFPROD-4819: And DHFPROD-4820 - New roles for entity models

rjrudin:
Created read/write-entity-model privileges, and adding them to the similarly-named existing DH roles. That avoids having to modify any of the other DH roles, like data-hub-developer/operator. 

Also improved HubProjectImpl so that we no longer need to record every role/privilege that we create.

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [ ] Added Tests
  

- ##### Reviewer:

- [ ] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/controllers/ModelController.java
+++ b/marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/controllers/ModelController.java
@@ -26,6 +26,7 @@ import io.swagger.annotations.ApiOperation;
 import io.swagger.annotations.ApiParam;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
+import org.springframework.security.access.annotation.Secured;
 import org.springframework.stereotype.Controller;
 import org.springframework.web.bind.annotation.*;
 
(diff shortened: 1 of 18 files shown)


marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/controllers/ModelController.java
```diff
@@ -40,20 +41,23 @@
...
+    @Secured("ROLE_readEntityModel")
     public ResponseEntity<List<JsonNode>> getLatestJobInfoForAllModels() {
         return ResponseEntity.ok(newModelManager().getLatestJobInfoForAllModels());
     }
 
     @RequestMapping(value = "/primaryEntityTypes", method = RequestMethod.GET)
     @ResponseBody
     @ApiOperation(value = "Get primary entity types; does not include entity definitions that are considered 'structured' types", response = PrimaryEntityTypeList.class)
+    @Secured("ROLE_readEntityModel")
     public ResponseEntity<JsonNode> getPrimaryEntityTypes() {
```
ryanjdew, line 61: As being discussed in other threads, we may need to leave the Secured annotation off of this endpoint for the sake of the Curate related roles.

PR #4005, opened 2020-05-27
DHFPROD-4388: Changes in Advanced Settings dialog

xnikhil08:
### Description
It contains front-end and backend changes for DHFPROD-4388.

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


marklogic-data-hub-central/ui/src/components/advanced-settings/advanced-settings-dialog.module.scss
```diff
@@ -76,6 +83,18 @@ span.readOnlyLabel {
     margin-bottom: 8px !important;
 }
 
+.formItemTgtCollections {
```
rjrudin, line 86: I mentioned this in another PR - I always recommend spelling out words, with the possible exception of local variables that have a very small scope. More verbose names are also useful when searching for references to functions/classes/etc, as you get fewer false positives. So let's not do "src" and "tgt" - let's use "source" and "target", the same words we normally use (using these abbreviations also raises the question of why we abbreviate some words but not others - like if we're doing src/tgt, why not do colls instead of collections? Not abbreviating means we don't have to think through that or explain it).

marklogic-data-hub-central/ui/src/components/advanced-settings/advanced-settings-dialog.tsx
```diff
@@ -56,7 +55,10 @@ const ActivitySettingsDialog = (props) => {
 
   const tgtDatabaseOptions = ['data-hub-STAGING','data-hub-FINAL'];
 
-  const provGranOptions = ['coarse', 'off'];
+  const provGranOptions = {
```
rjrudin, line 58: Same thing here - let's do provenanceGranularityOptions. A benefit - let's say someone wants to search the codebase for all references to granularity options. One approach would be to search for "granularity", ignoring case. That search won't find this code though, which could lead to a mistake being made. But if we use the same term throughout the codebase, it's easy to understand and easy to search for references.

marklogic-data-hub-central/ui/src/components/advanced-settings/advanced-settings-dialog.module.scss
xnikhil08: Done

marklogic-data-hub-central/ui/src/components/advanced-settings/advanced-settings-dialog.tsx
xnikhil08: Done.

ryanjdew:
Some of the tests were failing due to data-hub-operator not having the data-hub-common role, which I've now added. Also, there was clean up of the older separate settings that caused a couple of test failures. I've fixed those tests as well.

PR #4007, opened 2020-05-27
DHFPROD-4932:Add 'outputURIPrefix' in ingestion step, replace "OutputURI Replacement" from HC

srinathgit:
### Description
Add 'outputURIPrefix' in ingestion step, replace "OutputURI Replacement" from HC
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/src/components/load-data/new-data-load-dialog/new-data-load-dialog.test.tsx
+++ b/marklogic-data-hub-central/ui/src/components/load-data/new-data-load-dialog/new-data-load-dialog.test.tsx
@@ -17,8 +17,8 @@ describe('New/edit load data configuration', () => {
     jest.clearAllMocks();
   });
 
-  test('fields non-Delimited Text render', () => {
-    const { baseElement, container, debug, queryAllByPlaceholderText } = render(<BrowserRouter><NewDataLoadDialog newLoad={true}
+  test('fields non-Delimited Text render', async () => {
+    const { debug, baseElement, queryAllByText, getAllByLabelText, queryAllByPlaceholderText, getByText } = render(<BrowserRouter><NewDataLoadDialog newLoad={true}
                                                            title={'Title'}
                                                            setNewLoad={() => {}}
                                                            createLoadDataArtifact={() => {}}
@@ -33,7 +33,15 @@ describe('New/edit load data configuration', () => {
     expect(baseElement.querySelector('#otherSeparator')).not.toBeInTheDocument();
     expect(baseElement.querySelector('#fileUpload')).toBeInTheDocument();
     expect(baseElement.querySelector('#targetFormat')).toBeInTheDocument();
-    expect(baseElement.querySelector('#outputUriReplacement')).toBeInTheDocument();
+    expect(baseElement.querySelector('#outputUriPrefix')).toBeInTheDocument();
+    expect(queryAllByText("Target URI Preview:").length ).toEqual(0);
+    expect(queryAllByPlaceholderText('Enter URI Prefix')[0]).toBeInTheDocument();
+    let tooltip  = getAllByLabelText('icon: question-circle');
+    //should be the last field in the form
+    fireEvent.mouseOver(tooltip[tooltip.length-1]);
+    await waitForElement(() => getByText("The prefix you want for the URIs of the loaded documents. Example: If your prefix is /rawData/ and you load a file called customer1.json, the URI of the loaded document becomes /rawData/customer1.json."))
+    expect(getByText("Target Format:")).toHaveTextContent('Target Format: *')
+    expect(getByText("Output URI Prefix:")).toHaveTextContent('Output URI Prefix:')
   });
 
   test('fields with Delimited Text render', () => {
@@ -53,7 +61,8 @@ describe('New/edit load data configuration', () => {
     expect(baseElement.querySelector('#otherSeparator')).toBeInTheDocument();
     expect(baseElement.querySelector('#fileUpload')).toBeInTheDocument();
     expect(baseElement.querySelector('#targetFormat')).toBeInTheDocument();
-    expect(baseElement.querySelector('#outputUriReplacement')).toBeInTheDocument();
+    expect(baseElement.querySelector('#outputUriReplacement')).not.toBeInTheDocument();
+    expect(baseElement.querySelector('#outputUriPrefix')).toBeInTheDocument();
   });
 
   test('set data should only be called once when adding multiple files', async () => {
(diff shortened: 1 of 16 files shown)


marklogic-data-hub-central/ui/src/components/load-data/new-data-load-dialog/new-data-load-dialog.test.tsx
```diff
@@ -33,7 +33,16 @@ describe('New/edit load data configuration', () => {
     expect(baseElement.querySelector('#otherSeparator')).not.toBeInTheDocument();
     expect(baseElement.querySelector('#fileUpload')).toBeInTheDocument();
     expect(baseElement.querySelector('#targetFormat')).toBeInTheDocument();
-    expect(baseElement.querySelector('#outputUriReplacement')).toBeInTheDocument();
+    expect(baseElement.querySelector('#outputUriReplacement')).not.toBeInTheDocument();
```
rjrudin, line 36: Generally, I wouldn't bother with these assertions that verify that things were removed. Someone will wonder 6 months later - why is it important that we assert this? I think it's sufficient to just remove the feature.

marklogic-data-hub-central/ui/src/components/load-data/new-data-load-dialog/new-data-load-dialog.tsx
```diff
@@ -14,20 +14,17 @@ const NewDataLoadDialog = (props) => {
   const [inputFilePath, setInputFilePath] = useState(props.stepData && props.stepData.inputFilePath ? props.stepData.inputFilePath : '');
   const [srcFormat, setSrcFormat] = useState(props.stepData && props.stepData != {} ? props.stepData.sourceFormat : 'json');
   const [tgtFormat, setTgtFormat] = useState(props.stepData && props.stepData != {} ? props.stepData.targetFormat : 'json');
-  const [outUriReplacement, setOutUriReplacement] = useState(props.stepData && props.stepData != {} ? props.stepData.outputURIReplacement : '');
+  const [outUriPrefix, setOutUriPrefix] = useState(props.stepData && props.stepData != {} ? props.stepData.outputURIPrefix : '');
```
rjrudin, line 17: Can this be setOutputUriPrefix instead of setOutUriPrefix? I am generally against any sort of abbreviations like this - same goes for "src" and "tgt" instead of "source" and "target". If this were a local variable in a function that's just a few lines long, I think extremely short abbreviations can work. But otherwise, we should use regular words, no need for abbreviations that can potentially confuse.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/step/saveStep.sjs
```diff
@@ -35,7 +35,15 @@ let existingStep = fn.head(cts.search(cts.andQuery([
 
 if (existingStep) {
   xdmp.trace(consts.TRACE_STEP, `Step with name ${stepName} and type ${stepDefinitionType} already exists, so will update`);
-  let updatedStep = Object.assign(existingStep.toObject(), stepProperties);
+  const isIngestionStep = "ingestion" === stepDefinitionType.toLowerCase() ? true : false;
```
rjrudin, line 38: Can we do this in the loadData.sjs module instead? This endpoint should be as generic as possible; ideally, the step-type-specific logic is all in modules specific to a step type. 

Looks like validateArtifact would be a simple way to do it.

marklogic-data-hub/src/main/java/com/marklogic/hub/cli/client/CommandLineFlowInputs.java
```diff
@@ -111,10 +114,20 @@
                 runFlowString.append("\n\t\tInput File Path: " + inputFilePath);
                 fileLocations.put("inputFilePath", inputFilePath);
             }
+
+            if (outputURIPrefix != null) {
+                if(outputURIReplacement != null){
```
rjrudin, line 119: Won't WriteStepRunner throw this error? I think it's fine to just handle it in one place.

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/data-services/step/convertIngestionStep.sjs
```diff
@@ -70,6 +71,7 @@ assertions.push(
   test.assertEqual("xml", flowStep.fileLocations.inputFileType),
   test.assertEqual("data-sets/CustomerLoadJSON", flowStep.fileLocations.inputFilePath),
   test.assertEqual(".*CustomerLoadJSON,'/customer'", flowStep.fileLocations.outputURIReplacement),
+  test.assertEqual("/customer/", flowStep.fileLocations.outputURIPrefix),
```
rjrudin, line 74: Shouldn't outputURIReplacement be dropped here since both exist?

srinathgit:
Fixed failing tests

marklogic-data-hub/src/main/java/com/marklogic/hub/step/impl/WriteStepRunner.java
```diff
@@ -398,6 +400,10 @@ protected void loadStepRunnerParameters(){
             this.withStopOnFailure(Boolean.parseBoolean(stepConfig.get("stopOnFailure").toString()));
         }
 
+        if(outputURIPrefix != null && outputURIReplacement != null){
```
rjrudin, line 403: I think it'd be good to change these checks to use a StringUtils method - e.g. hasText, isNotEmpty - because if one has a value and the other just has "", we're fine with that.

marklogic-data-hub/src/main/java/com/marklogic/hub/step/impl/WriteStepRunner.java
```diff
@@ -652,14 +658,24 @@ private RunStepResponse runIngester(RunStepResponse runStepResponse, Collection<
     }
 
     private void processCsv(JacksonHandle jacksonHandle, File file) {
-        String uri = file.getParent();
-        if(SystemUtils.OS_NAME.toLowerCase().contains("windows")){
-            uri = "/" + FilenameUtils.separatorsToUnix(StringUtils.replaceOnce(uri, ":", ""));
+        String uri;
```
rjrudin, line 661: I frequently like to put logic like this into a protected method so that it can easily unit-tested. You can even pass in the "osName" so you can verify the windows-specific code while not on windows. I didn't check the test coverage of all the conditions in here, but it'd be really easy to get 100% coverage of it with unit tests that will of course run very quickly too (i.e. put them in a class that doesn't extend anything, as you don't need any of the HubTestBase plumbing).

srinathgit: Moved uri generation to a protected method and added test in WriteTestBatcherTest class

PR #4010, opened 2020-05-27
DHFPROD-4914: Adding support for hubDhs and hubSsl properties

rjrudin:
Reworked/simplified HubConfigImpl so that refreshProject / loadConfigurationFromProperties go through applyProperties now, ensuring that the propertyConsumerMap is the only thing that applies properties, and does so in a consistent fashion. 

One tricky thing was ensuring that DataHubPlugin uses the *Config objects constructed by ml-gradle. I think how that works is now more clear in the code. 

Most of the changes in here are to HubConfigImpl, and most of the changes involved deleting code. 

A number of test classes were updated as well based on some updated method signatures in HubConfigImpl. The HubConfig interface itself has not changed.

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


web/src/main/java/com/marklogic/hub/web/model/HubSettings.java
```diff
@@ -238,73 +202,63 @@ public void setProjectDir(String projectDir) {
...
-    @Value("${mlStagingForestsPerHost}")
-    Integer stagingForestsPerHost = null;
-
-    @Value("${mlStagingPort}")
-    Integer stagingPort = null;
-
-    @Value("${mlStagingAuth}")
-    String stagingAuthMethod = null;
+    public HubSettings() {
+        HubConfig config = new HubConfigImpl();
```
rjrudin, line 206: This just gets its default property values from a new HubConfigImpl instead of dhf-defaults.properties.

marklogic-data-hub-central/src/main/resources/application.properties
```diff
@@ -7,20 +7,8 @@ spring.thymeleaf.prefix=classpath:/static/
...
-mlJobAuth=basic
-mlStagingAuth=basic
-mlAppServicesSimpleSsl=true
-mlFinalSimpleSsl=true
-mlJobSimpleSsl=true
-mlStagingSimpleSsl=true
-# Need this for getting the version of MarkLogic via the Versions class
-mlAppServicesPort=8010
+hubDhs=true
+hubSsl=true
```
srinathgit, line 11: I am guessing ,we need 2 separate properties because we aren't sure if HC will need SSL to talk to ML. If we decide that we need SSL, one property should be sufficient ?

rjrudin: Exactly - hopefully we only need one soon, just not there yet.

PR #4017, opened 2020-05-28
DHFPROD-4452: Add editing capability for entity type description

akshaysonvane:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


marklogic-data-hub-central/ui/src/components/modeling/entity-type-modal/entity-type-modal.tsx
```diff
@@ -47,51 +57,74 @@ const EntityTypeModal: React.FC<Props> = (props) => {
...
+  };
+
+  const updateEntityDescription = async (name: string, description: string) => {
+    try {
+      const response = await updateModelInfo(name, description);
+      if (response['status'] === 200) {
+        props.updateEntityTypesAndHideModal();
+      }
+    } catch (error) {
+      if (error.response.status === 400) {
```
rjrudin, line 82: I keep seeing this code repeated in the UI - couldn't this be in some sort of "common" library and reused, perhaps as a callback/promise so that if hasOwnProperty('message') is true, then the caller determines what to do, with the common function providing the value of error['response']['data']['message']?

akshaysonvane: Good point.
@rjrudin, @brucean52, what do you think about this?
```
  const sendAndHandle = (requestFunction, errorFunction) => {
    try {
      requestFunction();
    } catch (error) {
      if (error.response.status === 400) {
        if (error.response.data.hasOwnProperty('message')) {
          errorFunction(error['response']['data']['message']);
        }
      } else {
        handleError(error);
      }
    } finally {
      resetSessionTime();
    }
  };
```
We can then use the requestFunction to make the request, open/close modal, etc. Similary we can also use errorFunction to set error message, open/close modal, etc.

brucean52: I think this could work, but make the errorFunction conditional more generic to support more customization

marklogic-data-hub-central/ui/src/components/modeling/entity-type-modal/entity-type-modal.tsx
```diff
@@ -102,7 +135,7 @@ const EntityTypeModal: React.FC<Props> = (props) => {
         <Form.Item
           className={styles.formItem}
           label={<span>
-            Name:&nbsp;<span className={styles.asterisk}>*</span>
+            Name:&nbsp;{props.isEditModal ? null : <span className={styles.asterisk}>*</span>}
```
brucean52, line 138: Going by the wireframe, Edit is supposed to have the Entity Type name as text, not as a disabled input

marklogic-data-hub-central/ui/src/components/modeling/entity-type-table/entity-type-table.tsx
```diff
@@ -23,15 +24,24 @@ const EntityTypeTable: React.FC<Props> = (props) => {
...
-      // TODO - Edit Entity DHFPROD-4452
       render: text => {
+        let parseText = text.split(',');
+        let entityName = parseText[0];
+        let entityDescription = parseText[1];
+
         return (
           <Tooltip title={ModelingTooltips.entityTypeName}>
-            <span data-testid={text + '-span'}>{text}</span>
+            <span data-testid={text + '-span'}
```
brucean52, line 34: Please add CSS to make is look similar to the instance link. ( color change, cursor: pointer )

marklogic-data-hub-central/ui/src/components/modeling/entity-type-modal/entity-type-modal.tsx
akshaysonvane: After a couple of discussions, @brucean52  and I came down with the following function:
```
function sendAndHandle(requestFunction, errorFunction, resetSessionTime) {
  try {
    requestFunction();
  } catch (error) {
    errorFunction(error);
  } finally {
    resetSessionTime();
  }
}
export default sendAndHandle;
```
The entire catch block is replaced with `errorFunction(error);` in order to support more customization and we need to pass in the `resetSessionTime` function since `useContext` hook for `resetSessionTime` only works in react components and this isn't one.

At this point, this just seems to be a try/catch handler and it doesn't seem worth adding it to a common library.

marklogic-data-hub-central/ui/src/components/modeling/entity-type-modal/entity-type-modal.tsx
akshaysonvane: Good catch. Updated in here: https://github.com/marklogic/marklogic-data-hub/pull/4017/commits/e8d869cc937370f183c2574b9e61293edd943442

marklogic-data-hub-central/ui/src/components/modeling/entity-type-table/entity-type-table.tsx
akshaysonvane: Updated in here: https://github.com/marklogic/marklogic-data-hub/pull/4017/commits/e8d869cc937370f183c2574b9e61293edd943442

marklogic-data-hub-central/ui/src/components/modeling/entity-type-modal/entity-type-modal.tsx
rjrudin: This seems like it could be part of that facade library we've discussed - i.e. where ideally, every HTTP call goes through a very narrow chunk of code that can handle generic concerns for every HTTP call.

PR #4019, opened 2020-05-28
DHFPROD-4672: Add hub-central-load-writer role

ryanjdew:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/src/pages/Load.test.tsx
+++ b/marklogic-data-hub-central/ui/src/pages/Load.test.tsx
@@ -27,9 +27,9 @@ describe('Load component', () => {
         const authorityService = new AuthoritiesService();
         authorityService.setAuthorities(['readIngestion']);
 
-        const { getByText, getByLabelText, getByTestId, debug } = render(<AuthoritiesContext.Provider value={authorityService}><Load/></AuthoritiesContext.Provider>);
-            debug();
-        expect(await(waitForElement(() => getByLabelText('load-list')))).toBeInTheDocument();
+        const { getByText, getByTitle, getByLabelText, getByTestId, queryByTestId, queryByText, queryByTitle } = render(<AuthoritiesContext.Provider value={authorityService}><Load/></AuthoritiesContext.Provider>);
+
+        expect(await(waitForElement(() => getByLabelText('switch-view-list')))).toBeInTheDocument();
 
         // Check for steps to be populated
         expect(axiosMock.get).toBeCalledWith('/api/steps/ingestion');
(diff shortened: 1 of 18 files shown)


marklogic-data-hub/src/main/resources/hub-internal-config/security/roles/data-hub-ingestion-writer.json
```diff
@@ -1,4 +1,7 @@
 {
   "role-name": "data-hub-ingestion-writer",
-  "description": "Permits reading a load data configuration"
+  "description": "Permits reading a load data configuration",
```
bsrikan, line 3: should data-hub-ingestion-writer get data-hub-common-writer role just like data-hub-mapping-writer

marklogic-data-hub/src/main/resources/hub-internal-config/security/roles/data-hub-common-writer.json
```diff
@@ -29,6 +29,11 @@
       "privilege-name": "rest-writer",
       "action": "http://marklogic.com/xdmp/privileges/rest-writer",
       "kind": "execute"
+    },
+    {
+      "privilege-name": "xdmp:invoke",
```
rjrudin, line 34: One downside to JSON is we don't have a way to have inline comments about why a privilege is needed. I've used README.md files in the same directory as JSON files before to provide those comments. But those can be difficult to keep in sync - someone has to remember to update them. 

In theory though, if someone wonders why this privilege is set, they could remove it and run the test suite and something should fail, which would point out what code needs the privilege, right?

marklogic-data-hub/src/main/resources/hub-internal-config/security/roles/data-hub-ingestion-writer.json
rjrudin: I think that's a good idea.

rjrudin: Oh wait I take that back - I think we're just using this as a "data" role, while the hub-central-* roles are "privilege" roles. 

The description does need to be modified though - change "reading" to "writing".

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/impl/security/getAuthoritiesTest.sjs
```diff
@@ -45,6 +45,15 @@ hubTest.runWithRolesAndPrivileges(['hub-central-load-reader'], [], function() {
     assertions.push(test.assertFalse(authorities.includes('writeIngestion'), 'hub-central-load-reader should not have "writeIngestion"'));
 });
 
+// Test hub-central-load-writer
+hubTest.runWithRolesAndPrivileges(['hub-central-load-writer'], [], function() {
```
rjrudin, line 49: Are these assertions necessary since you have the tests on the controller, which verifies that the ROLE_* stuff works, which means the right authorities are being returned?

marklogic-data-hub/src/main/resources/hub-internal-config/security/roles/data-hub-common-writer.json
ryanjdew: True. This one stems from the need to insert via hub-utils. It was missed when doing the mapping roles since the mapping reader already provided it as part of testing the mapping.

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/impl/security/getAuthoritiesTest.sjs
bsrikan: Will the controller test fail, if the role doesnt include loginToHubCentral or readIndestion authorities?

marklogic-data-hub/src/main/resources/hub-internal-config/security/roles/data-hub-ingestion-writer.json
ryanjdew: I added the data-hub-common-writer role to the data-hub-ingestion-writer. It makes sense to me that the data-hub-ingestion-writer should be enough provide enough privileges/permissions to write an ingestion step outside the context of hub-central.

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/impl/security/getAuthoritiesTest.sjs
rjrudin: Yes - the loginAsTestUser method in AbstractMvcTest is making a POST to /api/login, so it's going through all the same plumbing as when a user authenticates in HC. So if a call is made with a user that can't login, the test will immediately fail.

marklogic-data-hub-central/ui/src/pages/Load.test.tsx
```diff
@@ -27,58 +27,91 @@ describe('Load component', () => {
...
+        expect(queryByTitle('delete')).not.toBeInTheDocument();
     });
 
-    test('Verify can edit with readIngestion and writeIngestion authorities', async () => {
+    test('Verify edit with readIngestion and writeIngestion authorities', async () => {
         const authorityService = new AuthoritiesService();
         authorityService.setAuthorities(['readIngestion','writeIngestion']);
 
-        const { getByText, getByLabelText, getByTestId } = render(<AuthoritiesContext.Provider value={authorityService}><Load/></AuthoritiesContext.Provider>);
+        axiosMock.delete['mockImplementation']((url) => {
```
bsrikan, line 71: With Mike's latest changes, I believe we refactored all the mock responses for the test under config/mocks.config.ts.
Can this mock delete be moved as well?

ryanjdew: sure

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/impl/security/getAuthoritiesTest.sjs
bsrikan: OK Thanks. I will update the test plan likewise.

PR #4025, opened 2020-05-29
DHFPROD-4967: Grant role to allow users to save queries

rahulvudutala:
### Description

- Adding hub-central roles for saved query documents
- UI Changes
- Adding tests

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


marklogic-data-hub/src/test/java/com/marklogic/hub/core/HubProjectTest.java
```diff
@@ -92,8 +92,8 @@ public void testInit() throws IOException {
         assertTrue(new File(projectPath, "src/main/hub-internal-config/security/roles/data-hub-step-definition-writer.json").exists());
         assertTrue(new File(projectPath, "src/main/hub-internal-config/security/roles/data-hub-entity-model-reader.json").exists());
         assertTrue(new File(projectPath, "src/main/hub-internal-config/security/roles/data-hub-entity-model-writer.json").exists());
-        assertTrue(new File(projectPath, "src/main/hub-internal-config/security/roles/data-hub-saved-query-reader.json").exists());
-        assertTrue(new File(projectPath, "src/main/hub-internal-config/security/roles/data-hub-saved-query-writer.json").exists());
+        assertTrue(new File(projectPath, "src/main/hub-internal-config/security/roles/hub-central-saved-query-reader.json").exists());
```
rjrudin, line 95: It's fine to do this, but I think soon, we should remove all of these assertions. If the roles/privileges weren't correctly written out, that "bootstrap" would fail and none of the tests would run. So these assertions are redundant and not adding any value. Fine to keep them for now, I'll look into removing all of them soon.

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/data-services/lib/entitySearchService.sjs
```diff
@@ -1,5 +1,11 @@
+const hubTest = require("/test/data-hub-test-helper.sjs");
+
 function invokeSavedQuery(module, args) {
-  return fn.head(xdmp.invoke("/data-hub/5/data-services/entitySearch/" + module, args));
+  return fn.head(hubTest.runWithRolesAndPrivileges(
+      ['data-hub-developer', 'hub-central-saved-query-reader', 'hub-central-saved-query-writer'], [],
```
rjrudin, line 5: Shouldn't you be able to do this without the "data-hub-developer" role, since this is an HC-specific capability?

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/impl/security/getAuthoritiesTest.sjs
```diff
@@ -62,4 +49,13 @@ hubTest.runWithRolesAndPrivileges(['hub-central-mapping-writer'], [], function()
     assertions.push(test.assertTrue(authorities.includes('readMapping'), 'hub-central-mapping-writer should have "readMapping"'));
     assertions.push(test.assertTrue(authorities.includes('writeMapping'), 'hub-central-mapping-writer should not have "writeMapping"'));
 });
+
+// Test hub-central-saved-query-writer
```
rjrudin, line 53: I don't think we want these assertions here. Instead, what we care about is that the EntitySearchController endpoints are secured correctly. Those should have @Secure("ROLE_manageSavedQuery") on them. You can then more effectively test the logic below by trying to hit the endpoints in a subclass of AbstractMvcTest as a user with that the saved-query roles, and without them.

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/data-services/lib/entitySearchService.sjs
rahulvudutala: Yes. It should. The any-uri, any-collection privileges should be added to hub-central-saved-query-writer.

rjrudin: I think hub-central-saved-query-writer just needs to inherit data-hub-common then, right @ryanjdew ?

rahulvudutala: data-hub-common-writer has the any-uri and unprotected-collections privileges. Should I extend that role?

rjrudin: Yes, I think that's the way to go.

marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/controllers/EntitySearchControllerTest.java
```diff
@@ -45,6 +45,9 @@ void testCRUDOnSavedQuery() throws Exception {
             "  }\n" +
             "}";
 
+        setTestUserRoles("hub-central-saved-query-writer", "hub-central-saved-query-reader");
+        loginAsTestUser();
```
rjrudin, line 49: No need to change, but you can do this in one method - loginAsTestUserWithRoles.

marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/controllers/EntitySearchControllerTest.java
```diff
@@ -94,6 +97,14 @@ void testCRUDOnSavedQuery() throws Exception {
                 ObjectNode response = readJsonObject(result.getResponse().getContentAsString());
                 assertTrue(response.isEmpty());
             });
+
+        // Try save/update/delete without the required role "hub-central-saved-query-writer"
+        setTestUserRoles("hub-central-saved-query-reader");
+        loginAsTestUser();
+
+        postJson(SAVED_QUERIES_PATH, json).andExpect(status().isForbidden());
```
rjrudin, line 105: There's a helper method in AbstractMvcTest to make this a little easier, and it guarantees that the user's authenticated HttpSession is used - verifyRequestIsForbidden. Please use this instead.

marklogic-data-hub/src/main/resources/hub-internal-config/security/roles/data-hub-saved-query-user.json
```diff
@@ -0,0 +1,31 @@
...
+      "privilege-name": "unprotected-collections",
+      "action": "http://marklogic.com/xdmp/privileges/unprotected-collections",
+      "kind": "execute"
+    },
+    {
+      "privilege-name": "unprotected-uri",
+      "action": "http://marklogic.com/xdmp/privileges/unprotected-uri",
+      "kind": "execute"
+    },
+    {
```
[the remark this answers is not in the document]
rahulvudutala: Yeah. The LeagacyRunFlowCollector test was failing as data-hub-common-writer has the any-uri privilege. The flowOperator is expected to have unprotected-uri, so I checked with @ryanjdew and added the privileges. These are the similar privileges that data-hub-common-writer has except for any-uri and ps-user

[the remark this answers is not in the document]
rjrudin: Pinging @ryanjdew  - flow-operator-role.json has unprotected-uri, so I don't see how an error could have occurred here based on what's in data-hub-saved-query-user.

PR #4027, opened 2020-05-29
DHFPROD-5088:minor fixes to deploy users and display mapped data in e…

kkanthet:
…xplorer

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x ] JIRA_ID included in all the commit messages
- [x ] PR title is in the format JIRA_ID:Title
- [ x] Rebase the branch with upstream
- [x ] Squashed all commits into a single commit
- [n/a ] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/examples/reference-entity-model/gradle-dhs.properties
+++ b/examples/reference-entity-model/gradle-dhs.properties
@@ -17,3 +17,6 @@ mlAppServicesPort=8010
 
 mlUsername=hub-developer
 mlPassword=password
+
+mlFlowOperatorRole=flowOperator
+mlFlowDeveloperRole=flowDeveloper
--- a/examples/reference-entity-model/src/main/ml-config/security/users/hub-developer.json
+++ b/examples/reference-entity-model/src/main/ml-config/security/users/hub-developer.json
@@ -5,6 +5,7 @@
   "role": [
     "data-hub-developer",
     "hub-central-user",
+    "hub-central-entity-model-writer",
     "hub-central-load-reader",
     "hub-central-mapping-writer"
   ]
--- a/examples/reference-entity-model/steps/mapping/mapCustomersJSON.step.json
+++ b/examples/reference-entity-model/steps/mapping/mapCustomersJSON.step.json
@@ -8,7 +8,9 @@
   "selectedSource": "collection",
   "sourceQuery": "cts.collectionQuery(['loadCustomersJSON'])",
   "collections": [
-    "default-mapping"
+    "default-mapping",
+    "Customer",
+    "mapCustomersJSON"
   ],
   "additionalCollections": [],
   "sourceDatabase": "data-hub-STAGING",


examples/reference-entity-model/gradle-dhs.properties
```diff
@@ -17,3 +17,6 @@ mlAppServicesPort=8010
 
 mlUsername=hub-developer
 mlPassword=password
+
+mlFlowOperatorRole=flowOperator
```
rjrudin, line 21: Yuck - were there things breaking when using DHF in DHS because these weren't overridden? I guess that's because we still use this for some default permissions.

examples/reference-entity-model/src/main/ml-config/security/users/hub-developer.json
```diff
@@ -5,6 +5,7 @@
   "role": [
     "data-hub-developer",
     "hub-central-user",
+    "hub-central-entity-model-writer",
```
rjrudin, line 8: Note that this will be simplified soon once we have the hub-central-developer role. Will only need to add that then.

examples/reference-entity-model/steps/mapping/mapCustomersJSON.step.json
```diff
@@ -10,7 +10,7 @@
   "collections": [
     "default-mapping"
   ],
-  "additionalCollections": [],
+  "additionalCollections": ["Customer","mapCustomersJSON"],
```
rjrudin, line 13: Can you change "collections" to be this array? Check to see if any other step depends on "default-mapping", but that's the right fix here.

examples/reference-entity-model/gradle-dhs.properties
kkanthet: yes, when we deploy users its failing bcs of this. I think we said flowDeveloper and operator will continue to  work in 5.2.x

examples/reference-entity-model/src/main/ml-config/security/users/hub-developer.json
kkanthet: Yeah, just added so PM can use hub-developer  has access to everything for now, when we get hub-central-developer-role, will remove this.

examples/reference-entity-model/steps/mapping/mapCustomersJSON.step.json
kkanthet: Moved to collections.

examples/reference-entity-model/gradle-dhs.properties
SameeraPriyathamTadikonda: You should also add mlFlowDeveloperRole=flowDeveloper

SameeraPriyathamTadikonda: My bad. I missed it

examples/reference-entity-model/steps/mapping/mapCustomersJSON.step.json
```diff
@@ -8,7 +8,9 @@
   "selectedSource": "collection",
   "sourceQuery": "cts.collectionQuery(['loadCustomersJSON'])",
   "collections": [
-    "default-mapping"
+    "default-mapping",
```
rjrudin, line 11: Please remove "default-mapping" too, it's not needed for anything. We don't want to include anything that's not needed, as it'll raise the question of "Why is this here?"

PR #4033, opened 2020-06-01
DHFPROD-4168: Add Entity Property

brucean52:
DHFPROD-4168: added integration tests and design fixes

** Adding property only adds to the UI. There is another story for saving the properties to the DB.
One integration test is blocked by the bug DHFPROD-5096

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


marklogic-data-hub-central/ui/e2e/cypress/fixtures/users/architect.json
```diff
@@ -0,0 +1,6 @@
+{
+  "user-name": "hc-architect",
```
bsrikan, line 2: We have a testUserWithRoles() function that we can reuse to add any roles a test would need. So we have a developer, operator and a generic user. The purpose of generic user is to add roles to it as needed and test.

Do you think we can use that to test and get rid of this user?

brucean52: Thanks.  I removed the architect role and using the loginAsTestUserWithRoles() method instead.

PR #4035, opened 2020-06-02
DHFPROD-4904:Run load step with selected files

srinathgit:
### Description
Run load step with selected files
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/controllers/FlowMvcTest.java
+++ b/marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/controllers/FlowMvcTest.java
@@ -1,0 +1,60 @@
+package com.marklogic.hub.central.controllers;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.marklogic.client.DatabaseClient;
+import com.marklogic.client.document.JSONDocumentManager;
+import com.marklogic.client.io.JacksonHandle;
+import com.marklogic.hub.central.AbstractMvcTest;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.multipart;
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;
+
+import com.marklogic.hub.flow.RunFlowResponse;
+import com.marklogic.hub.job.JobStatus;
+import io.swagger.annotations.OAuth2Definition;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.mock.web.MockMultipartFile;
+
+import javax.ws.rs.core.MediaType;
+import java.nio.charset.StandardCharsets;
+
+
+
+public class FlowMvcTest extends AbstractMvcTest {
+
+    @Autowired
+    FlowController flowController;
+
+    private final static String PATH = "/api/flows";
+    private ObjectMapper mapper = new ObjectMapper();
+
+    @Test
+    void runIngestionStep() throws Exception {
+        installReferenceModelProject();
+        loginAsTestUserWithRoles("hub-central-user","data-hub-operator");
+
+        MockMultipartFile file1 = new MockMultipartFile("files","file1.json", MediaType.APPLICATION_JSON, "{\"name\": \"Joe\"}".getBytes(StandardCharsets.UTF_8));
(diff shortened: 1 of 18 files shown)


marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/controllers/FlowMvcTest.java
```diff
@@ -0,0 +1,35 @@
...
+
+public class FlowMvcTest extends AbstractMvcTest {
+
+    private final static String PATH = "/api/flows";
+
+    @Test
+    void runIngestionStep() throws Exception {
+        installReferenceModelProject();
+        setTestUserRoles("hub-central-user","data-hub-operator");
+        loginAsTestUser();
```
rjrudin, line 22: No need to change, but you can do these 2 lines in 1 line - loginAsTestUserWithRoles.

marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/controllers/FlowMvcTest.java
```diff
@@ -0,0 +1,35 @@
...
+        MockMultipartFile file1 = new MockMultipartFile("file1", "file1.json", MediaType.APPLICATION_JSON, "{\"name\": \"Joe\"}".getBytes(StandardCharsets.UTF_8));
+        MockMultipartFile file2 = new MockMultipartFile("file2","file2.json", MediaType.APPLICATION_JSON, "{\"name\": \"John\"}".getBytes(StandardCharsets.UTF_8));
+
+        mockMvc.perform(MockMvcRequestBuilders.multipart(PATH + "/{flowName}/steps/{stepNumber}", "ingestToFinal", "1")
+                        .file(file1)
+                        .file(file2)
+                        .contentType(MediaType.MULTIPART_FORM_DATA)
+                        .session(mockHttpSession))
+                .andDo(MockMvcResultHandlers.log())
+                .andExpect(status().isOk());
```
rjrudin, line 33: I think we need to verify that the two JSON documents were ingested as well. This is good in that it verifies no error occurred, but it should be a simple matter to verify that the two documents are in the staging database too - I imagine you can assert on the size of the collection that the documents were ingested into.

srinathgit: runFlow() starts the job and returns , so had to include a function to check for completion of job and verified if the docs were ingested.

marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/controllers/FlowMvcTest.java
```diff
@@ -0,0 +1,81 @@
...
+            "Verifying that 2 docs were ingested into the final database");
+
+        rawDoc = getFinalDoc("/customers/file2.json");
+        assertEquals("John", rawDoc.get("envelope").get("instance").get("name").asText(),
+            "Verifying that 2 docs were ingested into the final database");
+
+    }
+
+
+    private void waitForJobToFinish(String jobId) {
```
rjrudin, line 57: This is a common problem we have, which is obtaining a reference to the FlowRunner so we can call awaitCompletion instead of having to write a lot of code here, or subclass NewFlowController.

Here's a better approach - add a "private Consumer<FlowRunner> flowRunnerConsumer" to NewFlowController (and can you actually rename that to FlowController too? I forgot to do so before). Add a public setter for it.

If flowRunnerConsumer is non-null, then call accept on it after running the flow.

That way, a test can obtain the Autowired FlowController and then pass in a Consumer impl so it'll get a reference to the FlowRunner. The impl of Consumer can then call awaitCompletion.

marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/controllers/FlowMvcTest.java
```diff
@@ -0,0 +1,60 @@
...
+    private ObjectMapper mapper = new ObjectMapper();
+
+    @Test
+    void runIngestionStep() throws Exception {
+        installReferenceModelProject();
+        loginAsTestUserWithRoles("hub-central-user","data-hub-operator");
+
+        MockMultipartFile file1 = new MockMultipartFile("files","file1.json", MediaType.APPLICATION_JSON, "{\"name\": \"Joe\"}".getBytes(StandardCharsets.UTF_8));
+        MockMultipartFile file2 = new MockMultipartFile("files","file2.json", MediaType.APPLICATION_JSON, "{\"name\": \"John\"}".getBytes(StandardCharsets.UTF_8));
+        flowController.setFlowRunnerConsumer((flowRunner -> {
```
rjrudin, line 42: Nice! And we should be able to reuse this too.

PR #4037, opened 2020-06-02
DHFPROD-5094: Make model-writer inherit from common-writer role

akshaysonvane:
also, add xslt-invoke to common-writer

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [N/A] Added Tests
  

- ##### Reviewer:

- [N/A] Reviewed Tests


marklogic-data-hub/src/main/resources/hub-internal-config/security/roles/data-hub-common-writer.json
```diff
@@ -34,6 +34,11 @@
       "privilege-name": "xdmp:invoke",
       "action": "http://marklogic.com/xdmp/privileges/xdmp-invoke",
       "kind": "execute"
+    },
+    {
```
ryanjdew, line 38: Just adding the comment here that it was discovered this privilege was needed for validation and hopefully, we'll be expanding validation to all of the writer roles in the future so it makes sense to add it to data-hub-common-writer

PR #4049, opened 2020-06-03
DHFPROD-4984: Filter object/array type properties using entity model

akshaysonvane:
### Description
Moved the logic for filtering out array/object type properties to the data service.
Also, we now use the entity model to filter out array/object type properties instead of relying on the column names sent by the FE.
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/entities/search/FacetHandler.java
+++ b/marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/entities/search/FacetHandler.java
@@ -1,0 +1,26 @@
+/*
+ * Copyright 2012-2020 MarkLogic Corporation
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ */
+package com.marklogic.hub.central.entities.search;
+
+import com.marklogic.client.query.StructuredQueryBuilder;
+import com.marklogic.client.query.StructuredQueryDefinition;
+import com.marklogic.hub.central.entities.search.models.DocSearchQueryInfo;
+
+public interface FacetHandler {
+
+    StructuredQueryDefinition buildQuery(DocSearchQueryInfo.FacetData data, StructuredQueryBuilder queryBuilder);
+}
--- a/marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/entities/search/impl/CreatedOnFacetHandler.java
+++ b/marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/entities/search/impl/CreatedOnFacetHandler.java
@@ -1,0 +1,53 @@
+/*
+ * Copyright 2012-2020 MarkLogic Corporation
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ */
+package com.marklogic.hub.central.entities.search.impl;
+
+import com.marklogic.client.query.StructuredQueryBuilder;
+import com.marklogic.client.query.StructuredQueryDefinition;
+import com.marklogic.hub.central.entities.search.Constants;
+import com.marklogic.hub.central.entities.search.FacetHandler;
+import com.marklogic.hub.central.entities.search.models.DocSearchQueryInfo;
+
+import java.time.LocalDate;
+import java.time.ZoneId;
+import java.time.format.DateTimeFormatter;
+
+public class CreatedOnFacetHandler implements FacetHandler {
+
+    private static final DateTimeFormatter DATE_FORMAT = DateTimeFormatter.ofPattern("yyyy-MM-dd");
+    private static final DateTimeFormatter DATE_TIME_FORMAT = DateTimeFormatter
+            .ofPattern("yyyy-MM-dd'T'HH:mm:ss.SSSZ");
+
+    @Override
+    public StructuredQueryDefinition buildQuery(DocSearchQueryInfo.FacetData data, StructuredQueryBuilder queryBuilder) {
+        // Converting the date in string format from yyyy-MM-dd format to yyyy-MM-dd HH:mm:ss format
+        LocalDate startDate = LocalDate.parse(data.getRangeValues().getLowerBound(), DATE_FORMAT);
+        String startDateTime = startDate.atStartOfDay(ZoneId.systemDefault())
+                .format(DATE_TIME_FORMAT);
+
+        // Converting the date in string format from yyyy-MM-dd format to yyyy-MM-dd HH:mm:ss format
+        // Adding 1 day to end date to get docs harmonized on the end date as well.
+        LocalDate endDate = LocalDate.parse(data.getRangeValues().getUpperBound(), DATE_FORMAT)
+                .plusDays(1);
+        String endDateTime = endDate.atStartOfDay(ZoneId.systemDefault()).format(DATE_TIME_FORMAT);
+
+        return queryBuilder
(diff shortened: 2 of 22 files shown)


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/entitySearch/getOpticPlan.sjs
```diff
@@ -58,6 +80,15 @@ structuredQuery = fn.head(xdmp.unquote(structuredQuery)).root;
...
+/*
+ * Filtering out columns (properties) that are of object/array type since we don't support them for now.
+ * Also replacing hyphen with underscore for column names (entity property names), schema names and view names since TDE's do the same.
+ */
+const simplePropertySet = filterObjectAndArrayTypeProperties(schemaName);
+columns = columns.filter(column => simplePropertySet.has(column)).map(column => replaceHyphenWithUnderscore(column));
+viewName = replaceHyphenWithUnderscore(viewName);
+schemaName = replaceHyphenWithUnderscore(schemaName);
+
 const newOptions = fn.head(xdmp.xsltEval(stylesheet, queryOptions)).root;
```
ryanjdew, line 92: This is a tangent to this PR, but what is the minimum role assigned to the user that should be able to use this data service? I believe we'll need a ticket to add the xslt-eval privilege to that role as part of another ticket.

akshaysonvane: Good point. 
This means we need to add the xslt-eval privilege to hub-central-entity-exporter role.

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/managers/EntitySearchManager.java
```diff
@@ -190,33 +189,35 @@ private StructuredQueryDefinition buildQuery(QueryManager queryMgr, SearchQuery
...
-                    .plusDays(1);
-                String endDateTime = endDate.atStartOfDay(ZoneId.systemDefault()).format(DATE_TIME_FORMAT);
-
-                facetDef = queryBuilder
-                    .and(queryBuilder.rangeConstraint(facetType, Operator.GE, startDateTime),
-                        queryBuilder.rangeConstraint(facetType, Operator.LT, endDateTime));
-            }
-            else { // If a property is not a Hub property, then it is an Entity Property
-                facetDef = getEntityPropertyConstraints(facetType, data, queryBuilder);
+            switch (facetType) {
```
rjrudin, line 192: Something to consider as this grows - a switch statement is often a bad practice because it naturally creates a large block of code with explicit knowledge of the various choices for the value being switched on. Switch statements can usually be replaced by a more "open/closed" mechanism via an interface like this:

     public interface FacetHandler 
       StructuredQueryDefinition buildQuery(FacetData data, StructuredQueryBuilder queryBuilder)

This class then has a Map<String, FacetHandler>, where String is the constraint name. And for each selected facet, it finds the matching FacetHandler and asks it to build an SQD.

The benefit of this is that you can easily unit-test each implementation of FacetHandler in isolation. And you avoid an ungainly switch statement and have a much more extensible design - when it's time to support a new constraint, you just create a new instance of FacetHandler and add it to the map. 

It may be worth making that change now, particularly if there's not enough coverage of the code in each switch block here.

akshaysonvane:
Moved `EntitySearchManager` class, entity search related models and switch blocks for handling facets to their own classes under `com.marklogic.hub.central.entities.search`
Also, added unit tests for all the code that was in each switch block previously.

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/entities/search/FacetHandler.java
```diff
@@ -0,0 +1,30 @@
...
+ *
+ */
+package com.marklogic.hub.central.entities.search;
+
+import com.marklogic.client.query.StructuredQueryBuilder;
+import com.marklogic.client.query.StructuredQueryDefinition;
+import com.marklogic.hub.central.entities.search.models.DocSearchQueryInfo;
+
+public interface FacetHandler {
+    String COLLECTION_CONSTRAINT_NAME = "Collection";
```
rjrudin, line 24: I think it's better not to define these, because this ends up causing the interface to know about all the possible implementations, which runs counter to the purpose of an interface. 

In EntitySearchManager on the develop branch, there's only one reference to each of these constants. If that's still the case, I wouldn't even bother making them constants.

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/entities/search/impl/CreatedOnFacetHandler.java
```diff
@@ -0,0 +1,52 @@
...
+import com.marklogic.hub.central.entities.search.FacetHandler;
+import com.marklogic.hub.central.entities.search.models.DocSearchQueryInfo;
+
+import java.time.LocalDate;
+import java.time.ZoneId;
+import java.time.format.DateTimeFormatter;
+
+public class CreatedOnFacetHandler implements FacetHandler {
+
+    private static final DateTimeFormatter DATE_FORMAT = DateTimeFormatter.ofPattern("yyyy-MM-dd");
```
rjrudin, line 30: This is another benefit of having a bunch of small Handler implementations - instead of packing all this state into EntitySearchManager, it's tightly associated with the Handler it belongs to.

marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/entities/search/impl/CollectionFacetHandlerTest.java
```diff
@@ -0,0 +1,31 @@
...
+    public void testCollectionFacet() {
+        String uri = "myCollection";
+        String constraintName = "Collection";
+        DocSearchQueryInfo.FacetData facetData = new DocSearchQueryInfo.FacetData();
+        facetData.setStringValues(Collections.singletonList(uri));
+        CollectionFacetHandler collectionFacetHandler = new CollectionFacetHandler();
+        StructuredQueryBuilder queryBuilder = new StructuredQueryBuilder();
+
+        StructuredQueryDefinition queryDefinition = collectionFacetHandler.buildQuery(facetData, queryBuilder);
+        Fragment fragment = new Fragment(queryDefinition.serialize(), Namespace.getNamespace("sch", "http://marklogic.com/appservices/search"));
```
rjrudin, line 26: Sorry for the nit-pick here, but can you use "search" instead of "sch"? "search" is widely used in ML code as the prefix for this namespace.

marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/entities/search/impl/CreatedOnFacetHandlerTest.java
```diff
@@ -0,0 +1,41 @@
...
+        DocSearchQueryInfo.FacetData facetData = new DocSearchQueryInfo.FacetData();
+        DocSearchQueryInfo.RangeValues rangeValues = new DocSearchQueryInfo.RangeValues();
+        rangeValues.setLowerBound(lowerBound);
+        rangeValues.setUpperBound(upperBound);
+        facetData.setRangeValues(rangeValues);
+        CreatedOnFacetHandler createdOnFacetHandler = new CreatedOnFacetHandler();
+        StructuredQueryBuilder queryBuilder = new StructuredQueryBuilder();
+
+        StructuredQueryDefinition queryDefinition = createdOnFacetHandler.buildQuery(facetData, queryBuilder);
+        Fragment fragment = new Fragment(queryDefinition.serialize(), Namespace.getNamespace("sch", "http://marklogic.com/appservices/search"));
```
rjrudin, line 32: To avoid duplication, I recommend an AbstractFacetHandlerTest that has a "protected Fragment toFragment(StructuredQueryDefinition)" method which will construct the Fragment and initialize the needed Namespace.

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/entities/search/FacetHandler.java
akshaysonvane: That's right, my bad.

I'll add it to a Constants class specific to entity search as they are used in a couple of places.

marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/entities/search/impl/CollectionFacetHandlerTest.java
akshaysonvane: Sure.

marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/entities/search/impl/CreatedOnFacetHandlerTest.java
akshaysonvane: Good idea.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/entitySearch/getOpticPlan.sjs
akshaysonvane: Added a ticket: https://project.marklogic.com/jira/browse/DHFPROD-5139

PR #4062, opened 2020-06-08
DHFPROD-4912: Fixing QS's dependency on dhfversion

rjrudin:
Instead of reverting (https://github.com/marklogic/marklogic-data-hub/pull/4061), I think this may address QS's dependency on mlDHFVersion. 

This is now populated via versions.getLocalProjectVersion. And the header of QS has been 
changed to reference runningVersion instead of local project version.

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/impl/Versions.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/impl/Versions.java
@@ -121,18 +121,22 @@ public class Versions extends LoggingObject {
             if (fallbackToLocalProject) {
                 logger.info("Unable to determine installed version, likely because DH is not yet installed: " + ex.getMessage());
                 logger.info("Will try to determine version from local project");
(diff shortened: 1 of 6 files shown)


bsrikan:
On this PR tested following:
- With no DH installed ran 5.3 QS -- got expected message in QS about upgrade
- With 5.2 mapping example deployed ran 5.3 QS -- got expected message in QS about upgrade 
`Details of the message for above 2 in JIRA ticket.`

- Ran hubUpdate on 5.2 mapping project, removed mlDHFVersion from gradle.properties and then ran flows in QS -- flows running as expected
- Ran hubMigrateFlows, ran hubDeployAsDeveloper, Cleared user data ran flows from HC -- flows running as expected


Will document this in JIRA as well since this is going to be the upgrade path as well.

PR #4065, opened 2020-06-08
DHFPROD-4706: Update hubCreateEntity task

akshaysonvane:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/resources/scaffolding/Entity.json
+++ b/marklogic-data-hub/src/main/resources/scaffolding/Entity.json
@@ -1,13 +1,13 @@
 {
   "info" : {
     "title" : "placeholder",
-    "version" : "0.0.1",
-    "baseUri" : "http://example.com/",
-    "description" : "An placeholder entity"
+    "version" : "1.0.0",
+    "baseUri" : "http://example.org",
+    "description" : ""
   },
   "definitions" : {
     "placeholder" : {
-      "description" : "The placeholder entity root.",
+      "description" : "",
       "required" : [ ],
       "rangeIndex" : [ ],
       "elementRangeIndex" : [ ],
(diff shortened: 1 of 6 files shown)


ml-data-hub-plugin/src/test/groovy/com/marklogic/gradle/task/CreateEntityTaskTest.groovy
```diff
@@ -61,7 +61,7 @@ class CreateEntityTaskTest extends BaseTest {
         result.task(":hubDeployUserArtifacts").outcome == SUCCESS
 
         File entityFile = Paths.get(testProjectDir.root.toString(), "entities", "my-new-entity.entity.json").toFile()
-        entityFile.isFile() == true
+        entityFile.isFile()
```
rjrudin, line 64: What is this line of code doing - should it just be removed?

akshaysonvane: The Spock testing framework does not require a condition check for boolean assertions.

akshaysonvane:
Fixed failing tests.

marklogic-data-hub/src/main/resources/scaffolding/Entity.json
```diff
@@ -1,13 +1,13 @@
 {
   "info" : {
     "title" : "placeholder",
-    "version" : "0.0.1",
-    "baseUri" : "http://example.com/",
-    "description" : "An placeholder entity"
+    "version" : "1.0.0",
+    "baseUri" : "http://example.org",
```
rjrudin, line 5: I thought this needs to end with a "/" or else the trigger for validating entities will fail?

akshaysonvane: We validate using 
```
xdmp:json-validate-node($entity-def/info, xdmp:unquote('{type: "object", properties: {"baseUri":{type:"string", format:"uri"}}, required: ["baseUri"]}'), "strict")
```
It seems any string that has the following structure is a valid URI for ML
```
?://?
``` 

I changed the baseUri to "http://example.org" since that's what the AC in the ticket said https://project.marklogic.com/jira/browse/DHFPROD-4706
Should it be "http://example.org/" ?

PR #4070, opened 2020-06-08
DHFPROD-4881: Save table columns when saving a query with flat fields

timur-isangulov:
### Description
includes DHFPROD-5058
Apply query isn't implemented in this pr because the backend work of the DHFPROD-5100 is needed for supporting this. 
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/src/util/search-context.tsx
+++ b/marklogic-data-hub-central/ui/src/util/search-context.tsx
@@ -15,6 +15,7 @@ type SearchContextInterface = {
   selectedQuery: string,
   zeroState: boolean,
   manageQueryModal: boolean,
+  selectedTableProperties: any,
 }
 
 const defaultSearchOptions = {
@@ -30,6 +31,7 @@ const defaultSearchOptions = {
   selectedQuery: 'select a query',
   zeroState: false,
   manageQueryModal: false,
+  selectedTableProperties: [],
 }
 
 interface ISearchContextInterface {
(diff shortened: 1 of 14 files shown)


marklogic-data-hub-central/ui/src/util/search-context.tsx
```diff
@@ -385,7 +389,7 @@ const SearchProvider: React.FC<{ children: any }> = ({ children }) => {
     });
   }
 
-    const applySaveQuery = (searchText: string, entityTypeIds: string[], selectedFacets: {}, selectedQuery: string, zeroState = false, manageQueryModal = false) => {
+    const applySaveQuery = (searchText: string, entityTypeIds: string[], selectedFacets: {}, selectedQuery: string, propertiesToDisplay: string[], zeroState = false, manageQueryModal = false) => {
```
brucean52, line 392: I would consider refactoring the variables into an options object. You can create an interface that goes with it, and share it across components that call the applySaveQuery function.

timur-isangulov: Thanks, will implement interface in next story.

timur-isangulov:
This is refactoring pr I didn't add new tests. Will add tests in the next pr where apply query is implemented.

PR #4072, opened 2020-06-09
DHFPROD-4786: Facet by time window based on when an entity instance was created

rahulvudutala:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/entities/search/impl/CreatedOnFacetHandler.java
+++ b/marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/entities/search/impl/CreatedOnFacetHandler.java
@@ -21,33 +21,113 @@ import com.marklogic.client.query.StructuredQueryDefinition;
 import com.marklogic.hub.central.entities.search.Constants;
 import com.marklogic.hub.central.entities.search.FacetHandler;
 import com.marklogic.hub.central.entities.search.models.DocSearchQueryInfo;
+import com.marklogic.hub.central.exceptions.DataHubException;
 
 import java.time.LocalDate;
 import java.time.ZoneId;
+import java.time.ZoneOffset;
+import java.time.ZonedDateTime;
 import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
+import java.time.temporal.TemporalAdjusters;
+import java.util.HashMap;
+import java.util.Map;
 
 public class CreatedOnFacetHandler implements FacetHandler {
 
-    private static final DateTimeFormatter DATE_FORMAT = DateTimeFormatter.ofPattern("yyyy-MM-dd");
-    private static final DateTimeFormatter DATE_TIME_FORMAT = DateTimeFormatter
-            .ofPattern("yyyy-MM-dd'T'HH:mm:ss.SSSZ");
+    public static final DateTimeFormatter DATE_TIME_FORMAT = DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ssZ");
 
     @Override
     public StructuredQueryDefinition buildQuery(DocSearchQueryInfo.FacetData data, StructuredQueryBuilder queryBuilder) {
-        // Converting the date in string format from yyyy-MM-dd format to yyyy-MM-dd HH:mm:ss format
-        LocalDate startDate = LocalDate.parse(data.getRangeValues().getLowerBound(), DATE_FORMAT);
-        String startDateTime = startDate.atStartOfDay(ZoneId.systemDefault())
-                .format(DATE_TIME_FORMAT);
+        Map<String, String> dateRange = computeDateRange(data, new CreatedOnFacetInputs());
+        return queryBuilder.and(
+            queryBuilder.rangeConstraint(Constants.CREATED_ON_CONSTRAINT_NAME, StructuredQueryBuilder.Operator.GE, dateRange.get("startDateTime")),
+            queryBuilder.rangeConstraint(Constants.CREATED_ON_CONSTRAINT_NAME, StructuredQueryBuilder.Operator.LT, dateRange.get("endDateTime"))
+        );
+    }
 
-        // Converting the date in string format from yyyy-MM-dd format to yyyy-MM-dd HH:mm:ss format
-        // Adding 1 day to end date to get docs harmonized on the end date as well.
-        LocalDate endDate = LocalDate.parse(data.getRangeValues().getUpperBound(), DATE_FORMAT)
-                .plusDays(1);
-        String endDateTime = endDate.atStartOfDay(ZoneId.systemDefault()).format(DATE_TIME_FORMAT);
+    protected Map<String, String> computeDateRange(DocSearchQueryInfo.FacetData data, CreatedOnFacetInputs facetInputs) {
 
-        return queryBuilder
-                .and(queryBuilder.rangeConstraint(Constants.CREATED_ON_CONSTRAINT_NAME, StructuredQueryBuilder.Operator.GE, startDateTime),
-                        queryBuilder.rangeConstraint(Constants.CREATED_ON_CONSTRAINT_NAME, StructuredQueryBuilder.Operator.LT, endDateTime));
+        if(data.getStringValues().size() > 0) {
+            facetInputs.timeRange = data.getStringValues().get(0);
+        }
 
+        if(data.getStringValues().size() > 1) {
+            facetInputs.zoneOffset = Integer.parseInt(data.getStringValues().get(1));
+        }
+
+        facetInputs.zoneId = ZoneId.ofOffset("UTC", ZoneOffset.ofTotalSeconds(facetInputs.zoneOffset*60));
+        facetInputs.startDate = LocalDate.now().atStartOfDay(facetInputs.zoneId);
+        facetInputs.endDate = LocalDate.now().atStartOfDay(facetInputs.zoneId);
+
+        switch (facetInputs.timeRange) {
+            case "Today":
+                getTodayTimeWindow(facetInputs);
+                break;
+
+            case "This Week":
+                getThisWeekTimeWindow(facetInputs);
+                break;
+
+            case "This Month":
+                getThisMonthTimeWindow(facetInputs);
+                break;
+
+            case "Custom":
+                getCustomTimeWindow(data, facetInputs);
+                break;
+        }
+        return facetInputs.dateRange;
+    }
+
+    private void getTodayTimeWindow(CreatedOnFacetInputs facetInputs) {
+        facetInputs.startDateTime = facetInputs.startDate.toLocalDate().atStartOfDay(facetInputs.zoneId).format(DATE_TIME_FORMAT);
+        facetInputs.dateRange.put("startDateTime", facetInputs.startDateTime);
+        facetInputs.endDateTime = facetInputs.endDate.plusDays(1).toLocalDate().atStartOfDay(facetInputs.zoneId).format(DATE_TIME_FORMAT);
+        facetInputs.dateRange.put("endDateTime", facetInputs.endDateTime);
+    }
+
+    private void getThisWeekTimeWindow(CreatedOnFacetInputs facetInputs) {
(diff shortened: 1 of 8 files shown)


marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/entities/search/impl/CreatedOnFacetHandler.java
```diff
@@ -21,33 +21,91 @@
...
+        }
+
+        ZoneId zoneId = ZoneId.ofOffset("UTC", ZoneOffset.ofTotalSeconds(zoneOffset*60));
+        ZonedDateTime startDate = LocalDate.now().atStartOfDay(zoneId);
+        ZonedDateTime endDate = LocalDate.now().atStartOfDay(zoneId);
+        String startDateTime;
+        String endDateTime;
+
+        switch (timeRange) {
+            case "Today":
```
rjrudin, line 69: To make this code easier to understand, move the guts of each case into a private method. That'll slim down this method - which I also think should be protected (if you want to access it via a unit test) or private, definitely not public.

marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/entities/search/impl/CreatedOnFacetHandlerTest.java
```diff
@@ -32,10 +41,73 @@ public void testCreatedOnDateRangeFacet() {
...
     }
+
+    @Test
+    public void testComputeDateRangeTimeWindows() {
+        String zoneOffset = "-420";
+        ZoneId zoneId = ZoneId.ofOffset("UTC", ZoneOffset.ofTotalSeconds(Integer.parseInt(zoneOffset)*60));
+        DocSearchQueryInfo.FacetData facetData = new DocSearchQueryInfo.FacetData();
+        DocSearchQueryInfo.RangeValues rangeValues = new DocSearchQueryInfo.RangeValues();
+
+        // For time window Today.
```
rjrudin, line 58: Because you have 4 cases in your switch in the implementation, I'd expect at least 4 separate test methods. And because of the presence of "if" and "try" blocks for "Custom", I'd expect multiple test methods for that case, one for each scenario. This will be much more readable than one long test that covers different scenarios, and it'll be easier to debug if/when one of the scenarios breaks.

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/entities/search/impl/CreatedOnFacetHandler.java
```diff
@@ -21,33 +21,115 @@
...
+import java.util.Map;
 
 public class CreatedOnFacetHandler implements FacetHandler {
 
-    private static final DateTimeFormatter DATE_FORMAT = DateTimeFormatter.ofPattern("yyyy-MM-dd");
-    private static final DateTimeFormatter DATE_TIME_FORMAT = DateTimeFormatter
-            .ofPattern("yyyy-MM-dd'T'HH:mm:ss.SSSZ");
+    public static final DateTimeFormatter DATE_TIME_FORMAT = DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ssZ");
+
+    int zoneOffset = 0;
```
rjrudin, line 40: These can't be class fields, because EntitySearchManager is storing one instance of it in a static map that is used for all threads. So we'll get thread-safety issues on this and possibly incorrect results. 

A good rule of thumb is that if you have a set of fields that are associated with one another, toss them into a class. So instead of making these class fields of CreatedOnFacetHandler, make an e.g. CreatedOnFacetInputs class and create a new instance of that when buildQuery is called, and then pass it to each case method. This avoids thread-safety issues and reduces the amount of state (which is generally good - the more state, the more problems there are).

marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/entities/search/impl/CreatedOnFacetHandlerTest.java
```diff
@@ -4,38 +4,171 @@
...
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 public class CreatedOnFacetHandlerTest extends AbstractFacetHandlerTest {
 
+    List<String> stringValues;
+    Map<String, String> dateRange;
+
+    CreatedOnFacetHandler createdOnFacetHandler;
```
rjrudin, line 32: Storing these as class fields is fine because we don't have to worry about multiple threads running this same test instance.

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/entities/search/impl/CreatedOnFacetHandler.java
```diff
@@ -21,33 +21,113 @@
...
-        String startDateTime = startDate.atStartOfDay(ZoneId.systemDefault())
-                .format(DATE_TIME_FORMAT);
+        Map<String, String> dateRange = computeDateRange(data, new CreatedOnFacetInputs());
+        return queryBuilder.and(
+            queryBuilder.rangeConstraint(Constants.CREATED_ON_CONSTRAINT_NAME, StructuredQueryBuilder.Operator.GE, dateRange.get("startDateTime")),
+            queryBuilder.rangeConstraint(Constants.CREATED_ON_CONSTRAINT_NAME, StructuredQueryBuilder.Operator.LT, dateRange.get("endDateTime"))
+        );
+    }
+
+    protected Map<String, String> computeDateRange(DocSearchQueryInfo.FacetData data, CreatedOnFacetInputs facetInputs) {
```
rjrudin, line 49: It's fine to pass CreatedOnFacetInputs in here, but you don't really need to because the caller doesn't care about it. You can just instantiate it within this method. No need to change though.

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/entities/search/impl/CreatedOnFacetHandler.java
```diff
@@ -21,33 +21,113 @@
...
+        facetInputs.endDateTime = facetInputs.endDate.plusDays(1).toLocalDate().atStartOfDay(facetInputs.zoneId).format(DATE_TIME_FORMAT);
+        facetInputs.dateRange.put("endDateTime", facetInputs.endDateTime);
+    }
+
+    private void getThisMonthTimeWindow(CreatedOnFacetInputs facetInputs) {
+        facetInputs.startDate = facetInputs.startDate.with(TemporalAdjusters.firstDayOfMonth());
+        facetInputs.startDateTime = facetInputs.startDate.toLocalDate().atStartOfDay(facetInputs.zoneId).format(DATE_TIME_FORMAT);
+        facetInputs.dateRange.put("startDateTime", facetInputs.startDateTime);
+        facetInputs.endDateTime = facetInputs.endDate.plusDays(1).toLocalDate().atStartOfDay(facetInputs.zoneId).format(DATE_TIME_FORMAT);
+        facetInputs.dateRange.put("endDateTime", facetInputs.endDateTime);
```
akshaysonvane, line 102: Method `getTodayTimeWindow` also uses the same code, maybe we can extract common code into a method.

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/entities/search/impl/CreatedOnFacetHandler.java
```diff
@@ -21,33 +21,113 @@
...
+    }
 
-        return queryBuilder
-                .and(queryBuilder.rangeConstraint(Constants.CREATED_ON_CONSTRAINT_NAME, StructuredQueryBuilder.Operator.GE, startDateTime),
-                        queryBuilder.rangeConstraint(Constants.CREATED_ON_CONSTRAINT_NAME, StructuredQueryBuilder.Operator.LT, endDateTime));
+    private void getCustomTimeWindow(DocSearchQueryInfo.FacetData data, CreatedOnFacetInputs facetInputs) {
+        if(!data.getRangeValues().getLowerBound().isEmpty() && !data.getRangeValues().getUpperBound().isEmpty()) {
+            try {
+                facetInputs.startDateTime = ZonedDateTime.parse(data.getRangeValues().getLowerBound()).format(DATE_TIME_FORMAT);
+                facetInputs.endDateTime = ZonedDateTime.parse(data.getRangeValues().getUpperBound()).plusDays(1).format(DATE_TIME_FORMAT);
```
akshaysonvane, line 109: It seems that FE now sends both date and time. Do we still need to add a day any more if that's the case?
Can we use startTime >= && <=endTime for comparison making this more accurate?

Since the FE still uses a DatePicker and not DateTimePicker maybe we can add the day over there. This would make the BE not deal with the wonky logic of adding days. Also, this would mean when we introduce a DateTimePicker in the FE we don't have to change the BE to support it.

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/entities/search/impl/CreatedOnFacetHandler.java
```diff
@@ -21,33 +21,113 @@
...
+                facetInputs.endDateTime = facetInputs.endDate.plusDays(1).format(DATE_TIME_FORMAT);
+            }
+            facetInputs.dateRange.put("startDateTime", facetInputs.startDateTime);
+            facetInputs.dateRange.put("endDateTime", facetInputs.endDateTime);
+        } else {
+            throw new DataHubException("The date range is missing for createdOn in your request");
+        }
+    }
 
+    public static final class CreatedOnFacetInputs {
```
akshaysonvane, line 123: We can have protected access for the class.

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/entities/search/impl/CreatedOnFacetHandler.java
rahulvudutala: Date and time for entity properties. We still have a date picker for hub properties
It is more accurate doing < than <= as we don't know appropriate end time for a date.

The feature supports client zone offset. To work with zone it has to be date-time, So converted to date-time even if date picker is used

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/entities/search/impl/CreatedOnFacetHandler.java
rahulvudutala: No it doesn't. `facetInputs.startDate = facetInputs.startDate.with(TemporalAdjusters.firstDayOfMonth());` 
is used here.
`facetInputs.startDateTime = facetInputs.startDate.toLocalDate().atStartOfDay(facetInputs.zoneId).format(DATE_TIME_FORMAT);`
This is the only common line.

akshaysonvane: I meant lines 99 to 102.

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/entities/search/impl/CreatedOnFacetHandler.java
akshaysonvane: I meant can we add the day in the FE since the only reason we need to add a day here is that we use a date picker and not a date-time picker. Since this seems to be a FE dependency maybe the hack should also be in the FE.

Moving the logic to the FE would make BE work correctly for all date-time ranges and if/when we add the date-time picker in the FE we dont have to make any changes since we already support the right thing.

rahulvudutala: The date picker value will change, as we display the same object. Its not just FE, even if the endpoint is hit, it should return the same results no different from UI

PR #4083, opened 2020-06-10
DHFPROD-5170: Remove the word "documents" in search results

saarimrahman:
### Description
Changed the word "documents" to "results"/"records" in various areas in the UI visible to the user to remove MarkLogic specific language.

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/src/components/entities/mapping/source-entity-map/source-to-entity-map.test.tsx
+++ b/marklogic-data-hub-central/ui/src/components/entities/mapping/source-entity-map/source-to-entity-map.test.tsx
@@ -22,7 +22,7 @@ describe('RTL Source-to-entity map tests', () => {
         expect(getByText('Clear')).toBeDisabled;
         expect(getByTestId("entityContainer")).toBeInTheDocument();
         expect(getByTestId("srcContainer")).toBeInTheDocument();
-        expect(getByText("Unable to find source documents using the specified collection or query.")).toBeInTheDocument;
+        expect(getByText("Unable to find source records using the specified collection or query.")).toBeInTheDocument;
         expect(getByTestId("srcContainer")).toHaveClass("sourceContainer");
         expect(getByText('Entity: Person')).toBeInTheDocument();
         expect(getByRole('presentation').className).toEqual("Resizer vertical ");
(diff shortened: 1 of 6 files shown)


yasht2010:
Agreed with @rjrudin .. Let's drop the word documents everywhere

marklogic-data-hub-central/ui/src/components/entities/mapping/source-entity-map/source-to-entity-map.test.tsx
```diff
@@ -22,7 +22,7 @@ describe('RTL Source-to-entity map tests', () => {
         expect(getByText('Clear')).toBeDisabled;
         expect(getByTestId("entityContainer")).toBeInTheDocument();
         expect(getByTestId("srcContainer")).toBeInTheDocument();
-        expect(getByText("Unable to find source documents using the specified collection or query.")).toBeInTheDocument;
+        expect(getByText("Unable to find source records using the specified collection or query.")).toBeInTheDocument;
```
rjrudin, line 25: This is good, I think "records" works well in this context. FWIW, that's the term I found in the field to be most comfortable to users, who often think of "documents" as being things they create with Word and Powerpoint.

marklogic-data-hub-central/ui/src/config/guided-tour-steps.js
```diff
@@ -21,27 +21,27 @@ const browseTableViewSteps = [
...
   },
   {
     selector: '.ant-table-row',
-    content: 'Document results include entity name, primary key (or document URI), table information, and metadata'
+    content: 'Record results include entity name, primary key (or record URI), table information, and metadata'
   },
   {
     selector: '.ant-table-row-expand-icon-cell',
-    content: 'Expansion button will show all the properties of each document'
+    content: 'Expansion button will show all the properties of each record'
```
rjrudin, line 36: I think we can say "entity" here, but "record" is still safe, as "record" is always a safe generalization.

marklogic-data-hub-central/ui/src/components/entities/mapping/source-entity-map/source-to-entity-map.test.tsx
bsrikan: @rjrudin is the scope of this change marklogic-data-hub-central only?

rjrudin: That's really a question for @yasht2010 , but I'm pretty sure the answer is "yes".

yasht2010:
Yes. This is applicable to the entire Data Hub 5.3 release. Hub Central is one part of that release.

bsrikan:
Thanks @yasht2010 . Since the scope is all of DH code, there are references in web as well as marklogic-data-hub to "documents" that is visible to user. There could be more, I could find following. 
- log info/warn messages
- run time exception messages
- schema and parameter description
- tooltips,components in QuickStart

rjrudin:
@yasht2010 @bsrikan The scope of this is just Hub Central. Initially it was just one mention of "documents" that I noticed, but then I did a quick search of instances of "documents" that are user-facing in HC, and the story expanded. But analyzing all of the DHF codebase would take many hours, as "documents" is likely an appropriate choice in some places.

yasht2010:
We can restrict it to HC for now to limit scope.

PR #4085, opened 2020-06-11
DHFPROD-5063: Speeding up HC tests

rjrudin:
Added installOnlyReferenceModelEntities for tests that only need entities, which saves about 3 seconds for each test. Combined some test classes too so that permission tests are now in the *StepControllerTest classes, which are now defaulting to using the step-specific classes. 

Overall time on my machine went from 2m4s to 1m30s.

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


rjrudin:
Rebased

PR #4086, opened 2020-06-11
DHFPROD-4677: Add hub-central-custom-reader role

ryanjdew:
### Description

No UI tests needed, as the Custom tab in curate hasn't been added yet.

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


marklogic-data-hub/src/main/resources/hub-internal-config/security/roles/hub-central-custom-reader.json
```diff
@@ -0,0 +1,19 @@
+{
+  "role-name": "hub-central-custom-reader",
+  "description": "Permits reading/reviewing custom steps in Hub Central",
```
srinathgit, line 3: Since custom steps do not exist as a separate artifacts and only exist inline within flows, should "data-hub-flow-reader" role be added and roles "hub-central-load-reader", "hub-central-mapping-reader","data-hub-match-merge-reader" be dropped ?

rjrudin: I think this story is coming earlier than it should be, but it doesn't hurt anything. But there's also not really a way to test it from a user's perspective.

marklogic-data-hub/src/main/resources/hub-internal-config/security/privileges/data-hub-read-step-definition.json
```diff
@@ -0,0 +1,5 @@
+{
+  "privilege-name": "data-hub:readStepDefinition",
```
rjrudin, line 2: What's this needed for? Does it map to something in HC?

marklogic-data-hub/src/main/resources/hub-internal-config/security/roles/hub-central-custom-reader.json
srinathgit: If access to viewing custom steps have to be restricted in DHFPROD-4997, DHFPROD-4998, then, "data-hub-flow-reader" role should be added right ?

ryanjdew: @srinathgit Similar to the other step types, the user should have access to the step document, not the flow. The primary information of the custom step is contained in the step definition.

Since all Custom step types will appear under the Custom tab, I believe you are correct that we can drop the hub-central-loader-reader, hub-central-mapping-reader, etc., roles.

marklogic-data-hub/src/main/resources/hub-internal-config/security/privileges/data-hub-read-step-definition.json
ryanjdew: @rjrudin The privilege/authority is for custom steps. I can change it to readCustom if we think it makes more sense. It is listed as stepDefinition in the roles wiki and is similar to readIngestion being closer to what we have on the backed. Perhaps we should use readCustom because we'll create separate custom steps aside from step definitions in the future?

ryanjdew:
After some discussion with over the roles for hub-central-custom-reader, hub-central-load-reader and hub-central-mapping-reader aren't needed. Also moving the privileges to the namespace "hub-central:" to make their domain clear.

PR #4088, opened 2020-06-11
DHFPROD-5164: DH project for e2e tests in HC

bsrikan:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- N/A Added Tests
  

- ##### Reviewer:

- N/A Reviewed Tests


bsrikan:
@ryanjdew I tried adding a mastering step, but getting an error. Pinged you in slack.
Issues:

- not sure about the address reduce which uses Address.street.
- dob match results in notification and no merge. SM project merges the dob match

bsrikan:
Thanks Ryan. Pushed changes with flow including mastering step as well. 
The results of the match/merge and mastering seems ok to me.

bsrikan:
Update: The purpose of DHFPROD-5164 was to update the reference entity model project with mastering steps so e2e tests can be run. Rob suggested to keep the e2e project separate.

So the story is now just for creating a project for running e2e tests. We still want to use the entityTypes in the reference project and so copied over the same.

Updating entity reference model project is now not in scope of this ticket.

PR #4101, opened 2020-06-16
DHFPROD-5193: hubCreateStepDefinition only supports ingestion/custom

rjrudin:
Removed OBE files under custom-step-module (were no references to them).

Removed single-use custom Exception classes in favor of using GradleException within CreateStepDefinitionTask.

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [ ] Added Tests
  

- ##### Reviewer:

- [ ] Reviewed Tests


rjrudin:
@srinathgit  The modules being used are under ./scaffolding/custom-module; those are all still there. The ./scaffolding/custom-step-module directory contains the same files as ./scaffolding/custom-module, but it is not referenced anywhere and thus can and should be deleted.

rjrudin:
Jenkins complained about "No JIRA ID" found, which I don't understand since there's a JIRA ID there. So I amended the commit comment and force pushed.

bsrikan:
@rjrudin which task should Pari use to create match/merge/mastering steps?

rjrudin:
@bsrikan hubCreateStep is used for creating referenced steps, though that's only for ingestion and mapping steps in 5.3. 

If you mean step definitions - DHF 5.2 doesn't support custom merge/match step definitions, and we're getting rid of the concept of mapping/mastering step definitions as part of this story.

PR #4102, opened 2020-06-16
DHFPROD-5177: Changes to EntitySearch API to accept propertiesToDisplay for tabular view

rahulvudutala:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/entities/entity-search-lib.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/entities/entity-search-lib.sjs
@@ -23,20 +23,26 @@ const entityLib = require("/data-hub/5/impl/entity-lib.sjs");
  * If the entity instance cannot be found for any search result, that fact is logged instead of an error being thrown or
  * trace logging being used. This ensures that the condition appears in logging, but it should not throw an error
  * because other entities in the search results may be able to have properties added for them.
- * 
+ *
  * @param entityName
  * @param searchResponse
  */
-function addPropertiesToSearchResponse(entityName, searchResponse) {
+function addPropertiesToSearchResponse(entityName, searchResponse, propertiesToDisplay) {
   const maxDefaultProperties = 5;
-
+  const selectedPropertyNames = typeof propertiesToDisplay === 'string' ? propertiesToDisplay.split(",") :  propertiesToDisplay;
   const entityModel = entityLib.findModelByEntityName(entityName);
   if (!entityModel) {
     ds.throwServerError(`Could not add entity properties to search response; could not find an entity model for entity name: ${entityName}`);
   }
 
-  const propertyMetadata = buildPropertyMetadata("", entityModel, entityName);
-  const selectedPropertyMetadata = propertyMetadata.slice(0, maxDefaultProperties);
+  const allMetadata = buildAllMetadata("", entityModel, entityName);
+  const propertyMetadata = allMetadata["allPropertiesMetadata"];
+
+  let selectedPropertyMetadata = [];
+  if(selectedPropertyNames) {
+    selectedPropertyMetadata = buildSelectedPropertiesMetadata(allMetadata, selectedPropertyNames);
+  }
+  selectedPropertyMetadata = selectedPropertyMetadata.length > 0 ? selectedPropertyMetadata : propertyMetadata.slice(0, maxDefaultProperties);
 
   // Add entityProperties to each search result
   searchResponse.results.forEach(result => {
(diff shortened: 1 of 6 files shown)


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/entities/entity-search-lib.sjs
```diff
@@ -98,13 +104,19 @@ function buildPropertyMetadata(parentPropertyName, entityModel, entityName) {
...
+    if(selectedProperties && selectedProperties.includes(propertyMetadata.propertyPath)) {
+      selectedPropertyMetadata.push(propertyMetadata);
+    }
   }
   return allPropertyMetadata;
 }
 
-function getPropertyValues(currentProperty, entityInstance) {
+// Helper function used by getPropertyValuesFromInstance to fetch property values from a propertyPath
+function getPropertyPathValues(currentProperty, entityInstance) {
```
rjrudin, line 116: It'll help a lot to define method comments here that define currentProperty as an object along with the expected keys in that object, which I believe are propertyPath and datatype. Otherwise, I think a reader would not know if "currentProperty" is simply a property name/path or a complex object.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/entities/entity-search-lib.sjs
```diff
@@ -19,24 +19,28 @@ const ds = require("/data-hub/5/data-services/ds-utils.sjs");
 // TODO Will move this to /data-hub/5/entities soon
 const entityLib = require("/data-hub/5/impl/entity-lib.sjs");
 
+let selectedPropertyMetadata = null;
```
rjrudin, line 22: Let's make these parameters that are passed to functions instead of them being library-level variables. That makes the scope of functions easier to understand, as they won't be depending on library-level variables where it's not clear when the state of those variables is modified.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/entities/entity-search-lib.sjs
```diff
@@ -141,6 +153,81 @@ function getPropertyValues(currentProperty, entityInstance) {
...
+    resultObject.propertyValue = propertyValuesArray;
+  } else {
+    resultObject.propertyValue =  getPropertyPathValues(currentProperty, entityInstance).propertyValue;
+  }
+  return resultObject;
+}
+
+// This function fetches the appropriate instance for first level simple, structured property paths and other levels of
+// structured entity properties from the propertyPath.
+function getInstanceFromPropertyPath(currentProperty, entityInstance) {
```
rjrudin, line 183: This naming is confusing - the only time we say "instance" is when referring to an "entity instance". Is this really a "value instance", or simply "property value"?

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/entities/entity-search-lib.sjs
```diff
@@ -141,6 +153,81 @@ function getPropertyValues(currentProperty, entityInstance) {
...
+  let splitPropertyNames = currentProperty.propertyPath.split(".");
+  splitPropertyNames.pop();
+
+  if(splitPropertyNames.length > 0) {
+    splitPropertyNames.forEach((propertyName) => {
+      let propertyInstanceArray = [];
+      if(Array.isArray(entityInstance)) {
```
rjrudin, line 190: This raises the same question above - an entity instance is always an object, it'll never be an array. So what is this variable actually referring to?

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/entities/entity-search-lib.sjs
```diff
@@ -60,19 +66,21 @@ function addPropertiesToSearchResponse(entityName, searchResponse) {
...
+// This function builds the logical entityType property metadata for all entityType properties from an entityModel.
+function buildAllMetadata(parentPropertyName, entityModel, entityName) {
   const entityType = entityModel.definitions[entityName];
   if (!entityType) {
     ds.throwServerError("Could not build property metadata; could not find entity type with name: " + entityName);
   }
 
-  const allPropertyMetadata = [];
+  const allPropertiesMetadata = [];
+  let granularPropertyMetadata = {};
```
rjrudin, line 83: What does "granular" mean in this context? Is it really "selectedPropertyMetadata"?

rahulvudutala: `{
              "propertyPath": "billing.zip.plusFour",
              "propertyLabel": "plusFour",
              "datatype": "string",
              "multiple": false
            }`
The parent structure from where it is referred to is ignored. But its child properties do exist in the object if the object has them

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/entities/entity-search-lib.sjs
```diff
@@ -60,19 +66,21 @@ function addPropertiesToSearchResponse(entityName, searchResponse) {
       }
     }
   });
-
   // Make it easy for the client to know which property names were used, and which ones exist
+  // searchResponse.selectedPropertyDefinitions = selectedPropertyMetadata;
```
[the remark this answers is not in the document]
rahulvudutala: I will remove this. Thanks

PR #4103, opened 2020-06-16
DHFPROD-5139: User with hub-central-entity-exporter role can't export data

saarimrahman:
### Description
Added necessary roles and privileges for user with role "hub-central-entity-exporter" to export CSV data. Also correctly renamed a file and added "hub-central-saved-query-user" role to the bertrand user in reference-entity-model.

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/controllers/EntitySearchControllerTest.java
+++ b/marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/controllers/EntitySearchControllerTest.java
@@ -160,15 +160,15 @@ public class EntitySearchControllerTest extends AbstractMvcTest {
 
 
         // Try exporting without the required role "hub-central-entity-exporter"
-        loginAsTestUserWithRoles("hub-central-user","data-hub-operator");
+        loginAsTestUserWithRoles("hub-central-user");
         postWithParams(EXPORT_PATH, getRequestParams(limit, json))
             .andExpect(status().isForbidden());
         getJson(EXPORT_PATH + "/query/non-existent-query-id", getRequestParams(limit, null))
             .andExpect(status().isForbidden());
 
 
-        // Set the required role and re-login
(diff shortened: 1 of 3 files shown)


marklogic-data-hub/src/main/resources/hub-internal-config/security/roles/hub-central-entity-exporter.json
```diff
@@ -2,13 +2,25 @@
   "role-name": "hub-central-entity-exporter",
   "description": "Permits exporting entity instance documents",
   "role": [
-    "hub-central-user"
+    "hub-central-user",
```
rjrudin, line 5: As part of this, I think EntitySearchControllerTest needs to be modified to login a user with this role, instead of this role plus 2 other roles.

marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/controllers/EntitySearchControllerTest.java
```diff
@@ -214,6 +214,22 @@ void testRowExport() throws Exception {
...
+            .andExpect(request().asyncStarted())
+            .andDo(MvcResult::getAsyncResult)
+            .andExpect(status().isOk())
+            .andDo(result -> {
+                String response = result.getResponse().getContentAsString();
+                Set<Integer> actualRowSet = calculateHash(response);
+                assertRowsAndColumns(limit, totalColumns, response);
+                assertTrue(actualRowSet.contains(getHashCode(customer1Info)));
+                assertTrue(actualRowSet.contains(getHashCode(customer2Info)));
+            });
```
akshaysonvane, line 232: Instead of adding an extra call to export, we can modify the existing tests to remove "data-hub-operator" and "hub-central-saved-query-user" roles for the first test where we export using query document.

Next, we can add "hub-central-saved-query-user" for login when we test export using queryId which would require for a query to be saved.

PR #4105, opened 2020-06-16
DHFPROD-4970: add validate entity field to mapping step

AkshayBajajML:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/schemas/StepSchema.java
+++ b/marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/schemas/StepSchema.java
@@ -106,7 +106,7 @@ public class StepSchema {
      */
     @JsonProperty("validateEntity")
     @JsonPropertyDescription("Applicable to mapping steps only")
-    private Boolean validateEntity;
+    private String validateEntity;
     @JsonProperty("targetDatabase")
     private String targetDatabase;
     @JsonProperty("sourceDatabase")
@@ -335,7 +335,7 @@ public class StepSchema {
      * 
      */
     @JsonProperty("validateEntity")
-    public Boolean getValidateEntity() {
+    public String getValidateEntity() {
         return validateEntity;
     }
 
@@ -344,7 +344,7 @@ public class StepSchema {
      * 
      */
     @JsonProperty("validateEntity")
-    public void setValidateEntity(Boolean validateEntity) {
+    public void setValidateEntity(String validateEntity) {
         this.validateEntity = validateEntity;
     }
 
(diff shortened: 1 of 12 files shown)


srinathgit:
@AkshayBajajML ,
This PR needs a rebase.

marklogic-data-hub-central/ui/src/components/advanced-settings/advanced-settings-dialog.tsx
```diff
@@ -179,6 +185,7 @@ const AdvancedSettingsDialog = (props) => {
           setTargetPermissions(response.data.permissions);
           setTargetFormat(response.data.targetFormat);
           setProvGranularity(response.data.provenanceGranularityLevel);
+          response.data.validateEntity === false ? setValidateEntity(defaultValidateEntity):setValidateEntity(response.data.validateEntity) ;
```
xnikhil08, line 188: Can't we get the default value for 'validateEntity' from backend only, instead of UI figuring it out based on its value in the API response. We can set it in UI as well but we need to be sure that its value will be 'false' in the response only the first time when the step is created.
It will be a bit cleaner to just set its default value in backend where the settings object is created for any step. Then UI will only care about the its value in the GET response.

ryanjdew: @AkshayBajajML You can set default values for the mapping step here: https://github.com/marklogic/marklogic-data-hub/blob/c2e709789a211aea9c16339632f66ffe4361f948/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/artifacts/mapping.sjs#L84

marklogic-data-hub/src/main/resources/hub-internal-artifacts/step-definitions/mapping/marklogic/entity-services-mapping.step.json
```diff
@@ -15,7 +15,7 @@
       "default-mapping",
       "mdm-content"
     ],
-    "validateEntity": false,
+    "validateEntity": "notValidate",
```
srinathgit, line 18: I am not sure if "notValidate" sounds good for the value of this property. 
@ryanjdew , @rjrudin, 
Is this value good ? , if not, any suggestions for the value for this property ? How does "none" sound like ?

ryanjdew: "doNotValidate" would be more closely aligned with UI

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/schemas/StepSchema.java
```diff
@@ -332,7 +332,7 @@ public void setHeaders(Headers headers) {
 
     /**
      * Applicable to mapping steps only
-     * 
+     *
      */
     @JsonProperty("validateEntity")
     public Boolean getValidateEntity() {
```
srinathgit, line 338: You will have to update the data type in schema json file and the run 'generateJsonSchema2Pojo' task , so that it's updated from Boolean to String

srinathgit:
@AkshayBajajML 
This PR needs a rebase

PR #4109, opened 2020-06-18
DHFPROD-4997:View Custom ingestion step

srinathgit:
### Description
View Custom ingestion step. It should show up in "Load"  tile as usual and on opening "Settings", "Step Definition Name" should be present(which is not present for ootb ingestion step).
https://wiki.marklogic.com/display/ENGINEERING/Data+Load?preview=/125169351/171055005/CustomLoadDataSettings.png

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/src/components/advanced-settings/advanced-settings-dialog.tsx
+++ b/marklogic-data-hub-central/ui/src/components/advanced-settings/advanced-settings-dialog.tsx
@@ -26,6 +26,9 @@ const AdvancedSettingsDialog = (props) => {
   const usesSourceDatabase = stepType !== 'ingestion';
   const defaultSourceDatabase = usesSourceDatabase ? 'data-hub-STAGING' : 'data-hub-FINAL';
 
+  const [isCustomIngestion, setIsCustomIngestion] = useState(false);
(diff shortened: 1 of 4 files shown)


marklogic-data-hub-central/ui/src/components/advanced-settings/advanced-settings-dialog.tsx
```diff
@@ -166,6 +172,11 @@ const AdvancedSettingsDialog = (props) => {
         let response = await Axios.get(`/api/steps/${stepType}/${props.stepData.name}`);
 
         if (response.status === 200) {
+          if(stepType === 'ingestion' && response.data.stepDefinitionName !== 'default-ingestion'){
```
rjrudin, line 175: Normally, I'd say the UI shouldn't be bound to this knowledge, and the middle tier should instead insert e.g. a "isCustomStepDefinition" property for the UI to be bound to. But since I think "Step Definition Name" should always be present and that we should eventually do that, I think it's fine for the UI to know about this detail for now.

PR #4112, opened 2020-06-18
DHFPROD-4843: Display service name in header, system-info dialog

wooldridge:
Add service name to UI (replacing project name) using new systemInfo endpoint. 

Screencast: https://drive.google.com/file/d/1gkfchbMGSUZzR5UgqRpds5dHTvjsvtpU/view

Also:

- Refactor project-page component, use new /api/environment/systemInfo endpoint.
- Rename ProjectPage component to SystemInfo
- Move SystemInfo component into header component directory since only Header calls SystemInfo
- Refactor mocked user-context data and associated types.
- Fix some typing issues in test mocks.
- Test for service name in: header.test, system-info.test

From @rjrudin: 

- /api/environment/project-info is now /api/environment/systemInfo.
- Removing /api/info - sessionTimeout is now part of systemInfo.
- No longer including projectName in the authentication response either, to avoid redundancy.

Thinking that SystemService.getVersions should be renamed to SystemService.getSystemInfo too, but will tackle that later in a separate PR to minimize the amount of changes here.

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


marklogic-data-hub-central/ui/src/components/header/system-info.test.tsx
```diff
@@ -0,0 +1,60 @@
...
+    test('verify project info display, user with "Download" button enabled', async () => {
+        const authorityService = new AuthoritiesService();
+        authorityService.setAuthorities(['downloadProjectFiles']);
+        const { getByText } = render(<Router><AuthoritiesContext.Provider value={authorityService}>
+            <SystemInfo 
+                serviceName={data.environment.serviceName}
+                dataHubVersion={data.environment.dataHubVersion}
+                marklogicVersion={data.environment.marklogicVersion}
+                systemInfoVisible={true} 
+                setSystemInfoVisible={jest.fn()}
```
bsrikan, line 23: This is good. I would probably add systemInfoVisible: true, setSystemInfoVisible: jest.fn() in system-info.data.ts and use that here like so: {...data.environment}

wooldridge: @bsrikan:

> Can we also pls update "marklogic-data-hub-central/ui/e2e/cypress/support/commands.js" to include systemInfo

Can you do this? I haven't done anything with the e2e cypress stuff, not sure what you're looking for.

wooldridge: @bsrikan:

> This is good. I would probably add systemInfoVisible: true, setSystemInfoVisible: jest.fn() in system-info.data.ts and use that here like so: {...data.environment}

Would be nice for brevity, but I don't want to put those additional properties in the environment object since they aren't returned from the environment method in the app.

marklogic-data-hub-central/ui/src/util/environment.tsx
```diff
@@ -1,16 +1,15 @@
...
 const defaultEnv = {
-    projectName: '',
+    serviceName: '',
     dataHubVersion: '',
     markLogicVersion: ''
 }
 
 export function setEnvironment()  {
-    axios.get('/api/environment/project-info')
+    axios.get('/api/environment/systemInfo')
```
rjrudin, line 10: One thought I've had for the future is to standardize on (I think) "system" instead of using "environment" and "system" interchangeably, as having those separate terms raises the question of - What does each one mean? "serviceName" adds a little more confusion to because it implies there's the concept of a "service" that is separate from "system". Really, system = environment = service, so in the future, I think we'll simplify the naming here.

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/VersionInfo.java
```diff
@@ -16,18 +16,21 @@
 
     private String hubVersion;
     private String markLogicVersion;
+    private String clusterName;
 
     public static VersionInfo newVersionInfo(HubClient hubClient) {
```
rjrudin, line 21: I expect to rename this to SystemInfo in a future JIRA ticket as well, didn't want to increase the blast radius of this PR though. Also note that "clusterName" is appropriate here because this is a DHF class, and it doesn't know specifically about DHS, which is where the concept of "serviceName" exists.

marklogic-data-hub-central/ui/src/components/header/system-info.test.tsx
bsrikan: Sure @wooldridge makes sense for test data props. I will update the endpoint for the cypress command support with the new endpoint.

PR #4113, opened 2020-06-18
DHFPROD-4679: Adding hub-central-developer and hub-central-operator r…

ryanjdew:
…oles

### Description

Not a whole lot to be done in the way of tests, as we are testing the lower-level roles that are inherited already. 

Also note that there are a few roles listed in the wiki that haven't been implemented yet.

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/examples/reference-entity-model/src/main/ml-config/security/users/hub-developer.json
+++ b/examples/reference-entity-model/src/main/ml-config/security/users/hub-developer.json
@@ -3,12 +3,6 @@
   "description": "Generic developer account; will soon be superceded by persona-specific accounts",
   "password": "%%testPassword%%",
   "role": [
-    "data-hub-developer",
-    "hub-central-user",
-    "hub-central-entity-model-writer",
-    "hub-central-load-writer",
-    "hub-central-mapping-writer",
-    "hub-central-flow-writer",
-    "hub-central-downloader"
+    "hub-central-developer"
   ]
 }
--- a/examples/reference-entity-model/src/main/ml-config/security/users/pari.json
+++ b/examples/reference-entity-model/src/main/ml-config/security/users/pari.json
@@ -3,10 +3,7 @@
   "description": "Pari is a MarkLogic expert",
   "password": "%%testPassword%%",
   "role": [
-    "data-hub-developer",
-    "hub-central-downloader",
-    "hub-central-load-writer",
-    "hub-central-mapping-writer",
-    "hub-central-flow-writer"
+    "hub-central-developer",
+    "data-hub-developer"
   ]
 }
--- a/marklogic-data-hub/src/main/resources/hub-internal-config/security/roles/hub-central-developer.json
+++ b/marklogic-data-hub/src/main/resources/hub-internal-config/security/roles/hub-central-developer.json
@@ -1,0 +1,14 @@
+{
+  "role-name": "hub-central-developer",
+  "description": "Permits developing and operating a Data Hub application via Hub Central",
+  "role": [
+    "hub-central-operator",
+    "hub-central-entity-model-writer",
+    "hub-central-flow-writer",
+    "hub-central-load-writer",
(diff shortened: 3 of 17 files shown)


examples/reference-entity-model/src/main/ml-config/security/users/hub-developer.json
```diff
@@ -3,12 +3,6 @@
...
   "password": "%%testPassword%%",
   "role": [
-    "data-hub-developer",
-    "hub-central-user",
-    "hub-central-entity-model-writer",
-    "hub-central-load-writer",
-    "hub-central-mapping-writer",
-    "hub-central-flow-writer",
-    "hub-central-downloader"
+    "hub-central-developer"
```
rjrudin, line 6: I think we should just delete this user and hub-operator, since we have pari and melissa.

marklogic-data-hub/src/main/resources/hub-internal-config/security/roles/hub-central-developer.json
```diff
@@ -0,0 +1,12 @@
+{
+  "role-name": "hub-central-developer",
+  "description": "Permits developing and operating a Data Hub application via Hub Central",
+  "role": [
```
bsrikan, line 4: seems to be missing:
hub-central-entity-model-writer and
hub-central-clear-user-data(should be allowed, since downloader is added?)

marklogic-data-hub/src/main/resources/hub-internal-config/security/roles/hub-central-operator.json
```diff
@@ -0,0 +1,14 @@
+{
+  "role-name": "hub-central-operator",
+  "description": "Permits operating a Data Hub application via Hub Central",
+  "role": [
+    "hub-central-user",
```
bsrikan, line 5: seems to be missing:
hub-central-entity-model-reader

examples/reference-entity-model/src/main/ml-config/security/users/pari.json
```diff
@@ -3,10 +3,6 @@
   "description": "Pari is a MarkLogic expert",
   "password": "%%testPassword%%",
   "role": [
-    "data-hub-developer",
-    "hub-central-downloader",
-    "hub-central-load-writer",
-    "hub-central-mapping-writer",
-    "hub-central-flow-writer"
+    "hub-central-developer"
```
bsrikan, line 6: Pari should have data-hub-developer right and data-hub-developer should inherit hub-central-developer?

marklogic-data-hub/src/main/resources/hub-internal-config/security/roles/hub-central-developer.json
ryanjdew: Yes. I was missing the clear-user-data and explorer related roles from the list in the wiki. It should be resolved now.

examples/reference-entity-model/src/main/ml-config/security/users/pari.json
ryanjdew: No. hub-central-developer grants access to the Hub Central UI. Having having access to Hub Central shouldn't be granted via the data-hub-developer role.

bsrikan: ok makes sense. Although, I would still expect Pari a MarkLogic expert to have data-hub-developer role.

rjrudin: I missed this - pari should still have data-hub-developer, in addition to hub-central-developer, as she'll be doing things like running hubDeploy. 

Perhaps HCD should inherit DHD by default? Is there really a use case for someone to be an HCD but not a DHD?

PR #4114, opened 2020-06-18
DHFPROD-4180: Delete entity model

akshaysonvane:
### Description
* Added two new endpoints
    1. GET /api/models/{modelName}/references - Returns the usage of the model in steps and other models
    2. DELETE /api/models/{modelName} - Deletes the model
* Updated `hub-central-entity-model-writer` role to make it work with deletion.
* Added update permission for `data-hub-entity-model-writer` role on schema and TDE docs.
* Fixed the baseUri to include the closing `/`

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/controllers/ModelController.java
+++ b/marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/controllers/ModelController.java
@@ -90,6 +90,28 @@ public class ModelController extends BaseController {
         return new ResponseEntity<>(HttpStatus.OK);
     }
 
+    @RequestMapping(value = "/{modelName}", method = RequestMethod.DELETE)
+    @Secured("ROLE_writeEntityModel")
+    public ResponseEntity<Void> deleteModel(@PathVariable String modelName) {
+        newService().deleteModel(modelName);
+
+        /*
+        * We're not doing anything with indexes or protected paths here because we don't have a reliable way to
+        * identify the ones to delete and we only expect entity types to be deleted
+        * in a development environment where it's usually fine if some old indexes / protected paths hang around.
+        */
+        logger.info("Deploying search options");
+        deploySearchOptions(newService().generateModelConfig());
+
+        return emptyOk();
+    }
+
+    @RequestMapping(value = "/{modelName}/references", method = RequestMethod.GET)
+    @ApiOperation(value = "Get step and model names that refer to this model.", response = ModelReferencesInfo.class)
+    public ResponseEntity<JsonNode> getModelReferences(@PathVariable String modelName) {
+        return ResponseEntity.ok(newService().getModelReferences(modelName));
+    }
+
     @RequestMapping(value = "/{modelName}/entityTypes", method = RequestMethod.PUT)
     @ApiImplicitParam(required = true, paramType = "body", dataType = "ModelDefinitions")
     @Secured("ROLE_writeEntityModel")
@@ -252,5 +274,10 @@ public class ModelController extends BaseController {
 
     public static class UpdateModelInfoInput {
         public String description;
+    }
+
+    public static class ModelReferencesInfo {
+        public List<String> stepAndMappingNames;
+        public List<String> entityNames;
     }
 }
--- a/marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/controllers/model/DeleteModelTest.java
+++ b/marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/controllers/model/DeleteModelTest.java
@@ -1,0 +1,99 @@
+package com.marklogic.hub.central.controllers.model;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.marklogic.client.DatabaseClient;
+import com.marklogic.client.FailedRequestException;
+import com.marklogic.client.document.GenericDocumentManager;
+import com.marklogic.hub.DatabaseKind;
+import com.marklogic.hub.central.controllers.ModelController;
+import com.marklogic.hub.impl.EntityManagerImpl;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.security.test.context.support.WithMockUser;
+
+import java.util.function.Consumer;
+import java.util.stream.Stream;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+public class DeleteModelTest extends AbstractModelTest {
+
+    @Autowired
+    ModelController controller;
+
+    @Test
+    @WithMockUser(roles = {"writeEntityModel"})
+    void testModelDeletionEndpoints() {
+        runAsDataHubDeveloper();
+        installProjectInFolder("test-projects/delete-model-project");
+        new EntityManagerImpl(getHubConfig()).deployQueryOptions();
+
+        runAsTestUserWithRoles("hub-central-entity-model-writer");
+        verifyEntity2BasedArtifactsExist();
+        getModelReferences();
+        deleteModel();
+        verifyEntity2BasedArtifactsDontExist();
+        verifyReferencesToEntity2DontExistInEntity1();
+    }
+
+    private void getModelReferences() {
+        JsonNode jsonNode = controller.getModelReferences("Entity2").getBody();
+        assertNotNull(jsonNode);
+        assertEquals(2, jsonNode.get("stepAndMappingNames").size());
+        assertEquals(1, jsonNode.get("entityNames").size());
+        Stream.of("testMap2", "matching-step")
+                .forEach(s -> assertTrue(jsonNode.get("stepAndMappingNames").toString().contains(s)));
+        assertTrue(jsonNode.get("entityNames").toString().contains("Entity1"));
+    }
+
+    private void deleteModel() {
+        assertThrows(FailedRequestException.class, () -> controller.deleteModel("Entity2"), "Should throw an exception since the entity is referenced in steps.");
+
+        runAsDataHubDeveloper();
+        removeReferencesToEntity();
+        runAsTestUserWithRoles("hub-central-entity-model-writer");
+
+        assertTrue(controller.deleteModel("Entity2").getStatusCode().is2xxSuccessful(), "Should be ok since we deleted the steps that refer to the entity.");
+
+        // Needed because a post-commit trigger is executed after a model is deleted
+        runAsDataHubDeveloper();
+        waitForTasksToFinish();
+        runAsTestUserWithRoles("hub-central-entity-model-writer");
+    }
+
+    private void removeReferencesToEntity() {
+        DatabaseClient stagingDatabaseClient = getHubClient().getStagingClient();
+        DatabaseClient finalDatabaseClient = getHubClient().getFinalClient();
+        Stream.of(stagingDatabaseClient, finalDatabaseClient)
+                .forEach(databaseClient -> databaseClient
+                        .newDocumentManager()
+                        .delete("/flows/testFlow.flow.json", "/steps/mapping/testMap2.step.json"));
+    }
+
+    private void verifyReferencesToEntity2DontExistInEntity1() {
+        assertFalse(getFinalDoc("/entities/Entity1.entity.json").toString().contains("Entity2"), "Expected the properties that refer to Entity2 to be deleted from Entity1.");
+        assertFalse(getStagingDoc("/entities/Entity1.entity.json").toString().contains("Entity2"), "Expected the properties that refer to Entity2 to be deleted from Entity1.");
+    }
+
+    private void verifyEntity2BasedArtifactsDontExist() {
+        assertSearchOptions("Entity2", Assertions::assertFalse, false);
+        assertSchemasAndTDE(Assertions::assertNull);
+    }
+
+    private void verifyEntity2BasedArtifactsExist() {
+        assertSearchOptions("Entity2", Assertions::assertTrue, true);
+        assertSchemasAndTDE(Assertions::assertNotNull);
+    }
+
+    private void assertSchemasAndTDE(Consumer<Object> assertion) {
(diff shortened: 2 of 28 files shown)


marklogic-data-hub/src/main/resources/hub-internal-config/security/roles/data-hub-entity-model-writer.json
```diff
@@ -2,13 +2,19 @@
...
+    "tde-admin"
   ],
   "privilege": [
     {
       "privilege-name": "any-uri",
       "action": "http://marklogic.com/xdmp/privileges/any-uri",
       "kind": "execute"
+    },
+    {
+      "privilege-name": "rest-admin",
```
akshaysonvane, line 15: We need this for deploying search options.

marklogic-data-hub/src/main/resources/hub-internal-config/security/roles/data-hub-entity-model-writer.json
```diff
@@ -2,13 +2,19 @@
   "role-name": "data-hub-entity-model-writer",
   "description": "Permits updating entity model documents",
   "role": [
-    "data-hub-common-writer"
+    "data-hub-common-writer",
+    "tde-admin"
```
akshaysonvane, line 6: We need this so the TDE's can be updated once we delete the entity model.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/models/deleteModel.sjs
```diff
@@ -19,8 +19,23 @@ const ds = require("/data-hub/5/data-services/ds-utils.sjs");
...
+const entityModelUri = entityLib.getModelUri(entityName);
+
+const steps = entityLib.findModelReferencesInSteps(entityName, entityTypeId);
+if (steps.length) {
+  ds.throwServerError(`Cannot delete the entity type '${entityName}' because it is referenced by the following step names: ${steps}`);
+}
+
 entityLib.deleteModel(entityName);
+entityLib.deleteModelReferencesInOtherModels(entityModelUri, entityTypeId);
+entityLib.deleteModelRelatedTDE(entityName);
```
akshaysonvane, line 41: Moved TDE deletion from the trigger to this data service.

The previous TDE deletion code in the trigger module wasn't working as the URL for trigger document needs the model version which can only be obtained from the entity model that was originally deleted.
In order to make that work, we would have needed to perform a point in time query to fetch the entity model using something like this: `xdmp:request-timestamp()-1`.
As this is a post-commit trigger the task would be queued in the task server and run sometime after the original transaction commits. Since request timestamp value refers to when the task server executes the task and not the original deletion request, there wouldn’t be a reliable way to fetch the file in time.

marklogic-data-hub/src/test/java/com/marklogic/hub/hubcentral/ApplyDownloadedZipToProjectTest.java
```diff
@@ -36,7 +36,7 @@
      */
     @Test
     void deleteOneEntityThenDownloadAndApply() {
-        installProjectInFolder("test-projects/all-artifacts");
+        installProjectInFolder("test-projects/download-artifacts");
```
rjrudin, line 39: There are a few other tests that depend on this folder - let's hold off on renaming this, as it's adding to the number of files modified in this PR.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/models/getPrimaryEntityTypes.sjs
```diff
@@ -1,5 +1,5 @@
 /*
-  Copyright 2012-2019 MarkLogic Corporation
+  Copyright (c) 2020 MarkLogic Corporation
```
rjrudin, line 2: While this is a good change, let's hold off on in and use a separate JIRA/PR to clean up the copyright years. I did a quick search, and there are a lot more issues to fix than just these. It'll shrink the size of this PR which is good too, since it has some complex stuff in it.

marklogic-data-hub/src/test/java/com/marklogic/hub/hubcentral/ApplyDownloadedZipToProjectTest.java
akshaysonvane: Since we are deleting the entity model which is referenced by a couple of mapping artifacts in the `ApplyDownloadedZipToProjectTest` it used to fail based on the new changes in this PR.

I created a new project called `download-artifacts` based on the `all-artifacts`(not touching it since other tests depend on it) project, the difference here is that I've removed the steps that referenced the Entity that was being deleted so the tests can pass.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/models/getPrimaryEntityTypes.sjs
akshaysonvane: Done

akshaysonvane:
Updated the GET endpoint from 
/api/models/{modelName}/usage to /api/models/{modelName}/references

`CreateAndUpdateModelTest` still runs as a data hub developer user. I'll include the correct roles and privileges to the `hub-central-entity-model-writer` role in the upcoming task https://project.marklogic.com/jira/browse/DHFPROD-5156

akshaysonvane:
Fixed the test compilation issue on the develop branch.

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/controllers/ModelController.java
```diff
@@ -90,6 +90,21 @@
         return new ResponseEntity<>(HttpStatus.OK);
     }
 
+    @RequestMapping(value = "/{modelName}", method = RequestMethod.DELETE)
+    @Secured("ROLE_writeEntityModel")
+    public ResponseEntity<Void> deleteModel(@PathVariable String modelName) {
+        newService().deleteModel(modelName);
```
rjrudin, line 96: I think it's important to document why we're not doing anything with indexes or protected paths here because we don't have a reliable way to identify the ones to delete and we only expect entity types to be deleted in a development environment where it's usually fine if some old indexes / protected paths hang around. I say this because when I looked at this, I immediately thought - what about indexes? - before remembering that we're taking the approach above. 6 months from now, we'll be glad we have a comment here explaining why.

marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/controllers/model/DeleteModelTest.java
```diff
@@ -0,0 +1,93 @@
...
+
+import static org.junit.jupiter.api.Assertions.*;
+
+public class DeleteModelTest extends AbstractModelTest {
+
+    @Autowired
+    ModelController controller;
+
+    @Test
+    @WithMockUser(roles = {"writeEntityModel"})
```
rjrudin, line 26: WithMockUser shouldn't be needed anymore, right?

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/controllers/ModelController.java
```diff
@@ -253,4 +268,9 @@ private ModelsService newService() {
     public static class UpdateModelInfoInput {
         public String description;
     }
+
+    public static class ModelReferencesInfo {
+        public List<String> steps;
+        public List<String> models;
```
rjrudin, line 274: Since the "models" list contains entity names, this should really be called "entityNames", not "models". My assumption is that if something is called "models", it's a list of JSON entity models.

marklogic-data-hub/src/main/java/com/marklogic/hub/mapping/MappingImpl.java
```diff
@@ -61,7 +61,7 @@ public MappingImpl(String name, HubEntity entity) {
         if (entity != null) {
             InfoType info = entity.getInfo();
             if (info != null) {
-                String prefix = info.getBaseUri() != null ? info.getBaseUri() + "/" : "";
+                String prefix = info.getBaseUri() != null ? info.getBaseUri() : "";
```
rjrudin, line 64: Pinging @ryanjdew  on this - why the change here? I believe "/" was included to ensure that the base URI ends with a "/" which I think is a requirement of ES. Perhaps the "/" should only be included if info.getBaseUri() doesn't end with "/"?

marklogic-data-hub/src/testFixtures/java/com/marklogic/hub/test/Action.java
```diff
@@ -0,0 +1,11 @@
+package com.marklogic.hub.test;
+
+/**
+ * Represents an operation that accepts no input arguments and returns no
+ * result.
+ */
+@FunctionalInterface
+public interface Action {
+    
+    void execute();
```
rjrudin, line 10: Note that when you do need a no-arg no-return-value interface in the future, you can just use Runnable instead.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/entity-lib.sjs
```diff
@@ -208,6 +209,97 @@ function deleteModel(entityName) {
...
+ */
+function findModelReferencesInSteps(entityName, entityTypeId) {
+  const steps = [];
+  const mappingArtifacts = cts.search(cts.andQuery([
+    cts.collectionQuery('http://marklogic.com/data-hub/mappings'),
+    cts.jsonPropertyValueQuery("targetEntityType", [entityName, entityTypeId])
+  ]));
+  mappingArtifacts.toObject()
+    .forEach(mapping => {
+      mapping = JSON.parse(mapping);
```
rjrudin, line 228: I think you just need mapping.toObject() here? Could get fancy and do this (I think this works - I also often like to declare a query in a separate variable so it's easy to understand outside of the context where it's used, and the variable name can describe the intent of the query):

```
const mappingQuery = ...
const steps = cts.search(mappingQuery).toArray().map(mapping => mapping.toObject().name);
```

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/models/getModelReferences.sjs
```diff
@@ -0,0 +1,39 @@
...
+*/
+'use strict';
+
+const entityLib = require("/data-hub/5/impl/entity-lib.sjs");
+const ds = require("/data-hub/5/data-services/ds-utils.sjs");
+
+
+var entityName;
+if (!entityName) {
+  ds.throwBadRequest("Must specify a name in order to delete an entity model");
```
rjrudin, line 24: I think this should say "Must specify a name in order to get model references"

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/models/deleteModel.sjs
```diff
@@ -19,8 +19,22 @@ const ds = require("/data-hub/5/data-services/ds-utils.sjs");
...
+if (!entityModel) {
+  ds.throwServerError(`Could not find entity model with name: ${entityName}`);
+}
+
+const entityTypeId = entityLib.getEntityTypeId(entityModel, entityName);
+const entityModelUri = entityLib.getModelUri(entityName);
+
+const steps = entityLib.findModelReferencesInSteps(entityName, entityTypeId);
+if (steps.length) {
+  ds.throwServerError(`Cannot delete the entity type '${entityName}' because it is referenced by the following step names: ${steps}`);
```
rjrudin, line 36: I think we have to say "step and/or mapping names", since the list may contain legacy mapping names, which can be different from the names of the steps that reference them.

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/controllers/ModelController.java
```diff
@@ -253,4 +268,9 @@ private ModelsService newService() {
     public static class UpdateModelInfoInput {
         public String description;
     }
+
+    public static class ModelReferencesInfo {
+        public List<String> steps;
```
rjrudin, line 273: A "step" is a JSON object. Looking at the impl below, it looks like these are step names, with one exception - mapping names can be in here as well (sigh). To account for this, I think it's best to be verbose here and call this "stepAndMappingNames", because the list can contain a combination of step and mapping names. That's not pretty, but it captures the reality.

marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/controllers/model/DeleteModelTest.java
akshaysonvane: It's required to make spring security happy as this is not using the MockMvc test framework.

marklogic-data-hub/src/main/java/com/marklogic/hub/mapping/MappingImpl.java
akshaysonvane: This just reverts the changes made in https://github.com/marklogic/marklogic-data-hub/pull/4065 since we do need a baseUri to end with `/`

marklogic-data-hub/src/testFixtures/java/com/marklogic/hub/test/Action.java
akshaysonvane: I did not use Runnable as it is associated with threads and might cause confusion in the future.

marklogic-data-hub/src/main/java/com/marklogic/hub/mapping/MappingImpl.java
ryanjdew: @rjrudin Entity Services will add the slash ("/") appropriately whether or not the base URI ends with a slash.

PR #4123, opened 2020-06-22
DHFPROD-4421:After mapping step is run, I should be able to easily view my results in Explorer

Sanjeevani19:
### Description

- This PR has both backend and FE changes hence the two commits
- Fixed integration tests which were failing on develop previously
https://jenkins.marklogic.com/blue/organizations/jenkins/Datahub_CI/detail/develop/1046/tests 

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/src/pages/Run.tsx
+++ b/marklogic-data-hub-central/ui/src/pages/Run.tsx
@@ -5,6 +5,8 @@ import { Modal, Collapse } from 'antd';
 import axios from 'axios'
 import { AuthoritiesContext } from "../util/authorities";
 import { UserContext } from '../util/user-context';
+import {SearchContext} from "../util/search-context";
+import Browse from './Browse';
 
 const { Panel } = Collapse;
 
@@ -21,7 +23,12 @@ const Statuses = {
 }
 
 const Run = (props) => {
-   const { handleError, resetSessionTime } = useContext(UserContext)
+   const { handleError, resetSessionTime } = useContext(UserContext);
+
+    const {
+        setViewWithEntity,
+        setNextEntity,
+    } = useContext(SearchContext);
 
     const [isLoading, setIsLoading] = useState(false);
     const [flows, setFlows] = useState<any[]>([]);
@@ -71,7 +78,7 @@ const Run = (props) => {
                     setFlowsDefaultActiveKey(key);
                 }
                 setFlows(response.data);
-                
+
                 //console.log('GET flows successful', response);
             }
         } catch (error) {
@@ -173,12 +180,21 @@ const Run = (props) => {
         return stepType[0].toUpperCase() + stepType.substr(1);
     }
 
-    function showSuccess(stepName, stepType) {
-        Modal.success({
-        title: <p>{formatStepType(stepType)} "{stepName}" ran successfully</p>,
-            okText: 'Close',
-            mask: false
-        });
+    const goToExplorer = (entityName) => {
+        setViewWithEntity(<Browse/>, false, entityName)
+    }
+
+    function showSuccess(stepName, stepType, entityName) {
+          Modal.success({
+              title:<div><p style={{fontWeight: 400}}>{formatStepType(stepType)} step <strong>{stepName}</strong> ran successfully</p></div>,
+               okText: 'Close',
+               mask: false,
+               width:650,
+               content: stepType === 'mapping' ? <div onClick={()=> goToExplorer(entityName)} className={styles.exploreCuratedData}>
+                   <span className={styles.exploreIcon}></span>
+                   <span className={styles.exploreText}>Explore Curated Data</span>
+               </div> : ''
+           });
     }
 
     function getErrors(response) {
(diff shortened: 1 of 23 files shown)


marklogic-data-hub-central/ui/src/pages/Run.tsx
```diff
@@ -317,19 +338,25 @@ const Run = (props) => {
                         return axios.get('/api/jobs/' + jobId);
                     }, pollConfig.interval)
                     .then(function(response: any) {
+                        let entityName;
```
rjrudin, line 341: If the purpose of having targetEntityType on the Job document is to construct a link for exploring search results for a particular entity name, it'd be much easier for the backend to provide that in the initial RunFlowResponse that is returned by running a step. That avoids having to persist additional data, particularly since we'd only be persisting it for a very narrow UI concern. 

In addition, because a Job document may encompass multiple steps, and each step may be associated with a different entity type, we can't have a single "targetEntityType" on a Job document, because we may need multiple values.

But the RunFlowResponse returned by running a step has a Map of step responses, so we can have a targetEntityType on each step response (if one exists).

We also have multiple scenarios to account for here - 1) running a mapping step with a legacy mapping (can't happen in HC, but we need to account for it); 2) running a mapping step via HC; 3) running a matching/merging/mastering step that currently uses "targetEntity" instead of "targetEntityType". Regardless of the approach we take, we need to account for all of these scenarios so that users aren't confused when targetEntityType is only set some of the time.

I recommend opening up a backend task for this, as we'll need to properly define it for the scenarios listed above.

ryanjdew: @rjrudin This was done by me as a backend task and the commits have since been squashed. I can adjust to account for targetEntity as well. I had thought that was normalized at this point via the function we have on the backend for constructing the flows to run. 

As for the issue of multiple target entities, is that something that we need to accommodate for the running of a single job?

rjrudin: I feel more comfortable providing targetEntityType on a per-step basis as part of the API for running a step, as opposed to persisting it in a Job document where it's a lot more difficult to change. 

It looks like that this can be easily implemented/tested too by modifying this method in RunStepResponse:

```
public RunStepResponse withStep(String stepNum) {
```

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/jobs.sjs
```diff
@@ -343,11 +355,7 @@ module.exports.updateJob = module.amp(
           // a better error will be thrown later when the step is run and the module cannot be found.
         }
         if (jobsReportFun) {
-          let flowStep = fn.head(datahub.hubUtils.queryLatest(function () {
-              return datahub.flow.getFlow(stepResp.flowName).steps[step];
-            },
-            datahub.config.FINALDATABASE
-          ));
+          let flowStep = getFlowStep(datahub, stepResp.flowName, step);
```
rjrudin, line 358: Is there a reason to add caching here - do we know if there's a performance issue here? My understanding is that since this is only called by the mlJobs endpoint, it's only going to be called once in a given transaction, and thus the caching won't have any impact, correct?

ryanjdew: Removed the caching since it no longer made sense after dreverting other changes.

Sanjeevani19:
> Following issues must be fixed:
> 
> 1. Success message for load step doesnt seem to appear anymore with changes in this PR. Map/Match/merge/Mastering/Custom steps show success/failure message as expected.
> 2. Modal remains open even after redirecting. Modal should be closed when I click on "Explore Curated Data" link
> 3. Add a mapping config to new/existing flow. Run the step. Click on "Explore Curated Data" link doesnt do anything. This is when url is /tiles-run
> 4. Tile title should be Explore. Still shows up Run tile title and icon
> 
> image

@bsrikan fixed the first two issues in this PR, I created a bug for other two issues since this would take some time to fix and don't want to block anyone (https://project.marklogic.com/jira/browse/DHFPROD-5258), Also the other reason is this issue might be dependent on https://project.marklogic.com/jira/browse/DHFPROD-5225.

bsrikan:
You are right about DHFPROD-5258 being a side effect of DHFPROD-5225 and agree with you, that it can be dealt with outside of DHFPROD-4421.

Will the view id for Run tile be different if the url is /tiles-run? I was expecting it to redirect when I run a step after adding it to a new flow. I can see the link for issue 3 above. It doesnt do anything. Close button is responsive in that modal. Do you think this is related to DHFPROD-5225 as well?

Sanjeevani19:
> You are right about DHFPROD-5258 being a side effect of DHFPROD-5225 and agree with you, that it can be dealt with outside of DHFPROD-4421.
> 
> Will the view id for Run tile be different if the url is /tiles-run? I was expecting it to redirect when I run a step after adding it to a new flow. I can see the link for issue 3 above. It doesnt do anything. Close button is responsive in that modal. Do you think this is related to DHFPROD-5225 as well?

Sanjeevani19:
> You are right about DHFPROD-5258 being a side effect of DHFPROD-5225 and agree with you, that it can be dealt with outside of DHFPROD-4421.
> 
> Will the view id for Run tile be different if the url is /tiles-run? I was expecting it to redirect when I run a step after adding it to a new flow. I can see the link for issue 3 above. It doesnt do anything. Close button is responsive in that modal. Do you think this is related to DHFPROD-5225 as well?

@bsrikan I am thinking the view id will not change based on url, yes I think it is related.

bsrikan:
@Sanjeevani19 Further changes are required. The modal with success message should not disappear by itself. It should when Close button is clicked. For mapping step, the modal should close when user either clicks Close button or clicks on "Explore Curated Data"

PR #4126, opened 2020-06-22
DHFPROD-5178: Upgrading to Java Client 5.2.0

rjrudin:
Upgraded Gradle wrappers to 6.2.2, which also depends on kotlin-stdlib 1.3.61. 

Upgrading marklogic-data-hub and web to use the com.github.node-gradle.node plugin instead of the old com.moowork.node plugin. This is because if you clear your Gradle cache and try to build either project, it won't work - you'll get an error of "Could not find org.nodejs:node:8.10.0". That's presumably due to moowork being out-of-date. The error does not happen with the more recent plugin. 

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


SameeraPriyathamTadikonda:
Build failed as node is trying to get downloaded from a HTTPS URL. Please include 
distBaseUrl = 'http://nodejs.org/dist' to use HTTP URL instead under node to fix the issue with the build

SameeraPriyathamTadikonda:
Should we also use the latest node version?

rjrudin:
Adding distBaseUrl doesn't fix the problem. The problem seems to be that marklogic-data-hub and web are both using the old moowork plugin instead of the more modern node plugin that marklogic-data-hub-central is using. Changing those two projects to use the more recent plugin solves the problem. Changing the version of node does not appear necessary.

srinathgit:
There are a few test failures

rjrudin:
I converted this to draft, this thing has been a real pain so far. Lot of weird classpath things going on that I can't explain and have hacked around.

rjrudin:
All the tests are passing now, after working out some weird classpath issues.

PR #4130, opened 2020-06-23
DHFPROD-4998: Deploy custom steps, view them in HC

srinathgit:
### Description
Deploy custom steps, view them in HC
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/src/assets/mock-data/flows.data.ts
+++ b/marklogic-data-hub-central/ui/src/assets/mock-data/flows.data.ts
@@ -1,3 +1,4 @@
+import commonData from "../../assets/mock-data/common.data";
 const response = {"data":{"jobId": "350da405-c1e9-4fa7-8269-d9aefe3b4b9a"}, "status": 200};
 
 let stepFailedWithError = [
@@ -473,6 +474,16 @@ const loadSettings = {"data" :
     "status" :200
 };
 
+const customSteps = {"data" :{"stepsWithEntity":[{
+        "entityType": "Customer",
+        "entityTypeId": "http://example.org/Customer-0.0.1/Customer",
+        "artifacts": [{...commonData.customData[0]}]
+    }],
+    "stepsWithoutEntity": [{...commonData.customData[1]}]},
+    "status" :200
+}
+
+
 const data = {
     primaryEntityTypes,
     flows: flows,
@@ -487,7 +498,9 @@ const data = {
     flowsXML: flowsXML,
     loadsXML: loadsXML,
     loadSettings: loadSettings,
-    mappingSettings: mappingSettings
+    mappingSettings: mappingSettings,
+    customSteps
+
 };
 
 export default data;
--- a/marklogic-data-hub-central/ui/src/components/entities/entity-tiles.tsx
+++ b/marklogic-data-hub-central/ui/src/components/entities/entity-tiles.tsx
@@ -5,31 +5,49 @@ import { UserContext } from '../../util/user-context';
 import styles from './entity-tiles.module.scss';
 import MappingCard from './mapping/mapping-card';
 import MatchingCard from './matching/matching-card';
+import CustomCard from "./custom/custom-card";
 
 const EntityTiles = (props) => {
     const { resetSessionTime } = useContext(UserContext);
-    const [viewType, setViewType] = useState('map');
     const entityModels = props.entityModels || {};
+    const [viewData, setViewData] = useState<string[]>([]);
     const [mappingArtifacts, setMappingArtifacts] = useState<any[]>([]);
     const [matchingArtifacts, setMatchingArtifacts] = useState<any[]>([]);
+    const [customArtifactsWithEntity, setCustomArtifactsWithEntity] = useState<any[]>([]);
+    const [customArtifactsWithoutEntity, setCustomArtifactsWithoutEntity] = useState<any[]>([]);
     const { canReadMapping, canWriteMapping } = props;
     //For accordian within entity tiles
     const { Panel } = Collapse;
-
+    const [requiresNoEntityTypeTile, setRequiresNoEntityTypeTile]  = useState(false);
     const [isLoading, setIsLoading] = useState(false);
-
-    const mappingCardsView = () => {
-        setViewType('map');
-    }
-
-    const matchingCardsView = () => {
-        setViewType('matching');
-    }
 
     useEffect(() => {
         getMappingArtifacts();
         getMatchingArtifacts();
+        getCustomArtifacts();
     },[isLoading]);
+
+    useEffect(() =>{
+        let view;
+        if(props.canReadMapping){
+            view = 'map-';
+        }
+        else if(props.canReadCustom){
+            view = 'custom-'
+        }
+        let tempView: string[] = [];
+        Object.keys(props.entityModels).sort().forEach(ent => {
+            tempView.push(view + ent);
+        })
+        setViewData([...tempView])
+    }, [props])
+
+    const updateView = (index, artifactType, entityType) => {
+        let tempView : string[] ;
+        tempView = viewData;
+        tempView[index] = artifactType + '-' + entityType;
+        setViewData([...tempView])
+    }
 
     const getMappingArtifacts = async () => {
         try {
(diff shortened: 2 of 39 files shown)


examples/reference-entity-model/src/main/ml-config/security/users/hub-developer.json
```diff
@@ -8,6 +8,7 @@
     "hub-central-entity-model-writer",
     "hub-central-load-writer",
     "hub-central-mapping-writer",
+    "hub-central-custom-reader",
```
bsrikan, line 11: Ryan's PR just got merged and hub-central-operator inherits hub-central-custom-reader and so available to hub-central-developer as well.

marklogic-data-hub-central/ui/src/assets/mock-data/flows.data.ts
```diff
@@ -473,6 +474,18 @@ const loadSettings = {"data" :
...
 
+const customStepsWithEntity = {"data" :[{
+        "entityType": "Customer",
+        "entityTypeId": "http://example.org/Customer-0.0.1/Customer",
+        "artifacts": [{...commonData.customData[0]}]
+    }],
+    "status" :200
+}
+
+const customStepsWithoutEntity = {"data" :[{...commonData.customData[1]}],
```
bsrikan, line 485: Should this be customData[0]?

srinathgit: No, it should be customData[1] as it is not associated with any entity

marklogic-data-hub-central/ui/src/components/entities/entity-tiles.tsx
```diff
@@ -178,10 +196,35 @@ const EntityTiles = (props) => {
...
 
-    const outputCards = (entityType, mappingCardData, matchingCardData) => {
-        let output;
+    const getCustomArtifacts = async () => {
+        try {
+            if(props.canReadCustom){
+                let response = await axios.get('/api/steps/custom');
+                if (response.status === 200) {
+                    let entArt = response.data;
+                    entArt.sort((a, b) => (a.entityType > b.entityType) ? 1 : -1)
```
rjrudin, line 205: This seems like something the backend should do. I feel a single DS endpoint should do all the work here - return the object structure I mentioned above, and do any sorting necessary.

marklogic-data-hub-central/ui/src/config/tooltips.config.js
```diff
@@ -50,6 +50,10 @@ const AdvMapTooltips = {
     'provGranularity': 'The level of detail logged for provenance. Choose *coarse* for the default level or *off* for no provenance logging.',
  }
 
+const AdvCustomTooltips = {
+    'additionalSettings': 'A JSON object containing key-value pairs to pass as parameters to the custom step module.',
```
rjrudin, line 54: These are technically passed as "options" to the custom step module, right? I don't think we should say "parameters" here, as that term is used in the context of a custom hook module, and it may lead to confusion.

marklogic-data-hub/src/main/java/com/marklogic/hub/dataservices/CustomService.java
```diff
@@ -0,0 +1,110 @@
...
+
+import com.marklogic.client.DatabaseClient;
+import com.marklogic.client.io.marker.JSONWriteHandle;
+
+import com.marklogic.client.impl.BaseProxy;
+
+/**
+ * Provides a set of operations on the database server
+ */
+public interface CustomService {
```
rjrudin, line 16: I think this should really be "CustomStepService". While we have "MappingService", "Mapping" is able to be by itself and avoid confusion. But "Custom" by itself should be avoided unless we have no choice, as it could refer to a custom step, or a custom module, or a custom hook.

marklogic-data-hub/src/main/resources/hub-internal-config/security/roles/data-hub-custom-writer.json
```diff
@@ -0,0 +1,4 @@
+{
+  "role-name": "data-hub-custom-writer",
```
rjrudin, line 2: Pinging @ryanjdew  - I think it's fine to do this for now, but before 5.3 goes out, I lean towards renaming all of these to be "data-hub-(stepType)-step-writer" for clarity.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/artifacts/core.sjs
```diff
@@ -33,7 +34,8 @@ const registeredArtifactTypes = {
     flow: Flow,
     stepDefinition: StepDef,
     mapping: Mapping,
-    matching: Matching
+    matching: Matching,
+    custom: Custom
```
rjrudin, line 38: I think we need "customStep" and "CustomStep" here as well. "Custom" is just too vague by itself for the reason mentioned above (custom step, custom module, custom hook).

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/artifacts/core.sjs
```diff
@@ -345,6 +347,23 @@ function removeNullProperties(obj) {
   return obj
 }
 
+// Used in UI (currently only in custom step for showing 'Additional Settings' field)
+function getStepWithAdditionalSettings(artifactType, artifactName) {
+  let dataHubOptions = ["sourceQuery", "sourceQueryIsScript", "constrainSourceQueryToJob", "provenanceGranularityLevel", "acceptsBatch",
```
rjrudin, line 352: It doesn't seem like matchOptions and mergeOptions should really be here, because those are specific to concrete step types. Same with validateEntity (though technically that could be supported for other step types in the future, while it's just mapping today). 

I think this is also worth keeping specific to a custom step for now, and thus shouldn't be in "artifacts/core.sjs" which is starting to have a scope of "anything that a user can persist". I think this library needs to be focused on generic persistence functions as opposed to functions that are specific to a concrete type.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/custom/getCustomStep.sjs
```diff
@@ -0,0 +1,21 @@
...
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+'use strict';
+
+var stepName;
+require('/data-hub/5/artifacts/core.sjs').getStepWithAdditionalSettings("custom", stepName)
```
rjrudin, line 19: Yeah let's just implement this in here for now until it can be reused for something outside the scope of a custom step. I really want to avoid artifacts/core.sjs becoming a kitchen sink with a very broad scope.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/step/saveStep.sjs
```diff
@@ -43,7 +43,16 @@ if (existingStep) {
...
-  const stepDefinitionName = "mapping" === stepDefinitionType.toLowerCase() ? "entity-services-mapping" : "default-ingestion";
+  let stepDefinitionName;
+  if("mapping" === stepDefinitionType){
+    stepDefinitionName = "entity-services-mapping";
+  }
+  else if("ingestion" === stepDefinitionType){
+    stepDefinitionName = "default-ingestion";
+  }
+  else {
+    stepDefinitionName = stepProperties.stepDefinitionName;
```
rjrudin, line 54: I think an exception should be thrown here if stepDefinitionName is null.

srinathgit: In custom.sjs "stepDefinitionName" is a mandatory property and will throw an error if it is absent. Shouldn't that be sufficient ?

rjrudin: Good point, that works.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/artifacts/core.sjs
srinathgit: "custom" here refers to the step definition type. I have retained that. It is custom:CustomStep in the latest commit

marklogic-data-hub/src/main/resources/hub-internal-config/security/roles/data-hub-custom-writer.json
ryanjdew: I can get behind that. We need to should have a separate ticket for that.

PR #4131, opened 2020-06-23
DHFPROD-5081: Hub Central header and toolbar should be sticky

briantang:
### Description

- Added inner container wrapping footer and main content so users can scroll while header/toolbar remain fixed.
- Disabled browser scrollbar that was located inside the toolbar.
- New scrollbar is to the left of toolbar and outside of inner tile.

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/src/App.scss
+++ b/marklogic-data-hub-central/ui/src/App.scss
@@ -8,6 +8,17 @@
 
 body {
   font-family: "Helvetica Neue";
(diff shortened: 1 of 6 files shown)


briantang:
Code passes existing UI tests. No new tests needed for containers and CSS refactoring.

PR #4132, opened 2020-06-23
DHFPROD-5204: BackEnd - Add primary Key value field in the searchAPI results response

saarimrahman:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/entities/entity-search-lib.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/entities/entity-search-lib.sjs
@@ -44,9 +44,10 @@ function addPropertiesToSearchResponse(entityName, searchResponse, propertiesToD
   }
   selectedPropertyMetadata = selectedPropertyMetadata.length > 0 ? selectedPropertyMetadata : propertyMetadata.slice(0, maxDefaultProperties);
 
-  // Add entityProperties to each search result
+  // Add entityProperties and primaryKey to each search result
   searchResponse.results.forEach(result => {
     result.entityProperties = [];
+    result.primaryKey = {};
 
     let instance = null;
     try {
@@ -55,14 +56,25 @@ function addPropertiesToSearchResponse(entityName, searchResponse, propertiesToD
       console.log(`Unable to obtain entity instance from document with URI '${result.uri}'; will not add entity properties to its search result`);
     }
 
+    let entityDef = entityModel.definitions[entityName];
     if (instance != null) {
       const entityInstance = instance[entityName];
       if (!entityInstance) {
         console.log(`Unable to obtain entity instance from document with URI '${result.uri}' and entity name '${entityName}'; will not add entity properties to its search result`);
       } else {
         selectedPropertyMetadata.forEach(parentProperty => {
-          result.entityProperties.push(getPropertyValues(parentProperty, entityInstance));
-        });
+            result.entityProperties.push(getPropertyValues(parentProperty, entityInstance));
+            result.primaryKey = getPrimaryValue(entityInstance, entityDef);
+
+            // no primaryKey in entity instance, so use URI
+            if (result.primaryKey === null) {
+              result.primaryKey = {
+                "propertyPath": "uri",
+                "propertyValue": result.uri
+              }
+            }
+          }
+        );
       }
     }
   });
@@ -267,6 +279,30 @@ function getPropertyValues(currentProperty, entityInstance) {
         (currentProperty.multiple ? [] : "");
   }
   return resultObject;
+}
+
+// returns null to use uri
+function getPrimaryValue(entityInstance, entityDefinition) {
+  let primaryKeyData = {}
+  if (entityDefinition.hasOwnProperty("primaryKey")) {
+    let primaryKey = entityDefinition.primaryKey;
+
+    if (entityInstance.hasOwnProperty(primaryKey)) {
+      let primaryKeyValue = entityInstance[primaryKey]
+
+      if (primaryKeyValue === null || primaryKeyValue === "") {
+        return null;
+      } else {
+        primaryKeyData.propertyPath = primaryKey;
+        primaryKeyData.propertyValue = primaryKeyValue;
+        return primaryKeyData;
+      }
+    } else { // no primaryKey in entityInstance, so use uri
+      return null;
+    }
+  } else { // no primaryKey in entityDef, so use uri
+    return null;
+  }
 }
 
 module.exports = {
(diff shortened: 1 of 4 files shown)


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/entities/entity-search-lib.sjs
```diff
@@ -43,10 +43,13 @@ function addPropertiesToSearchResponse(entityName, searchResponse, propertiesToD
     selectedPropertyMetadata = buildSelectedPropertiesMetadata(allMetadata, selectedPropertyNames);
   }
   selectedPropertyMetadata = selectedPropertyMetadata.length > 0 ? selectedPropertyMetadata : propertyMetadata.slice(0, maxDefaultProperties);
+  let selectedPropertyMetadataFinal = JSON.parse(JSON.stringify(selectedPropertyMetadata));
```
ryanjdew, line 46: I'm unclear why it is stringifying and then parsing immediately after. Is it is to convert from a JSON node to a plain JSON object? If so, I think that is worth adding a comment for. Also,  `selectedPropertyMetadata.toObject()` may be a better way.

saarimrahman: This is to create a deep copy of selectedPropertyMetadata with its own object references, as I later modify the objects inside of selectedPropertyMetadataFinal. I will add a clarifying comment.

xnikhil08:
> All Entities

@rjrudin - Its required for the All Entities results display only. The subtask was created under 5152 by mistake i think. It should be under 5112. Once this is merged then i will create a new PR for both 5112 and 5152 combined as they involve same components in UI.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/entities/entity-search-lib.sjs
```diff
@@ -55,19 +60,56 @@ function addPropertiesToSearchResponse(entityName, searchResponse, propertiesToD
...
       const entityInstance = instance[entityName];
       if (!entityInstance) {
         console.log(`Unable to obtain entity instance from document with URI '${result.uri}' and entity name '${entityName}'; will not add entity properties to its search result`);
       } else {
         selectedPropertyMetadata.forEach(parentProperty => {
-          result.entityProperties.push(getPropertyValues(parentProperty, entityInstance));
-        });
+            result.entityProperties.push(getPropertyValues(parentProperty, entityInstance));
+
+            if (entityDef.hasOwnProperty("primaryKey")) {
```
rjrudin, line 72: I was able to comment out most of the new code here and the tests all still pass. I did this:

```
            if (entityDef.hasOwnProperty("primaryKey")) {
              let primaryKey = entityDef.primaryKey;
              if (entityInstance.hasOwnProperty((primaryKey))) {
                  result.primaryKey.propertyPath = primaryKey;
                  result.primaryKey.propertyValue = entityInstance[primaryKey];
              } else { // no primaryKey in entityInstance, so use uri
                result.primaryKey.propertyPath = "uri";
                result.primaryKey.propertyValue = result.uri;
              }
            }
```

So I'm not clear on why all the other logic is needed. If it's for the All Entities view, I don't think we care about selected properties there, right? The user shouldn't be able to do that for that feature. In that context, we know the identifier should always be first. 

Also, in the above logic, it's not handling the case where there's no primaryKey property defined, right? I think what we really want is a getPrimaryValue(entityInstance, entityDefinition) function that will return an object containing propertyPath/propertyValue, or it will return null. If it returns null, then use URI.

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/entities/search/addPropertiesToSearchResponse.sjs
```diff
@@ -373,8 +373,54 @@ function verifyResultsWithoutSelectedProperties() {
...
+        "uri": "/content/jane.json",
+      },
+      {
+        "index": 2,
+        "uri": "/content/sally.json",
+      }
+    ]
+  };
+  entitySearchLib.addPropertiesToSearchResponse(entityName, response);
+  test.assertEqual(101, response.results[0].primaryKey.propertyValue);
```
rjrudin, line 392: @saarimrahman @rahulvudutala All of the functions in here need to return an array of assertions so that they all get reported back (as opposed to JUnit, where you only have to call test functions). 

So this should be:

```
  return [
    test.assertEqual(101, response.results[0].primaryKey.propertyValue),
    test.assertEqual("customerId", response.results[0].primaryKey.propertyPath),
    test.assertEqual(101, response.results[1].primaryKey.propertyValue),
    test.assertEqual("customerId", response.results[1].primaryKey.propertyPath)
  ]
```

That needs to be done for all of the functions.

xnikhil08:
> > All Entities
> 
> @rjrudin - Its required for the All Entities results display only. The subtask was created under 5152 by mistake i think. It should be under 5112. Once this is merged then i will create a new PR for both 5112 and 5152 combined as they involve same components in UI.

@rjrudin - I just noticed that the primary key is also being used in the Detail view page in the Detail View header where if the primary key is available then we display that or else we display the URI. So, we would need primary key to be present either way under each element of results array whether user has selected 'All Entities' or any specific entity like 'Customer'.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/entities/entity-search-lib.sjs
rjrudin: Is this no longer needed now that selectedPropertyMetadata isn't being modified?

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/entities/entity-search-lib.sjs
```diff
@@ -55,19 +60,30 @@ function addPropertiesToSearchResponse(entityName, searchResponse, propertiesToD
...
+              }
+            }
+          }
+        );
       }
     }
   });
   // Make it easy for the client to know which property names were used, and which ones exist
-  searchResponse.selectedPropertyDefinitions = selectedPropertyMetadata;
+  searchResponse.selectedPropertyDefinitions = selectedPropertyMetadataFinal;
```
rjrudin, line 86: And I think this can just be selectedPropertyMetadata again?

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/entities/entity-search-lib.sjs
```diff
@@ -269,6 +285,23 @@ function getPropertyValues(currentProperty, entityInstance) {
...
 }
 
+// returns null to use uri
+function getPrimaryValue(entityInstance, entityDefinition) {
+  let primaryKeyData = {}
+  if (entityDefinition.hasOwnProperty("primaryKey")) {
+    let primaryKey = entityDefinition.primaryKey;
+    if (entityInstance.hasOwnProperty((primaryKey))) {
+      primaryKeyData.propertyPath = primaryKey;
+      primaryKeyData.propertyValue = entityInstance[primaryKey];
```
rjrudin, line 295: I think there's one more scenario to support here - the primaryKey property exists, but it does not have a value. Would be unusual, but could happen. So if entityInstance[primaryKey] is null or an empty string, return null here as well.

PR #4143, opened 2020-06-25
DHFPROD-5126: Support XML instances in search transform

rahulvudutala:
DHFPROD-5126: Support XML instances in search results transform
DHFPROD-5276: Change manageSavedQuery privilege to savedQueryUser

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/entities/entity-search-lib.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/entities/entity-search-lib.sjs
@@ -16,6 +16,8 @@
 'use strict';
 
 const ds = require("/data-hub/5/data-services/ds-utils.sjs");
+const es = require('/MarkLogic/entity-services/entity-services');
+
 // TODO Will move this to /data-hub/5/entities soon
 const entityLib = require("/data-hub/5/impl/entity-lib.sjs");
 
@@ -51,7 +53,7 @@ function addPropertiesToSearchResponse(entityName, searchResponse, propertiesToD
 
     let instance = null;
     try {
-      instance = cts.doc(result.uri).toObject().envelope.instance;
+      instance = getEntityInstance(result.uri);
     } catch (error) {
       console.log(`Unable to obtain entity instance from document with URI '${result.uri}'; will not add entity properties to its search result`);
     }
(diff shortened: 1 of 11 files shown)


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/entities/entity-search-lib.sjs
```diff
@@ -241,6 +247,10 @@ function buildAndCacheSelectedPropertyMetadata(selectedPropertyName, selectedPro
   return finalMetadataProperty;
 }
 
+function getEntityInstance(docUri) {
+  return es.instanceJsonFromDocument(cts.doc(docUri)).toArray();
```
rjrudin, line 251: How is this returning an array? http://docs.marklogic.com/es.instanceJsonFromDocument says it should return an objectNode. The name "getEntityInstance" also implies that a single thing is being returned, not an array. 

Perhaps you just need to call fn.head on the result of es.instanceJsonFromDocument, as shown in the docs page for that function?

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/entities/entity-search-lib.sjs
```diff
@@ -51,7 +54,11 @@ function addPropertiesToSearchResponse(entityName, searchResponse, propertiesToD
 
     let instance = null;
     try {
-      instance = cts.doc(result.uri).toObject().envelope.instance;
+      const instanceArray = getEntityInstance(result.uri);
+      instance = instanceArray.find(instance => Object.keys(instance)[0] === entityName &&
```
rjrudin, line 58: I'm hoping this code won't be necessary if you call fn.head on the result from es.instanceJsonFromDocument.

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/entities/search/getEntityInstance.sjs
```diff
@@ -0,0 +1,117 @@
...
+    test.assertEqual("Sally Hardin", results["name"]),
+    test.assertEqual(shippingResults, results["shipping"]),
+    test.assertEqual(["Sal", "din", "shh"], results["nicknames"]),
+    test.assertEqual(JSON.stringify(billingResults), JSON.stringify(results["billing"]))
+  ];
+}
+
+function getInstanceWithESInfo() {
+  let docUri = "/content/jane.json";
+  const instanceArray = entitySearchLib.getEntityInstance(docUri);
```
rjrudin, line 79: These looks like good tests, but hoping that getEntityInstance can be adjusted above to either return a single object or nothing, not an array.

PR #4146, opened 2020-06-25
DHFPROD-5268:Remove unused code from load artifact create/edit page

srinathgit:
### Description
Remove a lot of unused code from load artifact create/edit page. Works fine with my local tests. No tests needed.
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


bsrikan:
Was able to create/edit a load step and run using json/text/xml/binary documents into corresponding formats as well cross formats like json-xml, xml-json.
txt-json/xml fails, but it fails without this change as well and maybe expected since the text document is not in a valid xml/json format.

PR #4149, opened 2020-06-25
DHFPROD-5112: Display default properties for "all entities" search results

xnikhil08:
#### Description
This PR contains changes for the tickets mentioned below.
DHFPROD-5112: Display default properties for "all entities" search results
DHFPROD-5152: Detail view column in the tabular view is missing
DHFPROD-5101: Expanded row render in the table view is missing after changes to the new table

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/entities/search/EntitySearchManager.java
+++ b/marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/entities/search/EntitySearchManager.java
@@ -104,7 +104,9 @@ public class EntitySearchManager {
                 // We have some awkwardness here where the input is 'entityName', but as of 5.3.0, the "entityTypeIds"
                 // property is capturing entity names, which are expected to double as collection names as well
                 ServerTransform searchResultsTransform = new ServerTransform("hubEntitySearchTransform");
-                searchResultsTransform.put("entityName", entityTypeCollections[0]);
+                if(entityTypeCollections.length == 1) {
+                    searchResultsTransform.put("entityName", entityTypeCollections);
+                }
                 searchResultsTransform.put("propertiesToDisplay", searchQuery.getPropertiesToDisplay());
                 rcQueryDef.setResponseTransform(searchResultsTransform);
             }
--- a/marklogic-data-hub-central/ui/src/components/results-tabular-view/results-tabular-view.tsx
+++ b/marklogic-data-hub-central/ui/src/components/results-tabular-view/results-tabular-view.tsx
@@ -6,6 +6,11 @@ import styles from './results-tabular-view.module.scss';
 import ColumnSelector from '../../components/column-selector/column-selector';
 import { Tooltip } from 'antd';
 import { SearchContext } from '../../util/search-context';
+import { tableParser } from '../../util/data-conversion';
+import { Link } from 'react-router-dom';
+import { faExternalLinkAlt, faCode } from '@fortawesome/free-solid-svg-icons';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import { dateConverter } from '../../util/date-conversion';
 
 interface Props {
     data: any;
@@ -14,6 +19,44 @@ interface Props {
     columns: any;
     hasStructured: boolean;
 }
+
+const DEFAULT_ALL_ENTITIES_HEADER = [
+    {
+        title: 'Identifier',
+        dataIndex: 'identifier',
+        key: '0-i',
+        visible: true,
+        width: 150
+    },
+    {
+        title: 'Entity',
+        dataIndex: 'entityName',
+        key: '0-1',
+        visible: true,
+        width: 150
+    },
+    {
+        title: 'File Type',
+        key: '0-2',
+        dataIndex: 'fileType',
+        visible: true,
+        width: 150
+    },
+    {
+        title: 'Created',
+        dataIndex: 'createdOn',
+        key: '0-c',
+        visible: true,
+        width: 150
+    },
+    {
+        title: 'Detail View',
+        dataIndex: 'detailView',
+        key: '0-d',
+        visible: true,
+        width: 150
+    }
+];
 
 const ResultsTabularView = (props) => {
     const [popoverVisibility, setPopoverVisibility] = useState<boolean>(false);
@@ -25,6 +68,8 @@ const ResultsTabularView = (props) => {
 
     const authorityService = useContext(AuthoritiesContext);
     const canExportQuery = authorityService.canExportEntityInstances();
+    let counter = 0;
+    let parsedPayload = tableParser(props);
 
     let selectedTableColumns = props.selectedPropertyDefinitions;
 
@@ -86,10 +131,72 @@ const ResultsTabularView = (props) => {
         return columns;
     }
 
-    const tableHeaders = tableHeaderRender(selectedTableColumns);
+
+    const updatedTableHeader = () => {
+        let header = tableHeaderRender(selectedTableColumns);
+        let detailView = {
+            title: 'Detail View',
+            dataIndex: 'detailView',
+            key: '0-d'
+        }
+        header.push(detailView);
+        return header;
+    }
+
+    const tableHeaders = props.selectedEntities?.length === 0 ? DEFAULT_ALL_ENTITIES_HEADER : updatedTableHeader();
+
 
     const tableDataRender = (item) => {
         let dataObj = {};
+        let primaryKeyValue = item.primaryKey?.propertyValue;
+        let isUri = item.primaryKey?.propertyPath === 'uri';
+        let uri = encodeURIComponent(item.uri);
+        let path = { pathname: `/detail/${isUri ? '-' : primaryKeyValue}/${uri}` };
+        let options = {};
+        let detailView =
+            <div className={styles.redirectIcons}>
+                <Link to={{ pathname: `${path.pathname}`, state: { selectedValue: 'instance' } }} id={'instance'}
+                    data-cy='instance'>
+                    <Tooltip title={'Show detail on a separate page'}><FontAwesomeIcon icon={faExternalLinkAlt} size="sm" data-testid={`${primaryKeyValue}-detailOnSeparatePage`} /></Tooltip>
+                </Link>
+                <Link to={{ pathname: `${path.pathname}`, state: { selectedValue: 'source' } }} id={'source'}
+                    data-cy='source'>
+                    <Tooltip title={'Show source on a separate page'}><FontAwesomeIcon icon={faCode} size="sm" data-testid={`${primaryKeyValue}-sourceOnSeparatePage`} /></Tooltip>
+                </Link>
+            </div>
+        if (props.selectedEntities?.length === 0) {
+            let itemIdentifier = item.identifier?.propertyValue;
+            let itemEntityName = item.entityName;
+            let document = item.uri.split('/')[item.uri.split('/').length - 1];
+            let createdOn = item.createdOn;
+            options = {
+                primaryKey: primaryKeyValue,
+                identifier: <Tooltip title={isUri && item.uri}>{isUri ? '.../' + document : itemIdentifier}</Tooltip>,
+                entityName: <span data-testid={`${itemEntityName}-${primaryKeyValue}`}>{itemEntityName}</span>,
+                fileType: <span data-testid={`${item.format}-${primaryKeyValue}`}>{item.format}</span>,
+                createdOn: dateConverter(createdOn),
+                uri: item.uri,
+                primaryKeyPath: path,
+                detailView: detailView
+            }
+        } else {
+            options = {
+                primaryKey: primaryKeyValue,
+                uri: item.uri,
+                primaryKeyPath: path,
+                detailView: detailView
+            }
+        }
+
+        dataObj = { ...dataObj, ...options };
+        if (item?.hasOwnProperty('entityProperties')) {
+            generateTableData(item.entityProperties, dataObj)
+        }
+
+        return dataObj;
+    }
+
+    const generateTableData = (item, dataObj = {}) => {
         for (let subItem of item) {
             if (!Array.isArray(subItem)) {
                 if (!Array.isArray(subItem.propertyValue) || typeof (subItem.propertyValue[0]) === 'string') {
@@ -97,19 +204,20 @@ const ResultsTabularView = (props) => {
                 } else {
                     let dataObjArr: any[] = [];
                     for (let el of subItem.propertyValue) {
-                        dataObjArr.push(tableDataRender(el));
+                        dataObjArr.push(generateTableData(el));
                     }
                     dataObj[subItem.propertyPath] = dataObjArr;
                 }
             } else {
-                return tableDataRender(subItem)
+                return generateTableData(subItem)
             }
         }
+
         return dataObj;
     }
 
     const dataSource = props.data.map((item) => {
-        return tableDataRender(item.entityProperties);
+        return tableDataRender(item);
     });
 
     useEffect(() => {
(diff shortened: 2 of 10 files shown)


marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/entities/search/EntitySearchManager.java
```diff
@@ -104,7 +104,7 @@ public StringHandle search(SearchQuery searchQuery) {
                 // We have some awkwardness here where the input is 'entityName', but as of 5.3.0, the "entityTypeIds"
                 // property is capturing entity names, which are expected to double as collection names as well
                 ServerTransform searchResultsTransform = new ServerTransform("hubEntitySearchTransform");
-                searchResultsTransform.put("entityName", entityTypeCollections[0]);
+                searchResultsTransform.put("entityNames", entityTypeCollections);
```
rjrudin, line 107: I'd rather this still be "entityName", and then if the request has multiple entity names (which corresponds to "all entities", though in the future, this ideally will be captured via some other mechanism than the UI sending all entity names), then don't include this parameter. 

I think that keeps the search transform cleaner. Instead of throwing an error if entityName isn't specified, it'll say - Oh, there's no entityName selected, so the user is searching all entities. Thus, I'll return a generic set of properties instead of an entity-specific set of properties.

marklogic-data-hub-central/ui/src/components/results-tabular-view/results-tabular-view.tsx
```diff
@@ -43,42 +89,159 @@ const ResultsTabularView = (props) => {
...
+
+    const tableHeaders = props.selectedEntities?.length === 0 ? DEFAULT_ALL_ENTITIES_HEADER : updatedTableHeader();
+
 
     const tableDataRender = (item) => {
         let dataObj = {};
+        let primaryKeyValue = item.primaryKey.propertyValue;
+        let isUri = item.primaryKey.propertyPath === 'uri';
+        let uri = encodeURIComponent(item.uri);
+        let path = { pathname: `/detail/${isUri ? '-' : primaryKeyValue}/${uri}` };
```
rjrudin, line 112: I'm not clear on why the primaryKeyValue - which could be any string with any characters in it - would be used for anything except for displaying it. Why does this path differ based on whether the primary key is a URI or not?

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/entities/entity-search-lib.sjs
```diff
@@ -305,6 +302,70 @@ function getPrimaryValue(entityInstance, entityDefinition) {
   }
 }
 
+//Helper funciton to add properties to each result instance under results array in searchReponse
+function buildResult(result, selectedPropertyMetadata, entityInfo, selectedEntityOption) {
```
rjrudin, line 306: This function seems like it has two chunks of code in it - one for entity-specific properties, and one for generic properties. For that reason, I think this should be two separate functions with names that imply what they do - e.g.

- addEntitySpecificProperties
- addGenericEntityProperties

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/entities/entity-search-lib.sjs
```diff
@@ -24,60 +24,57 @@ const entityLib = require("/data-hub/5/impl/entity-lib.sjs");
...
-  if(selectedPropertyNames) {
-    selectedPropertyMetadata = buildSelectedPropertiesMetadata(allMetadata, selectedPropertyNames);
-  }
-  selectedPropertyMetadata = selectedPropertyMetadata.length > 0 ? selectedPropertyMetadata : propertyMetadata.slice(0, maxDefaultProperties);
+  let propertyMetadata = [];
+
+  if (Array.isArray(entityNames) && entityNames.length > 1) {
+    //'All Entities' option is selected
+    searchResponse.results.forEach(result => {
+      let entityInfo = {
```
rjrudin, line 39: This initialization of data in "result" should happen in the function that's called. I think the forEach loop should be very simple and just do this:

```
searchResponse.results.forEach(result => {
  if (entityName != null) {
    addEntitySpecificProperties(...);
  } else {
    addGenericEntityProperties(...);
  }
});
```

That way, only addEntitySpecificProperties cares about selected properties - that concept doesn't exist in addGenericEntityProperties, since we're using a fixed set of columns (I think?).

marklogic-data-hub-central/ui/src/components/results-tabular-view/results-tabular-view.tsx
xnikhil08: @rjrudin - This path property is just created to be sent over to the detail view page via the react Link router where its being used in a few places. Since this story was only about the navigation and not updating the detail view page so i didn't touched it.

rjrudin: This seems like it's asking for trouble though - the primaryKeyValue could have a "/" in it, would the path still work?? 

I feel that using the primary key value for anything other than display purposes is risky. We can't guarantee that it's unique, and it could contain any character in it. Only the URI is guaranteed to be unique.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/entities/entity-search-lib.sjs
xnikhil08: Done

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/entities/search/EntitySearchManager.java
xnikhil08: Done.

marklogic-data-hub-central/ui/src/components/results-tabular-view/results-tabular-view.tsx
xnikhil08: @rjrudin - I don't think we are handling the case currently, if there is a "/" in the primary key. I am going to create a separate bug to handle that case and also to remove primary key logic completely from Detail , Detail header and other components as well (if any). We might need a bit of refactoring over there.

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/entities/search/EntitySearchManager.java
```diff
@@ -100,11 +100,14 @@ public StringHandle search(SearchQuery searchQuery) {
 
             // If an entity has been selected, then apply this transform
             String[] entityTypeCollections = searchQuery.getQuery().getEntityTypeCollections();
+            System.out.println("entityTypeCollections-sysOut "+Arrays.toString(entityTypeCollections) + entityTypeCollections);
```
rjrudin, line 103: This should be removed

marklogic-data-hub-central/ui/src/components/results-tabular-view/results-tabular-view.tsx
rjrudin: A separate ticket works, thanks!

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/entities/entity-search-lib.sjs
```diff
@@ -315,6 +301,96 @@ function getPrimaryValue(entityInstance, entityDefinition) {
...
+    createdOnDate = xdmp.documentGetMetadata(result.uri).datahubCreatedOn;
+  } catch (error) {
+    console.log(`Unable to obtain document with URI '${result.uri}'; will not add createdOn date to its search result`);
+  }
+  let identifierValue = result.primaryKey.propertyPath === "uri" ? result.uri : result.primaryKey.propertyValue;
+  result.identifier = {
+    "propertyPath": "identifier",
+    "propertyValue": identifierValue
+  };
+  result.entityName = {
```
rjrudin, line 384: Is propertyPath/propertyValue included for entityName and createdOn to make it easier to include in the table? I ask because technically, neither is an entity property, so it's misleading to refer to them as having a propertyPath. So I'd expect to just see "entityName":entityName and "createdOn": createdOnDate. But if including propertyPath/propertyValue makes life simpler for the UI table, that's fine.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/entities/entity-search-lib.sjs
```diff
@@ -315,6 +301,96 @@ function getPrimaryValue(entityInstance, entityDefinition) {
...
+
+  if (instance != null) {
+    let entityDef = entityInfo.entityModel.definitions[entityTitle];
+    const entityInstance = instance[entityTitle];
+    if (!entityInstance) {
+      console.log(`Unable to obtain entity instance from document with URI '${result.uri}' and entity name '${entityTitle}'; will not add entity properties to its search result`);
+    } else {
+      selectedPropertyMetadata.forEach(parentProperty => {
+        result.entityProperties.push(getPropertyValues(parentProperty, entityInstance));
+        result.primaryKey = getPrimaryValue(entityInstance, entityDef);
```
rjrudin, line 324: Pinging @rahulvudutala - I don't think this should be in the forEach loop, it should be outside the loop. 

To avoid duplication with the same code in addGenericEntityProperties, you can then make an "addPrimaryKeyToResult(entityInstance, entityDef)" function to reuse between the two functions.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/entities/entity-search-lib.sjs
xnikhil08: @rjrudin - Well i added the propertyPath/propertyValue just to be consistent with other fields like primaryKey and Identifier. But we can keep it simple like "entityName":entityName and "createdOn": createdOnDate as it's not like we want it is any specific format for UI. I can make those changes.

rjrudin: If it doesn't make life significantly easier for the UI, let's just do "entityName":entityName and likewise for "createdOn" because they're definitely not properties, and thus we shouldn't refer to them having a propertyPath.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/entities/entity-search-lib.sjs
rahulvudutala: @rjrudin Agree with Rob. This is when an entityModel is selected. So the primary key and primaryKeyValue should be set for each result but not for each for each selectedMetadataProperty for each result

xnikhil08: Done.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/entities/entity-search-lib.sjs
xnikhil08: Done.

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/entities/search/EntitySearchManager.java
xnikhil08: Removed.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/entities/entity-search-lib.sjs
```diff
@@ -31,55 +31,41 @@ const entityLib = require("/data-hub/5/impl/entity-lib.sjs");
...
-    result.primaryKey = {};
-
-    let instance = null;
-    try {
-      instance = getEntityInstance(result.uri);
-    } catch (error) {
-      console.log(`Unable to obtain entity instance from document with URI '${result.uri}'; will not add entity properties to its search result`);
+  let propertyMetadata = [];
+
+  if (entityName != null) {
```
rahulvudutala, line 38: This can be `if(entityName)`

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/entities/entity-search-lib.sjs
```diff
@@ -315,6 +301,86 @@ function getPrimaryValue(entityInstance, entityDefinition) {
...
+  result.entityName = "";
+  result.createdOn = "";
+
+  try {
+    instance = getEntityInstance(result.uri);
+  } catch (error) {
+    console.log(`Unable to obtain entity instance from document with URI '${result.uri}'; will not add entity properties to its search result`);
+  }
+
+  if (instance != null) {
```
rahulvudutala, line 347: This can also be `if(instance)`

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/entities/entity-search-lib.sjs
```diff
@@ -315,6 +301,86 @@ function getPrimaryValue(entityInstance, entityDefinition) {
...
+    entityModel = entityLib.findModelByEntityName(entityTitle);
+    let entityDef = entityModel.definitions[entityTitle];
+    const entityInstance = instance[entityTitle];
+    if (!entityInstance) {
+      console.log(`Unable to obtain entity instance from document with URI '${result.uri}' and entity name '${entityTitle}'; will not add entity properties to its search result`);
+    } else {
+        addPrimaryKeyToResult(result, entityInstance, entityDef);
+    }
+  }
+  try {
```
rahulvudutala, line 358: If there is no instance, then is no point in returning the identifier, entityName and createdOn

probably we should check this first in the beginning of the function and then if the condition fails, we should process the identifier, createdOn and other details and return the object 

```
if(! instance) {
  return;
}
```

@rjrudin Can you confirm this

xnikhil08:
> I'm getting the error "entityName is required to generate search results containing entity properties" after clicking on the Explore tile icon
> Screen Shot 2020-06-26 at 1 18 36 PM
> .

@timur-isangulov - Did you reset the branch and published the changes to local maven ?

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/entities/entity-search-lib.sjs
xnikhil08: @rahulvudutala @rjrudin - Well we can restrict identifier and entityName but createdOn doesn't seem to depend on whether instance is available in the doc or not. It just captures the datahubCreatedOn date from document metadata.
Any thoughts ?

rjrudin: I think identifier and createdOn are all still relevant even if we can't find the instance in the document. This is assuming that the URI would be the identifier and createdOn comes from metadata.

xnikhil08: @rahulvudutala - Then let us not make any changes in this logic. We can update it later if required.

rahulvudutala:
I have resolved the merge conflicts and tested the PR. I am able to see data on the detail view for XML documents from AllEntities View and selected entityModel View. Attaching the screenshots for reference
Screen Shot 2020-06-26 at 3 55 01 PM
Screen Shot 2020-06-26 at 3 55 05 PM
Screen Shot 2020-06-26 at 3 55 15 PM
Screen Shot 2020-06-26 at 3 55 17 PM

xnikhil08:
@timur-isangulov @rahulvudutala @Sanjeevani19 - We don't want the Column Selector to be visible for All Entities Option.

rahulvudutala:
I rebased it off develop branch in my local and tested. It is working

PR #4162, opened 2020-06-26
DHFPROD-4812: UI: Clear user data from Hub Central

AkshayBajajML:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/src/components/header/system-info.test.tsx
+++ b/marklogic-data-hub-central/ui/src/components/header/system-info.test.tsx
@@ -1,9 +1,20 @@
 import React from 'react';
-import {render, cleanup, } from "@testing-library/react";
+import {render, cleanup, wait, fireEvent, waitForElement,} from "@testing-library/react";
 import SystemInfo from './system-info';
 import {AuthoritiesContext, AuthoritiesService} from "../../util/authorities";
 import {BrowserRouter as Router} from "react-router-dom";
 import data from '../../assets/mock-data/system-info.data';
+import axiosMock from "axios";
+import LoginForm from "../login-form/login-form";
+
+const getSubElements=(content,node, title)=>{
+    const hasText = node => node.textContent === title;
+    const nodeHasText = hasText(node);
+    const childrenDontHaveText = Array.from(node.children).every(
+        child => !hasText(child)
+    );
+    return nodeHasText && childrenDontHaveText;
+}
 
 describe('Update data load settings component', () => {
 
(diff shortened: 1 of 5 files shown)


marklogic-data-hub-central/ui/src/components/header/system-info.test.tsx
```diff
@@ -47,8 +49,36 @@ describe('Update data load settings component', () => {
...
+        expect(getByTestId('clearData')).toBeInTheDocument();
+        expect(getByText('Download a zip file containing flow definitions, step definitions and other user artifacts created or modified by Hub Central.')).toBeInTheDocument();
+        expect(getByText('Delete all user data in STAGING, FINAL, and JOBS databases.')).toBeInTheDocument();
+        expect(getByText('Download')).toBeDisabled();
+        expect(getByText('Clear')).toBeEnabled();
+        let clearBtn = getByText('Clear');
+        await wait (()=> {
+            fireEvent.submit(clearBtn);
+        });
+        expect(getByTestId('alertTrue')).toBeInTheDocument();
```
bsrikan, line 81: Clear data alert has display:'none', so the alert might show up in the DOM passing this assertion even when alert message doesnt appear.
toBeVisible() may be better suited in this situation. To verify using the text which has inline bold tags, refer getSubElements() in Run.test.tsx

marklogic-data-hub-central/ui/src/components/header/system-info.tsx
```diff
@@ -55,48 +77,58 @@ const SystemInfo = (props) => {
...
-                                    <Button type="primary">Clear</Button>
-                                </div>
-                            </Card>
-                        </Col>
+                            <Col>
+                                <Card size="small" className={styles.clearAll}>
+                                    {isLoading === true ? <div className={styles.spinRunning}>
+                                         <Spin size={"large"} />
+                                    </div>: ''}
+                                    <div className={styles.title} data-testid="clearData">Clear All User Data</div>
```
bsrikan, line 116: If getByText('Clear All User Data') works we can get rid of the data-testid here. data-testid should be used as a last resort. RTL encourages good testing practices with the primary guiding principle:
`The more your tests resemble the way your software is used, the more confidence they can give you.`

srinathgit: @AkshayBajajML ,
getByText('Clear All User Data') doesn't work ?

AkshayBajajML: Since Alert "Clear All User Data completed successfully" has initial words same as "Clear All User Data" button, due to this most of the test cases fail so had to use data-testid in order to avoid ambiguity.

srinathgit: But, doesn't the alert appears after the button is clicked ? So, before the button is clicked, getByText('Clear All User Data') refers to the button and there is no ambiguity right ?

AkshayBajajML: As Srikanth previously mentioned that alert is set to none when button is not clicked due to which its still present in DOM even it is not visible hence fails the test case using getByText.

PR #4165, opened 2020-06-29
DHFPROD-5066: Add routes and links for handling tile view

wooldridge:
Make the different tile views "URL accessible," i.e.:

/tiles/load -> Load view
/tiles/model -> Model view
/tiles/curate -> Curate view
/tiles/run -> Run view
/tiles/explore -> Explore view

Notes:

- Refactored routing in Add.tsx to add routes for the different tile views.
- Made the tool buttons links to the new routes to enable "open in new window" clicking etc.
- Refactored user-context to remove props no longer needed (redirect, etc).
- Removed some routes that are no longer relevant.
- Refactored add-step-to-flow feature based on new routes (use /tiles/run/add for that).

Updated tests (with @bsrikan's help), all passing.

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/src/pages/TilesView.test.tsx
+++ b/marklogic-data-hub-central/ui/src/pages/TilesView.test.tsx
@@ -2,7 +2,7 @@ import React from 'react';
 import { Router } from 'react-router'
 import { createMemoryHistory } from 'history'
 const history = createMemoryHistory()
-import { render, fireEvent, waitForElement, cleanup } from '@testing-library/react'
+import { render, fireEvent, waitForElement, cleanup, wait } from '@testing-library/react'
 import '@testing-library/jest-dom/extend-expect'
 import TilesView from './TilesView';
 import {AuthoritiesContext, AuthoritiesService} from '../util/authorities';
@@ -35,10 +35,12 @@ describe('Tiles View component tests for Developer user', () => {
         cleanup();
     })
 
-    test('Verify TilesView renders with the toolbar', async () => {
+    test('Verify TilesView renders with the toolbar', () => {
         const { getByLabelText } = render(<Router history={history}>
-            <AuthoritiesContext.Provider value={mockDevRolesService}><TilesView/></AuthoritiesContext.Provider> />
-          </Router>);
+            <AuthoritiesContext.Provider value={mockDevRolesService}>
+                <TilesView/>
+            </AuthoritiesContext.Provider> />
+        </Router>);
 
         expect(getByLabelText("toolbar")).toBeInTheDocument();
 
@@ -57,7 +59,7 @@ describe('Tiles View component tests for Developer user', () => {
 
     });
 
-    test('Verify tiles can be closed', async () => {
+    test('Verify tiles can be closed', () => {
         const tools = Object.keys(tiles);
         const { getByLabelText } = render(<Router history={history}>
             <AuthoritiesContext.Provider value={mockDevRolesService}>
@@ -80,11 +82,11 @@ describe('Tiles View component tests for Developer user', () => {
     });
 
     test('Verify Curate tile displays from toolbar', async () => {
-        const {getByLabelText, getByText, queryByText, debug} = render(<Router history={history}>
+        const {getByLabelText, getByText, queryByText } = render(<Router history={history}>
             <AuthoritiesContext.Provider value={mockDevRolesService}>
                 <SearchContext.Provider value={setViewCurateFunction}>
-                <TilesView/>
-            </SearchContext.Provider>
+                    <TilesView id='curate'/>
+                </SearchContext.Provider>
             </AuthoritiesContext.Provider>
         </Router>);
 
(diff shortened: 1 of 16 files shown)


wooldridge:
@xnikhil08, because I was updating the routes, I made some adjustments to your "add step to flow" feature, I was hoping you could review. Here are the relevant parts:

https://github.com/marklogic/marklogic-data-hub/pull/4165/files#diff-e602ad8b69fc625bc5de59ef0d38e3d2R118

https://github.com/marklogic/marklogic-data-hub/pull/4165/files#diff-750a2839784e3af75e131250d15214a3R106

I now use a /tiles/run/add route when adding a step and redirecting to the run view.

marklogic-data-hub-central/ui/src/pages/TilesView.tsx
```diff
@@ -83,8 +87,6 @@ const TilesView = (props) => {
             setView(views[id])
     }
 
-    const location: any = useLocation();
-
     useEffect(() => {
```
Sanjeevani19, line 90: @wooldridge, the default view for explorer should be zerostate page, right now if you click on explorer icon from the toolbar it is being redirected to overview page. I did some debugging and adding this code inside the useEffect in tilesview.tsx file is fixing the issue. Please push those changes and I will test them again, also let me know if you run into any issues regarding this.

```
if(props.id == 'explore')
                setView(views[props.id], true)
            else
                setView(views[props.id])
```

wooldridge: @Sanjeevani19, thanks so much for checking this! Thanks, @briantang, too!

marklogic-data-hub-central/ui/src/pages/TilesView.test.tsx
```diff
@@ -35,7 +35,7 @@ describe('Tiles View component tests for Developer user', () => {
         cleanup();
     })
 
-    test('Verify TilesView renders with the toolbar', async () => {
+    xtest('Verify TilesView renders with the toolbar', async () => {
```
bsrikan, line 38: Are these tests supposed to be skipped?

bsrikan:
@wooldridge per latest comment in JIRA by Yash, /detail and /browse should show up within Explore tile. In this PR, they browse shows up separately with no toolbar in the right.

marklogic-data-hub-central/ui/src/pages/TilesView.test.tsx
wooldridge: @bsrikan, those xtests should be removed, thanks. Now those tests are failing so I need to update those tests. :-(

wooldridge:
@bsrikan, thanks. With those additional /detail and /browse requirements, I'm skeptical that this can get finished for this sprint.

bsrikan:
@wooldridge have asked Yash for a comment on the scope of /detail and /browse for this story.

marklogic-data-hub-central/ui/src/pages/TilesView.tsx
```diff
@@ -93,6 +93,10 @@ const TilesView = (props) => {
             setCurrentNode(props.id); // TODO Handle multiple with nested objects
             setOptions(tiles[props.id]);
             setView(views[props.id]);
+            if(props.id == 'explore')
```
bsrikan, line 96: Do we still need setView(views[props.id]); before the if condition?

Sanjeevani19: @bsrikan I don't think so we need the setView before if condition

bsrikan: Also what if we just call setView(views[props.id], true) instead of "if" condition?

bsrikan:
Other than the comments I could verify that tiles open in new tab and with tile name in the url. After running mapping step Explore link takes us to the expected page. 
This PR fixes https://project.marklogic.com/jira/browse/DHFPROD-5225

marklogic-data-hub-central/ui/src/pages/TilesView.tsx
Sanjeevani19: @bsrikan The boolean parameter holds true only for explorer tile, for all other tiles we don't need the second parameter hence we need that condition

bsrikan: True and exactly for the same reason -- "zeroState: true", wont impact other tiles -- I thought we can achieve the behavior we want with 1 line change.

Sanjeevani19: @wooldridge I see that we are no longer using onSelect function in tilesview.tsx ? do we still need that for anything ?

Sanjeevani19: > True and exactly for the same reason -- "zeroState: true", wont impact other tiles -- I thought we can achieve the behavior we want with 1 line change.

@bsrikan right that works.

wooldridge: @Sanjeevani19 I'll remove onSelect, thanks! I'm not sure about how we want to handle the setView at this point.

wooldridge:
@bsrikan, I've updated the branch, changing xtest instances to test in TilesView.test, removing the unneeded onSelect(), and fixing the setView().

bsrikan:
@wooldridge  @brucean52 I think it should be safe to delete the private route for modeling:
```
<PrivateRoute path="/model" exact>
                <Modeling/>
              </PrivateRoute>
```
Thoughts?

bsrikan:
@Sanjeevani19 @briantang pls re-review.

wooldridge:
@bsrikan @Sanjeevani19 I updated and force-pushed, please review again. Tests are passing, thanks!

wooldridge:
OK, I'm seeing the following when hitting the Explore button, @Sanjeevani19 maybe you know what's going on?

POST http://localhost:3000/api/entitySearch 500 (Internal Server Error)

Screen Shot 2020-06-30 at 10 17 09 AM

PR #4172, opened 2020-06-30
DHFPROD-5151:Handle null values in json array

srinathgit:
### Description
Handle null values in json array. e2e tests will be added by @bsrikan , no unit tests
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/src/components/entities/mapping/mapping-card.tsx
+++ b/marklogic-data-hub-central/ui/src/components/entities/mapping/mapping-card.tsx
@@ -365,28 +365,29 @@ const MappingCard: React.FC<Props> = (props) => {
                         parentNamespace = tempNS;
                     }
                 } else if (val.constructor.name === "Array") {
-
                     val.forEach(obj => {
-                        if (obj.constructor.name == "String") {
-                            sourceTableKeyIndex = sourceTableKeyIndex + 1;
-                            let propty = {
-                                rowKey: sourceTableKeyIndex,
-                                key: key,
-                                val: obj
-                            };
-                            nestedDoc.push(propty);
-                        } else {
-                            let tempNS = parentNamespace;
-                            if(obj.constructor.name === "Object" && obj.hasOwnProperty('@xmlns')){
-                                parentNamespace = updateParentNamespace(obj);
-                            }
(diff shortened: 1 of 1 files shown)


srinathgit:
> When I load the following test doc, and then try to use it as a source document, it doesn't appear on the left side of the mapping UI:
> 
> `{ "id": null, "test": [ null, null, null ] }`
> 
> I don't get an error or anything. It just doesn't appear.
> 
> I tried regular docs (with no null values), and my mapping UI shows up OK.

@wooldridge , @bsrikan, @xnikhil08 
What should be displayed in the source column if the value of a property is null or "" ? Should it be "null" or empty without any value ?

xnikhil08:
> > When I load the following test doc, and then try to use it as a source document, it doesn't appear on the left side of the mapping UI:
> > `{ "id": null, "test": [ null, null, null ] }`
> > I don't get an error or anything. It just doesn't appear.
> > I tried regular docs (with no null values), and my mapping UI shows up OK.
> 
> @wooldridge , @bsrikan, @xnikhil08
> What should be displayed in the source column if the value of a property is null or "" ? Should it be "null" or empty without any value ?

@srinathgit - I think it should be empty without any value, as long as the column is not being used as a unique identifier(primary key) for the table or else it might be skipped like its happening for @wooldridge.

wooldridge:
@srinathgit @xnikhil08 @bsrikan the following are distinct, value JSON properties and are stored this way in STAGING:

```
"stringEmpty": "",
"null": null,
"nullAsString": "null"
```

I think they should displayed in the source column as they are stored in the database. So, for this PR/ticket, a `null` property value should show as `null` in the mapping UI. PM should comment on this: @yasht2010 @sbayatpur @jmakeig 

This brings up some other JSON display issues that I will document in a separate Jira ticket.

srinathgit:
@wooldridge , @bsrikan , @xnikhil08 
This PR will not display null values if they are present in array. This topic will. be revisited in story https://project.marklogic.com/jira/browse/DHFPROD-5372 but this will atleast prevent the mapping UI to not break if array has null values

PR #4176, opened 2020-06-30
DHFPROD-5305: (DH 5.2) Provide privileges required for MLCP copy

ryanjdew:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/util/MlcpRunner.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/util/MlcpRunner.java
@@ -50,6 +50,10 @@ public class MlcpRunner extends ProcessRunner {
     private DatabaseClient databaseClient;
     private String database = null;
 
+    public MlcpRunner(String mainClass, HubConfig hubConfig, JsonNode mlcpOptions) {
+        this(null, mainClass, hubConfig, null, null, mlcpOptions, null);
+    }
+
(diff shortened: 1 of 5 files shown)


marklogic-data-hub/src/main/java/com/marklogic/hub/util/MlcpRunner.java
```diff
@@ -73,26 +77,40 @@ public String getJobId() {
...
 
-            // Assume that the HTTP credentials will work for mlcp
-            bean.setUsername(hubConfig.getAppConfig().getAppServicesUsername());
-            bean.setPassword(hubConfig.getAppConfig().getAppServicesPassword());
 
-            File file = new File(mlcpOptions.get("input_file_path").asText());
-            String canonicalPath = file.getCanonicalPath();
-            bean.setInput_file_path(canonicalPath);
-            bean.setTransform_param("\"" + bean.getTransform_param() + ",job-id=" + jobId + "\"");
+            if (!"copy".equals(bean.getCommand().toLowerCase())) {
```
srinathgit, line 98: MlcpRunner was used to run legacy 4.x input flows. Are these changes meant to get the test running ?

rjrudin: I agree - I believe the only thing needed is the privilege addition so that MLCP from the command line will work. MlcpRunner is not in scope.

ryanjdew: The change is for a test to verify MLCP copy works with the operator role. The easiest way to test that was to leverage MlcpRunner.

PR #4177, opened 2020-07-01
DHFPROD-5305: (DH 5.3) Provide privileges required for MLCP copy

ryanjdew:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


ryanjdew:
@rjrudin I'll go ahead and try changing this test to use the MLCP main class on this PR. I don't think it worth changing on the 5.2.x branch, though. I don't see us going back and removing the MlcpRunner class on that branch, so I don't see what the benefit of that would be.

rjrudin:
@ryanjdew Okay, then I think it's fine to keep the updates to MlcpRunner, they seem safe. I was more concerned about making more updates than needed to the 5.2.x branch in particular.

PR #4181, opened 2020-07-06
DHFPROD-5267: Removing unused artifact endpoints

rjrudin:
Had to add one new DS endpoint to accomplish this - StepDefinitionService.deleteStepDefinition, though it's only used by a test. 

Reworked LoadUserArtifactsCommand as well so we have explicit methods for loading flows and step definitions, as we have different logic for each of those.

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/artifacts/core.sjs
```diff
@@ -38,39 +38,7 @@ const registeredArtifactTypes = {
...
-                userCanUpdate: userCanUpdate,
-                userCanRead: userCanRead
-            });
-        }
-    }
-    return typesInfo;
-}
-
-const entityServiceDrivenArtifactTypes = ['mapping', 'matching', 'merging', 'mastering', 'custom'];
+const entityServiceDrivenArtifactTypes = ['mapping', 'custom'];
```
rjrudin, line 41: Since matching/merging/mastering don't exist yet, I removed them from this constant. They can be added in once those step types are created.

marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/LoadUserArtifactsCommand.java
```diff
@@ -362,23 +290,65 @@ private void loadSteps(DatabaseClient stagingClient) throws IOException {
...
 
+    private void loadFlows(DatabaseClient stagingClient) throws IOException {
+        final Path flowsPath = hubConfig.getHubProject().getFlowsDir();
+        if (flowsPath.toFile().exists()) {
+            ObjectMapper objectMapper = new ObjectMapper();
+            ArtifactService service = ArtifactService.on(stagingClient);
+            for (File file : flowsPath.toFile().listFiles(f -> f.isFile() && f.getName().endsWith(".flow.json"))) {
+                JsonNode flow = objectMapper.readTree(file);
+                if (!flow.has("name")) {
+                    throw new RuntimeException("Unable to load flow from file: " + file + "; no 'name' property found");
```
srinathgit, line 316: /data-hub/5/artifacts/flow.sjs has 'name' as a requiredProperty. Do we need this validation in the client ?

marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/LoadUserArtifactsCommand.java
```diff
@@ -362,23 +290,65 @@ private void loadSteps(DatabaseClient stagingClient) throws IOException {
...
+            ObjectMapper objectMapper = new ObjectMapper();
+            ArtifactService service = ArtifactService.on(stagingClient);
+            for (File typeDir : stepDefsPath.toFile().listFiles(File::isDirectory)) {
+                final String stepDefType = typeDir.getName();
+                for (File defDir : typeDir.listFiles(File::isDirectory)) {
+                    final String stepDefName = defDir.getName();
+                    File stepDefFile = new File(defDir, stepDefName + ".step.json");
+                    if (stepDefFile.exists()) {
+                        JsonNode stepDef = objectMapper.readTree(stepDefFile);
+                        if (!stepDef.has("name")) {
```
srinathgit, line 337: Same as previous comment. /data-hub/5/artifacts/stepDefinition.sjs has 'name' as a requiredProperty. Do we need this additional validation in the client ?

marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/LoadUserArtifactsCommand.java
rjrudin: Yes, because if name isn't found, then we can't call setArtifact, which requires name as the second parameter.

marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/LoadUserArtifactsCommand.java
rjrudin: Same thing here too, we need it for the second parameter.

PR #4184, opened 2020-07-07
DHFPROD-5441: Adding a proxy nexus

SameeraPriyathamTadikonda:
### Description
Added an internal node mirror for downloading the node(http://node-mirror.eng.marklogic.com:8080/).  
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [N/A ] Added Tests
  

- ##### Reviewer:

- [N/A ] Reviewed Tests

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -2,3 +2,4 @@ version=5.3-SNAPSHOT
 publishUrl=file:../marklogic-data-hub/releases
 systemProp.file.encoding=utf-8
 org.gradle.jvmargs='-Dfile.encoding=UTF-8'
+disBaseUrl=http://nodejs.org/dist/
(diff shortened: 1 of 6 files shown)


gradle.properties
```diff
@@ -2,3 +2,4 @@ version=5.3-SNAPSHOT
 publishUrl=file:../marklogic-data-hub/releases
 systemProp.file.encoding=utf-8
 org.gradle.jvmargs='-Dfile.encoding=UTF-8'
+disBaseUrl=http://nodejs.org/dist/
```
rjrudin, line 5: Should be "distBaseUrl" - can you do a separate PR to update it? I also recommend verbose property names in general - so "nodeDistributionBaseUrl" would be better.

PR #4185, opened 2020-07-08
Update examples/dh-5-example for 5.3

wooldridge:
### Description

Code updates for examples/dh-5-example project after running hubUpdate for 5.3. @bsrikan for your review.

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [ ] Added Tests
  

- ##### Reviewer:

- [ ] Reviewed Tests

### DIFF
--- a/examples/dh-5-example/build.gradle
+++ b/examples/dh-5-example/build.gradle
@@ -1,5 +1,6 @@
 buildscript {
     repositories {
+        maven {url 'http://distro.marklogic.com/nexus/repository/maven-snapshots/'}
         mavenLocal()
         maven { url "https://plugins.gradle.org/m2/" }
         jcenter()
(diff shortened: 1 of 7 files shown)


examples/dh-5-example/build.gradle
```diff
@@ -1,5 +1,6 @@
 buildscript {
     repositories {
+        maven {url 'http://distro.marklogic.com/nexus/repository/maven-snapshots/'}
```
rjrudin, line 3: This is only for beta releases and won't work for anyone outside of the VPN - our example projects (per their README) assume that the user has published a local copy of Gradle in case testing=true.

examples/dh-5-example/build.gradle
```diff
@@ -8,53 +9,12 @@ buildscript {
...
     }
 }
-
 plugins {
     id "net.saliman.properties" version "1.5.1"
     id "java"
 }
-apply plugin: "com.marklogic.ml-data-hub"
-
-repositories {
```
rjrudin, line 22: We still need everything that's been deleted here - extractZip is needed to make sample data available, and runFlowWithoutProject is referenced in the README file.

examples/dh-5-example/src/main/entity-config/exp-final-entity-options.xml
```diff
@@ -43,6 +43,16 @@
       <search:element ns="http://marklogic.com/entity-services" name="title"/>
     </search:value>
   </search:constraint>
+  <search:constraint name="Order">
```
rjrudin, line 46: We actually want the files in entity-config to be deleted and then gitignored.

examples/dh-5-example/flows/ingestion_mapping-flow.flow.json
```diff
@@ -1,105 +1,95 @@
...
+{
+  "name" : "ingestion_mapping-flow",
+  "description" : "This is the default flow containing all of the default steps",
+  "batchSize" : 100,
+  "threadCount" : 4,
+  "stopOnError" : false,
+  "options" : {
+    "sourceQuery" : null
+  },
+  "version" : 0,
```
rjrudin, line 10: The flow files should be migrated via hubMigrateProjectFlows so that ingestion/mapping/custom steps are all converted into referenced steps instead of being inline steps.

rjrudin:
Closing this, as it will need to be redone once mastering steps are referenced.

PR #4194, opened 2020-07-09
DHFPROD-5047: Endpoint for list of all steps

ryanjdew:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/controllers/steps/StepController.java
+++ b/marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/controllers/steps/StepController.java
@@ -1,0 +1,39 @@
+package com.marklogic.hub.central.controllers.steps;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.marklogic.hub.central.controllers.BaseController;
+import com.marklogic.hub.dataservices.StepService;
+import io.swagger.annotations.ApiOperation;
+import org.springframework.http.ResponseEntity;
+import org.springframework.security.access.annotation.Secured;
+import org.springframework.stereotype.Controller;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestMethod;
+
+import java.util.List;
+import java.util.stream.Stream;
+
+@Controller
+@RequestMapping("/api/steps")
+public class StepController extends BaseController {
+    @RequestMapping(method = RequestMethod.GET)
+    @ApiOperation(value = "Get all steps grouped by type", response = Steps.class)
+    @Secured("ROLE_readFlow")
+    public ResponseEntity<JsonNode> get() {
+        return ResponseEntity.ok(newStepService().getStepsByType(Stream.of("stepId", "name")));
+    }
+
+    private StepService newStepService() {
+        return StepService.on(getHubClient().getStagingClient());
+    }
+
+    public static class Steps {
+        public List<StepReference> ingestionSteps;
+        public List<StepReference> mappingSteps;
+    }
+
+    public static class StepReference {
+        public String stepId;
+        public String name;
+    }
+}
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/artifacts/core.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/artifacts/core.sjs
@@ -40,7 +40,7 @@ const registeredArtifactTypes = {
 
 const entityServiceDrivenArtifactTypes = ['mapping', 'custom'];
 
-function getArtifacts(artifactType) {
+function getArtifacts(artifactType, groupByEntityType = entityServiceDrivenArtifactTypes.includes(artifactType)) {
     const queries = [];
     const artifactLibrary =  getArtifactTypeLibrary(artifactType);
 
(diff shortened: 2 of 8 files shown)


marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/controllers/steps/StepController.java
```diff
@@ -0,0 +1,46 @@
...
+@Controller
+@RequestMapping("/api/steps")
+public class StepController extends BaseController {
+    @RequestMapping(method = RequestMethod.GET)
+    @ApiOperation(value = "Get all steps grouped by type", response = Steps.class)
+    @Secured("ROLE_readFlow")
+    public ResponseEntity<JsonNode> get() {
+        ObjectNode response = new ObjectMapper().createObjectNode();
+        ArtifactService service = newArtifactService();
+        response.set("ingestionSteps", service.getList("ingestion", Stream.of("stepId", "name"), false));
```
rjrudin, line 27: Pinging @bsrikan and @srinathgit  about this - the issue with our use of "name" on a step document will impact an entity "name" property with a range index on it (flow documents do the same). 

I am thinking we could kill two birds with one stone here - I think it's very common to need the ID and name from a set of documents, and thus it's common to add range indexes to those. "stepId" is a good name, but "name" of course is not for a range index. 

How about if - as part of migration for 5.3 - we use "stepName" instead of "name" in step documents and add range indexes for stepId and stepName to the staging and final databases? I think that's a good use of indexes. We'd then just do a new DS endpoint that retrieves the ID and name properties for all steps (it looks like "false" is always the 3rd parameter, so that's not really needed?).

marklogic-data-hub/src/main/java/com/marklogic/hub/dataservices/ArtifactService.java
```diff
@@ -301,9 +302,11 @@ private ArtifactServiceImpl(DatabaseClient dbClient, JSONWriteHandle servDecl) {
    * Invokes the getList operation on the database server
    *
    * @param artifactType	provides input
+   * @param propertiesToReturn	List of properties to return. Default behavior returns all properties
+   * @param groupByEntityType	If true, artifacts are grouped by Entity Type; default varies by artifact. Artifact types with entity type relationships default to true.
    * @return	as output
    */
-    com.fasterxml.jackson.databind.JsonNode getList(String artifactType);
+    com.fasterxml.jackson.databind.JsonNode getList(String artifactType, Stream<String> propertiesToReturn, Boolean groupByEntityType);
```
srinathgit, line 309: Rather than changing the getList() method here, will it be easier to add a DS endpoint to StepService (something like getAllSteps.sjs), where we can get all ingestion, mapping (custom in the future) steps and return them ?

rjrudin: I agree with that - I think the artifact library is useful for reuse below the service level, but at the service level, life is generally easier with lots of fine-grained endpoints as opposed to fewer coarse-grained ones.

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/controllers/steps/StepController.java
srinathgit: @rjrudin ,
I am ok with changing "name" to "stepName" and having range index on "stepName" and "stepId". In addition to flow migration, we will also have to change it back to "name" in getFullFlow() method  (in core.sjs)

ryanjdew: Is this something to cover in this PR?

rjrudin: I think it would be a separate JIRA ticket so we could fully think through it. It does seem worthwhile to change "name" to "stepName" as part of 5.3.0.

marklogic-data-hub/src/main/java/com/marklogic/hub/dataservices/ArtifactService.java
ryanjdew: Moved to separate DS endpoint.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/artifacts/core.sjs
```diff
@@ -320,6 +320,17 @@ function convertStepReferenceToInlineStep(stepId) {
   return newFlowStep;
 }
 
+function cleanArtifact(artifact, propertiesToReturn) {
```
rjrudin, line 323: I think "clean" typically means - "remove bad/malformed stuff from it". But the intent here is to only return certain top-level properties of each step. I think "removeAllPropertiesExcept" would be a better function name.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/step/getStepsByType.sjs
```diff
@@ -0,0 +1,29 @@
...
+'use strict';
+
+const Artifacts = require('/data-hub/5/artifacts/core.sjs');
+
+var propertiesToReturn;
+
+const response = {};
+const stepTypes = ['ingestion', 'mapping'];
+
+for (const stepType of stepTypes) {
```
rjrudin, line 25: Because this is the only client that needs to only return certain properties for each step, I'd rather that logic be in here as opposed to being in artifacts/core.sjs. Once we have a second use case for this logic, we can consider how to reuse the logic. 

Also, since this endpoint is only used for returning stepId and name, I think that logic should go in here as well. If we later have a use case for getting all properties of steps or a different set, we can consider how to address it. 

In short, I'd postpone adding flexibility until we need it, because that flexibility comes with a cost.

wooldridge:
For the reference project, I'm seeing the following from the new endpoint:

`{"ingestionSteps":[{"stepId":"loadCustomersJSON-ingestion","name":"loadCustomersJSON"},{"stepId":"loadCustomersXML-ingestion","name":"loadCustomersXML"}],"mappingSteps":[{},{}]}`

The two mapping steps in that project are coming through as empty objects.

PR #4197, opened 2020-07-09
DHFPROD-5310:Fix the versions on which datahub 5.3.0 can be installed

srinathgit:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


marklogic-data-hub/src/main/java/com/marklogic/hub/impl/DataHubImpl.java
```diff
@@ -254,26 +254,26 @@ public boolean isServerVersionValid(String versionString) {
                 return false;
             }
             if(serverVersion.isNightly()){
-                //Support all 10.0-nightly on or after 6/20/2019 and 9.0-nightly on or after 7/25/2019
+                //Support all 10.0-nightly on or after 9/12/2019 and 9.0-nightly on or after 11/12/2019
```
rjrudin, line 257: Why are these dates important? This comment is just stating the same thing as the code - the comment should instead address what the code cannot, which is why 9/12 and 11/12 are the dates.

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/DataHubImpl.java
```diff
@@ -254,26 +254,26 @@ public boolean isServerVersionValid(String versionString) {
...
                         return false;
                     }
                 }
             }
-            //5.1.0 supports server versions 9.x >= 9.0-10 and 10.x >= 10.0.2
+            //5.3.0 supports server versions 9.x >= 9.0-11 and 10.x >= 10.0.2
             else {
                 if(serverVersion.getMajor() == 9){
-                    if(serverVersion.getMinor() < 1000) {
+                    if(serverVersion.getMinor() < 1100) {
```
rjrudin, line 276: Same thing - why do we require 9.0-11 and not 9.0-10?

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/DataHubImpl.java
srinathgit: These dates are the nightly servers on the day the server was released (looking into builds dir). I will add a comment to that effect.

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/DataHubImpl.java
srinathgit: @kkanthet mentioned in the ticket that the support for 5.3.0 starts from 9.0-11. I am not sure of the specific reason why 9.0-10 is not supported.

rjrudin: Please try to find the reason for it - this is likely the best place to have some documentation, since it's the code that's enforcing the requirement.

PR #4198, opened 2020-07-09
DHFPROD-5349: Various roles can access tiles they are not supposed to

timur-isangulov:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


brucean52:
We can add integration test to cover these cases.

timur-isangulov:
> We can add integration test to cover these cases.

I'm going to add tests for this pr. Thanks.

PR #4201, opened 2020-07-10
DHFPROD-5461: Fixing generation of protected paths

rjrudin:
Also added more exception handling for query rolesets, as the Manage API endpoint for them changed in ML 10.0-4 such that "PUT" is no longer available.

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/hub-entities.xqy
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/hub-entities.xqy
@@ -234,7 +234,27 @@ declare function hent:dump-search-options($entities as json:array, $for-explorer
 declare function hent:dump-pii($entities as json:array)
 {
   let $uber-model := hent:uber-model(json:array-values($entities) ! xdmp:to-json(.)/object-node())
-  return es:pii-generate($uber-model)
+  let $response := es:pii-generate($uber-model)
+
+  (: DHFPROD-5461 - Fix path expressions to support XML and JSON :)
+  let $json-response := xdmp:from-json($response)
+  let $_ :=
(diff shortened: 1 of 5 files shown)


rjrudin:
Fixed PiiE2E test - turns out es:pii-generate works correctly when a namespacePrefix is defined in the entity model.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/hub-entities.xqy
```diff
@@ -234,7 +234,27 @@ declare function hent:dump-search-options($entities as json:array, $for-explorer
...
+      let $ex := fn:string(map:get($path, "path-expression"))
+      (: The PiiE2E test is failing because when its entities are passed in, the es:pii-generate function generates a
+      path starting with /es: that uses the correct namespace prefixes. This appears to happen when a namespacePrefix
+      is defined in the entity model. So if the path starts with /es:, we can assume that es:pii-generate generated a
+      correct path that does not need modification. :)
+      where fn:not(fn:starts-with($ex, "/es:"))
+      return
+        let $ex := fn:replace($ex, "/", "/*:")
+        let $ex := fn:replace($ex, "\*:/", "/")
+        return map:put($path, "path-expression", $ex)
```
bsrikan, line 255: @rjrudin essentially we are adding a "*:" when there is no namespace, right?

rjrudin: Right, just need to account for there being "/" and "//" in the path returned by es:pii-generate.

PR #4210, opened 2020-07-13
DHFPROD-5360: Explorer breaks if entity instance's property of type a…

timur-isangulov:
…rray has one of the value as null

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/src/components/results-tabular-view/results-tabular-view.tsx
+++ b/marklogic-data-hub-central/ui/src/components/results-tabular-view/results-tabular-view.tsx
@@ -91,17 +91,19 @@ const ResultsTabularView = (props) => {
                         }
                     },
                     render: (value) => {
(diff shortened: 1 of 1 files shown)


bsrikan:
Added Srinath as reviewer. It is a good practice to add bug reporter as reviewer.

timur-isangulov:
> We need to add tests for all bug fixes to catch regressions later. We have some plumbing available in RTL test for this component already. Might just have to add a search snippet that returns null value and test.

This is about handling null values for the table. It looks like nulls are already handled on the ingest/mapping steps e.g. if I have a document with the null it will appear as empty string in the harmonized document. To reproduce it we added nulls to the harmonized document in the database.

PR #4213, opened 2020-07-14
DHFPROD-5478: Supporting facetable attribute on frontend

saarimrahman:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [X] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


examples/reference-entity-model/entities/Customer.entity.json
```diff
@@ -1,98 +1,111 @@
...
+      ],
+      "primaryKey": "customerId",
+      "properties": {
+        "customerId": {
+          "datatype": "integer",
+          "facet": true
+        },
+        "name": {
+          "datatype": "string",
+          "facet": false,
```
rjrudin, line 20: This needs to be "facetable", and if it's false, let's not include it (the omission of it will be the same as false).

examples/reference-entity-model/entities/Customer.entity.json
```diff
@@ -1,98 +1,111 @@
...
+  "definitions": {
+    "Customer": {
+      "required": [
+        "name"
+      ],
+      "primaryKey": "customerId",
+      "properties": {
+        "customerId": {
+          "datatype": "integer",
+          "facet": true
```
rjrudin, line 16: Let's make "status" a facet instead of customerId - a user would likely never have a facet on a unique value like this.

examples/reference-entity-model/entities/Customer.entity.json
```diff
@@ -1,98 +1,111 @@
...
+            "$ref": "#/definitions/Address"
+          }
+        },
+        "billing": {
+          "description": "Example of a single-value structured property",
+          "$ref": "#/definitions/Address"
+        },
+        "birthDate": {
+          "datatype": "date",
+          "facet": true
```
rjrudin, line 45: If the property is of type "date" and facet = true, does that translate to a date range selector in the UI, as opposed to a set of values?

examples/reference-entity-model/entities/Customer.entity.json
```diff
@@ -1,98 +1,111 @@
...
+          "datatype": "array",
+          "facet": false,
+          "items": {
+            "datatype": "string",
+            "collation": "http://marklogic.com/collation/codepoint"
+          }
+        },
+        "city": {
+          "datatype": "string",
+          "facet": true,
```
rjrudin, line 82: We're not going to support facets on structured properties for this first go-round.

examples/reference-entity-model/entities/Order.entity.json
```diff
@@ -1,31 +1,34 @@
...
+    "baseUri": "http://marklogic.com/example/"
+  },
+  "definitions": {
+    "Order": {
+      "required": [ ],
+      "primaryKey": "orderId",
+      "properties": {
+        "orderId": {
+          "datatype": "string",
+          "facet": true
```
rjrudin, line 14: Let's hold off on modifying this entity - I think making birthDate and status facetable on Customer is enough for now (this is another property that a user would rarely enable a facet on).

marklogic-data-hub-central/ui/src/components/modeling/property-table/property-table.tsx
```diff
@@ -321,6 +321,7 @@ const PropertyTable: React.FC<Props> = (props) => {
       let externalEntity = modelingOptions.entityTypeNamesArray.find(entity => entity.name === parseType[1])
       return {
         datatype: 'array',
+        facet: facetable,
```
rjrudin, line 324: Let's use the term "facetable" in the UI as well.

examples/reference-entity-model/entities/Customer.entity.json
rahulvudutala: @rjrudin Ideally it should work after generating the path range indexes for facetable properties. But there is a bug currently in UI that the facets for range path indexes are not shown.
The entitySearch endpoint though is returning the response.

marklogic-data-hub-central/ui/src/components/modeling/property-table/property-table.tsx
brucean52: "facet" is referenced a lot in the UI already, but go ahead and change it @saarimrahman

examples/reference-entity-model/entities/Customer.entity.json
rjrudin: Ignore my comment - I forgot this is only adjusting the modeling UI. How this translates to search options and the search UI is out of scope.

marklogic-data-hub-central/ui/src/components/modeling/property-table/property-table.tsx
rjrudin: I think it's important to distinguish between the two concepts, which are - 1) the thing that appears in search results with values and counts, and 2) whether or not a property should have a thing appear in search results for it. Thus, two separate terms are warranted.

marklogic-data-hub-central/ui/src/components/modeling/property-table/property-table.test.tsx
```diff
@@ -240,6 +240,10 @@ describe('Entity Modeling Property Table Component', () => {
     fireEvent.change(wildcardCheckbox, { target: { checked: true } });
     expect(wildcardCheckbox).toBeChecked();
 
+    const facetableCheckbox = screen.getByLabelText('Facetable');
```
rjrudin, line 243: I think the code should definitely use the term "facetable" to avoid confusion over what "facet" means based on the context it's used in. But verify with @yasht2010 that it's okay for the UI to say "Facetable" - that's a change in the design. I don't have any issue with the UI saying e.g. "Facet" and "Sort" - in that context, those can be seen as verbs. But in the entity model itself, I think we want adjectives.

marklogic-data-hub-central/ui/e2e/cypress/integration/modeling/scenarios.modeling.tsx
```diff
@@ -41,11 +41,13 @@ describe('Entity Modeling', () => {
     propertyModal.getYesRadio('multiple').click();
     propertyModal.getYesRadio('pii').click();
     propertyModal.clickCheckbox('wildcard');
+    propertyModal.clickCheckbox('facetable');
     propertyModal.getSubmitButton().click();
 
     propertyTable.getMultipleIcon('newID').should('exist');
     propertyTable.getPiiIcon('newID').should('exist');
     propertyTable.getWildcardIcon('newID').should('exist');
+    propertyTable.getWildcardIcon('newID').should('exist');
```
brucean52, line 50: This check should be for getFacetIcon

marklogic-data-hub-central/ui/src/components/modeling/property-table/property-table.test.tsx
yasht2010: I'd prefer to use the word "Facet" in the UI. "Facetable" is not a valid word.

@eltesoro-ml

eltesoro-ml: How about...
    `const isFacetCheckbox = screen.getByLabelText('Use as a facet');`

rjrudin: "facetable" has use in other products - for example, Azure uses it in their docs - https://docs.microsoft.com/en-us/azure/search/search-filters-facets , along with "filterable".

FWIW, I believe "facetable" is commonly used in the field too - I think it's common for a consultant to say e.g. "Let's make that property facetable" (often as a substitute for "searchable"). 

It's not in the dictionary, but - well, that hasn't stopped Microsoft from using it. :)

marklogic-data-hub-central/ui/src/components/modeling/property-table/property-table.tsx
```diff
@@ -55,7 +55,7 @@ const DEFAULT_SELECTED_PROPERTY_OPTIONS: PropertyOptions = {
   multiple: '',
   pii: '',
   sort: false,
-  facet: false,
+  facetable: false,
   wildcard: false
```
rjrudin, line 59: No need to change right now - but @brucean52  is "wildcard" used? I didn't think we were supporting that. Enabling a word lexicon alone won't enabling wildcarded search. I'd rather see this removed (separate JIRA ticket) if we're not doing anything with it.

brucean52: yeah, wildcard is hooked up to wordLexicon right now

rjrudin: Can you work with Allison on changing this? We can't release this, because wordLexicon alone will not enable wildcard search.

marklogic-data-hub-central/ui/src/components/modeling/property-table/property-table.test.tsx
eltesoro-ml: https://64.media.tumblr.com/7bc8be17beeed1dd44aa4e18b5e3e825/tumblr_inline_okb99j5SEa1sstivl_500.png
"FACETABLE ... when face palm just isn't enough."

PR #4216, opened 2020-07-15
DHFPROD-5018: Enable faceting for entity properties marked "facet-able" via modeling

rahulvudutala:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/hub-entities.xqy
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/hub-entities.xqy
@@ -223,6 +223,7 @@ declare %private function hent:fix-options-exp($nodes as node()*)
 
 declare function hent:dump-search-options($entities as json:array, $for-explorer as xs:boolean?)
 {
+  let $entities := hent:add-indexes-for-entity-properties($entities)
   let $uber-model := hent:uber-model(json:array-values($entities) ! xdmp:to-json(.)/object-node())
   return if ($for-explorer = fn:true())
     then
@@ -259,6 +260,7 @@ declare function hent:dump-pii($entities as json:array)
 
 declare function hent:dump-indexes($entities as json:array)
 {
+  let $entities := hent:add-indexes-for-entity-properties($entities)
   let $uber-model := hent:uber-model(json:array-values($entities) ! xdmp:to-json(.)/object-node())
 
   let $database-config := xdmp:from-json(es:database-properties-generate($uber-model))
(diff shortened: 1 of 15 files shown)


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/hub-entities.xqy
```diff
@@ -523,3 +525,42 @@ declare function hent:json-schema-generate($entity-title as xs:string, $uber-mod
   )
   return xdmp:to-json($uber-model)
 };
+
+declare %private function hent:add-range-indexes-to-model($entities as json:array) {
```
rjrudin, line 529: I think this function deserves a comment, because for a new developer, it begs the question - Wait, aren't range indexes already supported in the model?

We want to make it immediately clear that the intent here is to support the custom DHF properties of facetable (and eventually sortable) by adding range indexes based on their existence.

rahulvudutala: @rjrudin I will add the function level comment

rahulvudutala:
Renamed the method name to hent:add-indexes-for-entity-properties in hub-entities.xqy adn added method comments to the same method
Modified a few tests in the latest commit.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/hub-entities.xqy
```diff
@@ -523,3 +525,47 @@ declare function hent:json-schema-generate($entity-title as xs:string, $uber-mod
...
+          let $ref := map:get($entity-type-properties, $entity-type-property)=>map:get("$ref")
+          let $is-facetable :=
+            if (fn:empty($ref)) then
+              let $result :=
+                let $items := map:get($entity-type-properties, $entity-type-property)=>map:get("items")
+                return if(fn:empty($items)) then map:get($entity-type-properties, $entity-type-property)=>map:get("facetable")
+                else fn:not(fn:starts-with(map:get($items, "$ref"), "#")) and map:get($entity-type-properties, $entity-type-property)=>map:get("facetable")
+              return $result
+            else
+              let $items := map:get($entity-type-properties, $entity-type-property)=>map:get("items")
```
rjrudin, line 564: I couldn't figure out what this "else" block was doing. I removed it locally and then ran the updated tests in this PR, and they still passed. If the top-level property has a "$ref", then it's a structured property, in which case we aren't yet supporting facetable on its child properties, right? 

This was the impl of the for loop that I tested locally (I like using "where" to avoid empty "else" statements, also think it makes the code more self-documenting):

```
(: Iterate over each top-level property and if it's facetable, add it to the rangeIndex array:)
    let $_ :=
      for $entity-type-property in map:keys($entity-type-properties)
      let $ref := map:get($entity-type-properties, $entity-type-property)=>map:get("$ref")
      where fn:empty($ref)
      return 
        let $is-facetable :=
          let $items := map:get($entity-type-properties, $entity-type-property)=>map:get("items")
          return 
            if (fn:empty($items)) then 
              map:get($entity-type-properties, $entity-type-property)=>map:get("facetable")
            else 
              fn:not(fn:starts-with(map:get($items, "$ref"), "#")) and 
              map:get($entity-type-properties, $entity-type-property)=>map:get("facetable")
        where $is-facetable
        return json:array-push(map:get($entity-type, "rangeIndex"), $entity-type-property)
```

rahulvudutala: Addressed them in the latest commit

PR #4217, opened 2020-07-15
DHFPROD-4400:Step run notification in the run tile

srinathgit:
### Description
Step run notification in the run tile
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/flow/getFlowsWithStepDetails.sjs
```diff
@@ -71,6 +72,27 @@ flows.map(flow => {
       stepDetails.stepName = step.name;
       stepDetails.stepDefinitionType = step.stepDefinitionType;
       stepDetails.sourceFormat = step.sourceFormat;
+
+      const jobQueries = [];
+      jobQueries.push(cts.collectionQuery('Job'));
+      jobQueries.push(cts.jsonPropertyValueQuery("flow",flow.name));
+      jobQueries.push(cts.jsonPropertyValueQuery("stepName",step.name));
+
+      let latestJob = fn.head(datahub.hubUtils.queryLatest(function() {
```
rjrudin, line 81: Pinging @ryanjdew and @fsnow  about this - having to do a query on every iteration of a loop (and this is a nested loop) is often a recipe for slow performance in production. Now, in theory, the query shouldn't happen that often - it would seem rare for a user to have more than a dozen flows, and for each of those flows to average a dozen or more steps. So if we're running this query 10 to 20 times on each call, it shouldn't add that much overhead. 

But, perhaps the performance is slightly worse because we're hitting a separate database. And we're also getting this job data every single time we get flows for the UI. 

I think before we go forward, we need to review the requirement to verify that it's worth doing this. The "when" in the scenario isn't valid - "When I look at the step in the Run tile" - that's not an event that the computer knows about. The way this implemented, it's assuming a "when" of "When I click on the Run tile" (or do something that causes the Run tile to be displayed). And is it true that we want job data every time this "when" happens? I don't know. 

@srinathgit  Can you organize a quick call with myself and Ryan and Justin to determine what the "when" should really be here? (No need to involve fsnow, just wanted to start getting him exposed to some backend code reviews)

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/data-services/flow/setup.xqy
```diff
@@ -5,4 +5,23 @@ hub-test:reset-hub();
...
 import module namespace hub-test = "http://marklogic.com/data-hub/test" at "/test/data-hub-test-helper.xqy";
 import module namespace test = "http://marklogic.com/test" at "/test/test-helper.xqy";
-hub-test:load-artifacts($test:__CALLER_FILE__)
+hub-test:load-artifacts($test:__CALLER_FILE__);
+
+xquery version "1.0-ml";
+
+import module namespace test = "http://marklogic.com/test" at "/test/test-helper.xqy";
+
+xdmp:invoke-function(function() {
```
rjrudin, line 14: Please build this into the test plumbing itself - the concept of loading static job documents for a test will likely be useful for many other unit tests. The code should go into data-hub-test-helper.xqy - make a "load-jobs" function that loads anything under the "jobs" folder (similar to "content", but written to the jobs data). Then this test doesn't need to do anything other than stick those two job documents under "jobs". 

To account for "batch" documents at some point, it may be worth supporting "test-data/job-data/jobs" and "test-data/job-data/batches".

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/flow/getFlowsWithStepDetails.sjs
ryanjdew: @srinathgit @rjrudin This one is difficult to reduce to one cts search given we need to get the latest job for each step. 

However, reducing this to a single queryLatest call which will reduce the number of invokes to another database should be doable. That would mean iterating over the steps twice. Once to get the job info in the queryLatest call and a second time for the additional step details and incorporating the job info. Despite the additional looping, the performance would be better in most scenarios taking that direction.

rjrudin: We'd still have N queries against the jobs database, right? Where N = the number of steps across all the flows? 

The overhead of the eval should be much less than N queries. We'll use the meeting today to figure out if the requirements justify N queries - that's just been a common performance issue for me in the field - but it may be mitigated enough by the rarity of there being hundreds of steps in a DHF project.

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/data-services/flow/setup.xqy
srinathgit: I have made changes so that job and batch docs will be written to jobs db if they are added to test-data/jobs, test-data/batches respectively

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/flow/getFlowWithLatestJobInfo.sjs
```diff
@@ -0,0 +1,64 @@
...
+ limitations under the License.
+ */
+'use strict';
+const DataHubSingleton = require("/data-hub/5/datahub-singleton.sjs");
+const datahub = DataHubSingleton.instance();
+const Artifacts = require('/data-hub/5/artifacts/core.sjs');
+
+var name;
+
+const flow = Artifacts.getArtifact("flow", name);
```
rjrudin, line 25: I think you can get rid of a chunk of this code by moving the code in getFlowWithStepDetails.sjs into a library, which can go into the same folder as this endpoint - e.g. flow-lib.sjs (I think the impl/flow.sjs that has all the code for running a step should really be called something else besides "flow.sjs" at some point, but that's a separate matter). 

You can have a function of "getFlowsWithStepDetails(flowName)", where flowName is an optional parameter. If it exists, you call getArtifact, else getArtifacts. Then the same code can be used for constructing the flow with all the steps. And then in this endpoint, you take the result and add the job data for the single flow. 

That'll cut down on the duplication, and I think it'll be good to start a pattern of having a library in a DS folder for reusable code. That library could always move to a non-DS-specific location in the future, but I think the DS folder is a fine home for it for now.

PR #4220, opened 2020-07-15
DHFPROD-5123: Update Spring Framework version

ryanjdew:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- N/A Added Tests
  

- ##### Reviewer:

- N/A Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/gradle.properties
+++ b/marklogic-data-hub-central/gradle.properties
@@ -2,8 +2,8 @@ reactUiPath=./ui
 springBootUiPath=src/main/resources/static
 springBootJarName=marklogic-data-hub-central
 
-springBootVersion=2.2.6.RELEASE
-springSecurityVersion=5.2.3.RELEASE
+springBootVersion=2.2.8.RELEASE
+springSecurityVersion=5.2.5.RELEASE
 
 # Passed to bootRun as the value of spring.profiles.active
 # Defaults to "dev" since it's assumed that "./gradlew bootRun" is only used during development. For production, or any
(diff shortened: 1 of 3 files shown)


marklogic-data-hub-central/gradle.properties
```diff
@@ -2,8 +2,8 @@ reactUiPath=./ui
 springBootUiPath=src/main/resources/static
 springBootJarName=marklogic-data-hub-central
 
-springBootVersion=2.2.6.RELEASE
-springSecurityVersion=5.2.3.RELEASE
+springBootVersion=2.2.8.RELEASE
+springSecurityVersion=5.1.11.RELEASE
```
rjrudin, line 6: This seems off - we're moving to a lower version of Spring Security?

ryanjdew: That was the version bump intended for marklogic-data-hub core. I've adjusted.

marklogic-data-hub/build.gradle
```diff
@@ -57,9 +57,9 @@ dependencies {
         exclude group: 'com.marklogic', module: 'marklogic-client-api'
     }
 
-    compile group: 'org.springframework.boot', name: 'spring-boot', version: '2.1.3.RELEASE'
-    compile group: 'org.springframework.integration', name: 'spring-integration-http', version: '5.1.3.RELEASE'
-    compile group: 'org.springframework.boot', name: 'spring-boot-autoconfigure', version: '2.1.3.RELEASE'
+    compile group: 'org.springframework.boot', name: 'spring-boot', version: '2.1.15.RELEASE'
```
rjrudin, line 60: Just an FYI (no action needed) - I soon want to get rid of Spring Boot from the core library. Including it causes issues for applications that want to depend on DHF core as a dependency. Ideally, we only depend on the Spring core stuff in DHF core, and then QS and HC depend on Spring Boot, since they're Spring Boot applications. Maybe something for 5.4.

PR #4224, opened 2020-07-16
DHFPROD-5504: Generating function metadata now regenerates transforms

rjrudin:
Did a little cleanup too:

- LoadTestModules can now construct a HubConfigImpl using host/username/password, which means it can pass a HubConfig to GenerateFunctionMetadataCommand
- So got rid of obsolete constructors in GFMC
- And LoadTestModules doesn't need to call Versions(appConfig), so got rid of that from Versions

This was some overdue cleanup - I had created the above stuff before HubConfig(host,username,password) existed. But now that we have that constructor (which I realized needs to call applyProperties too, though that wasn't previously causing any problems), LoadTestModules doesn't need the stuff that existed in GFMC and Versions only for it.

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/dataservices/MappingService.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/dataservices/MappingService.java
@@ -48,6 +48,7 @@ public interface MappingService {
             private BaseProxy baseProxy;
 
             private BaseProxy.DBFunctionRequest req_testMapping;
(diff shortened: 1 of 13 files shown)


rjrudin:
Fixed the JIRA ticket - is 5504, not 5496.

PR #4227, opened 2020-07-16
DHFPROD-3671: Support more formats for parseDate and parseDateTime

ryanjdew:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/mapping-functions/core.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/mapping-functions/core.sjs
@@ -85,7 +85,8 @@ function parseDate(value, pattern) {
   let standardFormats = ["MM/DD/YYYY", "DD/MM/YYYY", "MM-DD-YYYY", "MM.DD.YYYY", "DD.MM.YYYY", "YYYYMMDD", "YYYY/MM/DD"];
   let nonStandardFormats = ["Mon DD,YYYY", "DD Mon YYYY", "DD-Mon-YYYY" ];
   let response;
-
+  let errorMsg = `The pattern '${pattern}' cannot be applied to the value '${value}'`;
+  const originalPattern = pattern;
   // Normalize both the picture and value a bit per DHFPROD-3121
   if (pattern != null) {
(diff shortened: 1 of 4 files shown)


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/mapping-functions/core.sjs
```diff
@@ -137,7 +138,11 @@ function parseDate(value, pattern) {
     }
   }
   else {
-    fn.error(null, "The given date pattern (" + pattern + ") is not supported.");
+    try {
+      response = xs.date(xdmp.parseYymmdd(originalPattern, value));
+    } catch(e) {
+      fn.error(null, "The given date pattern (" + originalPattern + ") is not supported.");
```
rjrudin, line 144: The story mentions that every error should be of the form - "Also, every Error result should be a message of "The pattern '$pattern' cannot be applied to the value '$value'"." - is that possible to do - which includes not showing the XSLT path where it failed, which a commenter on the story described as "user hostile"?

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/mapping-functions/core.sjs
```diff
@@ -137,24 +138,33 @@ function parseDate(value, pattern) {
     }
   }
   else {
-    fn.error(null, "The given date pattern (" + pattern + ") is not supported.");
+    try {
```
bsrikan, line 141: Should we still keep the earlier error message for the case when pattern is not supported.
For ex, if "Mon YYYY DD" is not supported and the value is "Jan 2019 16", the error message would say
`The pattern 'Mon YYYY DD' cannot be applied to the value 'Jan 2019 16'`

The value seems to match the pattern and we should probably say it is not supported(?)

rjrudin: I think that makes sense for the 3 non-standard formats.

PR #4229, opened 2020-07-17
DHFPROD-5186: Display multi-value property of simple values as a list in source table

briantang:
### Description

- Arrays of multi-values no longer show up in separate rows with duplicate names in 'Name' column.
- Multi values are now displayed as a list in a single row under a single 'Name'. 
- Hovering over values (truncated or not) in Source table now shows full text inside tooltip.
- Added tests for verifying truncated values and tooltip displaying all values on hover.

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/src/components/entities/mapping/mapping-card.tsx
+++ b/marklogic-data-hub-central/ui/src/components/entities/mapping/mapping-card.tsx
@@ -367,32 +367,31 @@ const MappingCard: React.FC<Props> = (props) => {
                         parentNamespace = tempNS;
                     }
                 } else if (val.constructor.name === "Array") {
-                    val.forEach(obj => {
-                        if(obj){
-                            if (obj.constructor.name == "String") {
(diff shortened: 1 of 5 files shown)


marklogic-data-hub-central/ui/src/components/entities/mapping/mapping-card.tsx
```diff
@@ -345,6 +345,7 @@ const MappingCard: React.FC<Props> = (props) => {
 
     // construct infinitely nested source Data
     const generateNestedDataSource = (respData, nestedDoc: Array<any>, parentNamespace = namespaceString) => {
+        console.log("DATA: ", respData)
```
rjrudin, line 348: I think these are typically removed in the UI code too before merging to develop?

briantang: Oops, accidentally left that in; thanks for catching it!

PR #4230, opened 2020-07-17
DHFPROD-5017: Enable sorting for entity properties marked sortable via modeling

rahulvudutala:
### Description

- Enable sorting for entity properties marked sortable via modeling
- removing src/main/entity-config/databases reference from gitignore in reference-entity-model
- Adding sortable properties to the Customer entity model

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/hub-entities.xqy
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/hub-entities.xqy
@@ -145,7 +145,7 @@ declare %private function hent:fix-options($nodes as node()*)
       default return $n
 };
 
-declare %private function hent:fix-options-exp($nodes as node()*)
+declare %private function hent:fix-options-exp($nodes as node()*, $sortable-properties as map:map)
 {
   for $n in $nodes
   return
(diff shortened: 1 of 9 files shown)


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/hub-entities.xqy
```diff
@@ -192,22 +192,42 @@ declare %private function hent:fix-options-exp($nodes as node()*)
...
+            for $attribute in $n/attribute()
+              where $attribute/name() = "name"
+              return $attribute
+          let $search-range-node := $n/search:range
+          return if (fn:empty($search-range-node) or fn:not(map:contains($sortable-properties, xs:string($constraint-name)))) then
+            hent:fix-options-exp($n/node(), $sortable-properties)
+            else
+              element {fn:node-name($search-range-node)} {
+                for $attribute in $search-range-node/attribute()
+                  let $name := $attribute/name()
```
rjrudin, line 210: For this, you can do $search-range-node/*[local-name(.) != "facet"] instead of a for loop

rahulvudutala: `$search-range-node/*[local-name(.) != "facet"]` is returning the child elements. But, `$search-range-node/attribute()[not(name()='facet')]` is returning the attributes of the element search-range-node. Replaced for loop with this xpath

PR #4238, opened 2020-07-20
DHFPROD-5497: Creating granular privileges will now reuse existing ones

rjrudin:
This is close to a re-write of the command. All privileges are now built in memory before any are saved. Each is then saved via the same process, which checks for an existing privilege with the same action.

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/CreateGranularPrivilegesCommand.java
```diff
@@ -95,160 +102,196 @@ public Integer getUndoSortOrder() {
...
-        } else {
-            p = new Privilege(null, name);
-            p.setKind("execute");
-            p.setAction(action);
-        }
-        for (String role : rolesToAdd) {
-            p.addRole(role);
-        }
-        return p;
+    protected void saveGranularPrivileges(ManageClient manageClient, Map<String, Privilege> granularPrivileges) {
```
rjrudin, line 233: This is the method that every in-memory privilege goes through, it's where the new logic is.

PR #4243, opened 2020-07-20
DHFPROD-5498: Remove Wildcard, DHFPROD-5276: PII Tooltip, design syst…

brucean52:
…em upgrade v0.3.1

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


brucean52:
> Any particular reason to comment out and not deleting the wildcard assertions? If there is, would it be worth mentioning via a comment?

Wildcard is being removed for 5.3 due to the server-side not being able to support it. If it's added back later, it'll be easy to add back to the UI

PR #4246, opened 2020-07-20
DHFPROD-5476: Download Configuration Files is not working in Windows machine

rahulvudutala:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/test/java/com/marklogic/hub/hubcentral/AllArtifactsProject.java
+++ b/marklogic-data-hub/src/test/java/com/marklogic/hub/hubcentral/AllArtifactsProject.java
@@ -45,35 +45,40 @@ public class AllArtifactsProject extends TestObject {
         }
     }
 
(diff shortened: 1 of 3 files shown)


marklogic-data-hub/src/test/java/com/marklogic/hub/hubcentral/AllArtifactsProject.java
```diff
@@ -45,35 +45,37 @@ public void writeProjectArtifactsToZipFile() {
         }
     }
 
+    // The path for the artifacts should not start "/" as it will not be picked up by Windows Winzip tool.
```
rjrudin, line 48: Can you add one more assertion to this - iterate over the paths of all zip entries and verify that none of them start with a "/". That way, if/when we add another path in the future, we'll have a safeguard in place to ensure that it doesn't start with "/".

PR #4255, opened 2020-07-21
DHFPROD-5521: Convert HC privileges into DH privileges

rjrudin:
This is being done in preparation for 5259, where each DS endpoint is secured with a privilege. 

getAuthorities now uses HC roles instead of privileges. 

To avoid deployment conflicts, I changed a couple HC privileges so that their names are consistent with the rest of privileges, but I had to change the action too to avoid a deployment conflict:
- hub-central:loginToHubCentral is now hub-central-login
- hub-central:exportEntityInstances is now hub-central-export-entities

I think "export-entities" implies that it's entity instances you're getting and not models. And we'll likely have e.g "export-jobs" in the future, and "export-entities" and "export-jobs" are consistent with each other.

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/data-services/hubCentral/getAuthoritiesTest.sjs
```diff
@@ -1,19 +0,0 @@
...
-
-function invokeService() {
-    return fn.head(xdmp.invoke(
-        "/data-hub/5/data-services/hubCentral/getAuthorities.sjs",
-        {}
-    ));
-}
-
-const response = invokeService();
-const minExpectedAuthorities = ["operator"];
```
rjrudin, line 11: I confirmed that "operator" is no longer used for anything.

PR #4257, opened 2020-07-22
DHFPROD-5211: Migrate entities <5.3 to 5.3 for hub central

rahulvudutala:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


marklogic-data-hub/src/main/java/com/marklogic/hub/EntityManager.java
```diff
@@ -75,6 +81,8 @@
 
     List<HubEntity> getEntities();
 
+    List<JsonNode> getEntitiesJsonNode();
```
rjrudin, line 84: Let's not add a method to this interface. We get almost no value out of the interfaces, except that we imply that the interface will stay constant for someone that wants to use the DHF Java API directly - but we provide no support for that other than publishing the javadocs. Just add the method to EntityManagerImpl and call that directly in the migration class.

marklogic-data-hub/src/main/java/com/marklogic/hub/hubcentral/migration/HubCentralMigrator.java
```diff
@@ -0,0 +1,126 @@
...
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+public class HubCentralMigrator extends LoggingObject {
+    private HubConfig hubConfig;
+    private EntityManager entityManager;
```
rjrudin, line 24: Just make this EntityManagerImpl so you can access the method in it.

marklogic-data-hub/src/main/java/com/marklogic/hub/hubcentral/migration/HubCentralMigrator.java
```diff
@@ -0,0 +1,126 @@
...
+    public void migrateUserArtifacts() {
+        migrateFlows();
+        migrateEntityModels();
+    }
+
+    /**
+     * Migrate the entity model files in a user's local project. Does not make any changes to what's stored in MarkLogic.
+     */
+    protected void migrateFlows() {
+        flowMigrator.migrateFlows();
```
rjrudin, line 46: Just call this directly in migrateUserArtifacts, the extra method adds a level of indirection without any benefit.

marklogic-data-hub/src/main/java/com/marklogic/hub/hubcentral/migration/HubCentralMigrator.java
```diff
@@ -0,0 +1,126 @@
...
+        logger.warn("Beginning migration of entity models in entities directory");
+
+        ObjectWriter writer = mapper.writerWithDefaultPrettyPrinter();
+        boolean atLeastOneEntityModelWasMigrated = false;
+
+        for (JsonNode entityModel : entityManager.getEntitiesJsonNode()) {
+            ObjectNode entityModelNode = (ObjectNode) entityModel;
+            if(entityModelRequiresMigration(entityModelNode)) {
+                atLeastOneEntityModelWasMigrated = true;
+                String title = entityModelNode.get("info").get("title").asText();
```
rjrudin, line 69: We can't assume that anything is valid about the entity models in a user's project. So verify that info exists first, then title exists. If neither exists, then log a warning that it couldn't be found and thus the entity model could not be migrated, and move on to the next entity.

marklogic-data-hub/src/main/java/com/marklogic/hub/hubcentral/migration/HubCentralMigrator.java
```diff
@@ -0,0 +1,126 @@
...
+
+                ObjectNode entityTypeNode = (ObjectNode) entityModelNode.get("definitions").get(title);
+                ObjectNode entityTypePropertiesNode = (ObjectNode) entityTypeNode.get("properties");
```
rjrudin, line 72: Same thing here - check to make sure things exist first, and if not, log a warning and move on.

marklogic-data-hub/src/main/java/com/marklogic/hub/hubcentral/migration/HubCentralMigrator.java
```diff
@@ -0,0 +1,126 @@
...
+
+                List<String> elementRangeIndex = mapper.convertValue(entityTypeNode.get("elementRangeIndex"), ArrayList.class);
+                List<String> rangeIndex = mapper.convertValue(entityTypeNode.get("rangeIndex"), ArrayList.class);
```
rjrudin, line 75: Need to support pathRangeIndex as well.

marklogic-data-hub/src/main/java/com/marklogic/hub/hubcentral/migration/HubCentralMigrator.java
```diff
@@ -0,0 +1,126 @@
...
+
+    protected boolean entityModelRequiresMigration(ObjectNode entityModelNode) {
+        String firstLevelEntityTypeName= entityModelNode.get("info").get("title").asText();
+        entityModelNode= (ObjectNode) entityModelNode.get("definitions").get(firstLevelEntityTypeName);
+        return entityModelNode.get("rangeIndex") != null || entityModelNode.get("elementRangeIndex") != null;
+    }
+
+    private boolean isStructuredTypeProperty(ObjectNode entityPropertyNode) {
+        // check for simple structured type property or simple relationship property
+        if (entityPropertyNode.get("datatype") == null && entityPropertyNode.get("$ref").asText() != null) {
```
rjrudin, line 115: I believe you need a "has" check on "$ref" first - I just did a quick check, and an NPE is thrown on asText() if "$ref" doesn't exist.

marklogic-data-hub/src/main/java/com/marklogic/hub/hubcentral/migration/HubCentralMigrator.java
```diff
@@ -0,0 +1,126 @@
...
+        if (atLeastOneEntityModelWasMigrated) {
+            logger.warn("Finished migrating entity models.");
+            logger.warn("Please examine the migrated entities and check the entity type properties defined as element range indexes and path range indexes are having facetable set to true");
+        } else {
+            logger.warn("No entity models required migration, so no project files were modified");
+        }
+    }
+
+    protected boolean entityModelRequiresMigration(ObjectNode entityModelNode) {
+        String firstLevelEntityTypeName= entityModelNode.get("info").get("title").asText();
```
rjrudin, line 108: Same thing here - be very pessimistic about things not existing. Also check for pathRangeIndex.

marklogic-data-hub/src/main/java/com/marklogic/hub/hubcentral/migration/HubCentralMigrator.java
```diff
@@ -0,0 +1,126 @@
...
+                    logger.warn(format("Entity Model '%s' was successfully migrated", entityModelFile));
+                } catch (IOException e) {
+                    logger.error(format("Entity Model '%s' migration failed; cause: %s", entityModelFile, e.getMessage()), e);
+                }
+            }
+        }
+
+        if (atLeastOneEntityModelWasMigrated) {
+            logger.warn("Finished migrating entity models.");
+            logger.warn("Please examine the migrated entities and check the entity type properties defined as element range indexes and path range indexes are having facetable set to true");
```
rjrudin, line 101: For the warning, let's do:

```
Please examine your entity model files to verify that properties that were listed in the rangeIndex, pathRangeIndex, or elementRangeIndex arrays now have "facetable":true in their property definition.
```

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/EntityManagerImpl.java
```diff
@@ -491,6 +490,10 @@ protected void addSubProperties(HubEntity entity, List<HubEntity> entityDefiniti
         return getEntities(Boolean.FALSE);
     }
 
+    public List<JsonNode> getEntitiesJsonNode() {
```
rjrudin, line 493: You can just call getAllEntities, right? No need for a new method?

marklogic-data-hub/src/test/resources/entity-migration-test/entities/Customer.entity.json
```diff
@@ -0,0 +1,97 @@
+{
+  "info": {
+    "title": "Customer",
+    "version": "0.0.1",
+    "baseUri": "http://example.org/"
+  },
+  "definitions": {
+    "Customer": {
+      "elementRangeIndex": [ "nicknames", "status" ],
+      "rangeIndex": [ "name", "birthDate" ],
```
rjrudin, line 10: Need to add pathRangeIndex. I recommend adding some "bad" entity files to the project too - ones that are missing info.title, or where the value of info/title is not found in "definitions".

ml-data-hub-plugin/src/main/groovy/com/marklogic/gradle/DataHubPlugin.groovy
```diff
@@ -115,14 +115,14 @@ class DataHubPlugin implements Plugin<Project> {
...
 
-        String flowMigrationGroup = "Data Hub Flow Migration"
-        project.task("hubDeleteInstalledLegacyMappings", group: flowMigrationGroup, type: DeleteInstalledLegacyMappingsTask,
+        String hubMigrationGroup = "Data Hub Migration"
+        project.task("hubDeleteInstalledLegacyMappings", group: hubMigrationGroup, type: DeleteInstalledLegacyMappingsTask,
             description: "Delete installed legacy mappings, which are mappings that have not been converted into the format required by Hub Central"
         ).mustRunAfter("hubDeployUserArtifacts")
-        project.task("hubMigrateProjectFlows", group: flowMigrationGroup, type: MigrateProjectFlowsTask,
+        project.task("hubMigrateForHubCentral", group: hubMigrationGroup, type: MigrateProjectForHubCentralTask,
             description: "Migrate flows and mappings in the local project that were created before version 5.3.0 into the new format required for usage within Hub Central"
```
rjrudin, line 123: For the description, change to "Migrate flows, mappings, and entity models in the local project..."

marklogic-data-hub/src/main/java/com/marklogic/hub/hubcentral/migration/HubCentralMigrator.java
```diff
@@ -0,0 +1,166 @@
...
+        if (!entityModelValidForMigration(entityModelNode)) {
+            return false;
+        }
+        String firstLevelEntityTypeName = entityModelNode.get("info").get("title").asText();
+        entityModelNode = (ObjectNode) entityModelNode.get("definitions").get(firstLevelEntityTypeName);
+        return entityModelNode.get("rangeIndex") != null || entityModelNode.get("elementRangeIndex") != null ||
+                entityModelNode.get("pathRangeIndex") != null;
+    }
+
+    protected boolean entityModelValidForMigration(ObjectNode entityModelNode) {
```
rjrudin, line 123: This is good - and logging the cause for each issue is good too.

marklogic-data-hub/src/main/java/com/marklogic/hub/hubcentral/migration/HubCentralMigrator.java
```diff
@@ -0,0 +1,166 @@
...
+        }
+    }
+
+    protected boolean entityModelRequiresMigration(ObjectNode entityModelNode) {
+        if (!entityModelValidForMigration(entityModelNode)) {
+            return false;
+        }
+        String firstLevelEntityTypeName = entityModelNode.get("info").get("title").asText();
+        entityModelNode = (ObjectNode) entityModelNode.get("definitions").get(firstLevelEntityTypeName);
+        return entityModelNode.get("rangeIndex") != null || entityModelNode.get("elementRangeIndex") != null ||
```
rjrudin, line 119: Going to be a little picky here - I think it's worth verifying that for each property, it both exists and the array has size of 1 or more. That's because I think it's common for entity definitions to be stubbed out with empty arrays. No need to process those.

marklogic-data-hub/src/main/java/com/marklogic/hub/hubcentral/migration/HubCentralMigrator.java
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    private boolean isStructuredTypeProperty(ObjectNode entityPropertyNode) {
+        // check for simple structured type property or simple relationship property
+        if (entityPropertyNode.get("datatype") == null && entityPropertyNode.get("$ref") != null) {
+            return true;
+        }
+
+        // check if structured type or relationship type property with array datatype
+        if (entityPropertyNode.get("datatype").asText().equals("array") && entityPropertyNode.get("items").get("$ref") != null) {
```
rjrudin, line 160: We unfortunately need null checks on both "datatype" and "items" here - one may exist, but the other doesn't. We just can't trust that anything about a user's entity definition is valid (I've seen a few support tickets recently where entity models can't be loaded, and the cause is completely invalid ES syntax).

marklogic-data-hub/src/main/java/com/marklogic/hub/hubcentral/migration/HubCentralMigrator.java
```diff
@@ -0,0 +1,166 @@
...
+            logger.warn("No entities directory exists, so no entity models will be migrated");
+            return;
+        }
+
+        logger.warn("Beginning migration of entity models in entities directory");
+
+        ObjectWriter writer = mapper.writerWithDefaultPrettyPrinter();
+        boolean atLeastOneEntityModelWasMigrated = false;
+
+        for (JsonNode entityModel : entityManager.getAllEntities()) {
```
rjrudin, line 59: Another problem I thought of is that a user may very well have an entity model file with invalid JSON in it. Have seen it happen in support tickets. So we need for "getAllEntities" to account for a file being bad. I think modifying that to try/catch an error on each file - instead of throwing it - is a good idea (it'll be beneficial to the current clients of this method too).

marklogic-data-hub/src/main/java/com/marklogic/hub/hubcentral/migration/HubCentralMigrator.java
rahulvudutala: 1. All the entity properties are iterated over to see if an entityPropertyName exist in one of the index arrays. So if something which is in an index array and not defined in entityTypeProperties will not cause an issue and the index arrays are deleted after adding facetable properties.
1. If the array is empty (size is 0), we delete those arrays on migration. (Since we dont want those references existing after migration)

marklogic-data-hub/src/main/java/com/marklogic/hub/hubcentral/migration/HubCentralMigrator.java
```diff
@@ -0,0 +1,167 @@
...
+                    if (mergedIndexArrays.contains(propertyName)) {
+                        ObjectNode entityPropertyNode = (ObjectNode) entityTypePropertiesNode.get(propertyName);
+                        if (!isStructuredTypeProperty(entityPropertyNode)) {
+                            entityPropertyNode.put("facetable", true);
+                        }
+                    }
+                });
+                entityTypeNode.remove(removableIndexArrays);
+
+                File entityModelFile = Paths.get(hubProject.getHubEntitiesDir().toString(), title + entityManager.ENTITY_FILE_EXTENSION).toFile();
```
rjrudin, line 93: This makes me nervous here, because we don't have a guarantee that this will overwrite the file that the entity was read from. The risk is that we write a new file and now the old file and migrated file both exist, and the user is then confused about what happened. 

For this reason, I think it's better to not use getAllEntities() from EntityManagerImpl. Instead, just do listFiles (and filter on name endsWith ".entity.json") on the entities dir and iterate over the list of files. That way, you know you're overwriting the correct File.

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/EntityManagerImpl.java
```diff
@@ -244,22 +244,22 @@ private HubModuleManager getPropsMgr() {
         return new HubModuleManager(timestampFile);
     }
 
-    private List<JsonNode> getAllEntities() {
+    public List<JsonNode> getAllEntities() {
         List<JsonNode> entities = new ArrayList<>(getAllLegacyEntities());
```
rjrudin, line 248: I think this is another reason not to use getAllEntities() - I don't think we want legacy entities included. According to our docs - https://docs.marklogic.com/datahub/upgrade.html - hubUpdate will move legacy entities to the ./entities directory. But in case there's anything left in the legacy entities directory, we don't want to touch those files.

PR #4261, opened 2020-07-22
DHFPROD-5493: entity-config files now have tokens replaced

rjrudin:
This isn't an issue yet, but it could be in the future. This ensures that after we merge an entity-config file with another database file, the output does not have any tokens in it that still need to be replaced.

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


bsrikan:
@rjrudin I still run into this issue:
Caused by: java.lang.RuntimeException: Unexpected error when trying to merge existing array properties for database '%%mlFinalDbName%%'; cause: Unable to build URI for path: /manage/v2/databases/%mlFinalDbName%/properties; cause: Malformed escape pair at index 77: https://2kjo0rqx1.ec2dentifier10.a.marklogicsvc.com:8002/manage/v2/databases/%mlFinalDbName%/properties

On a clean DH 5.3 project, I am able to run the installer no issues. 
When I add an entity and save indexes, entity-config directory is created. If you run the installer now, you are bound to hit above error.

PR #4265, opened 2020-07-23
DHFPROD-5223:Replace data-hub-operator with data-hub-common for certain default permissions

srinathgit:
### Description
Replace data-hub-operator with data-hub-common for certain default permissions

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/com.marklogic.hub/config.xqy
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/com.marklogic.hub/config.xqy
@@ -13,7 +13,7 @@ declare variable $HUB-VERSION := "%%mlHubVersion%%";
 declare variable $HUB-LOG-LEVEL := "%%mlHubLogLevel%%";
 declare variable $FLOW-OPERATOR-ROLE := "%%mlFlowOperatorRole%%";
 declare variable $FLOW-DEVELOPER-ROLE := "%%mlFlowDeveloperRole%%";
-declare variable $DEFAULT-DATA-HUB-PERMISSIONS := "data-hub-common,read,data-hub-common-writer,update";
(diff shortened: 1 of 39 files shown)


marklogic-data-hub/src/main/resources/ml-modules/root/com.marklogic.hub/config.xqy
```diff
@@ -13,7 +13,7 @@ declare variable $HUB-VERSION := "%%mlHubVersion%%";
 declare variable $HUB-LOG-LEVEL := "%%mlHubLogLevel%%";
 declare variable $FLOW-OPERATOR-ROLE := "%%mlFlowOperatorRole%%";
 declare variable $FLOW-DEVELOPER-ROLE := "%%mlFlowDeveloperRole%%";
-declare variable $DEFAULT-DATA-HUB-PERMISSIONS := "data-hub-common,read,data-hub-common-writer,update";
+declare variable $DEFAULT-DATA-HUB-PERMISSIONS := "data-hub-common,read,data-hub-common,update";
```
rjrudin, line 16: Pinging @ryanjdew about this - I think this is reasonable to do, because by requiring data-hub-common-writer, we'd be requiring that a user have all the privileges in data-hub-common-writer, which the customer may not want to enable - i.e. the user may want a custom role that inherits data-hub-common but not data-hub-common-writer, yet still has the ability to modify the documents protected by these permissions.

So for document permissions, I think it's best to stick with data-hub-common for both "read" and "update".

ryanjdew: I agree.

srinathgit:
Fixed failing tests

PR #4271, opened 2020-07-24
DHFPROD-5516: Rename conflicting logback configuration file

akshaysonvane:
### Description
- Renaming to logback-spring.xml now does not create a conflict with applications that use the data-hub core as a dependency. The installer logs still get written to the installer.log file.
- Also renamed the HC log file name from oneui.log to hub-central.log
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [N/A] Added Tests
  

- ##### Reviewer:

- [N/A] Reviewed Tests


akshaysonvane:
> Does this keep logback-spring.xml out of the DHF core jar, or is it there and it just doesn't have any impact?

It still gets packaged with the core jar but won't cause any conflicts now based on the logback config file's priority list.

PR #4280, opened 2020-07-27
DHFPROD-5460:Determine if DHF installer can be run against existing DHS instance

srinathgit:
### Description
Determine if DHF installer can be run against existing DHS instance

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/dhs/installer/command/AbstractInstallerCommand.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/dhs/installer/command/AbstractInstallerCommand.java
@@ -1,10 +1,14 @@
 package com.marklogic.hub.dhs.installer.command;
 
 import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.marklogic.client.ext.helper.LoggingObject;
 import com.marklogic.hub.dhs.installer.InstallerCommand;
 import com.marklogic.hub.dhs.installer.Options;
 import com.marklogic.hub.impl.HubConfigImpl;
+import com.marklogic.hub.impl.Versions;
+import com.marklogic.hub.impl.Versions.MarkLogicVersion;
 import com.marklogic.mgmt.util.ObjectMapperFactory;
 import org.springframework.context.ApplicationContext;
 import org.springframework.web.client.HttpClientErrorException;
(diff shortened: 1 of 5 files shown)


marklogic-data-hub/src/main/java/com/marklogic/hub/dhs/installer/command/AbstractInstallerCommand.java
```diff
@@ -81,4 +84,21 @@ protected String getServerMajorVersion() {
             return "9";
         }
     }
+
+    protected ObjectNode canInstallDhs(String installedHubVersion, MarkLogicVersion mlVersion){
```
rjrudin, line 88: Since this class has a reference to HubConfig, I don't think it needs any inputs here, and it should do the work of using Versions to get the needed data. That will simplify what the subclass has to do.

marklogic-data-hub/src/main/java/com/marklogic/hub/dhs/installer/command/AbstractInstallerCommand.java
```diff
@@ -81,4 +84,21 @@ protected String getServerMajorVersion() {
...
+        ObjectMapper mapper = new ObjectMapper();
+        ObjectNode node = mapper.createObjectNode();
+        if(Character.getNumericValue(installedHubVersion.charAt(0)) < 5){
```
rjrudin, line 91: A scenario I forgot in the story is - DHF isn't installed yet. Thus, we should have a try/catch on getInstalledVersion(), so that if we get an exception, we know DHF isn't installed, and thus we don't need to check anything here.

marklogic-data-hub/src/main/java/com/marklogic/hub/dhs/installer/command/AbstractInstallerCommand.java
```diff
@@ -81,4 +84,21 @@ protected String getServerMajorVersion() {
...
+            node.put("canBeInstalled", false);
+            node.put("message", "DHF cannot be upgraded when the major version of the existing DHF instance is 4");
+        }
+        else if(mlVersion.getMajor().equals(9) || (mlVersion.getMajor().equals(10) && mlVersion.getMinor() < 300)){
```
rjrudin, line 95: I think a better way of implementing this to future-proof it is to say:

- If major version is higher than 10, then can install
- If major version is 10, and minor version is 3 or higher, then can install 
- Else, cannot install

marklogic-data-hub/src/main/java/com/marklogic/hub/dhs/installer/command/InstallIntoDhsCommand.java
```diff
@@ -29,22 +32,34 @@
...
+            // Update the servers in the Curator group
+            groupName = "Curator";
+            modifyHubConfigForDhs(groupName);
+            DhsDeployServersCommand dhsDeployServersCommand = new DhsDeployServersCommand();
+            dhsDeployServersCommand.setServerVersion(serverVersion);
+            deployer.setCommands(Arrays.asList(dhsDeployServersCommand));
+            deployer.deploy(hubConfig.getAppConfig());
+        }
+        else {
+            logger.error("Failed to install DHF version " + hubConfig.getJarVersion());
```
rjrudin, line 59: I think this message should be "Unable to install ", as a failure hasn't occurred yet.

marklogic-data-hub/src/main/java/com/marklogic/hub/dhs/installer/command/CanInstallDhsCommand.java
```diff
@@ -0,0 +1,35 @@
...
+import com.marklogic.hub.impl.Versions.MarkLogicVersion;
+import org.springframework.context.ApplicationContext;
+
+@Parameters(commandDescription = "Verify if current version of DHF can be installed in a DHS environment")
+public class CanInstallDhsCommand extends AbstractInstallerCommand {
+
+    ObjectMapper mapper = new ObjectMapper();
+
+    @Override
+    public void run(ApplicationContext context, Options options) {
```
rjrudin, line 18: As a sanity check, I think all of what this method does should be put into a try/catch. And if any exception occurs, when create an ObjectNode with canBeInstalled=false and message = ex.getMessage(), and write that to stdout (which you can do with just node.toString(), I don't think we need to worry about pretty-printing here - or at least, we should write it in a way that avoids the chance of a JsonProcessingException).

marklogic-data-hub/src/main/java/com/marklogic/hub/dhs/installer/command/AbstractInstallerCommand.java
srinathgit: The only reason I had the 2 parameters for this method is for using it in the CanInstallDhsCommandTest. Without the 2 params, I can't test this method for various inputs.

rjrudin: Ah, that's a good reason. I think the way to go then is keep this method, and add a protected method above it that calls it and that the subclasses use. That eliminates the duplication in the subclasses.

PR #4287, opened 2020-07-27
DHFPROD-5259: Securing DS endpoints with privileges

rjrudin:
Most changes in here are just one-line additions to DS endpoints. 

Did some cleanup - the jobInfo/getLatestJobData endpoint is now part of ModelService, as it's a model-specific function. And the name "JobInfo" was inconsistent with all other service names. 

Added the data-hub-write-custom privilege so that endpoints for modifying custom steps could be secured. Only a data-hub-developer has the privilege, as no HC user can do this yet.

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/artifacts/setArtifact.sjs
```diff
@@ -16,7 +16,18 @@
...
+const ds = require("/data-hub/5/data-services/ds-utils.sjs");
 
 var artifactType, artifactName, artifact;
 
+if ("ingestion" === artifactType) {
+  xdmp.securityAssert("http://marklogic.com/data-hub/privileges/read-ingestion", "execute");
+} else if ("mapping" === artifactType) {
+  xdmp.securityAssert("http://marklogic.com/data-hub/privileges/read-mapping", "execute");
+} else if ("flow" === artifactType || "stepDefinition" === artifactType) {
+  xdmp.securityAssert("http://marklogic.com/data-hub/privileges/read-flow", "execute");
```
srinathgit, line 28: Shouldn't all of this be 'write-*' instead of 'read-*'  as setArtifact() is a write operation ?

rjrudin: Oops, that is correct, will fix.

PR #4289, opened 2020-07-27
DHFPROD-5182: Migrate match steps so they are independent of flows

fsnow:
Adding a library to convert matching steps to the new format, a data service that does that conversion and SJS unit tests

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [ ] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/dataservices/MasteringService.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/dataservices/MasteringService.java
@@ -48,6 +48,7 @@ public interface MasteringService {
             private BaseProxy baseProxy;
 
             private BaseProxy.DBFunctionRequest req_getDefaultCollections;
+            private BaseProxy.DBFunctionRequest req_updateMatchOptions;
 
             private MasteringServiceImpl(DatabaseClient dbClient, JSONWriteHandle servDecl) {
                 this.dbClient  = dbClient;
@@ -55,6 +56,8 @@ public interface MasteringService {
 
                 this.req_getDefaultCollections = this.baseProxy.request(
                     "getDefaultCollections.sjs", BaseProxy.ParameterValuesKind.SINGLE_ATOMIC);
+                this.req_updateMatchOptions = this.baseProxy.request(
+                    "updateMatchOptions.sjs", BaseProxy.ParameterValuesKind.SINGLE_NODE);
             }
 
             @Override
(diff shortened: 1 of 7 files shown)


Jenkinsfile
```diff
@@ -290,7 +290,7 @@ pipeline{
 		stage('code-review'){
 		when {
   			 allOf {
-    changeRequest author: '', authorDisplayName: '', authorEmail: '', branch: '', fork: '', id: '', target: 'develop', title: '', url: ''
+    changeRequest author: '', authorDisplayName: '', authorEmail: '', branch: '', fork: '', id: '', target: 'feature/5.4-develop', title: '', url: ''
```
rjrudin, line 293: If you do a pull on 5.4-develop locally and then rebase off it, these changes should go away.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/mastering/updateOptions.api
```diff
@@ -0,0 +1,14 @@
+{
+    "functionName": "updateOptions",
```
rjrudin, line 2: I am wondering if we need to distinguish between matching and merging in this function. Also, @ryanjdew - while we prefer matching/merging steps to mastering steps, is it reasonable to still refer to "mastering" as a subsystem, such that it makes sense to have a MasteringService as opposed to separate MatchingService and MergingService interfaces? (I kind of prefer just saying "mastering")

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/mastering/updateOptions.sjs
```diff
@@ -0,0 +1,24 @@
...
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+'use strict';
+
+const updateOptionsLib = require("/data-hub/5/data-services/mastering/updateOptionsLib.sjs");
```
rjrudin, line 18: I go back and forth on making a library vs just tossing all the code into the endpoint, and waiting until the need arises to move it into a library for reuse. Thoughts on that? I assume we'll revisit this when we handle merging steps.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/mastering/updateOptions.api
fsnow: Yes, I will rename this to updateMergeOptions.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/mastering/updateOptions.sjs
ryanjdew: Personally, I like the library approach. Both match and merge could use the same library and then if we anything other than the data service needs it, we can always move it into another folder.

PR #4295, opened 2020-07-28
DHFPROD-5242:Display array of simple values in source dropdown

briantang:
### Description

- Added icon in source dropdown to signify values with arrays and objects for JSON and XML mapping.
- Added tooltip for the corresponding icons so users know what it signifies (Temporary until we implement a legend).

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


briantang:
> Nice work! In reviewing this with the reference project, I'm seeing multiple menu selections on mouseover (e.g., Street, Shpping..., properties that appear multiple times in the menu). Is that expected based on the specs from UX? Also, occasionally, I'm seeing the menu "jump around," not sure what's going on but I think you can see it in the screencast:
> 
> https://drive.google.com/file/d/1E_lXH5B4XHUHZRKrOYRGdINMMwClQYDM/view

@wooldridge Yes, the duplicate selections are because of duplicate objects being displayed in the source dropdown. This will be resolved in DHFPROD-5315. That should also be the reason for the menu jumping around (because of duplicate names having different indexes in the menu). For this PR, the specs are just to make sure properties with array values (not objects) are only shown once in the source menu and a respective icon is alongside it.

PR #4301, opened 2020-07-28
DHFPROD-5258:After mapping step is run, Tile title should be Explore, shows up Run tile title and icon

Sanjeevani19:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/src/pages/Browse.tsx
+++ b/marklogic-data-hub-central/ui/src/pages/Browse.tsx
@@ -168,7 +168,9 @@ const Browse: React.FC<Props> = ({ location }) => {
             location.state['query'])
         location.state['tableView'] ? toggleTableView(true) : toggleTableView(false);
     }
-
+    if(location.state && location.state['entityName'] && location.state['jobId']){
+        setLatestJobFacet(location.state['jobId'], location.state['entityName']);
+    }
     if (location.state && location.state['entity']) {
       setEntityClearQuery(location.state['entity']);
     }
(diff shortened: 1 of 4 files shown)


marklogic-data-hub-central/ui/src/pages/Browse.tsx
```diff
@@ -168,7 +168,9 @@ const Browse: React.FC<Props> = ({ location }) => {
             location.state['query'])
         location.state['tableView'] ? toggleTableView(true) : toggleTableView(false);
     }
-
+    if(location.state && location.state['entityName'] && location.state['jobId']){
```
brucean52, line 171: Isn't this the same thing as line 178?

Sanjeevani19: @brucean52 if I have both entityName and jobId then only I am setting latestJobFacet, In the other scenario we are only checking for jobId, if that condition can be changed to include entityName too then i can remove this condition.

marklogic-data-hub-central/ui/src/pages/Run.test.tsx
```diff
@@ -631,7 +631,8 @@ describe('Verify map/match/merge/master step failures in a flow', () => {
             fireEvent.click(exploreButton);
         }
         wait(() => {
```
wooldridge, line 633: For my tests that follow in DHFPROD-5047, I need an await before wait, so:

`await wait(() => {`

marklogic-data-hub-central/ui/src/pages/Browse.tsx
brucean52: Yeah, it can be changed in the future.

PR #4307, opened 2020-07-29
DHFPROD-5047: Add Step menu for flows

wooldridge:
Implement "Add Step" dropdown menu in flow container that allows user to add a step to a flow.

Screencast (showing privileged and unprivileged views): 

https://drive.google.com/file/d/1QmtrBnJegwhzoVowwAC-cpfrK-BNGiT9/view

Also:
- Update existing tests to have /api/steps mocked
- Add new roles-based tests for flows component

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/src/api/__mocks__/mocks.data.ts
+++ b/marklogic-data-hub-central/ui/src/api/__mocks__/mocks.data.ts
@@ -64,6 +64,8 @@ const curateAPI = (axiosMock) => {
     switch (url) {
       case '/api/flows':
         return Promise.resolve(curateData.flows);
+      case '/api/steps':
+        return Promise.resolve(curateData.steps);
       case '/api/models/primaryEntityTypes':
         return Promise.resolve(curateData.primaryEntityTypes);
       case '/api/steps/ingestion':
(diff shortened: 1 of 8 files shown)


bsrikan:
Pre-build check is failing for all the runs in Jenkins and so tests are not being run at all. @wooldridge can you pls check.

bsrikan:
I see a blank screen after clicking on Yes to add a step to a flow in FF. Seems to work ok in Chrome. Console errors:

image

Another issue is the modal doesnt open relative to the "Add Step" button. The modal with list of steps hides the button itself. For ex below I clicked "Add Step" button in the third row. This problem exists in Chrome too:
image

wooldridge:
@briantang and @bsrikan, thanks for the feedback. Those issues should be fixed with my latest force-push. 

The following extra check should fix the async errors that you both were seeing:

https://github.com/marklogic/marklogic-data-hub/blob/feature/DHFPROD-5047/marklogic-data-hub-central/ui/src/components/flows/flows.tsx#L434

I've also added CSS for the Add Step dropdown which should result in a better display of the menu:

https://github.com/marklogic/marklogic-data-hub/blob/feature/DHFPROD-5047/marklogic-data-hub-central/ui/src/components/flows/flows.scss#L11

I modeled the style after the property and function menus in mapping.

wooldridge:
@briantang, thanks for the review, the DIV fix did the trick. Please review again.

PR #4309, opened 2020-07-29
DHFPROD-5460:Changing the log message when getting DHF version fails

srinathgit:
### Description
Changing the log message when getting DHF version fails.
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [N/A ] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/dhs/installer/command/AbstractInstallerCommand.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/dhs/installer/command/AbstractInstallerCommand.java
@@ -93,7 +93,8 @@ public abstract class AbstractInstallerCommand extends LoggingObject implements 
             installedHubVersion = versions.getInstalledVersion();
         }
         catch (Exception e){
-            logger.info("Data Hub is not installed yet");
+            logger.error("Unable to determine version of existing Data Hub installation: " + e.getMessage());
+            logger.info("Will only check MarkLogic version to determine if Data Hub can be installed");
         }
 
         MarkLogicVersion mlVersion = versions.getMLVersion(serverVersion);


marklogic-data-hub/src/main/java/com/marklogic/hub/dhs/installer/command/AbstractInstallerCommand.java
```diff
@@ -93,7 +93,8 @@ protected ObjectNode canInstallDhs(){
             installedHubVersion = versions.getInstalledVersion();
         }
         catch (Exception e){
-            logger.info("Data Hub is not installed yet");
+            logger.error("Unable to determine the version of DHF: " + e.getMessage());
+            logger.info("MarkLogic Server version will only be checked for determining compatibility");
```
rjrudin, line 97: I think we need to tweak this message - should say "Will only check MarkLogic version to determine if Data Hub can be installed" (instead of saying what the only thing we're going to do with ML version is).

marklogic-data-hub/src/main/java/com/marklogic/hub/dhs/installer/command/AbstractInstallerCommand.java
```diff
@@ -93,7 +93,8 @@ protected ObjectNode canInstallDhs(){
             installedHubVersion = versions.getInstalledVersion();
         }
         catch (Exception e){
-            logger.info("Data Hub is not installed yet");
+            logger.error("Unable to determine the version of DHF: " + e.getMessage());
```
rjrudin, line 96: Does this only happen when DH is not installed? Or could it be for other reasons? 

I think a better message is "Unable to determine version of existing Data Hub installation: " + e.getMessage(), unless we know that there is no existing DH installation.

PR #4311, opened 2020-07-29
DHFPROD-5492: Manage session timeout across multiple tabs

ryanjdew:
### Description

Using WebSockets to keep track of the session timeout. 

@bsrikan I'm setting this as a draft for now, as there are not any automated tests. I may need some assistance/direction with automated testing.

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/WebSocketConfig.java
+++ b/marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/WebSocketConfig.java
@@ -1,0 +1,40 @@
+/*
+ * Copyright 2012-2020 MarkLogic Corporation
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ */
+package com.marklogic.hub.central;
+
+import org.springframework.context.annotation.Configuration;
+import org.springframework.messaging.simp.config.MessageBrokerRegistry;
+import org.springframework.web.socket.config.annotation.EnableWebSocketMessageBroker;
+import org.springframework.web.socket.config.annotation.StompEndpointRegistry;
+import org.springframework.web.socket.config.annotation.WebSocketMessageBrokerConfigurer;
+
+//
+@Configuration
+@EnableWebSocketMessageBroker
+public class WebSocketConfig implements WebSocketMessageBrokerConfigurer {
+
+    @Override
+    public void configureMessageBroker(MessageBrokerRegistry config) {
+        config.enableSimpleBroker("/topic");
+        config.setApplicationDestinationPrefixes("/app");
+    }
+
+    @Override
+    public void registerStompEndpoints(StompEndpointRegistry registry) {
+        registry.addEndpoint("/websocket").withSockJS();
+    }
+}
--- a/marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/auth/SessionMonitorInterceptor.java
+++ b/marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/auth/SessionMonitorInterceptor.java
@@ -1,0 +1,50 @@
+package com.marklogic.hub.central.auth;
+
+import org.jetbrains.annotations.NotNull;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.messaging.simp.SimpMessagingTemplate;
+import org.springframework.stereotype.Component;
+import org.springframework.web.servlet.HandlerInterceptor;
+
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import javax.servlet.http.HttpSession;
+import java.util.HashMap;
+import java.util.Map;
+
+public class SessionMonitorInterceptor implements HandlerInterceptor {
+    static final Map<String, Object> stompHeaders = new HashMap<>() {{
+        put("content-type","application/json");
+    }};
+
+    private long timeBetweenSessionUpdates = 1000;
+    private SimpMessagingTemplate template;
+
+    public SessionMonitorInterceptor(SimpMessagingTemplate template) {
+        this.template = template;
+    }
+
+    @Override
+    public boolean preHandle(
+            HttpServletRequest request, @NotNull HttpServletResponse response, @NotNull Object handler) throws Exception {
+        HttpSession session = request.getSession();
+        // If last accessed was less than a second ago, don't bother with additional WebSocket traffic
+        if ((System.currentTimeMillis() - session.getLastAccessedTime()) >= timeBetweenSessionUpdates) {
+            int sessionTimeout = session.getMaxInactiveInterval();
+            template.convertAndSend("/topic/sessionStatus/" + session.getAttribute("hubCentralSessionToken"), new SessionStatus(sessionTimeout), stompHeaders);
+        }
+        return true;
+    }
+
+    protected void setTimeBetweenSessionUpdates(long timeInMillis) {
+        this.timeBetweenSessionUpdates = timeInMillis;
+    }
+
+    static class SessionStatus {
+        public int sessionTimeout;
+        public SessionStatus(int sessionTimeout) {
+            this.sessionTimeout = sessionTimeout;
+        }
+    }
+}
--- a/marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/auth/WebSecurityConfiguration.java
+++ b/marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/auth/WebSecurityConfiguration.java
@@ -19,6 +19,7 @@ import com.marklogic.hub.central.HttpSessionHubClientProvider;
 import com.marklogic.hub.central.HubCentral;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.context.annotation.Configuration;
+import org.springframework.core.env.Environment;
 import org.springframework.security.config.annotation.method.configuration.EnableGlobalMethodSecurity;
 import org.springframework.security.config.annotation.web.builders.HttpSecurity;
 import org.springframework.security.config.annotation.web.configuration.EnableWebSecurity;
@@ -41,6 +42,9 @@ public class WebSecurityConfiguration extends WebSecurityConfigurerAdapter {
     @Autowired
     HttpSessionHubClientProvider hubClientProvider;
 
+    @Autowired
+    Environment environment;
+
     @Override
     protected void configure(HttpSecurity http) throws Exception {
         http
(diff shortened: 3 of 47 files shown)


marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/exceptions/CustomExceptionHandler.java
```diff
@@ -41,6 +41,7 @@
     @ExceptionHandler(FailedRequestException.class)
     protected ResponseEntity<JsonNode> handleFailedRequestExceptionRequest(
         FailedRequestException failedRequestException) {
+        failedRequestException.printStackTrace();
```
rjrudin, line 44: These should be removed, as @akshaysonvane has a separate ticket for logging these via logger.error.

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/controllers/EnvironmentController.java
```diff
@@ -76,6 +81,11 @@ public SystemInfo getSystemInfo() {
         info.marklogicVersion = versionInfo.getMarkLogicVersion();
         info.host = hubCentral.getHost();
         info.sessionTimeout = environment.getProperty("server.servlet.session.timeout");
+        Optional.of(session).ifPresent((session) -> {
```
rjrudin, line 84: Autowiring in the HttpSession makes me nervous - I think it's simpler just to declare an HttpServletRequest as a parameter of getSystemInfo() (you may be able to declare HttpSession as a parameter too, I forget if Spring MVC will populate it automatically). You can get the session from the HttpServletRequest.

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/auth/WebSecurityConfiguration.java
```diff
@@ -71,6 +71,7 @@ protected void configure(HttpSecurity http) throws Exception {
      */
     protected String[] getAlwaysPermittedPatterns() {
         return new String[]{
+            "/websocket/**",
```
rjrudin, line 74: I feel it'd be safer for this to be explicitly named for its purpose today, which is for checking session status, right? So this could be "/websocket/sessionStatus", and we'd want to explicitly list that here instead of allowing anything under "/websocket/" (as we could add another path under "/websocket/" that does require authentication and forget to update this).

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/auth/SessionMonitorInterceptor.java
```diff
@@ -0,0 +1,46 @@
...
+
+    public SessionMonitorInterceptor(SimpMessagingTemplate template) {
+        this.template = template;
+    }
+
+    @Override
+    public boolean preHandle(
+            HttpServletRequest request, @NotNull HttpServletResponse response, @NotNull Object handler) throws Exception {
+        HttpSession session = request.getSession();
+        if (!(template == null || session == null)) {
```
rjrudin, line 31: How could the template be null, since it's passed into the constructor? If it could be null, then the constructor should throw an exception because the system is ostensibly misconfigured.

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/auth/SessionMonitorInterceptor.java
```diff
@@ -0,0 +1,46 @@
...
+            int sessionTimeout = session.getMaxInactiveInterval();
+            template.convertAndSend("/topic/sessionStatus/"+session.getAttribute("hubCentralSessionToken"), new SessionStatus(sessionTimeout), stompHeaders);
+        } else {
+            System.out.println("Template: " + template + "\nSession: " + session);
```
rjrudin, line 35: I think this "else" block goes away since we don't need to check for the template being null.

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/WebSocketConfig.java
```diff
@@ -0,0 +1,39 @@
...
+import org.springframework.messaging.simp.config.MessageBrokerRegistry;
+import org.springframework.web.socket.config.annotation.EnableWebSocketMessageBroker;
+import org.springframework.web.socket.config.annotation.StompEndpointRegistry;
+import org.springframework.web.socket.config.annotation.WebSocketMessageBrokerConfigurer;
+
+@Configuration
+@EnableWebSocketMessageBroker
+public class WebSocketConfig implements WebSocketMessageBrokerConfigurer {
+
+    @Override
```
rjrudin, line 29: I think both of these beans need comments to explain what's going on here. Referencing Spring docs on this is fine too. But the typical problem with Spring beans like this is they get a lot done in a small amount of code, but it's not clear why we're doing this.

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/exceptions/CustomExceptionHandler.java
ryanjdew: Yeah. Meant to back those changes out before pushing.

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/auth/SessionMonitorInterceptor.java
```diff
@@ -0,0 +1,46 @@
...
+    public SessionMonitorInterceptor(SimpMessagingTemplate template) {
+        this.template = template;
+    }
+
+    @Override
+    public boolean preHandle(
+            HttpServletRequest request, @NotNull HttpServletResponse response, @NotNull Object handler) throws Exception {
+        HttpSession session = request.getSession();
+        if (!(template == null || session == null)) {
+            int sessionTimeout = session.getMaxInactiveInterval();
```
akshaysonvane, line 32: Why do we need a WebSocket to send the session timeout value continuously?
Wouldn't `getMaxInactiveInterval` always return 300 (5 minutes) in our case?

marklogic-data-hub-central/ui/src/util/user-context.tsx
```diff
@@ -40,8 +41,27 @@ const UserProvider: React.FC<{ children: any }> = ({children}) => {
...
+    stompService.onConnected().then(() => {
+      const hubCentralSessionToken = localStorage.getItem('hubCentralSessionToken');
+      stompService.messages.subscribe((message) => {
+        sessionCount = parseInt(JSON.parse(message.body).sessionTimeout);
+      });
+      stompService.subscribe(`/topic/sessionStatus/${hubCentralSessionToken}`, (msgId: string) => {
+        unsubscribeId = msgId;
+      });
+    });
+  };
```
akshaysonvane, line 58: Do we need some sort of cross tab communication to propagate the timeout events on all tabs  https://dev.to/naismith/cross-tab-communication-with-javascript-1hc9

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/controllers/EnvironmentController.java
rjrudin: Just verified, you can just declare HttpSession as a parameter - https://www.logicbig.com/tutorials/spring-framework/spring-web-mvc/http-session-param.html - I think that's the better way to go, also makes testing easier. Note that the session will never be null too.

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/auth/SessionMonitorInterceptor.java
rjrudin: I'm puzzled by that too - this is a constant value. I thought the intent was for the middle tier to send a message when the session expired, and/or when there are N seconds remaining before session expiration. 

An HttpSessionListener can be used for sending a message when the session expires. 

If there's a requirement for sending a message N seconds before expiration - I'm not sure how to do that. I could imagine some object we store in the HttpSession that waits until N seconds before expiration to send a message. That object would be overridden/reset every time a request is received.

marklogic-data-hub-central/ui/src/util/user-context.tsx
ryanjdew: Each tab will be subscribed to the same sessionToken via WebSockets, so each tab will have sessionCount updated.

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/controllers/EnvironmentController.java
ryanjdew: Moved to a parameter.

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/auth/WebSecurityConfiguration.java
ryanjdew: There are a lot more endpoints called for setting up WebSockets and they include randomly generated paths, so determining all the patterns that should be permitted isn't really practical.

I was able to remove the WebSockets permitted pattern by refactoring the front-end to not attempt to connect to WebSockets when logged out, which was bring the browser its knees with connection attempts.

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/auth/SessionMonitorInterceptor.java
ryanjdew: As @rjrudin points out, the HtttpSessionListener doesn't allow us to catch the situation where we are N seconds before closing the connection. The only opportunity to catch that is with each request. So with each request, I'm using the interceptor to reset the session count down on the front-end side. 

I'll make an update to only send out the update if the session last accessed time was more than a second ago to cut down on some of the chatter.

bsrikan:
@ryanjdew both stomp.ts and user-context.ts are utilities and we dont yet have tests for any of the existing utilities. We can add a test to verify localStorage for hubCentralSessionToken.

ryanjdew:
@akshaysonvane Updated the session counter to make use of useRef and that appears to resolve the issue you were seeing. I temporarily set the SESSION_WARNING_COUNTDOWN to 299 for the following video and reverted that value before my last push. 

https://drive.google.com/file/d/1y8-LwRS8QzUND_reejuUf6orEPqXu1UI/view?usp=sharing

As for subscriptions, after looking at it further those are handle on the client-side. There isn't anything that we need to do on the server-side to manually clean up.

akshaysonvane:
The sync between the browser tabs looks good now but on auto logout I am seeing
Screen Shot 2020-08-05 at 12 50 19 PM
We might need to adjust the way we unsubscribe.

ryanjdew:
I adjusted the session modal to not show if the user isn't authenticated or if the session timeout is in the negative (adjusting the count come from userContext allowed for jumps so it could skip === 0). 

I'm not seeing repeated attempts to connect to websockets on logout my side. Depending on timing I may get a single attempt, but that shouldn't an issue.

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/auth/WebSecurityConfiguration.java
```diff
@@ -55,12 +59,19 @@ protected void configure(HttpSecurity http) throws Exception {
             .and()
             // Define requests that are always permitted, regardless of whether the user is authenticated or not
             .authorizeRequests()
+                // Needed for websocket tests
+//                .antMatchers("/websocket/**").permitAll()
```
akshaysonvane, line 63: Minor nitpick, no need to change, for now, we can get rid of these commented out lines.

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/auth/WebSecurityConfiguration.java
```diff
@@ -55,12 +59,19 @@ protected void configure(HttpSecurity http) throws Exception {
...
                 .antMatchers("/swagger-resources/**", "/swagger-ui.html", "/v2/api-docs", "/webjars/**").permitAll()
                 // Non-springfox patterns to permit
-                .antMatchers(getAlwaysPermittedPatterns()).permitAll().anyRequest().authenticated()
-            .and()
-            .logout().logoutUrl("/api/logout").logoutSuccessHandler(((request, response, authentication) -> request.getSession().invalidate()));
+                .antMatchers(getAlwaysPermittedPatterns()).permitAll();
+        // needed for WebSocket test
+        if (environment.getProperty("hub.websocket.securityDisabled","false").equals("true")) {
+            http.authorizeRequests().antMatchers("/websocket/**").permitAll();
+        }
```
akshaysonvane, line 71: @rjrudin, FYI, we are permitting access to ```/websocket/**``` only while running tests.  So the production app has all endpoints behind spring security.

PR #4315, opened 2020-07-30
DHFPROD-5484: Package hub-central as a RPM

akshaysonvane:
### Description
The package does not perform a dependency check on Java 11 because the java dependency could be satisfied by various packages (oracle java, openjdk java, etc). We can only add a dependency on one of the packages and if we change the java vendor or the package type in the future on DHS then the RPM will complain and not install. So I think its best not to include this dependency check.
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [N/A] Added Tests
  

- ##### Reviewer:

- [N/A] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/build-rpm.gradle
+++ b/marklogic-data-hub-central/build-rpm.gradle
@@ -1,0 +1,59 @@
+ospackage {
+    packageName = 'hub-central'
+    version = "${project.version}"
+    release = 1
+    os = LINUX
+    type = BINARY
+    arch = NOARCH
+    summary = 'Hub Central Application Server'
+    url = 'http://marklogic.com/'
+    vendor = 'MarkLogic Corporation'
+
+    preInstall file('src/rpm/scripts/pre-install.sh')
+    postInstall file('src/rpm/scripts/post-install.sh')
+    preUninstall file('src/rpm/scripts/pre-uninstall.sh')
+    postUninstall file('src/rpm/scripts/post-uninstall.sh')
+
+    requires('systemd')
+
+    user 'hc-runner'
+    permissionGroup 'hc-runner'
+
+    from(war.outputs.files) {
+        rename { String fileName ->
+            fileName.replace("marklogic-data-hub-central-${project.version}", "hub-central")
+        }
+        fileMode 0755
+        into '/opt/hub-central'
+    }
+
+    from('src/rpm/services/hub-central.service') {
+        addParentDirs(false)
+        fileMode 0755
+        into '/etc/systemd/system'
+    }
+
+    from('src/rpm/configs/hub-central.conf') {
+        fileType CONFIG | NOREPLACE
+        fileMode 0644
+        into '/opt/hub-central'
+    }
+
+    from('src/rpm/configs/hub-central.properties') {
+        addParentDirs(false)
+        fileType CONFIG | NOREPLACE
+        fileMode 0644
+        into '/etc/opt/hub-central'
+    }
+}
+
+buildRpm {
+    user 'hc-runner'
+    permissionGroup 'hc-runner'
+
+    directory('/var/log/hub-central', 0755)
+
+    doLast {
+        println "RPM file written to './build/distributions'"
+    }
+}
--- a/marklogic-data-hub-central/build.gradle
+++ b/marklogic-data-hub-central/build.gradle
@@ -13,6 +13,7 @@ buildscript {
 plugins {
     id "net.saliman.properties" version "1.5.1"
     id "java"
+    id "nebula.ospackage" version "8.4.1"
 }
 
 apply plugin: "org.springframework.boot"
@@ -21,6 +22,7 @@ apply plugin: "io.spring.dependency-management"
 apply plugin: "war"
 apply plugin: "com.github.node-gradle.node"
 apply plugin: "jsonschema2pojo"
+apply from: 'build-rpm.gradle'
 
 mainClassName = "com.marklogic.hub.central.Application"
 
@@ -38,6 +40,10 @@ ext.junitJupiterVersion = '5.3.1'
 
 bootWar {
     baseName = springBootJarName
+    // Used for running hub-central as a system service
+    // Also allows us to customize options like JAVA_HOME using a config file
+    // https://docs.spring.io/spring-boot/docs/current/reference/html/deployment.html#deployment-script-customization-when-it-runs
+    launchScript()
 }
 
 // To pass Gradle properties for HubCentral to Spring Boot, copy them into the JVM environment
@@ -186,6 +192,7 @@ test {
 	useJUnitPlatform()
 }
 
+tasks.buildRpm.dependsOn(buildUi, copyUiFiles, bootWar)
 
 task testUI(type: NpmTask, dependsOn: installDependencies, group: taskGroup){
     description = "Runs the UI unit tests"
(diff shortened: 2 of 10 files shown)


marklogic-data-hub-central/build.gradle
```diff
@@ -38,6 +39,7 @@ ext.junitJupiterVersion = '5.3.1'
 
 bootWar {
     baseName = springBootJarName
+    launchScript()
```
rjrudin, line 42: Can you include a comment here indicating why this is needed - i.e. so we can run it as a service? 

I think Spring bean declarations and Gradle configuration blocks are similar - both can get a lot done in a small amount of code, but there's often a question of why the code exists. It's really slick that all we need is "launchScript()" here, but it begs the question of why we need it.

marklogic-data-hub-central/build.gradle
```diff
@@ -186,6 +188,63 @@ test {
 	useJUnitPlatform()
 }
 
+ospackage {
+    packageName = 'hub-central'
+    version = "${project.version}"
+    release = 1
```
rjrudin, line 194: What does "release" mean/affect here? When would we change this to some other value?

marklogic-data-hub-central/build.gradle
```diff
@@ -186,6 +188,63 @@ test {
 	useJUnitPlatform()
 }
 
+ospackage {
```
rjrudin, line 191: Can you try moving this config to "build-rpm.gradle" in the same directory? I've found that build.gradle files can become very long over time - this seems like an easy way to split it up. On the other hand, sometimes extracting a separate file is harder than it should be.

Ideally, we can move the ospackage plugin declaration to build-rpm.gradle, and also the entire ospackage block. The trick is often where you apply the separate file within build.gradle. We may still need ospackage declared in build.gradle. 

The benefit would be that we'd have one separate Gradle file for everything pertaining to the rpm, instead of having to sift through hundreds of lines of config in build.gradle to find it.

SameeraPriyathamTadikonda:
Should we include a check if any other service is running on the default port where the application will run?

marklogic-data-hub-central/build.gradle
akshaysonvane: It refers to the RPM release version. 
We increment the release version only if the payload (hub-central) of the RPM remains the same but RPM itself undergoes some change.

Meaning if we publish the RPM with hub-central 5.3.0 and if the RPM scripts have some issue and need to be updated then we will increment the release number for the RPM.

The RPM name is generally of the type NVR (Name-Version-Release)
So in the above case it becomes 
hub-central-5.3.0-1.noarch.rpm -> hub-central-5.3.0-2.noarch.rpm

marklogic-data-hub-central/installer/configs/hub-central.conf
```diff
@@ -0,0 +1,6 @@
+# The location of the java executable is discovered by using the PATH by default,
```
rjrudin, line 1: One other comment - can you put these files under "src/rpm"? I like to avoid a lot of root-level directories, and these files all still count as "source code".

akshaysonvane:
> Should we include a check if any other service is running on the default port where the application will run?

I don't think so because the RPM does not start the service on installation so the user has the option to change the port number. We don't want to show errors or quit installation for something that is user-configurable.

marklogic-data-hub-central/build.gradle
```diff
@@ -13,6 +13,7 @@ buildscript {
 plugins {
     id "net.saliman.properties" version "1.5.1"
     id "java"
+    id "nebula.ospackage" version "8.4.1"
```
akshaysonvane, line 16: Fails if moved to build-rpm.gradle

marklogic-data-hub-central/build.gradle
akshaysonvane: Done

marklogic-data-hub-central/build.gradle
rjrudin: Got it, thanks. That may be worth adding as a comment above the "release" line, as I imagine other developers down the road will wonder how "version" and "release" interact.

marklogic-data-hub-central/build.gradle
rjrudin: Figured, ah well.

marklogic-data-hub-central/build-rpm.gradle
```diff
@@ -0,0 +1,55 @@
...
+
+    from(war.outputs.files) {
+        rename { String fileName ->
+            fileName.replace("marklogic-data-hub-central-${project.version}", "hub-central")
+        }
+        fileMode 0755
+        into '/opt/hub-central'
+    }
+
+    from('src/rpm/services/hub-central.service') {
```
rjrudin, line 30: This is really cool, thanks for getting this together. Makes me think we should opt for an rpm whenever possible now.

marklogic-data-hub-central/build.gradle
```diff
@@ -186,6 +192,7 @@ test {
 	useJUnitPlatform()
 }
 
+tasks.buildRpm.dependsOn(buildUi, copyUiFiles, bootWar)
```
rjrudin, line 195: Can this line not be in build-rpm.gradle? No problem if it has to be here.

marklogic-data-hub-central/build-rpm.gradle
```diff
@@ -0,0 +1,55 @@
...
+
+    from('src/rpm/configs/hub-central.properties') {
+        addParentDirs(false)
+        fileType CONFIG | NOREPLACE
+        fileMode 0644
+        into '/etc/opt/hub-central'
+    }
+}
+
+buildRpm {
```
rjrudin, line 50: Where does this write the rpm file to?

marklogic-data-hub-central/build.gradle
akshaysonvane: Nope, can't move it.

marklogic-data-hub-central/build-rpm.gradle
akshaysonvane: The RPM gets generated under `marklogic-data-hub-central/build/distributions`

rjrudin:
One little tweak - can you add this to the list of tasks in ./marklogic-data-hub/build.gradle where we don't call copyUIAssets? That's around lines 227. Will need to add:

    gradle.startParameter.taskNames*.toLowerCase().contains("buildrpm") ||

That avoids the processing time for the trace-ui stuff in core DHF, which we definitely don't need for HC.

marklogic-data-hub-central/build-rpm.gradle
rjrudin: Can you add a "doLast" block to buildRpm that logs that the rpm was written to "./build/distributions" (don't need to specify the filename, that'll be obvious)? Otherwise a developer testing this has no idea where the file is written to.

Also, when I ran this locally, I noticed that the hyphen in the version name was changed to a tilde: hub-central-5.3~SNAPSHOT-1.noarch.rpm . Do you know what that is? In theory, that won't matter in practice because our real version names don't have hyphens, just curious what's going on.

akshaysonvane:
> One little tweak - can you add this to the list of tasks in ./marklogic-data-hub/build.gradle where we don't call copyUIAssets? That's around lines 227. Will need to add:
> 
> ```
> gradle.startParameter.taskNames*.toLowerCase().contains("buildrpm") ||
> ```
> 
> That avoids the processing time for the trace-ui stuff in core DHF, which we definitely don't need for HC.

Awesome. I was hoping we'd have something to skip that but did not look into the build.gradle file.

marklogic-data-hub-central/build-rpm.gradle
akshaysonvane: I believe that's because of the RPM naming conventions:
`[name]-[version]-[release].[arch].rpm`
Since `-` is the only way to distinguish between 'version' and 'release' the Redline java library gets rid of all hyphens in the version section.

PR #4317, opened 2020-07-30
DHFPROD-3577: Global sort based on sortable properties

rahulvudutala:
Adding tests

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/entities/search/EntitySearchManager.java
+++ b/marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/entities/search/EntitySearchManager.java
@@ -31,9 +31,7 @@ import com.marklogic.client.io.DocumentMetadataHandle;
 import com.marklogic.client.io.Format;
 import com.marklogic.client.io.ReaderHandle;
 import com.marklogic.client.io.StringHandle;
-import com.marklogic.client.io.marker.StructureWriteHandle;
 import com.marklogic.client.query.QueryManager;
-import com.marklogic.client.query.RawCombinedQueryDefinition;
 import com.marklogic.client.query.StructuredQueryBuilder;
 import com.marklogic.client.query.StructuredQueryDefinition;
 import com.marklogic.client.row.RowManager;
@@ -50,7 +48,6 @@ import com.marklogic.hub.central.managers.ModelManager;
 import com.marklogic.hub.dataservices.EntitySearchService;
 import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang3.StringUtils;
-import org.apache.commons.text.StringEscapeUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -59,16 +56,17 @@ import java.io.IOException;
 import java.io.OutputStream;
 import java.time.ZonedDateTime;
 import java.time.format.DateTimeFormatter;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
 
 public class EntitySearchManager {
 
     private static final String MASTERING_AUDIT_COLLECTION_NAME = "mdm-auditing";
     private static final String[] IGNORED_SM_COLLECTION_SUFFIX = {"auditing", "notification"};
 
-    private static final String SEARCH_HEAD = "<search xmlns=\"http://marklogic.com/appservices/search\">\n";
-    private static final String SEARCH_TAIL = "</search>";
-    private static final Set<String> METADATA_FIELD_NAME = new HashSet<>(Arrays.asList("datahubCreatedOn"));
     private static final String CSV_CONTENT_TYPE = "text/csv";
     private static final String CSV_FILE_EXTENSION = ".csv";
     private static final String SEPARATOR = "_";
@@ -95,10 +93,9 @@ public class EntitySearchManager {
         resultHandle.setFormat(Format.JSON);
         try {
             //buildQuery includes datetime conversion which could cause DateTimeException or DateTimeParseException
-            StructuredQueryDefinition queryDef = buildQuery(queryMgr, searchQuery);
-            String query = queryDef.serialize();
-            StructureWriteHandle handle = new StringHandle(buildSearchOptions(query, searchQuery)).withMimetype("application/xml");
-            RawCombinedQueryDefinition rcQueryDef = queryMgr.newRawCombinedQueryDefinition(handle, queryDef.getOptionsName());
+            StructuredQueryDefinition queryDefinition = buildQuery(queryMgr, searchQuery);
+            buildSearchTextWithSortOperator(searchQuery);
+            queryDefinition.setCriteria(searchQuery.getQuery().getSearchText());
 
             // If an entity has been selected, then apply this transform
             String[] entityTypeCollections = searchQuery.getQuery().getEntityTypeCollections();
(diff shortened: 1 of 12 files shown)


marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/entities/search/EntitySearchManager.java
```diff
@@ -371,32 +346,14 @@ private String getFileNameForDownload(JsonNode queryDocument, String fileExtensi
...
-                else {
-                    sb.append(String.format("<element ns=\"\" name=\"%s\"/>\n", StringEscapeUtils.escapeXml10(o.getName())));
-                }
-                sb.append("</sort-order>");
-            });
-            sb.append("</options>");
-        });
+        sortOrders.ifPresent(sortOrderList -> sortOrderList.forEach(sortOrder -> {
+            String sortOperator = "sort".concat(sortOrder.getSortOperator().substring(0, 1).toUpperCase()).concat(sortOrder.getSortOperator().substring(1));
+            String sortDirection = sortOperator.concat(sortOrder.getSortDirection().substring(0, 1).toUpperCase()).concat(sortOrder.getSortDirection().substring(1));
```
akshaysonvane, line 354: You can use `StringUtils.capitalize()` here

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/hub-entities.xqy
```diff
@@ -235,6 +237,28 @@ declare %private function hent:fix-options-exp($nodes as node()*, $sortable-prop
       default return $n
 };
 
+declare %private function hent:add-sort-operators-to-search-options($sortable-properties as map:map)
+{
+  let $sort-operators :=
+    for $property-path in map:keys($sortable-properties)
+      let $property-path := fn:replace($property-path, "es:", "*:")
```
rjrudin, line 244: Are there tests for this line of code? This seems brittle - what if the the namespace prefix is "aes"? Then an occurrence of "aes:" would end up as "a*:", which I believe will fail. 

I recommend putting the conversion of a property path into a constraint name into a separate non-private function and writing unit tests for that. That will both verify and document the behavior of the function.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/hub-entities.xqy
```diff
@@ -235,6 +237,28 @@ declare %private function hent:fix-options-exp($nodes as node()*, $sortable-prop
...
+      let $constraint-name := fn:concat("sort", xdmp:initcap(fn:tokenize($property-path, "/")[last()]))
+      return
+        <search:operator name="{$constraint-name}">
```
rjrudin, line 247: We only need a single operator with a name of "sort". Then, for every sortable property, we have two states - "(propertyName)Ascending" and "(propertyName)Descending". 

I believe you can get away with "propertyName" and not property path because for 5.3.0, we're only supporting sorting on top-level properties. Thus, we do not need to worry about property paths (we may never have to).

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/hub-entities.xqy
rjrudin: Also, see my comment below - I don't think we need property paths here, because for 5.3.0, we're only supporting sorting on top-level properties. Thus, a simple property name will suffice. And I don't think that should have a namespace prefix in it, in which case there's no need to replace "es:".

marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/entities/search/EntitySearchManagerTest.java
```diff
@@ -122,41 +122,6 @@ public void testSearchResultsOnNoData() {
         assertTrue(new EntitySearchManager(getHubClient()).search(query).get().isEmpty());
     }
```
rjrudin, line 123: I believe we need two kinds of tests in this class:

1. One or more tests that verify that the correct "criteria" is constructed based on the SortOrder objects in the SearchQuery
1. A test that verifies that inserts two customer entity instances and verifies that sorting on a property in ascending and then descending order returns the 2 instances in the correct order.

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/impl/hub-entities/generateSearchOptions.sjs
```diff
@@ -135,7 +135,73 @@ function generateExplorerWithFacetableAndSortableProperties() {
...
+      }
+    }
+  }];
+
+  const expOptions = hent.dumpSearchOptions(input, true);
+  return [
+    test.assertEqual("sort", xs.string(fn.head(expOptions.xpath("/*:operator/@name")))),
+    test.assertExists(expOptions.xpath("/*:operator/*:state[@name = 'titleDescending']")),
+    test.assertEqual("descending", xs.string(fn.head(expOptions.xpath("/*:operator[@name = 'sort']/*:state[@name = 'titleDescending']/*:sort-order/@direction")))),
+    test.assertEqual("//*:instance/Book/title", xs.string(fn.head(expOptions.xpath("/*:operator[@name = 'sort']/*:state[@name = 'titleDescending']/*:sort-order/*:path-index")))),
```
rjrudin, line 161: This is fine to keep here - 5626 will turns this into "/*:envelope/*:instance/*:Book/*:title", just an FYI.

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/impl/hub-entities/generateSearchOptions.sjs
```diff
@@ -135,7 +135,73 @@ function generateExplorerWithFacetableAndSortableProperties() {
...
+    test.assertExists(expOptions.xpath("/*:operator/*:state[@name = 'authorsAscending']")),
+    test.assertEqual("ascending", xs.string(fn.head(expOptions.xpath("/*:operator[@name = 'sort']/*:state[@name = 'authorsAscending']/*:sort-order/@direction")))),
+    test.assertEqual("//*:instance/Book/authors", xs.string(fn.head(expOptions.xpath("/*:operator[@name = 'sort']/*:state[@name = 'authorsAscending']/*:sort-order/*:path-index")))),
+    test.assertNotExists(expOptions.xpath("/*:operator[@name = 'bookId']")),
+    test.assertNotExists(expOptions.xpath("/*:operator[@name = 'completedDate']"))
+  ];
+}
+
+function verifyReplaceEsNamespace() {
+  let propertyPath = "/es:instance/es:entityType/es:value";
```
rjrudin, line 177: I am hoping 5626 removes the need for this, as path expressions will now always have "/*:" in them instead of "/es:", but I'm not 100% sure of that yet.

marklogic-data-hub/src/test/resources/test-config/databases/staging-database.json
```diff
@@ -0,0 +1,5 @@
+{
+  "database-name": "data-hub-STAGING",
+  "range-element-index" : [],
```
rjrudin, line 3: Are these empty arrays intentionally here to clear out the indexes?

rahulvudutala: Yes. To restore the staging database index config to the previous state, just as for data-hub-FINAL database. Since there were no indexes previously, I have added empty arrays

PR #4318, opened 2020-07-30
DHFPROD-5315:Display fields in array of objects only once in source d…

briantang:
…ropdown

### Description

- Source dropdown in Entity table only shows fields for one object now, instead of multiple with the same name from array.
- Fixes weird menu jumping around when scrolling in dropdown.
- Refactored some existing tests to accommodate added data I used for new unit tests.

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/src/assets/mock-data/common.data.ts
+++ b/marklogic-data-hub-central/ui/src/assets/mock-data/common.data.ts
@@ -77,11 +77,19 @@ const jsonSourceDataMultipleSiblings = [
       ]}
     ]
   },
-  { rowKey: 7, key: 'proteinCat', val: 'commercial' },
   {
-    rowKey: 8, key: 'withNutsOrganism', children: [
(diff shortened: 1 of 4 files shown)


marklogic-data-hub-central/ui/src/assets/mock-data/common.data.ts
```diff
@@ -142,8 +142,14 @@ const xmlSourceData = [
         { rowKey: 6, key: 'LastName', val: 'Smith' }
       ]
       },
-      { rowKey: 7, key: 'proteinCat', val: 'commercial' },
-      { rowKey: 8, key: 'proteinDog', val: 'retriever, golden, labrador', array: true}
+      {
+        rowKey: 7, key: 'nutFree:name', children: [
```
bsrikan, line 146: Can you try using other existing mock data that end with `MultipleSiblings` than modifying this one.

PR #4319, opened 2020-07-30
DHFPROD-5265:Create step via Gradle

srinathgit:
### Description
Create step via Gradle

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/step/saveStep.sjs
```diff
@@ -76,6 +77,17 @@ if (existingStep) {
...
   }
 
+  if(stepProperties.entityType){
+    if(fn.docAvailable("/entities/"+ stepProperties.entityType +".entity.json")){
+      const entityTypeId = entityLib.getEntityTypeId(entityLib.findModelByEntityName(stepProperties.entityType), stepProperties.entityType);
+      stepProperties.targetEntityType = entityTypeId;
+      delete stepProperties.entityType;
+    }
+    else {
+      throw new Error(`The specified entity type ${stepProperties.entityType} is not found`);
```
rjrudin, line 87: This doesn't allow Pari to enter e.g. "http://example.org/Customer-1.0/Customer", which she may do. 

What we want to do instead is say - if the entity URI you're constructing exists, then we know Pari has entered "Customer" and she'd really appreciate it if we converted that into an entityTypeId for her. But if the URI doesn't exist, then we assume Pari entered a valid value. If it's not valid, she'll get an error when the step is saved - which is good, and the try/catch for that operation should make it very clear that saving the step failed.

PR #4323, opened 2020-07-30
DHFPROD-5630: Adding steps to reference project

rjrudin:
Also adding "hub-user" for testing a user that only has hub-central-user (it doesn't work, Explore blows up because the user can't hit the saved query endpoints, will write up a ticket for that).

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/examples/reference-entity-model/flows/CurateCustomerJSON.flow.json
+++ b/examples/reference-entity-model/flows/CurateCustomerJSON.flow.json
@@ -6,6 +6,113 @@
     },
     "2": {
       "stepId": "mapCustomersJSON-mapping"
+    },
+    "3": {
+      "name": "match-customers",
+      "stepDefinitionName": "default-matching",
+      "stepDefinitionType": "MATCHING",
+      "options": {
+        "sourceQuery": "cts.collectionQuery('mapCustomersJSON')",
+        "acceptsBatch": true,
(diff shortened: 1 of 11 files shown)


examples/reference-entity-model/flows/CurateCustomerJSON.flow.json
```diff
@@ -6,7 +6,114 @@
     },
     "2": {
       "stepId": "mapCustomersJSON-mapping"
+    },
+    "3": {
```
rjrudin, line 10: @ryanjdew  Let me know if there's a better way to do a very simple match config on one property.

One thing I noticed is that the two customers that are merged together remain in the "Customer" collection. Is it up to the user to move them out of this collection? This results in the new, merged customer and those two original customers all appearing in search results, which doesn't seem desirable.

ryanjdew: There isn't a simpler way. Part of the reason we're looking to refactor the options.

The user could change the options to remove the Customer collection on archive. In the case of Customer, "sm-Customer-mastered" would be the collection to signify the distinct documents after mastering.

PR #4326, opened 2020-07-31
DHFPROD-5371: Navigate away prompt for modeling

brucean52:
### Description
updated @testing-library/user-event to v12 for type method no longer being async 
https://github.com/testing-library/user-event/releases/tag/v12.0.0
added @testing-library/dom -> new dependency for user-event

added model reader e2e test

Persist model between tiles + navigation warn when logout: https://drive.google.com/file/d/15jSoV-fe7JgRWs4xeaYaJEBHTmNDrzQY/view

Navigate away with edits: https://drive.google.com/file/d/1eI-cDhw0GlTXd0YDSGvjnVzWsTWjmaOu/view

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/e2e/cypress/integration/modeling/reader.spec.tsx
+++ b/marklogic-data-hub-central/ui/e2e/cypress/integration/modeling/reader.spec.tsx
@@ -1,0 +1,102 @@
+/// <reference types="cypress"/>
+
+import modelPage from '../../support/pages/model';
+import {
+  entityTypeModal,
+  entityTypeTable,
+  propertyModal,
+  propertyTable,
+} from '../../support/components/model/index';
+import { confirmationModal, toolbar, tiles } from '../../support/components/common/index';
+import { Application } from '../../support/application.config';
+import browsePage from '../../support/pages/browse'
+import 'cypress-wait-until';
+
+describe('Entity Modeling: Reader Role', () => {
+
+  //login with valid account
+  beforeEach(() => {
+    cy.visit('/');
+    cy.contains(Application.title);
+    console.log(Cypress.env('mlHost'));
+    cy.loginAsTestUserWithRoles("hub-central-entity-model-reader", "hub-central-saved-query-user").withRequest()
+    cy.waitUntil(() => toolbar.getModelToolbarIcon()).click();
+    entityTypeTable.waitForTableToLoad();
+  });
+
+  after(() => {
+      //resetting the test user back to only have 'hub-central-user' role
+      cy.resetTestUser();
+  });
+
+  it('can navigate by clicking instance count and last processed, can not create, edit, or delete entity models', () => {
+    cy.waitUntil(() => entityTypeTable.getEntityLastProcessed('Person')).click();
+    tiles.getExploreTile().should('exist');
+    cy.waitUntil(() => browsePage.getSelectedEntity()).should('eq', 'Person');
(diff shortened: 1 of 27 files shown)


marklogic-data-hub-central/ui/e2e/cypress/integration/modeling/reader.spec.tsx
```diff
@@ -0,0 +1,93 @@
...
+    cy.url().should('include', '/tiles/explore');
+    tiles.getExploreTile().should('exist');
+    cy.waitUntil(() => browsePage.getSelectedEntity().should('eq', 'Order'));
+    browsePage.getClearAllButton().should('not.exist');
+
+    cy.go('back');
+    cy.url().should('include', '/tiles/model');
+    tiles.getModelTile().should('exist');
+
+    modelPage.getAddEntityButton().click({ force: true });
```
bsrikan, line 53: wondering why you had to use force: true option.

marklogic-data-hub-central/ui/src/components/confirmation-modal/confirmation-modal.test.tsx
```diff
@@ -304,5 +296,39 @@ describe('Confirmation Modal Component', () => {
...
+    rerender(<ConfirmationModal
+      isVisible={true}
+      type={ConfirmationType.NavigationWarn}
+      boldTextArray={[]}
+      toggleModal={toggleModal}
+      confirmAction={confirmAction}
+    />);
+
+    expect(getByText('Unsaved Changes')).toBeInTheDocument();
+    expect(getByLabelText('navigation-warn-text')).toBeInTheDocument();
```
bsrikan, line 325: can we verify the actual message here?

marklogic-data-hub-central/ui/e2e/cypress/integration/modeling/reader.spec.tsx
brucean52: The test will fail if I don't because they aren't actionable items. Even though the button is disabled, I'm forcing clicks to see if the functionality is there. For edit Entity/Property names in the table, model-reader role displays text instead of links (with the same selector). So for reader tests, I'm clicking the same selectors as writer role, but I don't expect any edit/ add modals to show up.

Running the without force: true
Entity Modeling Reader Role -- can navigate by clicking instance count and last processed, can not create, edit, or delete entity models (failed)

marklogic-data-hub-central/ui/src/components/confirmation-modal/confirmation-modal.test.tsx
brucean52: Yeah, I can change it.

marklogic-data-hub-central/ui/e2e/cypress/integration/modeling/reader.spec.tsx
bsrikan: Okay. interesting way to test a reader role. So you are forcing a click action to verify that the user is able to read?

brucean52: I'm testing that the reader role doesn't have any of the writer role privileges in the UI. If you add the "model writer" role to the reader test, the test will fail.

PR #4329, opened 2020-08-03
DHFPROD-5626: Fixing path expressions for entity models

rjrudin:
### Description

Most changes are to test files, as we had a lot of assertions (which is good) that had to be updated. 

Application changes of note:

- The mlPiiGenerator REST extension is gone, in favor of generateProtectedPathConfig, which is a custom approach. This was easier than trying to fix the output of es.piiGenerate. Added a bunch of tests for this, including namespace coverage, which we previously lacked.
- hub-entities.xqy now fixes path expressions in search options and database indexes (via fix-path-expression), and it no longer has dump-pii 
- hub-entities.xqy also handles XML path expressions correctly now, and tests have been added for that
- hub-entities.xqy now builds correct path-index elements with the required namespaces for indexes when it builds sort states
- Updated the reference model project so that its indexes are correct

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


rjrudin:
Tweaked GenerateProtectedPathsTest so that it didn't assume which protected path is written to which file, which doesn't matter.

PR #4332, opened 2020-08-03
DHFPROD-5311: Fix file type circles in Run Tile

AkshayBajajML:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/src/assets/mock-data/flows.data.ts
+++ b/marklogic-data-hub-central/ui/src/assets/mock-data/flows.data.ts
@@ -527,9 +527,28 @@ const customSteps = {"data" :{"stepsWithEntity":[{
     "status" :200
 }
 
+const flowProps = {
+    flows: flows.data,
+    deleteFlow: jest.fn(),
+    createFlow: jest.fn(),
+    updateFlow: jest.fn(),
+    deleteStep: jest.fn(),
+    runStep: jest.fn(),
+    canReadFlow: true,
+    canWriteFlow: true,
+    hasOperatorRole: true,
+    running: [],
+    uploadError: "",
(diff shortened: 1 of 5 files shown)


jbelonoj:
Sure, would look nicer all boxes with the same height, @wooldridge. I guess one thing to test as we are making that change would be a long step name that wraps to the next line.

wooldridge:
@jbelonoj, good point about the long text. Would you prefer we use CSS to add ellipses for long names that hit the end of the enclosing container?

jbelonoj:
@wooldridge I think we have to truncate and add "..." for the names that don't have space and we wrap to the next line for the ones that have. What do you think?

wooldridge:
Interesting point, @jbelonoj. I can't recall any other places where we've distinguished between long names with spaces and long names without spaces, and applied different formatting. The ellipses would come with a mouseover that shows the entire name text. I think having the ellipsis truncation for all long text is best as it keeps the containers a consistent size and is much, much easier to implement and much less complicated.

jbelonoj:
Sure, @wooldridge. That works, especially for these cards with a pretty rigid spacing. I think generally it's better usability - wise to wrap text with spaces in the tables, for example. We want to the user to see the actual data rather than a bunch of "...". If the text doesn't have spaces, we have no choice but truncate. Does this make sense?

wooldridge:
@jbelonoj, here's what I currently see:

Screen Shot 2020-08-04 at 3 02 16 PM

I will move the truncation issue to a separate ticket since it's not trivial and is different from what @AkshayBajajML fixed, here's the issue I created: https://project.marklogic.com/jira/browse/DHFPROD-5683

PR #4335, opened 2020-08-04
DHFPROD-5266:Add step to flow

srinathgit:
### Description

Add step to flow

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


marklogic-data-hub/src/main/java/com/marklogic/hub/impl/ScaffoldingImpl.java
```diff
@@ -173,6 +174,29 @@ public static String getAbsolutePath(String first, String... more) {
         }
     }
 
+    public Pair<File, String> addStepToFlow(String flowName, String stepName, String stepType) {
```
rjrudin, line 177: This looks good, but can you move it to FlowManagerImpl? I think the scope of ScaffoldingImpl should be limited to scaffolding out new file in a project, whereas this is a flow-specific operation.

PR #4339, opened 2020-08-04
DHFPROD-5678: Moving deployment of amps from RunMarkLogicUnitTestsTest to Installer and fixing 9.0-11 tests

rahulvudutala:
### Description

- Moving the deployment of data-hub-test-internal role and amps to Installer class so that tests can be run from http://localhost:8011/test/default.xqy without running the RunMarkLogicUnitTestsTest once. (Currently, the role and amp is deployed by RunMarkLogicUnitTestsTest)

- Adding version compatibility check for deploying indexes in EntitySearchManagerTest to work with 9.0-11 version

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/test/java/com/marklogic/hub_unit_test/RunMarkLogicUnitTestsTest.java
+++ b/marklogic-data-hub/src/test/java/com/marklogic/hub_unit_test/RunMarkLogicUnitTestsTest.java
@@ -65,15 +65,8 @@ public class RunMarkLogicUnitTestsTest extends HubTestBase {
     @Override
     protected void init() {
         if (!initialized) {
-            super.init();
-            try {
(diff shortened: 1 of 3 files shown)


marklogic-data-hub/src/test/java/com/marklogic/hub_unit_test/RunMarkLogicUnitTestsTest.java
```diff
@@ -65,15 +65,8 @@ public void resetDatabasesAfterAllTestsHaveRun() {
     @Override
     protected void init() {
         if (!initialized) {
```
ryanjdew, line 67: I believe this method can be removed at this point.

PR #4343, opened 2020-08-05
DHFPROD-5527:Changes to sort should be consideration for actions on Saved Query

Sanjeevani19:
### Description

- Also Includes https://project.marklogic.com/jira/browse/DHFPROD-5689

- Added an additional role for numeric slider e2e tests to pass

- Previously loadCustomersJSON name facet was shown on the sidebar under name entity properties but it seems to be a bug and loadCustomersJSON has been removed from the facets as part of https://project.marklogic.com/jira/browse/DHFPROD-5018, unfortunately most e2e tests are written based on loadCustomersJSON facet, Addressed all those issues in this PR

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


timur-isangulov:
Hi Sanjeevani, I set sort option for the array property e.g. nicknames and got error when sorting on the table. 
Screen Shot 2020-08-05 at 2 58 10 PM

PR #4345, opened 2020-08-05
DHFPROD-5637: Add entityInstance object to search results for row render

rahulvudutala:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/entities/entity-search-lib.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/entities/entity-search-lib.sjs
@@ -35,7 +35,7 @@ function addPropertiesToSearchResponse(entityName, searchResponse, propertiesToD
   let selectedPropertyMetadata = [];
   let propertyMetadata = [];
 
-  if (entityName != null) {
+  if (entityName) {
     //Single Entity is selected
 
     const entityModel = entityLib.findModelByEntityName(entityName);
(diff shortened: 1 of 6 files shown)


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/entities/entity-search-lib.sjs
```diff
@@ -342,17 +345,23 @@ function addEntitySpecificProperties(result, entityInfo, selectedPropertyMetadat
         result.entityProperties.push(getPropertyValues(parentProperty, entityInstance));
       });
 
+      propertyMetadata.forEach(parentProperty => {
+        result.entityInstance.push(getPropertyValues(parentProperty, entityInstance));
```
rjrudin, line 349: I thought the intent was that the entire entity instance would be included - i.e. the part at /*:envelope/*:instance ? Isn't that what the UI needs for the current row expander, where it's just showing a pretty JSON view of the entity? I know in the future, we want to show properties with labels for the row expander and detail view, but I think that's 5.4.0.

rahulvudutala: Yes, propertyMetadata has all the entityProperties of an entityModel.
Row expander shows a tabular column with entityTypePropertyName and its corresponding value.

rjrudin: Doesn't that result in a change of what's shown though? The row expander is showing a pretty JSON view of the entity instance. That's apparent because the user sees "shipping / Address / street" instead of just "shipping / street". 

I'll try this out locally, it just seems the easiest thing to do is return the entity instance so that the UI doesn't have to try to guess where it is in the extracted content array.

rahulvudutala:
Rebased so it gets the fix for the websocket test

PR #4348, opened 2020-08-06
DHFPROD-5425: Duplicate flow name error message now has a bolded flow name

bsrikan:
### Description
Fixes error message to have bolded flow name
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/src/pages/Run.tsx
+++ b/marklogic-data-hub-central/ui/src/pages/Run.tsx
@@ -85,9 +85,10 @@ const Run = (props) => {
     }
 
     const createFlow = async (payload) => {
+        let newFlow;
         try {
             setIsLoading(true);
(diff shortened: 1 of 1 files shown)


bsrikan:
Correct Ryan. All the error modal text where we have entity/step/property name etc in single quote is being updated to show in bold text to be consistent.
I will confirm with Julia again on the messaging and add a test and mark it Ready for review.
Thanks guys!

bsrikan:
I am not able to add an RTL test. The response we show is what we get from the server. Even though I mock the 400 response, the test fails with  `Cannot read property 'data' of undefined`.
Maybe later I will add an e2e test.

PR #4354, opened 2020-08-06
DHFPROD-5224:Fix issues with "outputURIPrefix" in ingestion step

srinathgit:
### Description
Fix issues with "outputURIPrefix" in ingestion step

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/data-services/step/convertIngestionStep.sjs
+++ b/marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/data-services/step/convertIngestionStep.sjs
@@ -37,7 +37,7 @@ let info = {
   additionalCollections: ["customerJSON"],
   inputFilePath: "data-sets/CustomerLoadJSON",
   outputURIReplacement: ".*CustomerLoadJSON,'/customer'",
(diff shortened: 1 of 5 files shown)


marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/data-services/step/convertIngestionStep.sjs
```diff
@@ -37,7 +37,7 @@ let info = {
   additionalCollections: ["customerJSON"],
   inputFilePath: "data-sets/CustomerLoadJSON",
   outputURIReplacement: ".*CustomerLoadJSON,'/customer'",
-  outputURIPrefix: "/customer/",
+  outputURIPrefix: "",
```
rjrudin, line 40: So is the fix here - if the prefix is just "", then treat that as a valid value?

srinathgit: yeah that's correct

srinathgit:
Fixed a failing test

PR #4362, opened 2020-08-07
DHFPROD-5692: hubDeployAsDeveloper now generates indexes and PII files

rjrudin:
### Description

Did some cleanup too:

- Got rid of GeneratePiiCommand, which didn't add any value. hubGeneratePii just calls EntityManager.savePii now. 
- Got rid of GeneratePiiCommandTest, as GenerateProtectedPathsTest supercedes it
- EntityManagerImpl honors its interface now and logs errors instead of throwing them when an error occurs when writing index or PII files

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


rjrudin:
Had a compile error, left an import in.

PR #4365, opened 2020-08-08
DHFPROD-5369: Usability bug: Mapping step should show a spinner while loading large docs

AkshayBajajML:
… loading large docs

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/src/components/entities/mapping/source-entity-map/source-to-entity-map.tsx
+++ b/marklogic-data-hub-central/ui/src/components/entities/mapping/source-entity-map/source-to-entity-map.tsx
@@ -10,8 +10,7 @@ import { getMappingValidationResp } from "../../../../util/manageArtifacts-servi
 import DropDownWithSearch from "../../../common/dropdown-with-search/dropdownWithSearch";
 import SplitPane from 'react-split-pane';
 import Highlighter from 'react-highlight-words';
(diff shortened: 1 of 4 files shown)


marklogic-data-hub-central/ui/src/components/entities/mapping/source-entity-map/source-to-entity-map.tsx
```diff
@@ -1196,7 +1196,10 @@ const SourceToEntityMap = (props) => {
                                 placement="right"
                             ><Icon type="question-circle" className={styles.questionCircle} theme="filled" /></Popover></p>
                         </div>
-                        {emptyData ?
+                        {props.isLoading === true ? <div className={styles.spinRunning}>
+                            <MLSpin size={"large"} data-testid="spinTest"/>
```
bsrikan, line 1200: Can we just call spinner instead of spinTest

AkshayBajajML: flow.tsx uses spinner as data-testid and it is also used in current cypress test cases. Would it impact it if we give same testId here?

PR #4371, opened 2020-08-10
DHFPROD-5359: Incorrect XPATH is generated when Mapping XML documents

briantang:
### Description
- Fix to XML source data format when there is a Namespace and the property is an array of values, causing invalid XPATH.
   (should be 'custOrderInfo: nicknames' instead of 'nicknames')

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


briantang:
Added test verifying XPATH for namespaced property with array of values.

PR #4382, opened 2020-08-11
DHFPROD-5043: Incorrect collection name shows up in mapping UI

AkshayBajajML:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


AkshayBajajML:
This patch was added to merged PR to reset collection name to null when user selects query as source query and saves it.
@bsrikan could not add test case to this patch as it will reflect change only after saving the update.

PR #4383, opened 2020-08-11
DHFPROD-4145: Ability to show multiple values in the table view when it conflicts with the associated property definition

rahulvudutala:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/entities/search/getEntityInstance.sjs
+++ b/marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/entities/search/getEntityInstance.sjs
@@ -160,6 +160,13 @@ function getInstanceFromTextDocument() {
   ]
 }
 
+function getInstanceWithInstanceInDifferentXpath() {
+  const assertions = [];
+  let docUri = "/content/instanceElementInDifferentXpath.xml";
+  let instance = entitySearchLib.getEntityInstance(docUri);
+  assertions.push(test.assertEqual(null, instance));
+}
+
 []
     .concat(getInstanceWithoutESInfo())
     .concat(getInstanceWithESInfo())
@@ -167,7 +174,5 @@ function getInstanceFromTextDocument() {
     .concat(getInstanceWithNonExistentModel())
     .concat(getInstanceWithNamespace())
     .concat(getInstanceWithNonExistingDocUri())
-    .concat(getInstanceFromTextDocument());
-
-
-
+    .concat(getInstanceFromTextDocument())
+    .concat(getInstanceWithInstanceInDifferentXpath());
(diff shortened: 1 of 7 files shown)


marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/entities/search/getEntityInstance.sjs
```diff
@@ -160,14 +160,39 @@ function getInstanceFromTextDocument() {
   ]
 }
 
+function getInstanceFromMergedJsonDocument() {
```
rjrudin, line 163: Are these functions really testing anything? Our current expectation of getEntityInstance is that it simply fetches the instance from the document, nothing more. So I don't think we need these tests, because we expect the instance to be returned as-is - we don't have any reason to think that the instance would be modified by getEntityInstance. If getEntityInstance were performing some logic on the instance, then we'd want tests, but that's not the case.

rahulvudutala: @rjrudin In getEntityInstance.sjs, we are validating if es:instance-json-from-document is returning the instance in various scenarios such as a mapped document with namespace, instance without info, instance with nonexistent modelName. Similarly, I added a test case for a merged document too.
Should I remove this test case or this test file?

rjrudin: I don't believe there's any difference in how the instance is obtained for a merged entity vs a non-merged entity - i.e. the instance is still at /*:envelope/*:instance. So I don't think a test is needed. We'd need one if a merged document instead had something like /*:envelope/*:merged/*:instance.

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/entities/search/getEntityInstance.sjs
```diff
@@ -160,14 +160,19 @@ function getInstanceFromTextDocument() {
   ]
 }
 
+function getInstanceWithInstanceInDifferentXpath() {
```
rjrudin, line 163: @rahulvudutala  We still don't need this test :) - I was just raising a hypothetical. Though I'm curious - does es.instanceJsonFromDocument actually return the instance when "es:instance" is at any depth? 

I would normally say - there's no need to support "/*:envelope/*/es:instance". But if es.instanceJsonFromDocument can handle that, then I suppose it's okay to include.

rahulvudutala: @rjrudin es.instanceJsonFromDocument is returning the instance when "es:instance" is at any depth
I tried on the document which has the path /es:envelope/es:merged/es:instance. The instance is returned (There is no  /es:envelope/es:instance in the document)

rjrudin: Huh - well I guess that's good? Definitely surprising behavior, but I don't think there's a real downside to it.

PR #4385, opened 2020-08-11
DHFPROD-5637: Adding entity def with namespace to reference project

rjrudin:
### Description

This is to facilitate testing of XML entity instances with data in a namespace. 

Also added "status" data for Customer entities so that a facet on it can be tested. 

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


examples/reference-entity-model/src/main/ml-config/security/query-rolesets/pii-reader.json
```diff
@@ -0,0 +1,3 @@
+{
```
bsrikan, line 1: I think we can get rid of this file. hubDeployAsDeveloper will generate this via hubGeneratePii anyways.

rjrudin: It either needs to be gitignored or added to version control. What I'll do is make one of the properties be PII, as we really ought to have one for testing anyway. I'll give pii-reader to one of the users too.

rjrudin: Just added "ssn" to Customer and made it PII. So now there's a path file and the roleset file.

PR #4386, opened 2020-08-11
DHFPROD-5700: Don't add unnecessary encoding for match queries

ryanjdew:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/com.marklogic.smart-mastering/survivorship/merging/base.xqy
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/com.marklogic.smart-mastering/survivorship/merging/base.xqy
@@ -53,7 +53,10 @@ import module namespace util-impl = "http://marklogic.com/smart-mastering/util-i
   at "/com.marklogic.smart-mastering/impl/util.xqy";
 import module namespace mem = "http://maxdewpoint.blogspot.com/memory-operations/functional"
   at "/mlpm_modules/XQuery-XML-Memory-Operations/memory-operations-functional.xqy";
-import module namespace es-helper = "http://marklogic.com/smart-mastering/entity-services" at "/com.marklogic.smart-mastering/sm-entity-services.xqy";
+import module namespace es-helper = "http://marklogic.com/smart-mastering/entity-services"
+  at "/com.marklogic.smart-mastering/sm-entity-services.xqy";
+import module namespace helper = "http://marklogic.com/smart-mastering/helper-impl"
+  at "/com.marklogic.smart-mastering/matcher-impl/helper-impl.xqy";
 
 declare namespace merging = "http://marklogic.com/smart-mastering/merging";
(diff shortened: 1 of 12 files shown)


marklogic-data-hub/src/main/resources/ml-modules/root/com.marklogic.smart-mastering/survivorship/merging/base.xqy
```diff
@@ -53,7 +53,10 @@ import module namespace util-impl = "http://marklogic.com/smart-mastering/util-i
   at "/com.marklogic.smart-mastering/impl/util.xqy";
 import module namespace mem = "http://maxdewpoint.blogspot.com/memory-operations/functional"
   at "/mlpm_modules/XQuery-XML-Memory-Operations/memory-operations-functional.xqy";
-import module namespace es-helper = "http://marklogic.com/smart-mastering/entity-services" at "/com.marklogic.smart-mastering/sm-entity-services.xqy";
+import module namespace es-helper = "http://marklogic.com/smart-mastering/entity-services"
+  at "/com.marklogic.smart-mastering/sm-entity-services.xqy";
+import module namespace helper = "http://marklogic.com/smart-mastering/helper-impl"
```
rjrudin, line 58: "helper-impl" is used as a prefix above - I find it's best to use the same prefix throughout the codebase to make it easier to find all references to a function via global search. I think "helper" is the appropriate prefix, unless there's an existing convention in the SM code for using "(something)-impl" for impl libraries.

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/smart-mastering/entity-matching/test-data/entities/Customer_V1.entity.json
```diff
@@ -0,0 +1,95 @@
+{
+  "info": {
+    "title": "Customer_V1",
+    "version": "0.0.1",
+    "baseUri": "http://example.org/"
+  },
+  "definitions": {
+    "Customer_V1": {
```
rjrudin, line 8: I recommend slimming this down to the bare minimum needed to produce the issue. Otherwise, there's an implication that all these properties are needed for the test. Also, don't you need a property with an underscore in its name too?

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/smart-mastering/entity-matching/test-data/flows/customerV1Flow.flow.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "customerV1Flow",
```
rjrudin, line 2: For simplicity, I'd just have name and steps, nothing more.

PR #4388, opened 2020-08-12
DHFPROD-5718: Entity name shouldn't appear as facet

saarimrahman:
Hub properties remain expanded when a hub property is selected.

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/e2e/cypress/integration/explorer/scenarios.json.tsx
+++ b/marklogic-data-hub-central/ui/e2e/cypress/integration/explorer/scenarios.json.tsx
@@ -75,6 +75,27 @@ describe('json scenario for snippet on browse documents page', () => {
     })
   });
 
+  it('select Person/Customer entity, verify no entityName in Collection facets and hub property persistence', () => {
+    browsePage.selectEntity('Person');
+    browsePage.getSelectedEntity().should('contain', 'Person');
+    browsePage.getFacetItemCheckbox('collection', 'mapPersonJSON').should('not.be.visible')
+    browsePage.getHubPropertiesExpanded();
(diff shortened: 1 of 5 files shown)


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/entities/entity-search-lib.sjs
```diff
@@ -437,6 +439,18 @@ function addPrimaryKeyToResult(result, entityInstance, entityDef) {
   }
 }
 
+function removeEntityNameFromCollection(searchResponse, entityName) {
+  if (searchResponse.hasOwnProperty('facets')) {
```
rjrudin, line 443: This is fine for now, but really, a custom constraint would be better here. That's because it would avoid having ML calculate the number of documents in the entity-name collection, which would speed up search. 

But really, the right long term solution is to not have a collection facet. That's on my kill list for 5.4.0. 

Pinging @fsnow  about this too, just for awareness.

marklogic-data-hub-central/ui/e2e/cypress/integration/explorer/scenarios.json.tsx
```diff
@@ -75,6 +75,27 @@ describe('json scenario for snippet on browse documents page', () => {
     })
   });
 
+  it('select Person/Customer entity, verify no entityName in Collection facets and hub property persistence', () => {
```
bsrikan, line 78: Probably ok for now, but its better for these scenarios to be tested in RTL.

PR #4395, opened 2020-08-13
DHFPROD-5380: datahubCreatedByStep is now set correctly

rjrudin:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/impl/flow-utils/getMetadata.sjs
+++ b/marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/impl/flow-utils/getMetadata.sjs
@@ -3,43 +3,25 @@ const test = require("/test/test-helper.xqy");
 const FlowUtils = require("/data-hub/5/impl/flow-utils.sjs");
 const flowUtils = new FlowUtils();
 
-const results = [];
-const metadata = [];
+const assertions = [];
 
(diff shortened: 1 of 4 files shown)


marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/impl/flow-utils/getMetadata.sjs
```diff
@@ -3,43 +3,25 @@ const test = require("/test/test-helper.xqy");
...
-                     "description": "This is the custom mapping step",
-                     "stepDefinitionName": "mastering",
-                     "stepDefinitionType": "MASTERING"
-                  }
-                }
-              };
-
-//this is to verify that stepDefName corresponding to the stepNumber is the value of metadata datahubCreatedByStep
-for (const stepNum in flowDoc.steps) {
-  metadata.push(flowUtils.createMetadata(metadata ? metadata : {}, flowDoc.name, flowDoc.steps[stepNum].stepDefinitionName, jobId).datahubCreatedByStep)
```
rjrudin, line 37: This test wasn't really testing the function, because it was passing in values - one of which was incorrect (the stepDefinitionName). I reworked the test to instead verify the interesting logic, which is that each bit is overwritten except for jobId.

PR #4403, opened 2020-08-13
DHFPROD-5565: Unauthorized user can view Model content via URL route,…

brucean52:
… DHFPROD-5619: disabled icon color

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/e2e/cypress/integration/modeling/writer.spec.tsx
+++ b/marklogic-data-hub-central/ui/e2e/cypress/integration/modeling/writer.spec.tsx
@@ -21,7 +21,7 @@ describe('Entity Modeling: Writer Role', () => {
     cy.contains(Application.title);
     console.log(Cypress.env('mlHost'));
     cy.loginAsTestUserWithRoles("hub-central-entity-model-reader", "hub-central-entity-model-writer", "hub-central-saved-query-user").withRequest()
-    cy.waitUntil(() => toolbar.getModelToolbarIcon()).click();
+    cy.waitUntil(() => toolbar.getModelToolbarIcon().should('have.css', 'cursor', 'pointer')).click();
     entityTypeTable.waitForTableToLoad();
   });
 
--- a/marklogic-data-hub-central/ui/src/pages/Modeling.test.tsx
+++ b/marklogic-data-hub-central/ui/src/pages/Modeling.test.tsx
@@ -20,6 +20,7 @@ const mockUpdateEntityModels = updateEntityModels as jest.Mock;
 
 const mockDevRolesService = authorities.DeveloperRolesService;
 const mockOpRolesService = authorities.OperatorRolesService;
+const mockHCUserRolesService = authorities.HCUserRolesService;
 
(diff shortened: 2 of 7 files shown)


marklogic-data-hub-central/ui/e2e/cypress/integration/modeling/writer.spec.tsx
```diff
@@ -21,7 +21,7 @@ describe('Entity Modeling: Writer Role', () => {
     cy.contains(Application.title);
     console.log(Cypress.env('mlHost'));
     cy.loginAsTestUserWithRoles("hub-central-entity-model-reader", "hub-central-entity-model-writer", "hub-central-saved-query-user").withRequest()
-    cy.waitUntil(() => toolbar.getModelToolbarIcon()).click();
+    cy.waitUntil(() => toolbar.getModelToolbarIcon().should('have.css', 'cursor', 'pointer')).click();
```
brucean52, line 24: Doing this to address this e2e test failure on windows. 
model-writer-fail

I believe this is due to toolbar component using inline style instead of CSS. 
https://reactjs.org/docs/faq-styling.html#are-inline-styles-bad

bsrikan: I doubt if that is the case. Other tool icons are clicked the same way and in windows, in this instance, load test didnt fail there. To me it seems to be the async issue. We are using wait(500) in withRequest() and the websocket get request might not have completed in that time resulting in that click timeout error. Will try to replace explicit wait with something else to ensure overview page is completely loaded with all XHR requests complete.

brucean52: It’s an intermittent failure where it sometimes fails and sometimes passes. The log output above shows the style has not been set correctly. But if you see the attached image below of where the video fails, the icon doesn’t have the style that’s in the error log.
video-fail

marklogic-data-hub-central/ui/src/pages/Modeling.test.tsx
```diff
@@ -85,6 +86,29 @@ describe("Modeling Page", () => {
     expect(getByLabelText("save-all")).toBeDisabled();
     expect(queryByLabelText('entity-modified-alert')).toBeNull();
   });
+
+  test("Modeling: can not see data if user does not have entity model reader role", async () => {
+    mockPrimaryEntityType.mockResolvedValueOnce({ status: 200, data: getEntityTypes });
+
+    const { queryByText, queryByLabelText } = render(
+      <AuthoritiesContext.Provider value={mockHCUserRolesService}>
```
bsrikan, line 94: This is ok. Another example is use an authority for ex readRole who doesnt have rights to model tile. If setting all the authority to false is same as a user with hub-central-user role, then this test looks good.
```
const authorityService = new AuthoritiesService();
authorityService.setAuthorities(['readLoad']);
```

PR #4405, opened 2020-08-14
DHFPROD-5723: Run build tasks in parallel

SameeraPriyathamTadikonda:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [N/A] Added Tests
  

- ##### Reviewer:

- [ N/A] Reviewed Tests

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -86,7 +86,7 @@ def dhfqsLinuxTests(String mlVersion,String type){
             cd $WORKSPACE/data-hub;
             rm -rf $GRADLE_USER_HOME/caches;
             ./gradlew clean;
-            ./gradlew build -x test -PnodeDistributionBaseUrl=http://node-mirror.eng.marklogic.com:8080/;
+            ./gradlew build -x test --parallel -PnodeDistributionBaseUrl=http://node-mirror.eng.marklogic.com:8080/;
             nohup ./gradlew bootRun >> $WORKSPACE/bootRun.out &
             sleep 60s;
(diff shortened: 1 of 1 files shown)


gradle.properties
```diff
@@ -3,3 +3,4 @@ publishUrl=file:../marklogic-data-hub/releases
 systemProp.file.encoding=utf-8
 org.gradle.jvmargs='-Dfile.encoding=UTF-8'
 nodeDistributionBaseUrl=http://nodejs.org/dist/
+org.gradle.parallel=true
```
rjrudin, line 6: How does this help? Reading https://guides.gradle.org/performance/#parallel_execution , I think we're in the category of projects that this won't help because the long pole is running the tests, and this won't have any impact on that. 

I'd rather not do this yet, as it'll impact every DHF engineer in ways I don't think we've thought through yet.

SameeraPriyathamTadikonda: It has no impact on running tests. But it improved the execution of clean and build tasks from about 11mins to 7mins. I can pass this as parameter instead while running in Jenkins. @rjrudin  Please let me know if we can proceed with that approach?

rjrudin: Let's pass it as a parameter for now, as I'm concerned about negative impacts it may have on developers locally.

SameeraPriyathamTadikonda: Yeah sure, I've tested both. Just wanted to do this as it improved the execution on my local from 10mins to 2mins.

I'll make that change

PR #4412, opened 2020-08-14
DHFPROD-5636: User without hub-central-saved-query-user can use Explorer and SavedQueries dropdown is not displayed on Zero state page

rahulvudutala:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/e2e/cypress/integration/explorer/scenarios.queries.tsx
+++ b/marklogic-data-hub-central/ui/e2e/cypress/integration/explorer/scenarios.queries.tsx
@@ -483,3 +483,44 @@ describe('manage queries modal scenarios on zero sate page, developer role', () 
 
 
 });
+
+describe('User without hub-central-saved-query-user role should not see saved queries drop down on zero sate page', () => {
+    beforeEach(() => {
+        cy.visit('/');
+        cy.contains(Application.title);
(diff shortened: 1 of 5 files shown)


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/entitySearch/getSavedQueries.sjs
```diff
@@ -15,7 +15,7 @@
  */
 'use strict';
 
-xdmp.securityAssert("http://marklogic.com/data-hub/hub-central/privileges/save-entity-query", "execute");
+xdmp.securityAssert("http://marklogic.com/data-hub/privileges/read-entity-model", "execute");
```
rjrudin, line 18: Why is this read-entity-model instead of save-entity-query? I think a comment would be helpful here, as it's unexpected that this isn't save-entity-query.

rahulvudutala: @rjrudin Reverted it back to normal. I misunderstood the scenario.

PR #4416, opened 2020-08-17
DHFPROD-5746: Migrating flows should be repeatable

fsnow:
### Description

In 5.3, FlowManagerImpl.getLocalFlow() does not know how to read referenced flows into Step objects. When you attempted to run migration twice, FlowMigrator would crash because of the mostly-empty Step object. (Specifically, trying to use the null result of step.getName().)

This PR is my attempt to make the smallest changes that will allow FlowMigrator.migrateFlows() to run twice.

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/hubcentral/migration/FlowMigrator.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/hubcentral/migration/FlowMigrator.java
@@ -149,6 +149,10 @@ public class FlowMigrator extends LoggingObject {
             ObjectNode newSteps = nodeFactory.objectNode();
             for (Map.Entry<String, Step> entry : steps.entrySet()) {
                 Step step = entry.getValue();
+                if (step.getStepId() != null) {
+                    newSteps.set(entry.getKey(), nodeFactory.objectNode().put("stepId", step.getStepId()));
+                    continue;
+                }
                 String stepId;
                 Path targetDir;
                 if (StepDefinitionType.INGESTION.equals(step.getStepDefinitionType())) {
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/step/impl/Step.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/step/impl/Step.java
@@ -17,6 +17,7 @@
 package com.marklogic.hub.step.impl;
 
 import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.node.ObjectNode;
@@ -28,6 +29,7 @@ import java.util.Map;
 
 @JsonInclude(JsonInclude.Include.NON_NULL)
 public class Step {
+    private String stepId;
     private String name;
     private String description;
     private Map<String, Object> options;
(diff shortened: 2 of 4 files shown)


marklogic-data-hub/src/main/java/com/marklogic/hub/hubcentral/migration/FlowMigrator.java
```diff
@@ -149,6 +149,10 @@ public void migrateFlows() {
             ObjectNode newSteps = nodeFactory.objectNode();
             for (Map.Entry<String, Step> entry : steps.entrySet()) {
                 Step step = entry.getValue();
+                if (step.getIsOnlyStepId()) {
```
rjrudin, line 152: While I'd normally say it's good to add methods like this to Step, I want to get away from classes like Step, which are incomplete representations of JSON objects. For the purpose of FlowMigrator, just check the original ObjectNode to see if it only has one key, and if the name of that key is stepId.

fsnow: There was no way to get to the JSON after getLocalFlow() executed, so I had to save a reference to the JsonNode in Step. (and hid it from serialization with an attribute)

marklogic-data-hub/src/main/java/com/marklogic/hub/hubcentral/migration/FlowMigrator.java
```diff
@@ -149,6 +149,11 @@ public void migrateFlows() {
             ObjectNode newSteps = nodeFactory.objectNode();
             for (Map.Entry<String, Step> entry : steps.entrySet()) {
                 Step step = entry.getValue();
+                JsonNode stepJson = step.getStepJson();
+                if (stepJson.has("stepId") && stepJson.size() == 1) {
```
rjrudin, line 153: Argh, sorry, I didn't realize the ObjectNode wasn't available. I think it's fine then to add stepId to the Step class. FlowMigrator can just say - if stepId is populated, then I don't need to migrate it. Then you don't need the stepJson JsonNode in Step.

fsnow: ok, I modified it to test for stepId not null.

marklogic-data-hub/src/main/java/com/marklogic/hub/hubcentral/migration/FlowMigrator.java
```diff
@@ -149,6 +149,10 @@ public void migrateFlows() {
             ObjectNode newSteps = nodeFactory.objectNode();
             for (Map.Entry<String, Step> entry : steps.entrySet()) {
                 Step step = entry.getValue();
+                if (step.getStepId() != null) {
+                    newSteps.set(entry.getKey(), nodeFactory.objectNode().put("stepId", step.getStepId()));
```
srinathgit, line 153: The flow document is migrated and hence steps will already have  "stepId" . So, do we have to set "stepId" again in the flow ?

rjrudin: I believe so, since "newSteps" is assigned to the flow.

srinathgit: But since the flow is already migrated, shouldn't it already have 'stepId' rather than inline steps ? so, we don't have to set 'stepId's in flow again right ?

fsnow: Yes, because it is building a new flow. You have to transfer the minimal {"1": "stepId"} to the new flow.

PR #4422, opened 2020-08-17
DHFPROD-5627:Add 'namespaces' to mapping step when xml docs are opened

srinathgit:
### Description
Add 'namespaces' to mapping step when xml docs are opened. No additional tests needed as we don't immediately save the updated mapping but wait for users to edit the mapping before saving it
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [n/a ] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/src/components/entities/mapping/mapping-card.tsx
+++ b/marklogic-data-hub-central/ui/src/components/entities/mapping/mapping-card.tsx
@@ -223,6 +223,10 @@ const MappingCard: React.FC<Props> = (props) => {
                 let sDta = generateNestedDataSource(docRoot,nestedDoc);
                 setSourceData([]);
                 setSourceData([...sDta]);
+                if(typeof(srcDocResp.data) === 'string'){
+                    let mData = await props.getMappingArtifactByMapName(props.entityModel.entityTypeId,props.data[index].name);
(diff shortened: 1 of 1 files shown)


marklogic-data-hub-central/ui/src/components/entities/mapping/mapping-card.tsx
```diff
@@ -241,13 +245,7 @@ const MappingCard: React.FC<Props> = (props) => {
     const updateMappingWithNamespaces = async (mapDataLocal) => {
```
rjrudin, line 245: @xnikhil08  @srinathgit  Does this support many namespaces in the source document? It looks like fetchSrcDocFromUri is parsing the value of "@xmlns"?? This really seems like something the backend should be doing. If this is just preserving QS functionality, I suppose it's fine for now. But can you create a ticket to rework this so that the backend does this properly, which include support namespaces at multiple levels of the document?

srinathgit: yeah, based on preliminary analysis, it does currently support multiple namespaces (in multiple levels) in the source doc.

PR #4424, opened 2020-08-17
DHFPROD-5695: Modals aren't closed when HC auto logs off

brucean52:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/src/pages/Run.tsx
+++ b/marklogic-data-hub-central/ui/src/pages/Run.tsx
@@ -51,6 +51,9 @@ const Run = (props) => {
         retryLimit: 10  // Timeout after retries
     }
 
+    useEffect(() => {
+        return () =>  Modal.destroyAll()
+    }, []);
 
     useEffect(() => {
         getFlows();


marklogic-data-hub-central/ui/src/pages/Run.tsx
```diff
@@ -51,6 +51,9 @@ const Run = (props) => {
         retryLimit: 10  // Timeout after retries
     }
 
+    useEffect(() => {
+        return () =>  Modal.destroyAll()
+    }, []);
```
bsrikan, line 56: any reason this is applied only to run.tsx. does curate/model/load tile modals get destroyed on log off?

brucean52: I tested the other tile modals, but I only saw the issue on the run page.

brucean52:
No tests to add.

PR #4425, opened 2020-08-18
DHFPROD-5372: Handle strings and null values correctly in mapping source

wooldridge:
Updates to source value display in mapping:

- Strings surrounded with double quotes
- Properties equal to null or "" are displayed in source table
- Properties equal to null or "" are displayed correctly in source dropdown in entity table

More details in the Jira ticket: https://project.marklogic.com/jira/browse/DHFPROD-5372

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/src/components/entities/mapping/mapping-card.tsx
+++ b/marklogic-data-hub-central/ui/src/components/entities/mapping/mapping-card.tsx
@@ -49,6 +49,7 @@ const MappingCard: React.FC<Props> = (props) => {
     const [mappingVisible, setMappingVisible] = useState(false);
     const [sourceData, setSourceData] = useState<any[]>([]);
     const [sourceURI,setSourceURI] = useState('');
+    const [sourceFormat,setSourceFormat] = useState('');
     const [sourceDatabaseName, setSourceDatabaseName] = useState('data-hub-STAGING')
     const [docNotFound, setDocNotFound] = useState(false);
     const [flowName, setFlowName] = useState('');
@@ -104,7 +105,6 @@ const MappingCard: React.FC<Props> = (props) => {
     const OpenMappingSettingsDialog = (index) => {
         setMapData(prevState => ({ ...prevState, ...props.data[index]}));
         setOpenMappingSettings(true);
-        console.log('Open settings')
     }
 
     //Custom CSS for source Format
@@ -188,7 +188,7 @@ const MappingCard: React.FC<Props> = (props) => {
         }
         catch(error)  {
             let message = error;
-            console.log('Error While loading the source data!', message);
+            console.error('Error While loading the source data!', message);
             setIsLoading(false);
             setDocNotFound(true);
         }
@@ -203,8 +203,10 @@ const MappingCard: React.FC<Props> = (props) => {
                 let parsedDoc: any;
                 if(typeof(srcDocResp.data) === 'string'){
                     parsedDoc = getParsedXMLDoc(srcDocResp);
+                    setSourceFormat('xml');
                 } else {
                     parsedDoc = srcDocResp.data;
+                    setSourceFormat('json');
                 }
                 if(parsedDoc['envelope']){
                     if(parsedDoc['envelope'].hasOwnProperty('@xmlns')){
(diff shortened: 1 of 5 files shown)


marklogic-data-hub-central/ui/src/components/entities/mapping/mapping-card.tsx
```diff
@@ -203,8 +203,10 @@ const MappingCard: React.FC<Props> = (props) => {
                 let parsedDoc: any;
                 if(typeof(srcDocResp.data) === 'string'){
                     parsedDoc = getParsedXMLDoc(srcDocResp);
+                    setSourceFormat('xml');
```
rjrudin, line 206: This is fine, but the backend should really provide all of this data - for 5.4, we've gotta do a better job of shifting more of the burden to the backend so that life is as easy as possible for the UI.

marklogic-data-hub-central/ui/src/components/entities/mapping/mapping-card.tsx
```diff
@@ -368,7 +378,7 @@ const MappingCard: React.FC<Props> = (props) => {
                     if(parentNamespace !== tempNS){
                         parentNamespace = tempNS;
                     }
-                } else if (val.constructor.name === "Array") {
+                } else if (val && val.constructor && val.constructor.name === "Array") {
```
jmakeig, line 381: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray

marklogic-data-hub-central/ui/src/components/entities/mapping/source-entity-map/source-to-entity-map.scss
```diff
@@ -7,3 +7,14 @@
 .ant-table-filter-icon.ant-dropdown-trigger.ant-table-filter-selected{
     background: #394494; 
 }
+
+.datatype-string::before {
```
jmakeig, line 11: 👍🏽

marklogic-data-hub-central/ui/src/components/entities/mapping/source-entity-map/source-to-entity-map.test.tsx
```diff
@@ -29,14 +29,25 @@ describe('RTL Source-to-entity map tests', () => {
...
         expect(getByTestId("srcContainer")).toBeInTheDocument();
         expect(getByTestId("srcContainer")).toHaveClass("sourceContainer");
         expect(getByText('Entity: Person')).toBeInTheDocument();
         expect(getByText('Test')).toBeEnabled();
+
+        // Check datatype class names for source values
+        expect(getByTestId("emptyString-srcValue").children[0].className.includes('datatype-string')).toBe(true);
+        expect(getByTestId("nullValue-srcValue").children[0].className.includes('datatype-null')).toBe(true);
+        expect(getByTestId("numberValue-srcValue").children[0].className.includes('datatype-number')).toBe(true);
+        expect(getByTestId("booleanValue-srcValue").children[0].className.includes('datatype-boolean')).toBe(true);
```
bsrikan, line 49: @wooldridge can we pls also add a test for white space here.

marklogic-data-hub-central/ui/src/components/entities/mapping/mapping-card.tsx
wooldridge: Agreed, it would be better to have the backend tell the front end the source format explicitly.

marklogic-data-hub-central/ui/src/components/entities/mapping/mapping-card.tsx
wooldridge: Thanks, updated with latest force-push.

marklogic-data-hub-central/ui/src/components/entities/mapping/source-entity-map/source-to-entity-map.test.tsx
wooldridge: Thanks, added test for whitespace with latest force-push.

wooldridge:
@briantang, thanks for pointing this out. When XML source has an element like this (with whitespace):

`<Whitespace>               </Whitespace>`

The whitespace is getting dropped during the parsing step here:

https://github.com/marklogic/marklogic-data-hub/blob/develop/marklogic-data-hub-central/ui/src/components/entities/mapping/mapping-card.tsx#L240

And the converted JSON shows this:

`Whitespace: ""`

So this is not a UI formatting issue, which is what I'm addressing, but an XML parsing issue. Can you file that as a separate bug and we can take a look in 5.4?

PR #4426, opened 2020-08-18
DHFPROD-5728: Sort entities based on datatype-specific index

rahulvudutala:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/hub-entities.xqy
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/hub-entities.xqy
@@ -255,14 +255,13 @@ declare %private function hent:build-sort-operator(
 {
   let $states :=
     for $path-expression in map:keys($sortable-properties)
-    let $state-name-prefix :=
-      let $sort-info := map:get($sortable-properties, $path-expression)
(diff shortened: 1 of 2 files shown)


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/hub-entities.xqy
```diff
@@ -255,14 +255,13 @@ declare %private function hent:build-sort-operator(
 {
   let $states :=
     for $path-expression in map:keys($sortable-properties)
-    let $state-name-prefix :=
-      let $sort-info := map:get($sortable-properties, $path-expression)
-      return fn:concat(map:get($sort-info, "entity-title"), "_", map:get($sort-info, "property-name"))
-
+    let $sort-info := map:get($sortable-properties, $path-expression)
+    let $propert-data-type := hent:get-indexable-datatype(map:get($sort-info, "property-data-type"))
```
rjrudin, line 259: Got a typo here - should be "property" - but let's call this "indexable-datatype", because I think "property-datatype" implies that it's the datatype as defined in the property definition. Also, use "datatype" throughout for consistency instead of "data-type".

PR #4435, opened 2020-08-18
DHFPROD-5762: Adding sources object in Entity Headers to search result

rahulvudutala:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/entities/entity-search-lib.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/entities/entity-search-lib.sjs
@@ -17,6 +17,7 @@
 
 const ds = require("/data-hub/5/data-services/ds-utils.sjs");
 const es = require('/MarkLogic/entity-services/entity-services');
+const esInstance = require('/MarkLogic/entity-services/entity-services-instance');
 
 // TODO Will move this to /data-hub/5/entities soon
 const entityLib = require("/data-hub/5/impl/entity-lib.sjs");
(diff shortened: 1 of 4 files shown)


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/entities/entity-search-lib.sjs
```diff
@@ -263,12 +264,28 @@ function getEntityInstance(docUri) {
 }
 
 function getEntityInstanceHeaders(docUri) {
-    let doc = cts.doc(docUri);
+  const doc = cts.doc(docUri);
```
rjrudin, line 267: What's the actual requirement here? The JIRA ticket mentions that "headers are not extracted", and this function is called "getEntityInstanceHeaders", but the code is specifically interested in the sources part of headers. 

What is the UI looking to do with the sources? 

If the intent is to provide an array of sources for an entity instance, I think this function should be called "getEntitySources" (I think the word 'instance' adds a little confusion because 'instance' and 'headers' are two separate parts of the envelope).

rahulvudutala: Sources are displayed in the instance and snippet view in the Explorer tile

Updated the function name to getEntitySources as suggested

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/entities/entity-search-lib.sjs
```diff
@@ -370,7 +387,7 @@ function addEntitySpecificProperties(result, entityInfo, selectedPropertyMetadat
     console.log(`Unable to obtain document with URI '${result.uri}'; will not add document metadata to its search result`);
   }
   result.entityInstance = entityInstance;
-  result.sources = getEntityInstanceHeaders(result.uri) ? getEntityInstanceHeaders(result.uri) : result.sources;
+  result.sources = getEntitySources(result.uri) ? getEntitySources(result.uri) : result.sources;
```
rjrudin, line 390: You can remove the "?" operator here because getEntitySources always returns an array, right?

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/entities/entity-search-lib.sjs
```diff
@@ -423,7 +440,7 @@ function addGenericEntityProperties(result) {
     "propertyValue": identifierValue
   };
   result.entityInstance = entityInstance;
-  result.sources = getEntityInstanceHeaders(result.uri) ? getEntityInstanceHeaders(result.uri) : result.sources;
+  result.sources = getEntitySources(result.uri) ? getEntitySources(result.uri) : result.sources;
```
rjrudin, line 443: Same here, don't need the "?".

PR #4440, opened 2020-08-19
DHFPROD-5735:Enable writing temporal documents

srinathgit:
### Description
Enable writing temporal documents
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [ ] Added Tests
  

- ##### Reviewer:

- [ ] Reviewed Tests


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/temporal/hub-temporal.sjs
```diff
@@ -0,0 +1,27 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+'use strict';
+const temporal = require("/MarkLogic/temporal.xqy");
+
+module.exports.getTemporalCollections = module.amp(
+  function getTemporalCollections() {
```
rjrudin, line 20: I think it'd be better for this to just call temporal.collections(). Returning an object with each collection as the key and a value of "true" is very specific to what invoke-queue-write wants. invoke-queue-write should be responsible for doing that; this should just call temporal.collections() and return the result.

marklogic-data-hub/src/test/java/com/marklogic/hub/flow/TemporalWriteTest.java
```diff
@@ -0,0 +1,128 @@
...
+
+import static com.marklogic.client.io.DocumentMetadataHandle.Capability.*;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class TemporalWriteTest extends AbstractHubCoreTest {
+
+    @BeforeEach
+    void setUp() {
+        runAsDataHubDeveloper();
+        installProjectInFolder("temporal-test");
```
rjrudin, line 28: Can you move the new project files to test-projects/temporal-test? I'd eventually like for all of the test projects to be in that folder so that we don't have so many folders directly under src/test/resources.

marklogic-data-hub/src/test/resources/temporal-test/step-definitions/custom/temporal-test/temporal-test.step.json
```diff
@@ -0,0 +1,9 @@
+{
+  "lang": "zxx",
+  "name": "temporal-test",
+  "type": "CUSTOM",
+  "options": {
+    "outputFormat": "json"
+  },
+  "modulePath": "/custom-modules/custom/temporal-step/main.sjs"
```
rjrudin, line 8: This is not related, but pinging @fsnow and @ryanjdew and @bsrikan about this - I've wondered if step definitions should go away. They certainly seem like unnecessary work when making a custom step. The only thing of interest here is the modulePath, and if I'm making a custom step, it's easier to make one new file - the custom step with a modulePath in it - vs making two new files. 

Perhaps step definitions don't need to go away, but should just become optional. They are a useful way to refer to OOTB step modules... though arguably, just using the OOTB step modulePath would be just as easy. The indirection added by having a step definition name and type just doesn't seem to buy much. The odds of the module path changing seem as low as the odds of the step definition name/type changing, so tossing the OOTB module path into a step seems fine.

srinathgit:
Just changed the path of "hub-temporal.sjs" from relative to absolute, unit tests run fine locally after the change

PR #4449, opened 2020-08-19
DHFPROD-5428: Reset URI nav in source mapping

wooldridge:
Reset URI index to 0 for source doc navigation when the mapping modal is closed. Next mapping modal opened will start at the first doc.

Also remove unused onOk() code for modal.

Screencast: https://drive.google.com/file/d/10KCpDVYJxj4Eso9PdbRnVYr-ud_hKflu/view

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [ ] Added Tests
  

- ##### Reviewer:

- [ ] Reviewed Tests


marklogic-data-hub-central/ui/src/components/entities/mapping/source-entity-map/source-to-entity-map.test.tsx
```diff
@@ -1084,3 +1084,29 @@ describe('RTL Source Selector/Source Search tests', () => {
...
+        let uriNext = within(getByLabelText('uriIndex'));
+        expect(uriNext.getByText('2')).toBeInTheDocument();
+
+        // Close mapping modal
+        fireEvent.click(getByLabelText('icon: close'));
+
+        // URI index reset to 1
+        let uriReset = within(getByLabelText('uriIndex'));
+        expect(uriReset.getByText('1')).toBeInTheDocument();
+    });
```
bsrikan, line 1109: This is a good test. Couple of nitpicky items:
1. Do we need a separate test suite? We tried to test URI navigation and verify source table data, didnt work out at the time. If we are able to assert doc content on navigation, then we can keep this in a separate suite and add more tests later to assert source data.
2. uriStart, uriNext, uriReset are really aliases to the same locator?

PR #4451, opened 2020-08-20
DHFPROD-5771: Stabilize e2e and handle async XHR requests

bsrikan:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/e2e/cypress/support/pages/browse.tsx
+++ b/marklogic-data-hub-central/ui/e2e/cypress/support/pages/browse.tsx
@@ -22,6 +22,7 @@ class BrowsePage {
     this.waitForSpinnerToDisappear();
     cy.get('#entity-select').click();
     cy.get(`[data-cy="entity-option-${entity}"]`).click();
+    cy.waitForAsyncRequest();
     this.waitForSpinnerToDisappear();
   }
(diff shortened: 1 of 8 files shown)


marklogic-data-hub-central/ui/e2e/cypress/support/pages/browse.tsx
```diff
@@ -185,7 +186,7 @@ class BrowsePage {
   }
 
   getHubPropertiesExpanded() {
-    return cy.get("#hub-properties > div").eq(1).invoke('show').click();
```
Sanjeevani19, line 188: @bsrikan I remember adding this line fixed some of the scenarios.json and xml e2e tests because some times it's hard to find the expand icon while the page is still loading and the test fails on click, that's the reason we thought of clicking on hub properties text. Any reason for this change ?

bsrikan: Added info: I could have changed it to use data-testid="toggle", but there is a duplicate data-testid for entity properties, which should be fixed, since locator id should be unique.

PR #4452, opened 2020-08-20
DHFPROD-5182: Transform match steps to new model

fsnow:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/test/java/com/marklogic/hub/hubcentral/conversion/AllStepsReferencedTest.java
+++ b/marklogic-data-hub/src/test/java/com/marklogic/hub/hubcentral/conversion/AllStepsReferencedTest.java
@@ -1,9 +1,7 @@
 package com.marklogic.hub.hubcentral.conversion;
 
 import com.fasterxml.jackson.databind.JsonNode;
-import com.marklogic.hub.AbstractHubCoreTest;
-import com.marklogic.hub.HubConfig;
-import com.marklogic.hub.HubProject;
+import com.marklogic.hub.*;
 import com.marklogic.hub.flow.FlowInputs;
 import com.marklogic.hub.flow.FlowRunner;
 import com.marklogic.hub.flow.RunFlowResponse;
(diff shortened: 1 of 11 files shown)


marklogic-data-hub/src/main/java/com/marklogic/hub/StepManager.java
```diff
@@ -0,0 +1,81 @@
+package com.marklogic.hub;
```
rjrudin, line 1: Can you toss this into the com.marklogic.hub.impl package? I think there was an intent for com.marklogic.hub to define public APIs that would be well-documented for use, but I don't think that has materialized. Because we're not making that a priority, I want to avoid the perception of creating public APIs, and I think we can thus "hide" stuff in the impl package to avoid this.

marklogic-data-hub/src/test/java/com/marklogic/hub/hubcentral/conversion/AllStepsReferencedTest.java
```diff
@@ -12,17 +11,16 @@
...
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Iterator;
+import java.util.*;
 
 import static org.junit.jupiter.api.Assertions.*;
 
 class AllStepsReferencedTest extends AbstractHubCoreTest {
 
+    ObjectMapper mapper = new ObjectMapper();
```
rjrudin, line 22: There's an objectMapper already on each instance of AbstractHubCoreTest (inherited from TestObject).

rjrudin (dismissed):
Looks good to me, though I just skimmed over the actual update-match code as I'm not yet familiar with the old and new option schemas.

marklogic-data-hub/src/test/java/com/marklogic/hub/hubcentral/conversion/AllStepsReferencedTest.java
fsnow: ok, it wasn't used so I removed it.

marklogic-data-hub/src/main/java/com/marklogic/hub/StepManager.java
fsnow: done

PR #4458, opened 2020-08-20
DHFPROD-5356: Fields in source selector are not visible in case of de…

saarimrahman:
…eply nested source documents

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


saarimrahman:
> I can still see the ellipsis in chrome. Although it looks like the field is not truncated, just that the ellipsis is visible.

@xnikhil08 Thanks for noticing this. I've pushed a fix. It should display uniformly across browsers now.

akshaysonvane:
If the tests have been reviewed can one of the reviewers check the "Reviewed Tests" checkbox?

PR #4461, opened 2020-08-21
DHFPROD-4663: Added test for sourceQueryIsScript

rjrudin:
### Description

The test verifies that content.uri can be converted into an array. Also converted flow-runner-test into a normal test project so we don't need special code for it any longer. 

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


marklogic-data-hub/src/test/resources/flow-runner-test/flows/testValuesFlow.flow.json
```diff
@@ -10,6 +10,16 @@
         "targetDatabase": "data-hub-FINAL",
         "sourceQueryIsScript": true
       }
+    },
+    "2": {
+      "name": "expectsArrayStep",
+      "stepDefinitionName": "expectArray",
+      "stepDefinitionType": "CUSTOM",
+      "options": {
```
srinathgit, line 18: Should there be a "sourceQuery" option here with a source query ?

rjrudin: It's added by the test - the test uses different values for sourceQuery. Which ends up verifying that sourceQuery can be specified via runtime options too.

PR #4474, opened 2020-08-24
DHFPROD-5183: Transform merge steps to new model

fsnow:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/hubcentral/conversion/FlowConverter.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/hubcentral/conversion/FlowConverter.java
@@ -165,7 +165,13 @@ public class FlowConverter extends LoggingObject {
                         stepManager.saveLocalStep(refStepNode);
                         int i = 0;
                     }
-
+                    // Need to convert merge steps to new model, if they have not been already
+                    else if (stepDefType.equals(StepDefinitionType.MERGING) && !stepNode.has("mergeRules")) {
+                        ObjectNode refStepNode = stepManager.getLocalStepAsJSON(stepId);
+                        transformMergingOptions(refStepNode);
+                        stepManager.saveLocalStep(refStepNode);
+                        int i = 0;
+                    }
                     continue;
                 }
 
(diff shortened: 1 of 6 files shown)


marklogic-data-hub/src/main/java/com/marklogic/hub/hubcentral/conversion/FlowConverter.java
```diff
@@ -514,4 +524,23 @@ protected void transformMatchingOptions(ObjectNode stepNode) {
...
     }
+
+    protected void transformMergingOptions(ObjectNode stepNode) {
+        JsonNode beforeNode = stepNode.get("mergeOptions");
+        // if mergeOptions is null, this conversion has already run
+        if (beforeNode == null) {
+            return;
+        }
+
+        JsonNode afterNode = MasteringService.on(hubConfig.newHubClient().getStagingClient()).updateMergeOptions(beforeNode);
```
rjrudin, line 535: No need to change this, but just an FYI, pinging @srinathgit  and @ryanjdew  on this too - one thing I'd like to do, particularly in our tests, is get a HubClient as early as possible and use that for all connections. The benefit is that it doesn't create a new DatabaseClient each time the way HubConfig does when you call one of its new*Client methods. That also means less and less code that depends on HubConfig - I'd really like our code to depend on HubClient and (if needed) HubProject, with HubConfig becoming something that's discarded once those two things are obtained. But again, no need to change this, just a note.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/mastering/updateMergeOptionsLib.sjs
```diff
@@ -0,0 +1,209 @@
+'use strict';
+
+function updateMergeOptions(opt)
```
rjrudin, line 3: Can you add unit tests for these functions?

fsnow: Oops, misplaced that file and found it in an auto-stash. Added in latest push.

PR #4479, opened 2020-08-25
DHFPROD-5783:String facets on the search popover don't indicate the current selection state.

Sanjeevani19:
### Description
Also includes fix for failing scenarios.json test in the pipeline

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/e2e/cypress/support/pages/browse.tsx
+++ b/marklogic-data-hub-central/ui/e2e/cypress/support/pages/browse.tsx
@@ -116,6 +116,11 @@ class BrowsePage {
     return cy.get('[data-cy="clear-' + facet + '"]');
   }
 
+  clickClearFacetSearchSelection(facet: string){
+      cy.findByTestId(`clear-${facet}`).click();
+      this.waitForSpinnerToDisappear();
+  }
+
   getFacetSearchSelectionCount(facet: string) {
     return cy.get('[data-cy="' + facet + '-selected-count"]').invoke('text');
   }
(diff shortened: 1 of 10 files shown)


marklogic-data-hub-central/ui/e2e/cypress/support/pages/browse.tsx
```diff
@@ -165,6 +170,19 @@ class BrowsePage {
    return cy.get('#selected-facets [data-cy="clear-'+ lowerBound +'"]');
   }
 
+  clickPopoverSearch(facetName: string){
+      return cy.get(`[data-testid=${facetName}-search-input]`).click();
```
bsrikan, line 174: we can remove return here and in the locator below. No need to return when action is performed in the locator itself. We only "return" an element locator.

marklogic-data-hub-central/ui/e2e/cypress/support/pages/browse.tsx
```diff
@@ -165,6 +170,19 @@ class BrowsePage {
...
+  }
+
+  getInputField(facetName: string,str: string){
+   return cy.get(`[data-testid=${facetName}-popover-input-field]`).type(str);
```
bsrikan, line 178: We are setting the input field, so its better to have the method name reflect that. Lets call it setInputField.
can we use cy.findByTestId here and in the locator below.

Will the input field always be empty. Should we clear first before typing?

PR #4492, opened 2020-08-27
DHFPROD-5803: The save query icon should not persist when switching between entities. 

timur-isangulov:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


timur-isangulov:
> @timur-isangulov switching between entities popup shows up even though there are no changes on the screen.

Thanks for noticing, it happened for the gray facets, I fixed it. Can you please review again.

PR #4495, opened 2020-08-27
DHFPROD-5826: Array of booleans and numbers not being handled in Mapp…

briantang:
…ing source data

### Description

- Update to Cust1.json in reference-entity-model project to accommodate edge cases
- Fixed issue with rendering of source data for cases where value is an array of booleans or array of numbers.

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/src/assets/mock-data/common.data.ts
+++ b/marklogic-data-hub-central/ui/src/assets/mock-data/common.data.ts
@@ -181,6 +181,8 @@ const jsonSourceData = [
   { rowKey: 12, key: 'booleanValue', val: 'true', datatype: 'boolean' },
   { rowKey: 13, key: 'whitespaceValue', val: ' ', datatype: 'string' },
   { rowKey: 14, key: 'emptyArrayValue', val: '[ ]', datatype: 'object' },
+  { rowKey: 15, key: 'numberArray', val: '1, 2, 3', array: true, datatype: 'number' },
+  { rowKey: 16, key: 'booleanArray', val: 'true, false, true', array: true, datatype: 'boolean' },
 ];
 
 const entityTypeProperties = [
--- a/marklogic-data-hub-central/ui/src/components/entities/mapping/mapping-card.tsx
+++ b/marklogic-data-hub-central/ui/src/components/entities/mapping/mapping-card.tsx
@@ -389,16 +389,16 @@ const MappingCard: React.FC<Props> = (props) => {
                         };
                         nestedDoc.push(propty);
                     }
-                    else if (val[0].constructor.name === "String") {
-                            let stringValues = val.join(', ')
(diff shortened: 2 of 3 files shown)


examples/reference-entity-model/input/json/Cust1.json
```diff
@@ -40,5 +40,22 @@
   "PIN": 2172,
   "Created": "04/03/2017",
   "Status": "New Customer",
-  "SSN": "111-11-1111"
+  "SSN": "111-11-1111",
+  "numbersArray" : [
```
srinathgit, line 44: I think it would be better to have UI tests for these values rather than adding it to reference project input data. These new data don't have relevance to "Customer" data

briantang: Thanks, @srinathgit  that's fair. I initially checked with @bsrikan beforehand so I'm curious what he thinks? I could also make it a new input file unrelated to "Customer" data.

srinathgit: I think we would like to keep reference project as clean as possible. @bsrikan , let us know your thoughts

bsrikan: When we were reviewing the test plan we wanted any enhanced test data to be added to reference project, because it is pushed to HC nightly and most of our engineers and PM use it for reviewing/testing PR.

bsrikan: As for tests, Brian has added unit tests for the fix.

wooldridge: I agree with @srinathgit, the example reference project should be more like a demo with (Customer) data that ties together (although it is also intended to cover many data use cases).

briantang: Looks like we're leaning towards taking out this enhancement to test data with certain edge cases to keep the 'Customer' data clean. @rjrudin could you confirm? And I can go ahead and take that out of the PR.

wooldridge: @briantang you did add more test data to common.data.ts to cover these cases and I think that's fine. Let's pull those examples out of the reference project for now and push this PR through. I can't see @rjrudin having a problem with this. :-)

briantang: Thanks @wooldridge, will leave it out for now so we can get this bug fix through.

rjrudin: The key to me is that the entity model isn't being used to determine how data in the source panel is displayed - that's not the case, right?

wooldridge: @rjrudin that is not the case, display of source data is based on the source data.

PR #4511, opened 2020-09-01
DHFPROD-5794: Speeding up saving models

rjrudin:
### Description

Added saveModels.sjs endpoint that saves models to one database only to avoid the combo of xdmp.invoke plus pre/post triggers, which causes performance delays. 

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


rjrudin:
> So, the artifact causing bottleneck is only entity models ? Rest all good as is ?

Yes, you can run the LoadUserArtifactsPerformanceTest to see the times. I thought mapping steps would require special handling, as they have a pre-commit trigger. But it appears to be specific to entity models - a presumed combo of xdmp.invoke, plus pre and post commit triggers. The project you attached to the JIRA ticket now loads all artifacts in a few seconds total.

rjrudin:
Added declareUpdate to a few endpoints to fix failing tests

akshaysonvane:
We still need to add declareUpdate to deleteModel.sjs to fix the failing test.

rjrudin:
Fixed deleteModel test

fsnow (approved):

PR #4518, opened 2020-09-02
DHFPROD-5851: Fix mapping tool namespacing edge case

ryanjdew:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/src/api/__mocks__/mocks.data.ts
+++ b/marklogic-data-hub-central/ui/src/api/__mocks__/mocks.data.ts
@@ -46,6 +46,7 @@ const curateAPI = (axiosMock) => {
       case '/api/steps/mapping/' + curateData.mappings.data[0].artifacts[0].name:
         return Promise.resolve(loadData.genericSuccess);
       default:
+        console.log('no DELETE defined: ' + url);
         return Promise.reject(new Error('not found'))
     }
   });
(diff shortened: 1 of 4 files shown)


xnikhil08:
> image
> 
> Here, OrderDetails should get **OrderNS** namespace from its parent node, but getting **OD**. This will load incorrect value in the xpath expression field when selected.
> 
> I am using attached file to test it. If the namespace is declared with a prefix to xmlns then the node is assuming the same namespace, even if the prefix is not used in its nodeName.
> 
> [10289.xml.zip](https://github.com/marklogic/marklogic-data-hub/files/5163850/10289.xml.zip)

@ryanjdew - Looks like it was a cache issue for me. I reset my environment and getting **OrderNS** only now for OrderDetails node. 
But all the child nodes of **p:OrderDetail**, except UnitPrice, are getting wrong namespaces as mentioned by @srinathgit . The other OrderDetail node and its children seems to be fine.

ryanjdew:
@xnikhil08 @srinathgit Missed passing the default namespace down recursively in some cases. I'm getting the expected namespaces set for the lower-level elements in the attached XML doc.

xnikhil08 (approved):
Looks good!

wooldridge (approved):
Looks good:

Screen Shot 2020-09-02 at 12 09 39 PM

srinathgit (approved):
Looks good

PR #4522, opened 2020-09-02
DHFPROD-5239: Creating a new ingestion step may result in it being in…

briantang:
… second page of table results

### Description
- Fixed all column sorters in Load List Table view as none were working properly, see my comment in https://project.marklogic.com/jira/browse/DHFPROD-5239 for more details.
- Load List Table now default sorts by Last Updated so newly created steps are seen at the top and do not disappear in later pages.
- Added tests that were missing for all sorting options in Load List component

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/src/util/test-utils.tsx
+++ b/marklogic-data-hub-central/ui/src/util/test-utils.tsx
@@ -34,7 +34,7 @@ const onClosestTableBody:any = command => command.closest('tbody');
 const onClosestTable:any = command => command.closest('table');
 const onClosestDiv:any = command => command.closest('div');
 
-const validateExplorerResultsTableRow = (dataTable, uris) => {
+const validateTableRow = (dataTable, uris) => {
     let rowKey = 0;
     
     dataTable.forEach(item => {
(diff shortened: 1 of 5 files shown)


wooldridge:
@rjrudin, yeah, we're NOT letting the backend handle sorting for this:

https://github.com/marklogic/marklogic-data-hub/blob/develop/marklogic-data-hub-central/ui/src/pages/Load.tsx#L60

I think Explorer results is the only place where we leverage ML's sorting. With the table components offering column sort features OOTB, I think we're fine using that to handle sorting for result sets that are small. such as those on the Curation side, as you say.

wooldridge (approved):
Sorting looks good, checked on Mac and Win, tests pass.

marklogic-data-hub-central/ui/src/util/test-utils.tsx
```diff
@@ -34,7 +34,7 @@ const onClosestTableBody:any = command => command.closest('tbody');
 const onClosestTable:any = command => command.closest('table');
 const onClosestDiv:any = command => command.closest('div');
 
-const validateExplorerResultsTableRow = (dataTable, uris) => {
+const validateTableRow = (dataTable, uris) => {
```
bsrikan, line 37: Nice!

bsrikan (approved):

PR #4526, opened 2020-09-03
DHFPROD-5877:Write flow artifact created by 'hubCreateFlow' to db, add better logging

srinathgit:
## Description
Write flow artifact created by 'hubCreateFlow' to db, add better logging

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/impl/ScaffoldingImpl.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/impl/ScaffoldingImpl.java
@@ -145,7 +145,7 @@ public class ScaffoldingImpl implements Scaffolding {
                 + "/custom-modules/" + stepType.toLowerCase() + "/" + stepDefName + "/main.sjs" + ". \n");
             messageBuilder.append("It is recommended to run './gradlew -i mlWatch' so that as you modify the module, it will be automatically loaded into your application's modules database.\n");
         }
-        messageBuilder.append("Created step '" + stepName + "' of type '" + stepType + "' with default properties. It will need to be modified before usage.");
+        messageBuilder.append("Created step '" + stepName + "' of type '" + stepType + "' with default properties. The step has been deployed to staging and final databases.");
         DatabaseClient stagingClient = hubConfig.newHubClient().getStagingClient();
         try {
             if(stepDefinition != null) {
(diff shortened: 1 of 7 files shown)


bsrikan (approved):

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/ScaffoldingImpl.java
```diff
@@ -145,7 +145,8 @@ public static String getAbsolutePath(String first, String... more) {
                 + "/custom-modules/" + stepType.toLowerCase() + "/" + stepDefName + "/main.sjs" + ". \n");
             messageBuilder.append("It is recommended to run './gradlew -i mlWatch' so that as you modify the module, it will be automatically loaded into your application's modules database.\n");
         }
-        messageBuilder.append("Created step '" + stepName + "' of type '" + stepType + "' with default properties. It will need to be modified before usage.");
+        messageBuilder.append("Created step '" + stepName + "' of type '" + stepType + "' with default properties. The step has been deployed to staging and final databases. " +
```
rjrudin, line 148: Do we know for sure that the step needs to be modified? I don't think that's true in all cases. Like I can see a custom step working as-is, though of course the user would need to modify the custom step module. I would remove this line of logging.

ml-data-hub-plugin/src/main/groovy/com/marklogic/gradle/DataHubPlugin.groovy
```diff
@@ -143,14 +143,14 @@ class DataHubPlugin implements Plugin<Project> {
                 "a type (either 'ingestion' or 'custom'; defaults to 'custom') via -PstepDefType=ingestion|custom, " +
                 "and a format (either 'sjs' or 'xqy'; defaults to 'sjs') via -Pformat=sjs|xqy")
         project.task("hubCreateStep", group: developGroup, type: CreateStepTask,
-            description: "Create a new step file in staging,final databases and write it to your project; specify a step name via -PstepName=YourStepName, and specify " +
-                "a step type via -PstepType=(ingestion|mapping|custom)")
+            description: "Create a new step file and write it to staging,final databases and to your project; specify a step name via -PstepName=YourStepName, " +
```
rjrudin, line 146: For consistency, say "write it to the staging and final databases"

ml-data-hub-plugin/src/main/groovy/com/marklogic/gradle/task/CreateFlowTask.groovy
```diff
@@ -40,9 +40,20 @@ class CreateFlowTask extends HubTask {
...
             println "Created new flow at: " + file.getAbsolutePath()
         }
+        ObjectMapper mapper = new ObjectMapper()
+        try{
+            ArtifactService service = ArtifactService.on(hubConfig.newStagingClient(null))
+            service.setArtifact("flow", flowName, mapper.readTree(file))
+        }
+        catch (Exception e){
+            println "Unable to write flow to database;cause: " + e.getMessage()
+            throw e
```
rjrudin, line 54: I am thinking we should just print the error and not throw it. The error will likely only occur for one reason - the user cannot connect to ML. And the message will make that obvious. The task can still be considered to have succeeded in that a new flow was created.

rjrudin (changes requested):

bsrikan (approved):

rjrudin (approved):

PR #4534, opened 2020-09-05
DHFPROD-5888: Handling null values in user preferences for saved queries

xnikhil08:
### Description
This PR fixes the bug DHFPROD-5888. e2e tests are already present for saved queries.
Also updated few e2e tests to fix certain intermittent failures.

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [N/A] Added Tests
  

- ##### Reviewer:

- [ N/A] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/src/components/queries/queries.tsx
+++ b/marklogic-data-hub-central/ui/src/components/queries/queries.tsx
@@ -148,7 +148,7 @@ const Query: React.FC<Props> = (props) => {
                 (currentQuery.savedQuery.query.searchText !== searchOptions.query) ||
                 (JSON.stringify(currentQuery.savedQuery.sortOrder) !== JSON.stringify(searchOptions.sortOrder)) ||
                 (JSON.stringify(currentQuery.savedQuery.propertiesToDisplay) !== JSON.stringify(searchOptions.selectedTableProperties)) ||
-                (props.greyFacets.length > 0) || props.isColumnSelectorTouched) && 
+                (props.greyFacets.length > 0) || props.isColumnSelectorTouched) &&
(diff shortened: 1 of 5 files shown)


rjrudin (approved):

brucean52 (approved):

marklogic-data-hub-central/ui/src/components/queries/queries.tsx
```diff
@@ -222,6 +224,7 @@ const Query: React.FC<Props> = (props) => {
     }
 
     const onNoClick  = () => {
+        toggleEntityConfirmation(false);
```
bsrikan, line 227: @xnikhil08 @brucean52 @Sanjeevani19 @timur-isangulov  I think we should close the modal first and then apply current query on entity change.
Earlier the modal was closed after setting the query, hence modal was hiding the spinner and hub properties was getting clicked on prematurely. So the fix was required in the code than to the test. This fixed scenario json/xml and queries test that were failing intermittently.
Another thing to keep in mind is careful usage of {force: true}. The action is performed if the element is available in the DOM and doesnt wait for animations and this could result in unnecessary failures. Tests ran ok 4 times successively. Hopefully it does so in Jenkins too.

Additionally pls review if this move to call toggleEntityConfirmation(false) within onNoClick wont impact the places where setCurrentQueryOnEntityChange() gets called.

Sanjeevani19: Verified scenarios for adding the toggleEntityConfirmation(false). Looks good to me, e2e tests passed too.

Sanjeevani19 (approved):

brucean52 (approved):

PR #4536, opened 2020-09-08
DHFPROD-5870: Fix namespaced path range indexes on export

ryanjdew:
### Description

In the process of adding tests.

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/entities/search/EntitySearchManager.java
+++ b/marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/entities/search/EntitySearchManager.java
@@ -34,7 +34,6 @@ import com.marklogic.client.io.StringHandle;
 import com.marklogic.client.query.QueryManager;
 import com.marklogic.client.query.StructuredQueryBuilder;
 import com.marklogic.client.query.StructuredQueryDefinition;
-import com.marklogic.client.row.RowManager;
 import com.marklogic.hub.HubClient;
 import com.marklogic.hub.central.entities.search.impl.CollectionFacetHandler;
 import com.marklogic.hub.central.entities.search.impl.CreatedOnFacetHandler;
@@ -50,10 +49,13 @@ import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.springframework.util.FileCopyUtils;
 
 import javax.servlet.http.HttpServletResponse;
 import java.io.IOException;
 import java.io.OutputStream;
+import java.io.OutputStreamWriter;
+import java.io.Reader;
 import java.time.ZonedDateTime;
 import java.time.format.DateTimeFormatter;
 import java.util.ArrayList;
@@ -61,6 +63,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.stream.Stream;
 
 public class EntitySearchManager {
 
@@ -246,17 +249,13 @@ public class EntitySearchManager {
         List<SearchQuery.SortOrder> sortOrder = searchQuery.getSortOrder().orElse(new ArrayList<>());
         ArrayNode sortOrderNode = sortOrderToArrayNode(sortOrder);
 
-
-        JsonNode opticPlanNode = EntitySearchService.on(finalDatabaseClient).getOpticPlan(structuredQuery, searchText, queryOptions, entityTypeId, entityTypeId, limit, sortOrderNode, columns.stream());
-        StringHandle stringHandle = new StringHandle(opticPlanNode.toString());
-        RowManager rowManager = finalDatabaseClient.newRowManager();
-        try (ReaderHandle readerHandle = new ReaderHandle()) {
-            rowManager.resultDoc(rowManager.newRawPlanDefinition(stringHandle), readerHandle.withMimetype(CSV_CONTENT_TYPE));
-            readerHandle.write(out);
+        EntitySearchService entitySearchService = EntitySearchService.on(finalDatabaseClient);
+        // Exporting directly from Data Service to avoid bug https://bugtrack.marklogic.com/55338 related to namespaced path range indexes
+        Reader export = entitySearchService.exportSearchAsCSV(structuredQuery, searchText, queryOptions, entityTypeId, entityTypeId, limit, sortOrderNode, columns.stream());
+        try {
+            FileCopyUtils.copy(export, new OutputStreamWriter(out));
         } catch (IOException e) {
             throw new RuntimeException(e);
-        } finally {
-            IOUtils.closeQuietly(out);
         }
     }
 
--- a/marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/controllers/EntitySearchControllerTest.java
+++ b/marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/controllers/EntitySearchControllerTest.java
@@ -245,9 +245,85 @@ public class EntitySearchControllerTest extends AbstractMvcTest {
             });
     }
 
-    private void assertRowsAndColumns(int limit, int totalColumns, String response) {
+    @Test
+    void testRowExportWithPathRangeQuery() throws Exception {
+        runAsDataHubDeveloper();
+        ReferenceModelProject project = installOnlyReferenceModelEntities(true);
+        deployEntityIndexes(getHubConfig());
+
+        Customer customer1 = new Customer();
+        customer1.setCustomerId(1);
+        customer1.setName("Jane");
+        customer1.setCustomerNumber(123456789);
+        customer1.setCustomerSince("2012-05-16");
+        project.createCustomerInstance(customer1);
+
+        Customer customer2 = new Customer();
+        customer2.setCustomerId(2);
+        customer2.setName("John");
+        customer2.setCustomerNumber(987654321);
+        customer2.setCustomerSince("2013-06-16");
+        project.createCustomerInstance(customer2);
+
+        String json = "{\n" +
+                "    \"savedQuery\": {\n" +
+                "        \"id\": \"\",\n" +
+                "        \"name\": \"some-query\",\n" +
+                "        \"description\": \"some-query-description\",\n" +
+                "        \"query\": {\n" +
+                "            \"searchText\": \"Customer.customerId:2\",\n" +
+                "            \"entityTypeIds\": [\n" +
+                "                \"Customer\"\n" +
+                "            ],\n" +
+                "            \"selectedFacets\": {\n" +
+                "                \"Collection\": {\n" +
+                "                    \"dataType\": \"xs:string\",\n" +
+                "                    \"stringValues\": [\n" +
+                "                        \"Customer\"\n" +
+                "                    ]\n" +
+                "                }\n" +
+                "            }\n" +
+                "        },\n" +
+                "        \"propertiesToDisplay\": [\n" +
+                "            \"customerId\",\n" +
+                "            \"name\",\n" +
+                "            \"customerNumber\",\n" +
+                "            \"shipping\",\n" +
+                "            \"orders\"\n" +
+                "        ],\n" +
+                "        \"sortOrder\": [\n" +
+                "          { \"propertyName\":  \"customerId\",\n" +
+                "            \"sortDirection\": \"ascending\"\n" +
+                "          }\n" +
+                "        ]\n" +
+                "    }\n" +
+                "}";
+
+        // Even though "propertiesToDisplay" has 5 columns we only export 3 since we dont export object and array type properties for now.
+        int totalColumns = 3;
+        int limit = 99;
+        int expectedCount = 1;
+        Object[] customer2Info = {customer2.getCustomerId(), customer2.getName(), customer2.getCustomerNumber()};
+
+        // Log in with role "hub-central-entity-exporter"
+        loginAsTestUserWithRoles("hub-central-entity-exporter");
+
+        // Export using query document
+        postWithParams(EXPORT_PATH, getRequestParams(limit, json))
+                .andExpect(request().asyncStarted())
+                .andDo(MvcResult::getAsyncResult)
+                .andExpect(status().isOk())
+                .andDo(result -> {
+                    String response = result.getResponse().getContentAsString();
+                    Set<Integer> actualRowSet = calculateHash(response);
+                    assertRowsAndColumns(expectedCount, totalColumns, response);
+                    assertTrue(actualRowSet.contains(getHashCode(customer2Info)));
+                });
+    }
+
+    private void assertRowsAndColumns(int expectedRowCount, int totalColumns, String response) {
         int headerRow = 1;
-        int totalRows = limit + headerRow;
+        int totalRows = expectedRowCount + headerRow;
 
         assertEquals(totalRows, response.chars().filter(i -> i == '\n').count() + 1);
         assertEquals(totalColumns, response.substring(0, response.indexOf("\n")).chars().filter(i -> i == ',').count() + 1);
--- a/marklogic-data-hub-central/src/test/resources/test-projects/reference-project/entities/Customer.entity.json
+++ b/marklogic-data-hub-central/src/test/resources/test-projects/reference-project/entities/Customer.entity.json
@@ -11,7 +11,8 @@
       ],
       "properties": {
         "customerId": {
-          "datatype": "integer"
+          "datatype": "integer",
+          "sortable": true
         },
         "name": {
           "datatype": "string",
--- a/marklogic-data-hub/src/test/resources/entity-reference-model/entities/Customer.entity.json
+++ b/marklogic-data-hub/src/test/resources/entity-reference-model/entities/Customer.entity.json
@@ -12,7 +12,8 @@
       "primaryKey": "customerId",
       "properties": {
         "customerId": {
-          "datatype": "integer"
+          "datatype": "integer",
+          "sortable": true
         },
         "name": {
           "datatype": "string",
--- a/marklogic-data-hub/src/testFixtures/java/com/marklogic/hub/test/AbstractHubTest.java
+++ b/marklogic-data-hub/src/testFixtures/java/com/marklogic/hub/test/AbstractHubTest.java
@@ -5,22 +5,18 @@ import com.marklogic.appdeployer.AppConfig;
 import com.marklogic.appdeployer.ConfigDir;
 import com.marklogic.appdeployer.command.Command;
 import com.marklogic.appdeployer.command.CommandContext;
+import com.marklogic.appdeployer.command.databases.DeployOtherDatabasesCommand;
 import com.marklogic.appdeployer.impl.SimpleAppDeployer;
 import com.marklogic.client.DatabaseClient;
 import com.marklogic.client.FailedRequestException;
 import com.marklogic.client.MarkLogicIOException;
 import com.marklogic.client.io.JacksonHandle;
-import com.marklogic.hub.DatabaseKind;
-import com.marklogic.hub.HubClient;
-import com.marklogic.hub.HubConfig;
-import com.marklogic.hub.HubProject;
-import com.marklogic.hub.deploy.commands.DeployDatabaseFieldCommand;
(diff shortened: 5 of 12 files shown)


marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/entities/search/impl/HubRowManager.java
```diff
@@ -0,0 +1,13 @@
+package com.marklogic.hub.central.entities.search.impl;
```
rjrudin, line 1: What is this for? I don't see any references to it.

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/entities/search/EntitySearchManager.java
```diff
@@ -246,18 +247,23 @@ public void exportRows(JsonNode queryDocument, Long limit, OutputStream out) {
...
-            rowManager.resultDoc(rowManager.newRawPlanDefinition(stringHandle), readerHandle.withMimetype(CSV_CONTENT_TYPE));
-            readerHandle.write(out);
-        } catch (IOException e) {
-            throw new RuntimeException(e);
-        } finally {
-            IOUtils.closeQuietly(out);
-        }
+        EntitySearchService entitySearchService = EntitySearchService.on(finalDatabaseClient);
+        // Exporting directly from Data Service to avoid bug https://bugtrack.marklogic.com/55338 related to namespaced path range indexes
+        Stream<Reader> export = entitySearchService.exportSearch(structuredQuery, searchText, queryOptions, entityTypeId, entityTypeId, limit, sortOrderNode, columns.stream(), CSV_CONTENT_TYPE);
```
rjrudin, line 252: If the intent of this operation is to export search results as CSV, I think the method name should imply that - e.g. "exportSearchResultsAsCsv".

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/entitySearch/exportSearch.api
```diff
@@ -36,11 +36,16 @@
...
+            "datatype": "string"
         }
     ],
     "return": {
-        "datatype": "jsonDocument",
-        "$javaClass": "com.fasterxml.jackson.databind.JsonNode",
-        "description": "Returns a json representation (AST) of an optic plan."
+        "datatype": "textDocument",
+        "$javaClass": "java.io.Reader",
+        "description": "Returns the results of an optic plan",
```
rjrudin, line 48: Is one Reader returned for every row? If so, I think this description should make that clear (the fact that an optic plan is used is an implementation detail) - e.g. "Returns an array of Reader objects, one for each search result represented as CSV"

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/entitySearch/exportSearch.api
```diff
@@ -1,5 +1,5 @@
 {
-    "functionName": "getOpticPlan",
+    "functionName": "exportSearch",
```
rjrudin, line 2: I think we really need a description for this - based on the code, I'm thinking: "Evaluates the given search query against the given TDE view, returning the results as CSV"

rjrudin (changes requested):
We need a test too that reproduces the original problem - i.e. a test that tries to export XML data where the entity definition has a namespace in it (I think that's the most complete test, as it verifies that both the "es" prefix and the user-defined prefix are resolved).

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/entities/search/EntitySearchManager.java
ryanjdew: It is passed the acceptType and the server then determines if it should be CSV. Right now CSV is the only format that will work, but I was trying to open that to be configurable. Similar to the REST rows endpoint.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/entitySearch/exportSearch.api
ryanjdew: It only does one reader for the CSV, but the signature allows for multiple, should we want to export them as separate JSON documents for example.

marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/controllers/EntitySearchControllerTest.java
```diff
@@ -245,9 +245,85 @@ void testRowExport() throws Exception {
             });
     }
 
-    private void assertRowsAndColumns(int limit, int totalColumns, String response) {
+    @Test
+    void testRowExportWithPathRangeQuery() throws Exception {
+        runAsDataHubDeveloper();
+        ReferenceModelProject project = installOnlyReferenceModelEntities(true);
+        deployEntityIndexes(getHubConfig());
```
rjrudin, line 252: I believe dataHub.updateIndexes(); will work here too - can you try that? I recently used it in DeployProtectedPaths..Test and it worked well.

marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/controllers/EntitySearchControllerTest.java
```diff
@@ -245,9 +245,85 @@ void testRowExport() throws Exception {
...
+
+        Customer customer2 = new Customer();
+        customer2.setCustomerId(2);
+        customer2.setName("John");
+        customer2.setCustomerNumber(987654321);
+        customer2.setCustomerSince("2013-06-16");
+        project.createCustomerInstance(customer2);
+
+        String json = "{\n" +
+                "    \"savedQuery\": {\n" +
```
rjrudin, line 269: @rahulvudutala  @akshaysonvane We've got several instances of inline JSON like this. Thoughts on simplifying this? I see a maintenance burden here where if we want to change something about this data structure, we'll need to update all of these tests. Ideally, @ryanjdew  can obtain a saved query here without having to copy/paste one from another test. (This would be a future cleanup, not something to address in this ticket)

marklogic-data-hub-central/src/test/resources/test-projects/reference-project/entities/Customer.entity.json
```diff
@@ -9,9 +9,11 @@
       "required": [
         "name"
       ],
+      "rangeIndex": ["customerId"],
```
rjrudin, line 12: You shouldn't need this - the fact that sortable=true means a path range index will be added automatically.

marklogic-data-hub/src/test/resources/entity-reference-model/entities/Customer.entity.json
```diff
@@ -9,6 +9,9 @@
       "required": [
         "name"
       ],
+      "rangeIndex": [
```
rjrudin, line 12: Same thing here, shouldn't need this - should just be able to add sortable=true or facetable=true to customerId. That's important because as part of converting to HC, rangeIndex/pathRangeIndex/elementRangeIndex are all removed so that HC only deals with sortable and facetable.

rjrudin (changes requested):

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/entitySearch/exportSearch.api
akshaysonvane: Although I understand the intention behind this, I feel like this creates some confusion wrt the returned reader and its relationship with the CSV. For the sake of simplicity should we change the signature to represent what's currently being returned?

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/entities/search/EntitySearchManager.java
```diff
@@ -246,18 +247,23 @@ public void exportRows(JsonNode queryDocument, Long limit, OutputStream out) {
...
+                        readerHandle.set(reader);
+                        readerHandle.write(out);
+                    } catch (IOException e) {
+                        IOUtils.closeQuietly(out);
+                        throw new RuntimeException(e);
+                    } finally {
+                        IOUtils.closeQuietly(reader);
+                    }
+                });
+        IOUtils.closeQuietly(out);
```
akshaysonvane, line 266: Changing the DS signature will also allow us to simplify this.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/entitySearch/exportSearch.api
rjrudin: I feel endpoints are cheap, and reuse can always be done via library modules. So I lean towards fine-grained, explicitly-named endpoints - e.g. "exportSearchResultsAsCsv" - that reuse libraries to avoid duplication. We can always create more generic endpoints in the future as we learn more.

marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/controllers/EntitySearchControllerTest.java
ryanjdew: dataHub isn't available to the hub central tests, unless I'm missing a function someplace. I believe the DataHub class is still depending on the auto-wiring of the HubConfig.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/entitySearch/exportSearch.sjs
```diff
@@ -115,13 +116,23 @@ if (sortOrder) {
...
+}
+if (orderDefinitions.length > 0) {
+  opticPlan = opticPlan.orderBy(orderDefinitions);
+}
+opticPlan = opticPlan.select(columnIdentifiers);
+// Not using the rows REST API due to https://bugtrack.marklogic.com/55338
+let result = opticPlan.result('object');
+switch(acceptType) {
+  case 'text/csv':
+    if (!(result instanceof Sequence)) {
```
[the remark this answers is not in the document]
ryanjdew: Changed to be a CSV specific endpoint.

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/entities/search/EntitySearchManager.java
```diff
@@ -246,16 +247,16 @@ public void exportRows(JsonNode queryDocument, Long limit, OutputStream out) {
...
-        RowManager rowManager = finalDatabaseClient.newRowManager();
-        try (ReaderHandle readerHandle = new ReaderHandle()) {
-            rowManager.resultDoc(rowManager.newRawPlanDefinition(stringHandle), readerHandle.withMimetype(CSV_CONTENT_TYPE));
+        EntitySearchService entitySearchService = EntitySearchService.on(finalDatabaseClient);
+        // Exporting directly from Data Service to avoid bug https://bugtrack.marklogic.com/55338 related to namespaced path range indexes
+        Reader export = entitySearchService.exportSearchAsCSV(structuredQuery, searchText, queryOptions, entityTypeId, entityTypeId, limit, sortOrderNode, columns.stream());
+        try (ReaderHandle readerHandle = new ReaderHandle(export).withMimetype(CSV_CONTENT_TYPE)) {
             readerHandle.write(out);
         } catch (IOException e) {
+            IOUtils.closeQuietly(out);
```
akshaysonvane, line 256: We don't need this here since the "finally" block will always close the resource.

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/entities/search/EntitySearchManager.java
```diff
@@ -246,16 +247,16 @@ public void exportRows(JsonNode queryDocument, Long limit, OutputStream out) {
...
             throw new RuntimeException(e);
         } finally {
+            IOUtils.closeQuietly(export);
```
akshaysonvane, line 259: We can get rid of this since ReaderHandle will internally close the "export" reader in the try-with-resource above.

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/data-services/entitySearch/exportSearchTest.sjs
```diff
@@ -1,18 +1,29 @@
...
+        results.push(obj);
+    }
+    return Sequence.from(results);
+}
+
+function invokeExportSearchService(schemaName, viewName, limit, structuredQuery, searchText, sortOrder, columns, retries=0) {
+    try {
         return fn.head(xdmp.invoke(
-            "/data-hub/5/data-services/entitySearch/getOpticPlan.sjs",
+            "/data-hub/5/data-services/entitySearch/exportSearch.sjs",
```
akshaysonvane, line 26: Needs to be updated to the new name `exportSearchAsCSV`

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/data-services/entitySearch/exportSearchTest.sjs
```diff
@@ -21,7 +32,8 @@ function invokeOpticPlanService(schemaName, viewName, limit, structuredQuery, se
                 searchText,
                 queryOptions,
                 sortOrder: xdmp.toJSON(sortOrder).root,
-                columns: Sequence.from(columns)
+                columns: Sequence.from(columns),
+                acceptType: 'text/csv'
```
akshaysonvane, line 36: Also can get rid of the `acceptType` param since we elected to drop it.

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/entities/search/EntitySearchManager.java
```diff
@@ -246,12 +247,10 @@ public void exportRows(JsonNode queryDocument, Long limit, OutputStream out) {
...
-
-        JsonNode opticPlanNode = EntitySearchService.on(finalDatabaseClient).getOpticPlan(structuredQuery, searchText, queryOptions, entityTypeId, entityTypeId, limit, sortOrderNode, columns.stream());
-        StringHandle stringHandle = new StringHandle(opticPlanNode.toString());
-        RowManager rowManager = finalDatabaseClient.newRowManager();
-        try (ReaderHandle readerHandle = new ReaderHandle()) {
-            rowManager.resultDoc(rowManager.newRawPlanDefinition(stringHandle), readerHandle.withMimetype(CSV_CONTENT_TYPE));
+        EntitySearchService entitySearchService = EntitySearchService.on(finalDatabaseClient);
+        // Exporting directly from Data Service to avoid bug https://bugtrack.marklogic.com/55338 related to namespaced path range indexes
+        Reader export = entitySearchService.exportSearchAsCSV(structuredQuery, searchText, queryOptions, entityTypeId, entityTypeId, limit, sortOrderNode, columns.stream());
+        try (ReaderHandle readerHandle = new ReaderHandle(export).withMimetype(CSV_CONTENT_TYPE)) {
```
rjrudin, line 253: @akshaysonvane I know this code existed already, but I don't think it's necessary to call withMimetype. I looked at the underlying code, and I don't think it cares. It's using a Java Client class called Utilities, which does the same thing as the Spring class FileCopyUtils. I think to avoid confusion about the necessity of withMimetype (I'm 99% sure it's not needed) and for consistency (FileCopyUtils is a favorite class of mine since it solves the terrible user experience in Java of copying data around), it'd be better to just do:

FileCopyUtils.copy(export, new OutputStreamWriter(out)); 

(which will still need to be in a try/catch block)

marklogic-data-hub/src/testFixtures/java/com/marklogic/hub/test/AbstractHubTest.java
```diff
@@ -374,6 +371,27 @@ protected void installUserModulesAndArtifacts(HubConfig hubConfig, boolean force
 
     }
 
+    protected void deployEntityIndexes(HubConfig hubConfig) {
+        EntityManager entityManager = new EntityManagerImpl(hubConfig);
+        entityManager.saveDbIndexes();
+        List<Command> commands = new ArrayList<>();
+
+        Path dir = hubConfig.getEntityDatabaseDir();
+        File finalFile = Paths.get(dir.toString(), HubConfig.FINAL_ENTITY_DATABASE_FILE).toFile();
```
rjrudin, line 380: Let's not put this here yet, as it only deals with the final file but the method name implies that both staging and final would be updated. I'll add a ticket to the 5.4.0 backend tech debt epic to make it easier to save indexes / deploy them in a test, as a quick search showed that we have a number of tests looking to do this. 

One simplification you can do is this:

dbCommand.execute(newCommandContext());

That avoids having to create a list of commands and a SimpleAppDeployer.

rjrudin (changes requested):

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/entities/search/EntitySearchManager.java
akshaysonvane: Yes, that's right. I looked and it seems withMimetype was required earlier when we were sending the optic plan to the RowManager as the default was JSON. Now that we are generating the CSV in the DS itself FileCopyUtils can be used for copying.

ryanjdew: Updated to use FileCopyUtils.copy.

marklogic-data-hub/src/testFixtures/java/com/marklogic/hub/test/AbstractHubTest.java
ryanjdew: Updated to use dbCommand.execute(newCommandContext()) and deploy both staging and final indexes

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/data-services/entitySearch/exportSearchTest.sjs
ryanjdew: renamed

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/entities/search/EntitySearchManager.java
ryanjdew: updated

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/data-services/entitySearch/exportSearchTest.sjs
ryanjdew: removed

rjrudin (approved):

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/entities/search/EntitySearchManager.java
```diff
@@ -246,17 +249,16 @@ public void exportRows(JsonNode queryDocument, Long limit, OutputStream out) {
...
+        // Exporting directly from Data Service to avoid bug https://bugtrack.marklogic.com/55338 related to namespaced path range indexes
+        Reader export = entitySearchService.exportSearchAsCSV(structuredQuery, searchText, queryOptions, entityTypeId, entityTypeId, limit, sortOrderNode, columns.stream());
+        try {
+            FileCopyUtils.copy(export, new OutputStreamWriter(out));
         } catch (IOException e) {
-            throw new RuntimeException(e);
-        } finally {
+            // Ensure both streams are closed if an error occurs
+            IOUtils.closeQuietly(export);
             IOUtils.closeQuietly(out);
```
akshaysonvane, line 260: I believe FileCopyUtils.copy will close the streams for us even if an error occurs.

akshaysonvane (approved):

akshaysonvane (approved):

rjrudin (approved):

PR #4537, opened 2020-09-08
DHFPROD-5811: Hopefully fixing how schemas dbs are cleared in tests

rjrudin:
### Description

The use of newAppServicesDatabaseClient is failing intermittently. Since we don't need to retain any documents, we can use the Manage API to clear the database. 

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/testFixtures/java/com/marklogic/hub/test/AbstractHubTest.java
+++ b/marklogic-data-hub/src/testFixtures/java/com/marklogic/hub/test/AbstractHubTest.java
@@ -7,6 +7,8 @@ import com.marklogic.appdeployer.command.Command;
 import com.marklogic.appdeployer.command.CommandContext;
 import com.marklogic.appdeployer.impl.SimpleAppDeployer;
 import com.marklogic.client.DatabaseClient;
+import com.marklogic.client.FailedRequestException;
+import com.marklogic.client.MarkLogicIOException;
 import com.marklogic.client.io.JacksonHandle;
 import com.marklogic.hub.DatabaseKind;
 import com.marklogic.hub.HubClient;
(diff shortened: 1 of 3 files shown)


akshaysonvane (approved):

SameeraPriyathamTadikonda (dismissed):

SameeraPriyathamTadikonda (approved):

akshaysonvane (approved):

rjrudin:
Added some primitive retry support for when resetDatabases gets an intermittent connect exception

SameeraPriyathamTadikonda (approved):

marklogic-data-hub/src/testFixtures/java/com/marklogic/hub/test/AbstractHubTest.java
```diff
@@ -100,15 +104,41 @@ protected void resetDatabases() {
...
+        DatabaseManager databaseManager = new DatabaseManager(hubClient.getManageClient());
+        databaseManager.clearDatabase(hubClient.getDbName(DatabaseKind.STAGING_SCHEMAS));
+        databaseManager.clearDatabase(hubClient.getDbName(DatabaseKind.FINAL_SCHEMAS));
+    }
+
+    /**
+     * Adding this to handle some intermittent connection failures that occur when Jenkins runs the tests.
+     * @param r
+     */
+    protected void retryIfNecessary(Runnable r) {
```
akshaysonvane, line 122: Should we create our own no-arg/no-return functional interface to avoid confusion related to Runnable being related to threads?
https://stackoverflow.com/a/23958916/4460877

rjrudin: I'm hoping I can remove this soon - it's just here for now to see if it helps at all. ML app servers shouldn't be unavailable for stretches of time during the test suite run - this is just to see if it helps. Hopefully the real problem can be fixed.

akshaysonvane: I see. If we do need to add this in, I think it would be worthwhile to add a comment explaining that Runnable is only being used as a no-arg/no-return functional interface and does not have anything to do with Threads.

srinathgit (approved):

PR #4544, opened 2020-09-10
DHFPROD-5597: Show search link when facets > 20

wooldridge:
### Description

Popover search link only shown for a facet when number of facet items is great than a limit (20).

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/src/components/facet/facet.test.tsx
+++ b/marklogic-data-hub-central/ui/src/components/facet/facet.test.tsx
@@ -1,26 +1,12 @@
 import React from 'react';
 import { render, fireEvent } from '@testing-library/react';
 import Facet from './facet';
-import { facetValues } from '../../assets/mock-data/entity-table';
+import { facetProps } from '../../assets/mock-data/facet';
 
 describe("Facet component", () => {
   it("Facet component renders with data properly" , () => {
-    const { getByTestId, getByText, queryByLabelText } = render(
-        <Facet
-          name="sales_region"
-          constraint="sales_region"
-          facetValues={facetValues}
-          key=""
-          tooltip=""
-          facetType="xs:string"
-          facetCategory="entity"
-          updateSelectedFacets={jest.fn()}
-          addFacetValues={jest.fn()}
-          referenceType="element"
-          entityTypeId=""
-          propertyPath="sales_region"
-        />
-    )
+    const { getByTestId, getByText, queryByLabelText } = render(<Facet {...facetProps} />)
+
     expect(getByText(/sales_region/i)).toBeInTheDocument();
 
     expect(getByText(/Customer/i)).toBeInTheDocument();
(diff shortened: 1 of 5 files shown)


bsrikan:
@wooldridge this is a 5.4 story and should go into feature/5.4-develop branch

marklogic-data-hub-central/ui/src/components/facet/facet.test.tsx
```diff
@@ -97,4 +99,35 @@ describe("Facet component", () => {
     expect(document.querySelector('[data-testid=sales_region-toggle] i svg')).toHaveStyle('transform: rotate(180deg);');
   });
 
+  it("Search link shown when facet number greater than limit" , () => {
```
bsrikan, line 102: I would recommend 3 things:
1. Create a default facetProps under assets/mock-data to be used by all the facet tests
2. Move the < 20 test into this one. facetValues is already mocked out and has 12 items. Can use the default props to verify for < 20 condition
3. Do facetValues.push to get it above 20, rerender with updated facetValues and test for > 20 conddition. 

Refer https://wiki.marklogic.com/display/ENGINEERING/Writing+Data+Hub+Tests#WritingDataHubTests-RTLTests for guidance.

bsrikan (changes requested):

marklogic-data-hub-central/ui/src/components/facet/facet.test.tsx
wooldridge: @bsrikan for tests of this functionality, I'd prefer to be able to set the limit as a constant and then key off of that for the tests (to make it clear what the limit is and make it easy to change). I can rewrite to test facet values for LIMIT-1, rerender, and then test values for LIMIT.

bsrikan: You can still test the way you want to. You are going to use the default props which has less than 20 values, which is how `expect(queryByLabelText('popover-search-label')).not.toBeInTheDocument();` is being tested now. Only difference with my recommendation is both conditions will be asserted in same test making the expected behavior more apparent.

```
    const { getByLabelText, rerender } = render(<Facet {...data.facetProps} />)
    // Search link not shown for facets < 20
    expect(queryByLabelText('popover-search-label')).not.toBeInTheDocument();
    
    // Search link shown for facets >= 20
    const LIMIT =  20;
    const f = {
      "name": "facetName",
      "count": 123,
      "value": "facetValue"
    };
    let facetVals: any = [];
    for (let i = 0; i < LIMIT; i++) {
      facetVals.push(f); //run time appending to facetValues in assets/mock-data/entity-table
    }
    rerender(<Facet {...data.facetProps} facetValues={facetVals}/>)
    expect(getByLabelText('popover-search-label')).toBeInTheDocument();
```

wooldridge: @bsrikan, thanks! I refactored the tests using a new facetProps default object, much cleaner.

Sanjeevani19 (approved):

Sanjeevani19 (approved):
@wooldridge Any reason for making it as two commits, if not could you please make them into one ?

brucean52 (dismissed):

bsrikan (approved):
Thanks Mike for refactoring the test. Looks good.

bsrikan:
@Sanjeevani19 @timur-isangulov seems to be failed e2e for the intermittent issue reported in this bug:
https://project.marklogic.com/jira/browse/DHFPROD-5392

SameeraPriyathamTadikonda:
Looks like we need to fix the e2e tests.

xnikhil08:
> Looks like we need to fix the e2e tests that failed on this PR

@wooldridge @timur-isangulov @Sanjeevani19 - Pushed a fix for those failing tests. The popover search related code can be removed from one of the test as its already being verified in RTL integration test. 
I removed the other test completely because its supposed to be handled in an RTL integration test and created a separate JIRA ticket for the same.
[https://project.marklogic.com/jira/browse/DHFPROD-5958]

rjrudin:
This is marked as FixVersion=5.4.0 in JIRA, but it's pointed at the develop branch, so one of those is wrong - which one?

wooldridge:
@rjrudin thanks for noticing this, this should go on feature/5.4-develop. I will create a new PR.

PR #4545, opened 2020-09-10
DHFPROD-5748: Radio group for switching views in Explore is not consistent with Load.

timur-isangulov:
Includes DHFPROD-3628: Remove unneeded tooltips for Explorer facet values

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/e2e/cypress/support/pages/browse.tsx
+++ b/marklogic-data-hub-central/ui/e2e/cypress/support/pages/browse.tsx
@@ -89,12 +89,18 @@ class BrowsePage {
     return this.getDocument(index).find('[data-cy=instance]');
   }
 
-
+  getTooltip(tooltip: string) {
+    return cy.get(`#${tooltip}-tooltip`);
+  }
 
   /**
    * facet search
    * available facets are 'collection', 'created-on', 'job-id', 'flow', 'step'
    */
+
+  getFacetName(facet: string) {
+    return cy.get('.ml-tooltip-container').contains(facet);
+  }
 
   getFacet(facet: string) {
     return cy.get('[data-cy="' + facet + '-facet"]');
@@ -220,12 +226,11 @@ class BrowsePage {
     this.waitForSpinnerToDisappear();
   }
 
-
     //table, facet view
   clickFacetView() {
     this.waitForSpinnerToDisappear();
     this.waitForTableToLoad();
-    return cy.get('[data-cy=facet-view]').click();
+    cy.get('[data-cy=facet-view]').click().trigger('mouseout');
   }
 
   getFacetView(){
@@ -398,7 +403,7 @@ class BrowsePage {
   }
 
   getRadioOptionSelected() {
-    return cy.get('[type="radio"]').first().check();
+    return cy.get('.ant-modal [type="radio"]').first().check();
   }
 
   getEditSaveChangesButton() {
--- a/marklogic-data-hub-central/ui/src/components/facet/facet-element.tsx
+++ b/marklogic-data-hub-central/ui/src/components/facet/facet-element.tsx
@@ -1,0 +1,30 @@
+import React, { useState } from 'react';
+import styles from './facet.module.scss';
+import { numberConverter } from '../../util/number-conversion';
+import { stringConverter } from '../../util/string-conversion';
+import { MLTooltip, MLCheckbox } from '@marklogic/design-system';
+import { OverflowDetector } from 'react-overflow';
+
+
+export const FacetName = (props) => {
+  const [isOverflowed, setisOverflowed] = useState(false)
+  const handleOverflowChange = (isOverflowed) => setisOverflowed(isOverflowed)
+
+  return (
+    <div className={styles.checkContainer} key={props.index} data-testid={props.facet.value} data-cy={stringConverter(props.name) + "-facet-item"}>
+      <OverflowDetector onOverflowChange={handleOverflowChange} style={{ width: '180px' }}>
+        <MLCheckbox
+          value={props.facet.value}
+          onChange={(e) => props.handleClick(e)}
+          checked={props.checked.includes(props.facet.value)}
+          className={styles.value}
+          data-testid={`${stringConverter(props.name)}-${props.facet.value}-checkbox`}
+        >
+          <MLTooltip title={isOverflowed && props.facet.value} id={props.facet.value + '-tooltip'} >{props.facet.value}</MLTooltip>
+        </MLCheckbox>
+      </OverflowDetector>
+      <div className={styles.count}
+        data-cy={`${stringConverter(props.name)}-${props.facet.value}-count`}>{numberConverter(props.facet.count)}</div>
+    </div>
+  )
+}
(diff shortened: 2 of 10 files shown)


bsrikan (dismissed):
Can you please split the commit based on corresponding JIRA ticket, when there are multiple tickets involved in a PR.

wooldridge:
Hi, @timur-isangulov. I'm seeing tooltip presence being keyed off of string length but the ellipsis presence applied via CSS, which considers container width (not string length). As a result I see inconsistent tooltip application at edge cases:

https://drive.google.com/file/d/1-RLilaWYHBm4MSL9TIkH2KF0h2H31plg/view

The isEllipsisActive() function here might be a solution:

https://stackoverflow.com/questions/7738117/html-text-overflow-ellipsis-detection

bsrikan:
We should all decide on a fixed indent size, or maybe even start using prettier to avoid formatting diffs show up in PRs.

wooldridge:
> We should all decide on a fixed indent size, or maybe even start using prettier to avoid formatting diffs show up in PRs.

This will help enforce such standards: https://project.marklogic.com/jira/browse/DHFPROD-4777

wooldridge:
@timur-isangulov, it seems to be ALMOST fixed. (Nice use of react-overflow package, by the way.) I have a bunch of examples where I use long collection names and the only one that doesn't work as expected (shows tooltip when not needed) is:

20-11111111111111111

("20-" followed by 17 1's.)

See: https://drive.google.com/file/d/1N4gI8fOJNJtticMNcBUfJF5HDEiYq76d/view

wooldridge (changes requested):
See comment.

timur-isangulov:
> @timur-isangulov, it seems to be ALMOST fixed. (Nice use of react-overflow package, by the way.) I have a bunch of examples where I use long collection names and the only one that doesn't work as expected (shows tooltip when not needed) is:
> 
> 20-11111111111111111
> 
> ("20-" followed by 17 1's.)
> 
> See: https://drive.google.com/file/d/1N4gI8fOJNJtticMNcBUfJF5HDEiYq76d/view

Thanks Mike, I was able to fix it locally https://drive.google.com/file/d/16u4xDUCohCQh8DcttNe-rtKdtPniCDV7/view and pushed commit, can you please review again.

Sanjeevani19 (commented):
@timur-isangulov Could you please make them into one commit ?

Sanjeevani19:
> @timur-isangulov Could you please make them into one commit ?

Thanks @timur-isangulov for clarifying since it's two different tickets, it makes sense to have two different commits

bsrikan:
@timur-isangulov should zero state page also have the changes from this PR for table/snippet views:
image

timur-isangulov:
> @timur-isangulov should zero state page also have the changes from this PR for table/snippet views:
> image

I don't think so. It's radio button component but there was a separate design for it.

Sanjeevani19 (dismissed):
Looks good to me. Works as expected

wooldridge (changes requested):
@timur-isangulov, I'm seeing a minor style difference between the switch on Explore vs. Load. On Load, the purple border is only around the selected icon whereas on Explore the purple border remains around both icons:

https://drive.google.com/file/d/1lkwFDxMyg1TJCEx6li4Suk4eqeQyzi6J/view

Based on the ticket description, the Load is the correct style. Can you fix the CSS?

xnikhil08:
@timur-isangulov - Looks good, except the change suggested by @wooldridge 
Also, can you please squash your commits to one ?  I will approve once the issue is fixed with single commit.

timur-isangulov:
> @timur-isangulov, I'm seeing a minor style difference between the switch on Explore vs. Load. On Load, the purple border is only around the selected icon whereas on Explore the purple border remains around both icons:
> 
> https://drive.google.com/file/d/1lkwFDxMyg1TJCEx6li4Suk4eqeQyzi6J/view
> 
> Based on the ticket description, the Load is the correct style. Can you fix the CSS?

Hi Mike, thanks for noticing that. I pushed commit with the fix, can you please review again. 
Screen Shot 2020-09-14 at 5 47 15 PM

timur-isangulov:
> @timur-isangulov - Looks good, except the change suggested by @wooldridge. I will approve once the issue is fixed.
> Also, If you can't avoid multiple commits then update the checklist option "Squashed all commits into a single commit" in the PR description, with "N/A" probably, so that the pre-build check doesn't fail.
> Pinging @SameeraPriyathamTadikonda to understand more options, if there are any.

Hi @xnikhil08, I just pushed the commit with the fix for the issues. Can you please review again.

marklogic-data-hub-central/ui/e2e/cypress/support/pages/browse.tsx
```diff
@@ -34,7 +34,7 @@ class BrowsePage {
   }
 
   clickPaginationItem(index: number) {
-    return cy.get(`#top-search-pagination-bar .ant-pagination-item-${index}`).click();
+    return cy.get(`#top-search-pagination-bar .ant-pagination-item-${index}`).click({force:true});
```
bsrikan, line 37: It wasnt too hard to find a solution that will work without using `{force: true}`. We just have to look for it. Can you pls try "mouseleave" trigger if it works.

timur-isangulov: mouseleave doesn't work but mouseout seems to work for me locally, I pushed it to verify if it passes on pipeline.

Sanjeevani19 (approved):

marklogic-data-hub-central/ui/e2e/cypress/support/pages/browse.tsx
Sanjeevani19: Tests passed with the new changes, looks good to me.

marklogic-data-hub-central/ui/e2e/cypress/support/pages/browse.tsx
```diff
@@ -220,12 +226,11 @@ class BrowsePage {
     this.waitForSpinnerToDisappear();
   }
 
-
     //table, facet view
   clickFacetView() {
     this.waitForSpinnerToDisappear();
     this.waitForTableToLoad();
-    return cy.get('[data-cy=facet-view]').click();
+    cy.get('[data-cy=facet-view]').click().trigger('mouseout');
```
bsrikan, line 233: Thanks for removing "return" and adding "mouseout" trigger

marklogic-data-hub-central/ui/src/components/facet/facet-element.tsx
```diff
@@ -0,0 +1,30 @@
+import React, { useState } from 'react';
+import styles from './facet.module.scss';
+import { numberConverter } from '../../util/number-conversion';
+import { stringConverter } from '../../util/string-conversion';
+import { MLTooltip, MLCheckbox } from '@marklogic/design-system';
+import { OverflowDetector } from 'react-overflow';
+
+
+export const FacetName = (props) => {
```
bsrikan, line 9: I thought there was a RTL test added for this initially, now I dont find it. Was it removed?

wooldridge (approved):

marklogic-data-hub-central/ui/src/components/facet/facet-element.tsx
timur-isangulov: I was having issues testing tooltip functionality using RTL. Probably related to the CSS visibility properties. I will do research and try to add the test in my next pr.

PR #4555, opened 2020-09-11
DHFPROD-5691: Fix outstanding issues found by coverity static scan

akshaysonvane:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [N/A] Added Tests
  

- ##### Reviewer:

- [N/A] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/flow/impl/FlowRunnerImpl.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/flow/impl/FlowRunnerImpl.java
@@ -2,23 +2,15 @@ package com.marklogic.hub.flow.impl;
 
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
-import com.fasterxml.jackson.databind.node.ObjectNode;
-import com.marklogic.client.DatabaseClient;
-import com.marklogic.client.FailedRequestException;
-import com.marklogic.client.io.JacksonHandle;
 import com.marklogic.hub.FlowManager;
 import com.marklogic.hub.HubClient;
 import com.marklogic.hub.HubConfig;
 import com.marklogic.hub.flow.Flow;
-import com.marklogic.hub.flow.FlowInputs;
-import com.marklogic.hub.flow.FlowRunner;
-import com.marklogic.hub.flow.FlowStatusListener;
-import com.marklogic.hub.flow.RunFlowResponse;
+import com.marklogic.hub.flow.*;
 import com.marklogic.hub.impl.FlowManagerImpl;
 import com.marklogic.hub.impl.HubConfigImpl;
 import com.marklogic.hub.job.JobDocManager;
 import com.marklogic.hub.job.JobStatus;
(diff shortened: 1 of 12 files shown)


ryanjdew (approved):

marklogic-data-hub/src/main/java/com/marklogic/hub/flow/impl/FlowRunnerImpl.java
```diff
@@ -254,18 +228,17 @@ private void initializeFlow(String jobId) {
...
-        synchronized (stepsMap) {
-            if(stepsMap.get(jobId) != null){
-                stepsMap.get(jobId).clear();
-                stepsMap.remove(jobId);
-                isJobCancelled.set(true);
-            }
-            else {
-                throw new RuntimeException("Job not running");
-            }
+        if (stepsMap.get(jobId) != null) {
```
rjrudin, line 231: What was the issue with "synchronized"? I don't know if it's actually needed, but I also don't know the impact of removing it.

marklogic-data-hub/src/main/java/com/marklogic/hub/hubcentral/HubCentralManager.java
```diff
@@ -160,9 +160,10 @@ protected void extractZipToProject(HubProject hubProject, File zipFile) {
                 outputFile.getParentFile().mkdirs();
                 try (InputStream inputStream = zip.getInputStream(entry);
                      FileOutputStream fileOut = new FileOutputStream(outputFile)) {
-                    inputStream.read(buffer, 0, entrySize);
-                    logger.info("Writing file: " + outputFile);
-                    FileCopyUtils.copy(buffer, fileOut);
+                    if (inputStream.read(buffer, 0, entrySize) > 0) {
```
rjrudin, line 163: I think the fileOut declaration needs to be moved into the "if" block - otherwise, it never gets closed. Could just do:

    FileCopyUtils.copy(buffer, new FileOutputStream(outputFile));

rjrudin (changes requested):

marklogic-data-hub/src/main/java/com/marklogic/hub/flow/impl/FlowRunnerImpl.java
akshaysonvane: Coverity had raised a concurrency issue earlier which is why @ryanjdew added this in. Right now it's raising an issue that we are synchronizing over an object of java.util.concurrent package.

I checked and also conformed with Ryan that its safe to remove the synchronized block since both stepsMap and isJobCancelled are part of java.util.concurrent package.

Ryan and I decided to remove the synchronized block for now and see if the Coverity scan raises the earlier issue again.

marklogic-data-hub/src/main/java/com/marklogic/hub/hubcentral/HubCentralManager.java
akshaysonvane: In the event when `FileCopyUtils.copy(buffer, fileOut)` does not run the fileOut stream should be closed via the try-with-resources block.

rjrudin (approved):

PR #4558, opened 2020-09-11
DHFPROD-5945: Initial cut at an envelope schema

rjrudin:
### Description

This will be updated as part of Source Data Exploration

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [ ] Added Tests
  

- ##### Reviewer:

- [ ] Reviewed Tests

### DIFF
--- a/specs/models/DataHubEnvelope.schema.json
+++ b/specs/models/DataHubEnvelope.schema.json
@@ -1,0 +1,141 @@
+{
+  "title": "DataHubEnvelope",
+  "description": "Defines the envelope structure with DHF-specific additions to headers",
+  "type": "object",
+  "properties": {
+    "headers": {
+      "description": "The unique path to this property. For a structured property, uses dot notation - e.g. billing.street.fiveDigit",
+      "type": "object",
+      "properties": {
+        "createdBy": {
+          "description": "MarkLogic user that created this document",
+          "type": "string"
+        },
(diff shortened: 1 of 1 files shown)


specs/models/DataHubEnvelope.schema.json
```diff
@@ -0,0 +1,141 @@
+{
+  "title": "DataHubEnvelope",
+  "description": "Defines the envelope structure with DHF-specific additions to headers",
+  "type": "object",
+  "properties": {
+    "headers": {
+      "description": "The unique path to this property. For a structured property, uses dot notation - e.g. billing.street.fiveDigit",
```
ryanjdew, line 7: looks like a copy/paste left from another description

specs/models/DataHubEnvelope.schema.json
```diff
@@ -0,0 +1,141 @@
...
+              "import-id": {
+                "description": "Appears to be specific to mastering",
+                "type": "string"
+              },
+              "name": {
+                "description": "Name of a source that provides instance data in this envelope",
+                "type": "string"
+              },
+              "user": {
+                "description": "May be specific to mastering",
```
ryanjdew, line 98: This isn't used by mastering for anything. Other properties can exist in the source object, but name and dateTime are the only properties that have features with some dependency on them.

PR #4566, opened 2020-09-14
DHFPROD-5691: Fix outstanding issue found by coverity static scan

akshaysonvane:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [N/A] Added Tests
  

- ##### Reviewer:

- [N/A] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/flow/impl/FlowRunnerImpl.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/flow/impl/FlowRunnerImpl.java
@@ -55,7 +55,11 @@ public class FlowRunnerImpl implements FlowRunner{
 
     private StepRunner stepRunner;
 
-    private final Map<String, Queue<String>> stepsMap = new ConcurrentHashMap<>();
+    /*
+     * Using concrete type ConcurrentHashMap instead of Map so coverity static scan does not complain about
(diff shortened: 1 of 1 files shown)


marklogic-data-hub/src/main/java/com/marklogic/hub/flow/impl/FlowRunnerImpl.java
```diff
@@ -55,7 +55,7 @@
 
     private StepRunner stepRunner;
 
-    private final Map<String, Queue<String>> stepsMap = new ConcurrentHashMap<>();
+    private final ConcurrentHashMap<String, Queue<String>> stepsMap = new ConcurrentHashMap<>();
```
rjrudin, line 58: Can you include a brief comment to explain why this map has a concrete type but the others are just Map? Otherwise some well-meaning developer in the future will change this to Map to be consistent with the others.

rjrudin (changes requested):

marklogic-data-hub/src/main/java/com/marklogic/hub/flow/impl/FlowRunnerImpl.java
akshaysonvane: Added the comment

rjrudin (approved):

srinathgit (approved):

PR #4574, opened 2020-09-15
DHFPROD-4854: Removed spring-boot dependency from core project

rjrudin:
### Description

The irony here is this mostly involves removing code!

For the core project, we only need spring-context, and we now obtain a ConfigurableApplicationContext where needed by just instantiating an AnnotationConfigurableApplicationContext with the appropriate Configuration class. 

Replaced the spring-boot bootJar task with a custom jar task that is almost identical to the existing one for clientJar - just has a different main class. 

Renamed logback-spring.xml to logback.xml, and this no longer is included in the DHF library jar, since processResources no longer copies it in. It's only included in the executable jars that need it.

Removed EnableAutoConfiguration from DataHubPlugin, which wasn't even doing anything. The Spring container is created manually in that class. 

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [ ] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/build.gradle
+++ b/marklogic-data-hub/build.gradle
@@ -5,10 +5,7 @@ plugins {
     id 'maven-publish'
     id 'com.jfrog.bintray' version '1.7.2'
     id 'com.marklogic.ml-gradle' version '4.0.4'
-    //id 'com.moowork.node' version '1.1.1'
     id "com.github.node-gradle.node" version "2.2.4"
-    id 'org.springframework.boot' version '2.1.15.RELEASE'
-    id "io.spring.dependency-management" version "1.0.7.RELEASE"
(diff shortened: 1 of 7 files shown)


SameeraPriyathamTadikonda:
@rjrudin Can we remove the trace-ui too?

rjrudin:
@SameeraPriyathamTadikonda  Can't remove trace-ui yet, as some people might be using that. That's just a pile of modules tucked into the DHF library jar which are then deployed to the modules database. I wish it didn't exist, but it only causes performance delays at build time. Doesn't cause any runtime issues that I'm aware of. Whereas spring-boot does cause runtime errors.

PR #4576, opened 2020-09-15
DHFPROD-5608: Placement of new card in Load

briantang:
### Description

- Load cards now ordered by last updated.
- Pointed for develop as story is marked for 5.3

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/src/components/load/load-card.test.tsx
+++ b/marklogic-data-hub-central/ui/src/components/load/load-card.test.tsx
@@ -58,7 +58,7 @@ describe('Load Card component', () => {
     //Verify if the flow related options are availble on mouseOver
     expect(getByTestId('testLoadXML-toNewFlow')).toBeInTheDocument(); // check if option 'Add to a new Flow' is visible
     expect(getByTestId('testLoadXML-toExistingFlow')).toBeInTheDocument(); // check if option 'Add to an existing Flow' is visible
-
+ 
     //Click on the select field to open the list of existing flows.
     fireEvent.click(getByTestId('testLoadXML-flowsList'));
 
@@ -73,10 +73,10 @@ describe('Load Card component', () => {
       expect(mockHistoryPush).toHaveBeenCalledWith('/tiles/run/add');
     })
     //TODO- E2E test to check if the Run tile is loaded or not.
-
+    
   });
 
-  test('Load Card - Add step to a new Flow', async () => {
+  test('Load Card - Verify card sort order and Add step to a new Flow', async () => {
     const authorityService = new AuthoritiesService();
     authorityService.setAuthorities(['readIngestion', 'writeIngestion', 'writeFlow']);
     const { getByText, getByLabelText, getByTestId } = render(
@@ -92,6 +92,13 @@ describe('Load Card component', () => {
       </MemoryRouter>
     )
 
+    //Verify cards get sorted by last updated
+    let loadCards: any = document.querySelectorAll('.ant-col');
+    expect(loadCards[0]).toHaveTextContent('Add New');
+    expect(loadCards[1]).toHaveTextContent('testLoadXML');
+    expect(loadCards[2]).toHaveTextContent('testLoad123');
+    expect(loadCards[3]).toHaveTextContent('testLoad');
+  
     //Check if the card is rendered properly
     expect(getByText('Add New')).toBeInTheDocument();
     expect(getByText('testLoadXML')).toBeInTheDocument();
--- a/marklogic-data-hub-central/ui/src/components/load/load-card.tsx
+++ b/marklogic-data-hub-central/ui/src/components/load/load-card.tsx
@@ -1,4 +1,4 @@
-import React, {CSSProperties, useContext, useState} from 'react';
+import React, {CSSProperties, useContext, useState, useEffect} from 'react';
 import styles from './load-card.module.scss';
 import { useHistory } from 'react-router-dom';
 import {Card, Icon, Tooltip, Popover, Row, Col, Modal, Select} from 'antd';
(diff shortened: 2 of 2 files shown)


wooldridge:
@bsrikan, is there anywhere in the e2e tests that we test adding a new Load? That would be a good place to check that the added Load step is placed in the correct location.

bsrikan:
> @bsrikan, is there anywhere in the e2e tests that we test adding a new Load? That would be a good place to check that the added Load step is placed in the correct location.

Yes we do. defaultIngestion.spec covers that

wooldridge:
> > @bsrikan, is there anywhere in the e2e tests that we test adding a new Load? That would be a good place to check that the added Load step is placed in the correct location.
> 
> Yes we do. defaultIngestion.spec covers that

Would a test for this go here?

https://github.com/saarimrahman/marklogic-data-hub/blob/DHFPROD-5896/marklogic-data-hub-central/ui/e2e/cypress/integration/curation/load/defaultIngestion.spec.tsx#L168

@bsrikan or should be keep this testing in the RTL unit tests?

briantang:
> > > @bsrikan, is there anywhere in the e2e tests that we test adding a new Load? That would be a good place to check that the added Load step is placed in the correct location.
> > 
> > 
> > Yes we do. defaultIngestion.spec covers that
> 
> Would a test for this go here?
> 
> https://github.com/saarimrahman/marklogic-data-hub/blob/DHFPROD-5896/marklogic-data-hub-central/ui/e2e/cypress/integration/curation/load/defaultIngestion.spec.tsx#L168
> 
> @bsrikan or should be keep this testing in the RTL unit tests?

My thinking was that sorting/ordering was more of a unit test to be handled in RTL rather than cypress. Please let us know your thoughts @bsrikan

marklogic-data-hub-central/ui/src/components/load/load-card.test.tsx
```diff
@@ -67,6 +67,14 @@ describe('Load Card component', () => {
     //Click on 'Yes' button
     fireEvent.click(getByTestId('testLoadXML-to-FlowA-Confirm'));
 
+    //verify Load cards are ordered by Last Updated
```
bsrikan, line 70: Is this the right place to add these assertions. Previous step is clicking on Yes to add to existing flow and redirected to run tile, so this is like breaking the chain of events. This scenario probably deserves to be a separate test. You could add it to this one before load step is added to a flow, but then you also would have to update the test description.

briantang: Okay, I will work on refactoring the test here in the same file then. Thanks!

wooldridge: > Is this the right place to add these assertions. Previous step is clicking on Yes to add to existing flow and redirected to run tile, so this is like breaking the chain of events. This scenario probably deserves to be a separate test. You could add it to this one before load step is added to a flow, but then you also would have to update the test description.

@bsrikan but generally are we doing roundtrip data CRUD in the unit tests? Seems like when we test POSTs, we're just testing that the POST gets called and not going about checking the state of the existing payload for that view. I could be wrong. Is there an example similar to this? I will look...

bsrikan: You can test for something like this:

```
//before update
    expect(loadCards[0]).toHaveTextContent('Add New');
    expect(loadCards[1]).toHaveTextContent('testLoadXML');
    expect(loadCards[2]).toHaveTextContent('testLoad123');
    expect(loadCards[3]).toHaveTextContent('testLoad');

//update timestamp for testLoadXML
data.loadData.data[1].lastUpdated = "1990-04-15T14:22:54.057519-07:00"

//after update
    expect(loadCards[0]).toHaveTextContent('Add New');
    expect(loadCards[2]).toHaveTextContent('testLoad123');
    expect(loadCards[3]).toHaveTextContent('testLoad');
    expect(loadCards[1]).toHaveTextContent('testLoadXML');
```

briantang: good idea, thanks. Will work on that.

wooldridge: @bsrikan, nice. We can avoid mucking with the CRUD and just change the source data and see that the UI adjusts.

wooldridge: @briantang, I like @bsrikan's suggestion of making a separate test.

briantang: Noted, will do. Thanks all :)

bsrikan: @wooldridge you are right about the roundtrip. For RTL unit tests, our mock data is setting up the response for POST request. So its like testing with the assumption that server responds as expected. Now its upto the UI to render that correctly. The same is tested with scenario above. We check for the order of load cards as is. Then update the timestamp, which is to mock the edit/save load configuration, when the timestamp is updated and verify the order of load cards again.
The only roundtrip from server for this ticket is the timestamp update on edit/save. But is it worth a Cypress test, probably not IMO. If a scenario comes up in the future where we edit the settings or step details, we could additionally test for placement based on last updated.

briantang:
RTL test modification is pushed, ready for re-review.

bsrikan (approved):

marklogic-data-hub-central/ui/src/components/load/load-card.test.tsx
bsrikan: @wooldridge we were not able to test as suggested; rerender was not calling useEffect and so we were not able to verify the sort order after timestamp was updated within the test. Instead, the mock data itself has unordered list, so we verify that it renders the cards in the expected order.

Sanjeevani19 (approved):

xnikhil08 (approved):

wooldridge (approved):

PR #4583, opened 2020-09-16
DHFPROD-2962:Add tests to ensure invalid entity model and without baseUri aren't inserted

srinathgit:
### Description
Add tests to ensure invalid entity model and without baseUri aren't inserted
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [ ] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/test/java/com/marklogic/hub/entity/EntityServicesAlignmentTest.java
+++ b/marklogic-data-hub/src/test/java/com/marklogic/hub/entity/EntityServicesAlignmentTest.java
@@ -1,5 +1,6 @@
 package com.marklogic.hub.entity;
 
+import com.fasterxml.jackson.databind.ObjectMapper;
 import com.marklogic.client.DatabaseClient;
 import com.marklogic.client.document.DocumentWriteSet;
 import com.marklogic.client.document.GenericDocumentManager;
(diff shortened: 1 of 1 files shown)


marklogic-data-hub/src/test/java/com/marklogic/hub/entity/EntityServicesAlignmentTest.java
```diff
@@ -42,10 +47,85 @@ public void testDeployTDEWithNoEntities() {
...
+            "        }\n" +
+            "      }\n" +
+            "    }\n" +
+            "  }\n" +
+            "}";
+        File entityFile = getHubProject().getHubEntitiesDir().resolve("Customer.entity.json").toFile();
+        try {
+            mapper.writerWithDefaultPrettyPrinter().writeValue(entityFile, mapper.readTree(entity));
+            installUserArtifacts();
+            assertTrue(false, "Inserting entity without baseUri should fail");
```
akshaysonvane, line 74: Can we use `assertThrows` with the Exception that we are expecting? 
Currently, the test will pass in multiple cases like network failure, failure to write the entity file, etc.

rjrudin: Use "fail" instead - e.g. fail("Expected a validation error to be thrown").

Note though that a missing baseUri is fine, it defaults to example.org by design.

marklogic-data-hub/src/test/java/com/marklogic/hub/entity/EntityServicesAlignmentTest.java
```diff
@@ -42,10 +47,85 @@ public void testDeployTDEWithNoEntities() {
...
+            "        }\n" +
+            "      }\n" +
+            "    }\n" +
+            "  }\n" +
+            "}";
+        File entityFile = getHubProject().getHubEntitiesDir().resolve("Customer.entity.json").toFile();
+        try {
+            mapper.writerWithDefaultPrettyPrinter().writeValue(entityFile,  mapper.readTree(entity));
+            installUserArtifacts();
+            assertTrue(false, "Inserting invalid entity should fail");
```
rjrudin, line 116: Same thing, fail here.

marklogic-data-hub/src/test/java/com/marklogic/hub/entity/EntityServicesAlignmentTest.java
```diff
@@ -42,10 +47,85 @@ public void testDeployTDEWithNoEntities() {
...
+        } catch (IOException e) {
+            logger.info("Cannot insert invalid entity");
+        }
+        finally {
+            FileUtils.deleteQuietly(entityFile);
```
rjrudin, line 121: No need to do this, the hub project directory gets cleared out at the start of every test.

marklogic-data-hub/src/test/java/com/marklogic/hub/entity/EntityServicesAlignmentTest.java
```diff
@@ -42,10 +47,85 @@ public void testDeployTDEWithNoEntities() {
...
+            "    }\n" +
+            "  }\n" +
+            "}";
+        File entityFile = getHubProject().getHubEntitiesDir().resolve("Customer.entity.json").toFile();
+        try {
+            mapper.writerWithDefaultPrettyPrinter().writeValue(entityFile,  mapper.readTree(entity));
+            installUserArtifacts();
+            assertTrue(false, "Inserting invalid entity should fail");
+        } catch (IOException e) {
+            logger.info("Cannot insert invalid entity");
```
rjrudin, line 118: For catch expected exceptions, I recommend: "Caught expected exception due to invalid entity: " + e.getMessage(). 

In fact, it's preferable to make an assertion on the contents of the message so you got the expected error. I'm surprised an IOException is being thrown - don't we want to catch something like an ML FailedRequestException?

rjrudin (changes requested):

PR #4585, opened 2020-09-17
DHFPROD-5962: Set default session timeout value to 10 minutes

akshaysonvane:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [N/A] Added Tests
  

- ##### Reviewer:

- [N/A] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/src/assets/mock-data/user-context-mock.ts
+++ b/marklogic-data-hub-central/ui/src/assets/mock-data/user-context-mock.ts
@@ -1,4 +1,5 @@
 import { UserContextInterface, IUserContextInterface } from '../../types/user-types';
+import {MAX_SESSION_TIME} from "../../config/application.config";
 
 const defaultUserData: UserContextInterface = {
   name: '',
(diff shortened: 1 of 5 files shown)


marklogic-data-hub-central/ui/src/assets/mock-data/user-context-mock.ts
```diff
@@ -9,7 +9,7 @@ const defaultUserData: UserContextInterface = {
     type: ''
   },
   pageRoute: '/',
-  maxSessionTime: 300
+  maxSessionTime: 600
```
rjrudin, line 12: Seems like this would be worth making into a constant somewhere so it doesn't have to be changed in many places the next time we change this value

akshaysonvane: Good idea; updated the PR

rjrudin (approved):

srinathgit (approved):

ryanjdew (approved):

PR #4587, opened 2020-09-17
DHFPROD-5990: Creating schemas for Job and Batch documents

rjrudin:
### Description

These aren't intended to be 100% complete yet in terms of descriptions for every property, but rather to at least provide visibility to what's in these documents without having to run a flow and look at the jobs database. 

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


anu3990 (approved):

SameeraPriyathamTadikonda:
Where can we dump all the metadata from the spark into the job document. I assume we will not have any steps from the glue connector.

Also where will we update the errors from glue/spark will it be in job document or should we create a batch document for that?

rjrudin:
This is just documenting the schema as-is, so it doesn't show any plans for the spark metadata. But we'd likely just toss a "sparkMetadata" top-level property in and throw whatever want in there. That will most likely include the error, as there's not an existing spot for an error to be captured.

SameeraPriyathamTadikonda (approved):

srinathgit (approved):

PR #4589, opened 2020-09-17
DHFPROD-5991:Add test to run flow with stemming searches set to advanced

srinathgit:
## Description
Add test to run flow with stemming searches set to "advanced"
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/test/java/com/marklogic/hub/flow/impl/RunFlowWithStemmingEnabledTest.java
+++ b/marklogic-data-hub/src/test/java/com/marklogic/hub/flow/impl/RunFlowWithStemmingEnabledTest.java
@@ -1,0 +1,67 @@
+package com.marklogic.hub.flow.impl;
+
+import com.marklogic.hub.AbstractHubCoreTest;
+import com.marklogic.hub.DatabaseKind;
+import com.marklogic.hub.HubConfig;
+import com.marklogic.hub.flow.FlowInputs;
+import com.marklogic.hub.flow.RunFlowResponse;
+import com.marklogic.hub.impl.Versions;
+import com.marklogic.hub.job.JobStatus;
+import com.marklogic.mgmt.api.API;
+import com.marklogic.mgmt.api.database.Database;
+import com.marklogic.mgmt.resource.databases.DatabaseManager;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assumptions.assumeTrue;
+
+public class RunFlowWithStemmingEnabledTest extends AbstractHubCoreTest {
+
+    /*
+     https://bugtrack.marklogic.com/54003
+     When stemmed search is set to "advanced", fetching legacy mappings (and/or step definition) fails and hence running
+     testFlow would fail. After that if db is reindexed, fetching of artifacts succeeds. In order to apply the fix for
+     bug 54003 in server, "Reindex JSON language property” trace event has to be added and the test should be run on
+     nightly servers. With the fix applied, the flow will fail after reindexing as well as queries requires 'lang=zxx'
+     to be included so as to fetch artifacts.
+     */
+    @Test
+    void runFlowWithStemmingSearch() {
+        assumeTrue(new Versions(getHubConfig()).getMLVersion().getMajor() >= 10);
+        try{
+            enableAdvancedStemming(true);
+            installProjectInFolder("flow-runner-test");
+
+            RunFlowResponse resp = runFlow(new FlowInputs("testFlow"));
+            assertEquals(JobStatus.STOP_ON_ERROR.toString(), resp.getJobStatus().toLowerCase());
+
+            reindexDatabase();
+
+            resp = runFlow(new FlowInputs("testFlow"));
+            assertEquals(JobStatus.FINISHED.toString(), resp.getJobStatus().toLowerCase());
+        }
+        finally {
+            enableAdvancedStemming(false);
+        }
+    }
+
+    private void reindexDatabase(){
+        HubConfig hubConfig = runAsAdmin();
+        DatabaseManager dbManager = new DatabaseManager(hubConfig.getManageClient());
+        dbManager.reindexDatabase(hubConfig.getDbName(DatabaseKind.STAGING));
+        logger.info("Starting to reindex staging database");
+        waitForReindex(hubConfig, HubConfig.DEFAULT_STAGING_NAME);
+    }
+
+    private void enableAdvancedStemming(boolean stemming){
+        HubConfig hubConfig = runAsDataHubDeveloper();
+        Database db = new Database(new API(hubConfig.getManageClient()), hubConfig.getDbName(DatabaseKind.STAGING));;
+        if(stemming){
+            db.setStemmedSearches("advanced");
+        }
+        else {
+            db.setStemmedSearches("off");
+        }
+        db.save();
+    }
+}
--- a/marklogic-data-hub/src/testFixtures/java/com/marklogic/hub/test/AbstractHubTest.java
+++ b/marklogic-data-hub/src/testFixtures/java/com/marklogic/hub/test/AbstractHubTest.java
@@ -529,6 +529,35 @@ public abstract class AbstractHubTest extends TestObject {
         return response;
     }
 
+    protected void waitForReindex(HubConfig hubConfig, String database){
+        String query = "(\n" +
+            "  for $forest-id in xdmp:database-forests(xdmp:database('" + database + "'))\n" +
+            "  return xdmp:forest-status($forest-id)//*:reindexing\n" +
+            ") = fn:true()";
+        boolean currentStatus;
+        int attempts = 125;
+        boolean previousStatus = false;
+        do{
+            sleep(200L);
+            currentStatus = Boolean.parseBoolean(hubConfig.newStagingClient().newServerEval().xquery(query).evalAs(String.class));
+            if(currentStatus){
+                logger.info("Reindexing staging database");
+            }
+            if(!currentStatus && previousStatus){
+                logger.info("Finished reindexing staging database");
+                return;
+            }
+            else{
+                previousStatus = currentStatus;
+            }
+            attempts--;
+        }
+        while(attempts > 0);
+        if(currentStatus){
+            logger.warn("Reindexing is taking more than 25 seconds");
+        }
+    }
+
     protected HubProjectImpl getHubProject() {
         return (HubProjectImpl) getHubConfig().getHubProject();
     }


marklogic-data-hub/src/test/java/com/marklogic/hub/flow/impl/FlowRunnerTest.java
```diff
@@ -81,6 +83,32 @@ public void testRunFlow() {
...
+            testRunFlow();
+        }
+        finally {
+            updateStemmingOnStagingDb(false);
+        }
+
+    }
+
+    private void updateStemmingOnStagingDb(boolean enableStemming) {
+        HubConfig adminConfig = runAsDataHubDeveloper();
```
rjrudin, line 99: Just call this "hubConfig", as "adminConfig" implies it's an admin user.

marklogic-data-hub/src/test/java/com/marklogic/hub/flow/impl/FlowRunnerTest.java
```diff
@@ -81,6 +83,32 @@ public void testRunFlow() {
         Assertions.assertTrue(file.contains("ingest.csv"));
     }
 
+    @Test
+    void runFlowWithStemmingSearch(){
+        try{
+            updateStemmingOnStagingDb(true);
+            testRunFlow();
```
rjrudin, line 90: Does this break on the nightly?

rjrudin (changes requested):

marklogic-data-hub/src/test/java/com/marklogic/hub/flow/impl/FlowRunnerTest.java
srinathgit: Initially, I thought I needed to get runAsAdmin(), so had it named 'adminConfig' but later realized I could do it as developer, but forgot to rename it. Will do

marklogic-data-hub/src/test/java/com/marklogic/hub/flow/impl/FlowRunnerTest.java
srinathgit: Yes, it breaks it today's nightly

rjrudin (approved):

ryanjdew (approved):

marklogic-data-hub/src/test/java/com/marklogic/hub/flow/impl/FlowRunnerTest.java
```diff
@@ -81,6 +84,89 @@ public void testRunFlow() {
...
+            dbManager.reindexDatabase(hubConfig.getDbName(DatabaseKind.STAGING));
+            String query = "(\n" +
+                "  for $forest-id in xdmp:database-forests(xdmp:database('data-hub-STAGING'))\n" +
+                "  return xdmp:forest-status($forest-id)//*:reindexing\n" +
+                ") = fn:true()";
+            StringHandle stringHandle;
+            String status;
+            int attempts = 5;
+            do{
+                Thread.sleep(5000L);
```
rjrudin, line 114: You can use "sleep" in the parent class so you don't have to deal with InterruptedException.

rjrudin: Also, let's get this all into a separate method called "waitForReindexingToFinish". And limit the sleep to maybe 500ms or so - the call to check on it is extremely fast, so we want this loop to finish as soon as the reindexer is done.

marklogic-data-hub/src/test/java/com/marklogic/hub/flow/impl/StemmingTest.java
```diff
@@ -0,0 +1,102 @@
...
+import com.marklogic.mgmt.resource.databases.DatabaseManager;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Assumptions;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.springframework.beans.factory.annotation.Autowired;
+
+import static org.junit.jupiter.api.Assertions.fail;
+
+public class StemmingTest extends AbstractHubCoreTest {
```
rjrudin, line 21: For consistency, can you name this "RunFlowWithStemmingEnabledTest"?

marklogic-data-hub/src/test/java/com/marklogic/hub/flow/impl/StemmingTest.java
```diff
@@ -0,0 +1,102 @@
...
+
+    Database db;
+    @Autowired
+    FlowRunnerImpl flowRunner;
```
rjrudin, line 25: Should this go to develop or 5.4-develop - is there any benefit having it on develop? 

If you go to 5.4-develop, you can use the runFlow method that handles everything for you. Actually, that's on develop too. 

In any case, don't use Autowired here - that can causes problems with parallel tests since FlowRunnerImpl isn't a thread-safe class.

marklogic-data-hub/src/test/java/com/marklogic/hub/flow/impl/StemmingTest.java
```diff
@@ -0,0 +1,102 @@
...
+
+    @BeforeEach
+    public void setupEach() {
+        HubConfig hubConfig = runAsAdmin();
```
rjrudin, line 29: Just curious, shouldn't a data-hub-developer be able to do this?

marklogic-data-hub/src/test/java/com/marklogic/hub/flow/impl/StemmingTest.java
```diff
@@ -0,0 +1,102 @@
...
+     to be included so as to fetch artifacts.
+     */
+    @Test
+    void runFlowWithStemmingSearch() {
+        Versions.MarkLogicVersion mlVersion = versions.getMLVersion();
+        Assumptions.assumeTrue(mlVersion.getMajor() >=10);
+
+        try{
+            RunFlowResponse resp = runFlow();
+            Assertions.assertTrue(JobStatus.STOP_ON_ERROR.toString().equalsIgnoreCase(resp.getJobStatus()));
```
rjrudin, line 50: assertTrue/False should really only be used when you're testing against a boolean variable. The problem is that if this fails, you don't see what the two values are. 

When you're not sure about case, just lowercase both, though you don't need to do that since JobStatus is already lowercased - so e.g. assertEquals(JobStatus.STOP_ON_ERROR.toString(), resp.getJobStatus().toLowerCase())

marklogic-data-hub/src/test/java/com/marklogic/hub/flow/impl/StemmingTest.java
```diff
@@ -0,0 +1,102 @@
...
+        }
+        catch (Exception e){
+            logger.error("Flow run failed;cause: "+ e.getMessage());
+        }
+
+        flowRunner.awaitCompletion();
+        return resp;
+    }
+
+    private void waitForReIndex(HubConfig hubConfig){
```
rjrudin, line 81: If this is reliable, I think it's a good thing to move up to AbstractHubTest. But for consistency, do "Reindex" instead of "ReIndex", as ML treats it as one word. 

I'm also going to open a ticket for this in ml-app-deployer, this would be a nice feature to have in there - e.g. new Database(...).waitForReindex().

marklogic-data-hub/src/test/java/com/marklogic/hub/flow/impl/StemmingTest.java
```diff
@@ -0,0 +1,102 @@
...
+            "  for $forest-id in xdmp:database-forests(xdmp:database('data-hub-STAGING'))\n" +
+            "  return xdmp:forest-status($forest-id)//*:reindexing\n" +
+            ") = fn:true()";
+        StringHandle stringHandle;
+        String status;
+        int attempts = 1;
+        do{
+            sleep(1000L);
+            stringHandle = new StringHandle();
+            hubConfig.newStagingClient().newServerEval().xquery(query).eval().next().get(stringHandle);
```
rjrudin, line 92: One tweak to make this simpler:

    String status = hubConfig...xquery(query).evalAs(String.class);

Then you don't need a StringHandle, nor do you need to fiddle with an iterator.

rjrudin (changes requested):

marklogic-data-hub/src/test/java/com/marklogic/hub/flow/impl/RunFlowWithStemmingEnabledTest.java
```diff
@@ -0,0 +1,61 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assumptions.assumeTrue;
+
+public class RunFlowWithStemmingEnabledTest extends AbstractHubCoreTest {
+
+    Database db;
+
+    @BeforeEach
+    public void setupEach() {
+        HubConfig hubConfig = runAsDataHubDeveloper();
```
rjrudin, line 25: I believe there's an issue here in that if the version of ML isn't supported, then the call to set stemmed searches to "off" won't be executed. So I'd just move this to the test. 

In fact, I think private methods can make this read nicely:

- enableAdvancedStemming();
- run the flow and verify an error occurs (I think it'd be good to assert on the contents of stepOutput too, as we can verify there are words in there indicating that a step definition was not found)
- reindexDatabase()
- run the flow again and verify it succeeds

rjrudin (changes requested):

marklogic-data-hub/src/test/java/com/marklogic/hub/flow/impl/RunFlowWithStemmingEnabledTest.java
srinathgit: I think moving assumeTrue(new Versions(getHubConfig()).getMLVersion().getMajor() >= 10);  to the first line of setupEach () should resolve that issue right ?

rjrudin: Since you only have the one test method, I think it's simpler to just put everything in that method. But if you want to keep the BeforeEach method, then I think assumeTrue will do the trick there.

rjrudin (approved):

marklogic-data-hub/src/testFixtures/java/com/marklogic/hub/test/AbstractHubTest.java
```diff
@@ -529,6 +529,26 @@ protected RunFlowResponse runFlow(FlowInputs flowInputs) {
...
 
+    protected void waitForReindex(HubConfig hubConfig, String database){
+        String query = "(\n" +
+            "  for $forest-id in xdmp:database-forests(xdmp:database('" + database + "'))\n" +
+            "  return xdmp:forest-status($forest-id)//*:reindexing\n" +
+            ") = fn:true()";
+        String status;
+        int attempts = 0;
+        do{
+            sleep(15000L);
```
srinathgit, line 540: The test keeps failing in the pipeline even though it passed in my local. I am increasing the sleep time to see of if the test passes. I will then work to see why the query isn't reliable.

rjrudin: You could try running this as a little main program (not the test itself, just this code - we don't want to clear the database) against the HC nightly index, just to test against an ML instance that isn't running on your laptop. I recommend checking with the DB team too to see if there's a recommended script for checking to see if the database is reindexing or not.

ryanjdew (dismissed):

rjrudin (approved):

fsnow (approved):

PR #4590, opened 2020-09-18
DHFPROD-5943: Renaming the connector to Spark Connector

SameeraPriyathamTadikonda:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/sql/sources/v2/HubDataSource.java
```diff
@@ -13,17 +13,33 @@
...
 import org.apache.spark.sql.sources.v2.writer.DataWriterFactory;
 import org.apache.spark.sql.sources.v2.writer.WriterCommitMessage;
 import org.apache.spark.sql.types.StructType;
 
 import java.util.Map;
+import java.util.Optional;
 
-public class MarkLogicWriter implements DataSourceWriter {
+public class HubDataSource extends LoggingObject implements WriteSupport {
+
```
anu3990, line 33: HubDataSource should be inside the writer folder because it implements WriteSupport. It cannot be used as a generic class. For Reader we will have to create a new class that implements ReadSupport.

marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/sql/sources/v2/HubDataSource.java
```diff
@@ -13,17 +13,33 @@
...
+    @Override
+    public Optional<DataSourceWriter> createWriter(String writeUUID, StructType schema, SaveMode mode, DataSourceOptions options) {
+        logger.info("Creating MarkLogicWriter");
+        return Optional.of(new MarkLogicWriter(options.asMap(), schema){
+
+        });
+    }
+}
+class MarkLogicWriter implements DataSourceWriter {
```
anu3990, line 42: I am not sure if we can have MarkLogicWriter as an inner class. Is this setup working.?

anu3990 (changes requested):

marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/sql/sources/v2/HubDataSource.java
rjrudin: I am thinking that since WriteSupport is in the "v2" package and not "v2.writer", we'll eventually have this implement ReadSupport as well. Here's an example of that - https://www.bugdbug.com/post/speed-up-apache-spark-operations-using-a-custom-data-source . 

I think that makes sense conceptually - we have a single DataSource which supports Read and Write. That makes life easier for Ernie, who only has to worry about one DataSource class now.

marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/sql/sources/v2/HubDataSource.java
```diff
@@ -13,26 +13,42 @@
...
 import java.util.Map;
+import java.util.Optional;
 
-public class MarkLogicWriter implements DataSourceWriter {
+public class HubDataSource extends LoggingObject implements WriteSupport {
+
+    @Override
+    public Optional<DataSourceWriter> createWriter(String writeUUID, StructType schema, SaveMode mode, DataSourceOptions options) {
+        logger.info("Creating HubDataSourceWriter");
+        return Optional.of(new HubDataSourceWriter(options.asMap(), schema){
```
rjrudin, line 37: I believe you can remove the empty braces here.

marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/sql/sources/v2/HubDataSource.java
```diff
@@ -13,26 +13,42 @@
...
+
+        });
+    }
+}
+class HubDataSourceWriter implements DataSourceWriter {
```
rjrudin, line 42: I like this approach of tossing this class in here as a non-public one. I think we'll want to do that for HubDataSourceReader too in the future.

rjrudin (approved):

anu3990 (approved):

PR #4591, opened 2020-09-18
DHFPROD-5531: Adding e2e test for mapping and load steps having custom Hook

xnikhil08:
### Description
Added e2e tests mapping and load steps having custom Hook.

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/e2e/cypress/integration/curation/curate/mapping.spec.tsx
+++ b/marklogic-data-hub-central/ui/e2e/cypress/integration/curation/curate/mapping.spec.tsx
@@ -27,15 +27,15 @@ describe('Mapping', () => {
 
   after(() => {
     cy.loginAsDeveloper().withRequest();
-    cy.deleteSteps('ingestion', 'loadOrderProcessor');
-    cy.deleteSteps('mapping', 'mapOrderProcessor');
+    cy.deleteSteps('ingestion', 'loadOrder');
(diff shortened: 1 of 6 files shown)


marklogic-data-hub-central/ui/e2e/cypress/integration/curation/curate/mapping.spec.tsx
```diff
@@ -112,5 +121,9 @@ describe('Mapping', () => {
     browsePage.getTableViewSourceIcon().click();
     cy.contains('mappedOrderDate');
     cy.contains('categoryCode');
+
+    //Verifying the properties added by load and mapping custom hooks respectively
+    cy.contains('primaryKey');
+    cy.contains('uriFromCustomHook');
```
bsrikan, line 127: can we pls add .should('be.visible') to all the four cy.contains()
I could be wrong, but I think, contains is a query and doesnt assert, so these steps might not fail if there is no mappedOrderDate or primaryKey in the DOM.

brucean52: cy.contains() will fail if the element is not found.

brucean52 (approved):

timur-isangulov (approved):

PR #4596, opened 2020-09-18
DHFPROD-5974: Ingest documents with user-defined prefix and guaranteed unique URI

anu3990:
### Description
This pull request focuses on the changes needed for - https://project.marklogic.com/jira/browse/DHFPROD-5974

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-spark-connector/src/test/java/com/marklogic/hub/spark/sql/sources/v2/WriteDataTest.java
+++ b/marklogic-data-hub-spark-connector/src/test/java/com/marklogic/hub/spark/sql/sources/v2/WriteDataTest.java
@@ -1,7 +1,6 @@
 package com.marklogic.hub.spark.sql.sources.v2;
 
-import com.fasterxml.jackson.databind.JsonNode;
-import com.marklogic.client.io.JacksonHandle;
+import com.marklogic.client.eval.EvalResultIterator;
 import org.apache.spark.sql.SaveMode;
 import org.apache.spark.sql.catalyst.InternalRow;
 import org.apache.spark.sql.catalyst.expressions.GenericInternalRow;
@@ -20,7 +19,7 @@ import java.io.IOException;
 import java.util.Map;
 import java.util.Optional;
 
-import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.*;
 
 public class WriteDataTest extends AbstractSparkConnectorTest {
 
@@ -48,6 +47,27 @@ public class WriteDataTest extends AbstractSparkConnectorTest {
         verifyFruitCount(3, "The commit call should result in the 3rd fruit being ingested");
     }
 
+
+    @Test
+    public void testBulkIngestWithoutUriPrefix() throws IOException {
+        DataWriter<InternalRow> dataWriter = buildDataWriter("1", "");
+        dataWriter.write(buildRow("pineapple", "green"));
+
+        String uriQuery = "cts.uris('', null, cts.andQuery([\n" +
+            "  cts.jsonPropertyValueQuery('fruitName', 'pineapple')\n" +
+            "]))";
+
+        EvalResultIterator uriQueryResult = getHubClient().getStagingClient().newServerEval().javascript(uriQuery).eval();
+        assertTrue(uriQueryResult.hasNext());
+        String uri = uriQueryResult.next().getString();
+
+        assertTrue(uri.endsWith(".json"));
+
+        assertFalse(uri.startsWith("/"), "If the user wants the URI to start with a forward slash, the user must provide one. " +
+            "If the user doesn't, then it's assumed that the user doesn't want a forward slash at the start of the URI, so the endpoint will not add one automatically.");
+        assertFalse(uriQueryResult.hasNext());
+    }
+
     /**
      * Spark will do all of this in the real world - i.e. a user will specify the entry class and the set of options.
      * But in a test, we need to do that ourselves. So we create the DataSource class, build up the params, and then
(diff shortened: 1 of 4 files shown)


SameeraPriyathamTadikonda (approved):

rjrudin (dismissed):

anu3990:
Rebased with the develop branch and also fixed an additional test that was failing - "BulkIngestTest.java"

marklogic-data-hub-spark-connector/src/test/java/com/marklogic/hub/spark/sql/sources/v2/WriteDataTest.java
```diff
@@ -91,10 +94,24 @@ private void verifyFruitCount(int expectedCount, String message) {
...
-            assertEquals("apple", doc.get("envelope").get("instance").get("fruitName").asText());
-            assertEquals("red", doc.get("envelope").get("instance").get("fruitColor").asText());
+        String uriQuery = "cts.uriMatch('/testFruit/**')";
+        EvalResultIterator uriQueryResult = getHubClient().getStagingClient().newServerEval().javascript(uriQuery).eval();
+        AtomicBoolean foundRedApple = new AtomicBoolean(false);
+        AtomicBoolean foundYellowBanana = new AtomicBoolean(false);
+        uriQueryResult.iterator().forEachRemaining(item -> {
+            final String expectUri = item.getString();
+            JsonNode doc = getHubClient().getStagingClient().newJSONDocumentManager().read(expectUri, new JacksonHandle()).get();
+            if((doc.get("envelope").get("instance").get("fruitName").asText().equals("apple")) &&
```
rjrudin, line 104: This test is a little brittle in that if the structure of the document doesn't match, we'll get a null-pointer exception. 

Instead of doing a cts.uriMatch, I recommend doing a cts.uris query like this:

    cts.uris(null, [limit='1'], cts.andQuery([
      cts.directoryQuery("/testFruit/"),
      cts.jsonPropertyValueQuery("fruitName", someVariable)
    ]))

You can use newServerEval to execute that and then verify that you get 1 result back and that it starts with "/testFruit/". 

That results in a much simpler test and also much less chance of a null-pointer occurring.

rjrudin (changes requested):

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/ingestion/bulkIngester.sjs
```diff
@@ -36,7 +36,7 @@ const inputs =
       [{UNKNOWN: input}];
 inputs.forEach(record => {
   state.next = state.next + 1;
-  const uri = (state.prefix) + '/' + (work.taskId) + '/' + (state.next) + '.json';
+  const uri = (state.uriprefix) + '/' + sem.uuidString() + '.json';
```
rjrudin, line 39: Just realized that we should leave out "/". That allows Ernie to have URIs of the form "(uuid).json" if he wants. That's consistent with how MLCP works as well - i.e. no "/" is added unless the user configures one.

rjrudin (changes requested):
Looks good, just identified one small tweak.

rjrudin (approved):

SameeraPriyathamTadikonda:
I think we should write another test to verify how connector behaves when the user does not provide "uriprefix"

rjrudin (changes requested):
That's a good point by @SameeraPriyathamTadikonda - we want another test scenario (separate test method) where uriprefix is not set at all, and then we verify that the doc is still written and the URI ends with ".json" and does not start with "/".

SameeraPriyathamTadikonda (approved):

SameeraPriyathamTadikonda (dismissed):

marklogic-data-hub-spark-connector/src/test/java/com/marklogic/hub/spark/sql/sources/v2/WriteDataTest.java
```diff
@@ -48,6 +47,25 @@ void ingestThreeFruitsWithBatchSizeOfTwo() throws IOException {
...
+    @Test
+    public void testBulkIngestWithoutUriPrefix() throws IOException {
+        DataWriter<InternalRow> dataWriter = buildDataWriter("1", "");
+        dataWriter.write(buildRow("pineapple", "green"));
+
+        String uriQuery = "cts.uris('', null, cts.andQuery([\n" +
+            "  cts.jsonPropertyValueQuery('fruitName', 'pineapple')\n" +
+            "]))";
+
+        EvalResultIterator uriQueryResult = getHubClient().getStagingClient().newServerEval().javascript(uriQuery).eval();
```
rjrudin, line 60: No need to change here, but most of the time, it's sufficient to just do this:

    String uri = getHubClient().getStagingClient().newServerEval().javascript(query).evalAs(String.class);

If that fails because there's not a URI, that's totally fine. And we don't need to worry about verifying that there's not another URI - we know the database was cleaned before the test ran, so it's safe to assume there's one or zero URIs.

marklogic-data-hub-spark-connector/src/test/java/com/marklogic/hub/spark/sql/sources/v2/WriteDataTest.java
```diff
@@ -48,6 +47,25 @@ void ingestThreeFruitsWithBatchSizeOfTwo() throws IOException {
...
+        assertTrue(uriQueryResult.hasNext());
+        String uri = uriQueryResult.next().getString();
+
+        assertTrue(uri.endsWith(".json"));
+        assertFalse(uri.startsWith("/"));
```
rjrudin, line 65: Assertion calls are often the best place to document *why* something should be the way it is. In this case, there's a question of - well why doesn't it start with "/"? So I'd do:

    assertFalse(uri.startsWith("/"), "If the user wants the URI to start with a forward slash, the user must provide one. If the user doesn't, then it's assumed that the user doesn't want a forward slash at the start of the URI, so the endpoint will not add one automatically.");

That makes it very clear why it's expected. This isn't needed for every assertion, but this is a good one to document in the above manner.

rjrudin (changes requested):

rjrudin (approved):

SameeraPriyathamTadikonda (approved):

PR #4598, opened 2020-09-18
DHFPROD-5915: Deploying search options when model is created

rahulvudutala:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


rjrudin (approved):

akshaysonvane (commented):
After our discussion on this @rahulvudutala, I realized that since the explorer options are so integral to the app, I think we should generate and deploy them during data hub installation.

Currently, EntityManagerImpl (as part of LoadUserModulesCommand) will not deploy the options if there aren't any entities. I think we can get rid of this optimization and always generate and deploy the options.

An additional advantage here would be to not waste cycles in generating and deploying empty options when we hit the createModel endpoint.

@rjrudin, @ryanjdew do you see any issues if we always generate the options file?

rjrudin:
@akshaysonvane That sounds good to me

rahulvudutala:
will es.searchOptionsGenerate(model) generate search options when there are no entityModels?

akshaysonvane:
> will es.searchOptionsGenerate(model) generate search options when there are no entityModels?

Yes, we can pass in an empty map.

ryanjdew (approved):

PR #4609, opened 2020-09-22
DHFPROD-5942: Assume DHS Usage by default.

SameeraPriyathamTadikonda:
### Description
- Assume DHS usage by default.
- Ability to run tests on DHS
- Running tests using gradle.

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-spark-connector/gradle.properties
+++ b/marklogic-data-hub-spark-connector/gradle.properties
@@ -1,3 +1,7 @@
 junitJupiterVersion=5.4.0
 junitPlatformVersion=1.4.0
 sparkVersion=2.4.5
+
+#Properties to be used for testing locally using gradle
+mlHost=localhost
+isDhs=false
--- a/marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/sql/sources/v2/writer/HubDataWriterFactory.java
+++ b/marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/sql/sources/v2/writer/HubDataWriterFactory.java
@@ -34,6 +34,7 @@ public class HubDataWriterFactory extends LoggingObject implements DataWriterFac
     private HubClient hubClient;
     private Map<String, String> params;
 
+
     /**
      * @param params a map of parameters containing both DHF-supported properties (most likely prefixed with ml* or
      *               hub*) and connector-specific properties. The DHF-supported properties will be used to construct a
@@ -43,13 +44,7 @@ public class HubDataWriterFactory extends LoggingObject implements DataWriterFac
     public HubDataWriterFactory(Map<String, String> params, StructType schema) {
         this.params = params;
         this.schema = schema;
-
-        Properties props = new Properties();
-        params.keySet().forEach(key -> props.setProperty(key, params.get(key)));
-        logger.info("Creating HubClient for host: " + props.getProperty("mlHost"));
-        HubConfigImpl hubConfig = new HubConfigImpl();
-        hubConfig.registerLowerCasedPropertyConsumers();
-        hubConfig.applyProperties(new SimplePropertySource(props));
+        HubConfigImpl hubConfig = buildHubConfig(params);
         this.hubClient = hubConfig.newHubClient();
     }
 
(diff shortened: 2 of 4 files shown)


marklogic-data-hub-spark-connector/gradle.properties
```diff
@@ -1,3 +1,6 @@
 junitJupiterVersion=5.4.0
 junitPlatformVersion=1.4.0
 sparkVersion=2.4.5
+#Properties to be used for testing locally using gradle
```
rjrudin, line 4: Toss a line above this just to space it out from the version numbers. 

Then - these properties are going to be used for testing in any environment, right?

marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/sql/sources/v2/writer/HubDataWriterFactory.java
```diff
@@ -43,13 +44,7 @@
...
         this.params = params;
         this.schema = schema;
-
-        Properties props = new Properties();
-        params.keySet().forEach(key -> props.setProperty(key, params.get(key)));
-        logger.info("Creating HubClient for host: " + props.getProperty("mlHost"));
-        HubConfigImpl hubConfig = new HubConfigImpl();
-        hubConfig.registerLowerCasedPropertyConsumers();
-        hubConfig.applyProperties(new SimplePropertySource(props));
+        HubConfigImpl hubConfig=buildHubConfig(params);
```
rjrudin, line 47: Assuming you're using Intellij, be sure to do a Reformat Code on every file - I have option-command-L as a shortcut for that, I forget if it's a default in Intellij. That'll ensure that we don't get a bunch of changes to this file in another PR when someone does run Reformat Code.

marklogic-data-hub-spark-connector/src/test/java/com/marklogic/hub/spark/sql/sources/v2/AbstractSparkConnectorTest.java
```diff
@@ -52,10 +52,26 @@ protected File getTestProjectDirectory() {
     @Override
     protected HubConfigImpl runAsUser(String username, String password) {
         hubProperties = new Properties();
-        // Assumes localhost for now; this will be improved once this project is developed on the 5.4 branch
-        hubProperties.setProperty("mlHost", "localhost");
+        String mlHost="localhost";
+        String hubDHS="false";
```
rjrudin, line 56: For consistency with other DHF props, do "hubDhs" and "isDhs" (that's consistent with "Ssl", for example).

rjrudin (changes requested):
Looks good, just a few little renames

marklogic-data-hub-spark-connector/gradle.properties
SameeraPriyathamTadikonda: Yes, We can use these in any environment. Should I create a new gradle-dhs.properties file?

rjrudin: Keep them in here - they have to exist in gradle.properties so that you can reference them in build.gradle. The only reason for a "gradle-env.properties" file is if there's an "env" that we want to configure properties for. That isn't the case and likely won't be for this project. We'll always use e.g. "-P" to pass in env-specific properties for things like a Jenkins environment.

marklogic-data-hub-spark-connector/src/test/java/com/marklogic/hub/spark/sql/sources/v2/AbstractSparkConnectorTest.java
SameeraPriyathamTadikonda: Addressed in new the commit.

marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/sql/sources/v2/writer/HubDataWriterFactory.java
SameeraPriyathamTadikonda: Formatted the code

marklogic-data-hub-spark-connector/src/test/java/com/marklogic/hub/spark/sql/sources/v2/AbstractSparkConnectorTest.java
```diff
@@ -52,10 +52,26 @@ protected File getTestProjectDirectory() {
...
+            isDhs = Boolean.parseBoolean(System.getProperty("isDhs"));
+        }
+        if(isDhs){
+            hubDhs="true";
+            hubSsl="true";
+        }
+        hubProperties.setProperty("mlHost",mlHost);
         hubProperties.setProperty("mlUsername", username);
         hubProperties.setProperty("mlPassword", password);
+        hubProperties.setProperty("hubDHS",hubDhs);
```
rjrudin, line 73: This should be "hubDhs", right? I believe "hubDHS" will still work because it gets lowercased, but we should still use hubDhs.

rjrudin (approved):
Looks good, just one tweak.

anu3990 (approved):

PR #4614, opened 2020-09-24
DHFPROD-6030: Prototype of running a step without a flow

rjrudin:
### Description

This depends on two additions:

1. A new endpoint for building an "adhoc" flow based on a stepId
1. That adhocFlow must be included in the options object so that the backend can use it instead of calling getFlow/getFullFlow

This otherwise looks very simple. I don't know that "adhoc" is the right term here, there's likely a more descriptive term for "a temporary flow only containing the one step a user wants to run". 

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [ ] Reviewed Tests


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/endpoints/collector.sjs
```diff
@@ -42,7 +42,13 @@ if (!step) {
 }
 let options = requestParams.options ? JSON.parse(requestParams.options) : {};
 
-let flowDoc= datahub.flow.getFlow(flowName);
+let flowDoc = null;
+if (options.datahubAdHocFlow) {
+  flowDoc = options.datahubAdHocFlow;
+} else {
+  datahub.flow.getFlow(flowName);
```
akshaysonvane, line 49: I believe this should be
```suggestion
  flowDoc = datahub.flow.getFlow(flowName);
```

rjrudin: Thanks, I just fixed this - this won't get merged yet, just a prototype for now.

PR #4618, opened 2020-09-24
DHFPROD-5665: Run match step with new schema

ryanjdew:
### Description

Added a test demostrating at a high level the various parts of match schema. Will be adding additional tests for the lower-level functions.

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


fsnow:
Rebase ?

fsnow (approved):

fsnow (approved):

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/smart-mastering/entity-driven/test-data/entities/Customer.entity.json
```diff
@@ -0,0 +1,98 @@
+{
```
rjrudin, line 1: I highly recommend slimming down the test data to the bare minimum needed. For example, only include properties used for matching - anything else is just noise that makes the test more difficult to understand. Same thing goes for test data - only insert the documents you need to verify that e.g. 2 docs match, but a 3rd doc does not match.

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/smart-mastering/entity-driven/test-data/householdAction.sjs
```diff
@@ -0,0 +1,15 @@
+'use strict';
+
+const sem = require("/MarkLogic/semantics.xqy");
+const householdPredicateIRI = sem.iri('http://marklogic.com/datahub/test/household#memberOf');
+
+function householdAction(uri, matches, options) {
+    declareUpdate();
+    const allURIs = [uri].concat(matches.map((match) => match.uri)).sort();
+    const householdIRI = sem.iri(`http://marklogic.com/datahub/test/household/${xdmp.md5(allURIs)}`);
+    sem.rdfInsert(allURIs.map((uri) => sem.triple(sem.iri(uri), householdPredicateIRI, householdIRI)));
```
rjrudin, line 10: Is inserting data via a match action something we'd encourage? I ask because I'm wondering if we can eventually shift to a model where everything to be written gets stored in-memory first and then is written after all step processing has finished.

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/smart-mastering/matching/include-matches.xqy
```diff
@@ -19,7 +19,7 @@ return (
     test:assert-equal(1, fn:count($def-match/@threshold[. = "Definitive Match"])),
     test:assert-equal(1, fn:count($def-match/@action[. = $constants:MERGE-ACTION])),
     test:assert-exists($def-match/matches),
-    test:assert-equal(7, fn:count($def-match/matches/match))
+    test:assert-equal(8, fn:count($def-match/matches/match), "Wrong count of matches: " || xdmp:describe($def-match, (), ()) )
```
rjrudin, line 22: If this test fails, it'll be very helpful to know why 8 matches are expected, and which 8 they are. Otherwise, a developer has to figure out themselves what should match and what should not.

rjrudin (commented):
I added a few comments about the tests. I'm not familiar enough with the app code to know the correctness of any of the changes. But if this isn't introducing functionality other than supporting the new match schema, I'm fine with the app changes. 

The main idea behind my test comments is that the simpler the tests are in terms of data and entity definitions, the simpler it will be for other developers to understand and debug them when they fail.

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/smart-mastering/matching/include-matches.xqy
ryanjdew: Added tests for the individual matches as well that makes it more clear.

rjrudin (approved):
Looks good, but one thing to consider - in the test content files, I would remove "attachments" and the structured properties unless they're needed by the tests. When we support matching on structured properties, we'll definitely need test docs with structured properties. But until then, I think having structured properties just adds complexity to the tests. Same goes for attachments, unless our tests are verifying something pertaining to the existence of attachments.

fsnow (dismissed):

rjrudin (dismissed):

rjrudin (dismissed):

fsnow (dismissed):

fsnow (approved):

rjrudin (approved):

PR #4619, opened 2020-09-25
DHFPROD-5823: Update amp with "data-hub-security-admin" on servers > 10.0-4.4

srinathgit:
### Description
Update amp with "data-hub-security-admin" on servers > 10.0-4.4
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/dhs/installer/deploy/DeployHubAmpsCommand.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/dhs/installer/deploy/DeployHubAmpsCommand.java
@@ -2,6 +2,7 @@ package com.marklogic.hub.dhs.installer.deploy;
 
 import com.marklogic.appdeployer.command.CommandContext;
 import com.marklogic.appdeployer.command.security.DeployAmpsCommand;
+import com.marklogic.hub.impl.Versions;
 import com.marklogic.mgmt.ManageClient;
 import com.marklogic.mgmt.SaveReceipt;
 import com.marklogic.mgmt.resource.ResourceManager;
@@ -38,12 +39,13 @@ public class DeployHubAmpsCommand extends DeployAmpsCommand {
 
         @Override
         public SaveReceipt updateResource(String payload, String resourceId) {
+
             SaveReceipt receipt = null;
             try{
                 receipt =  super.updateResource(payload, resourceId);
             }
             catch (HttpClientErrorException ex){
-                if (HttpStatus.FORBIDDEN.equals(ex.getStatusCode())) {
+                if (HttpStatus.FORBIDDEN.equals(ex.getStatusCode()) && cannotUpdateAmps()) {
                     logger.error("An error occurred while trying to update an amp: " + ex.getMessage()
                         + ". Updates to amps as a user with the data-hub-security-admin role are not allowed " +
                         "in this version of MarkLogic, so this error is being logged and not thrown. " +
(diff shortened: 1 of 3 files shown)


srinathgit:
@rjrudin 
1. Fayez had mentioned in https://bugtrack.marklogic.com/55309 that dhsa needs "amp-set-roles" to add/remove roles to amps. I added it to dhsa and am able to update the amps. I have modified the test accordingly.
2. If we have to modify "DeployHubAmpsCommand" and throw message based on version then currently dhsa lacks privilege as a versions.getMarkLogicVersion() will fail for a dhsa user (doesn't have eval privilege)

marklogic-data-hub/src/test/java/com/marklogic/hub/dhs/DeployUserAmpsTest.java
```diff
@@ -32,13 +36,16 @@
...
     @BeforeEach
     void checkIfTestsCanBeRun(){
-        Versions.MarkLogicVersion mlVersion = versions.getMLVersion();
+        if(mlVersion == null){
+            mlVersion = versions.getMLVersion();
+        }
         assumeTrue(
             (mlVersion.isNightly() && mlVersion.getMajor() >= 10) ||
-            (mlVersion.getMajor() > 10) ||(mlVersion.getMajor() == 10 && mlVersion.getMinor() >= 440)
+            (mlVersion.getMajor() > 10) ||(mlVersion.getMajor() == 10 && mlVersion.getMinor() >= 404)
```
srinathgit, line 48: 10.0-4.4 corresponds to minor version 404 not 440.

marklogic-data-hub/src/main/resources/hub-internal-config/security/roles/data-hub-security-admin.json
```diff
@@ -35,6 +35,11 @@
       "action": "http://marklogic.com/xdmp/privileges/remove-amp",
       "kind":"execute"
     },
+    {
+      "privilege-name": "amp-set-roles",
```
rjrudin, line 39: Is this privilege only needed on 10.0-4.4 or higher in order for updating an amp to work? If so, add a message to one of the assertions below to state that.

rjrudin (changes requested):

marklogic-data-hub/src/main/java/com/marklogic/hub/dhs/installer/deploy/DeployHubAmpsCommand.java
```diff
@@ -38,12 +39,16 @@ protected String getResourceName() {
 
         @Override
         public SaveReceipt updateResource(String payload, String resourceId) {
+            Versions.MarkLogicVersion mlVersion = new Versions().getMLVersion(getManageClient().getXml("/manage")
```
rjrudin, line 42: For the path, use "/node()/c:version" since we know exactly where to find the version. This prevents us from ever getting two values back, which would cause a break.

Also, I recommend moving all of this into a "private boolean canUpdateAmps" method. And only call it inside the catch block here - we don't need to check this unless we get an exception. And in that private method, put everything in a try/catch block, as we don't ever want an exception being thrown by the version-check code. Instead, do something like this:

    logger.warn("Could not determine MarkLogic version; cause: " + e.getMessage() + "; will assume that user can update amps and will throw original exception");

marklogic-data-hub/src/test/java/com/marklogic/hub/dhs/DeployUserAmpsTest.java
```diff
@@ -32,10 +37,13 @@
     ObjectMapper mapper = new ObjectMapper();
     HubClient operatorHubClient;
     HubClient adminHubClient;
+    Versions.MarkLogicVersion mlVersion;
 
     @BeforeEach
     void checkIfTestsCanBeRun(){
-        Versions.MarkLogicVersion mlVersion = versions.getMLVersion();
+        if(mlVersion == null){
```
rjrudin, line 44: mlVersion will always be null here, so this check should be removed.

rjrudin (changes requested):

rjrudin (approved):

akshaysonvane (approved):

PR #4630, opened 2020-09-28
DHFPROD-5989: Configure workUnit and endpointState in Spark connector

anu3990:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/sql/sources/v2/writer/HubDataWriter.java
+++ b/marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/sql/sources/v2/writer/HubDataWriter.java
@@ -15,10 +15,15 @@
  */
 package com.marklogic.hub.spark.sql.sources.v2.writer;
 
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.marklogic.client.dataservices.InputEndpoint;
 import com.marklogic.client.ext.helper.LoggingObject;
+import com.marklogic.client.io.JacksonHandle;
 import com.marklogic.client.io.StringHandle;
 import com.marklogic.hub.HubClient;
+import org.apache.commons.lang.StringUtils;
 import org.apache.spark.sql.catalyst.InternalRow;
 import org.apache.spark.sql.catalyst.json.JSONOptions;
 import org.apache.spark.sql.catalyst.json.JacksonGenerator;
@@ -28,6 +33,7 @@ import org.apache.spark.sql.sources.v2.writer.WriterCommitMessage;
 import org.apache.spark.sql.types.StructType;
 
 import java.io.ByteArrayInputStream;
+import java.io.IOException;
 import java.io.InputStream;
 import java.io.StringWriter;
 import java.util.ArrayList;
@@ -45,24 +51,25 @@ public class HubDataWriter extends LoggingObject implements DataWriter<InternalR
     /**
      *
      * @param hubClient
-     * @param taskId
      * @param schema
      * @param params contains all the params provided by Spark, which will include all connector-specific properties
      */
-    public HubDataWriter(HubClient hubClient, long taskId, StructType schema, Map<String, String> params) {
+    public HubDataWriter(HubClient hubClient, StructType schema, Map<String, String> params) {
         this.records = new ArrayList<>();
         this.schema = schema;
         this.batchSize = params.containsKey("batchsize") ? Integer.parseInt(params.get("batchsize")) : 100;
 
-        final String apiModulePath = params.containsKey("apipath") ? params.get("apipath") : "/data-hub/5/data-services/ingestion/bulkIngester.api";
-        logger.info("Will write to endpoint defined by: " + apiModulePath);
+        JsonNode endpointParams = determineIngestionEndpointParams(params);
+
+        final String apiPath = endpointParams.get("apiPath").asText();
+        logger.info("Will write to endpoint defined by: " + apiPath);
         this.loader = InputEndpoint.on(
             hubClient.getStagingClient(),
-            hubClient.getModulesClient().newTextDocumentManager().read(apiModulePath, new StringHandle())
+            hubClient.getModulesClient().newJSONDocumentManager().read(apiPath, new StringHandle())
         ).bulkCaller();
 
-        loader.setWorkUnit(new ByteArrayInputStream(("{\"taskId\":" + taskId + "}").getBytes()));
-        loader.setEndpointState(new ByteArrayInputStream(("{\"next\":" + 0 + ", \"uriprefix\":\"" + params.get("uriprefix") + "\"}").getBytes()));
+        this.loader.setEndpointState(new JacksonHandle(endpointParams.get("endpointState")));
+        this.loader.setWorkUnit(new JacksonHandle(endpointParams.get("workUnit")));
     }
 
     @Override
@@ -114,4 +121,41 @@ public class HubDataWriter extends LoggingObject implements DataWriter<InternalR
         return jsonObjectWriter.toString();
     }
 
+    protected JsonNode determineIngestionEndpointParams(Map<String, String> params) {
+        ObjectMapper objectMapper = new ObjectMapper();
+
+        ObjectNode endpointParams;
+        if (params.containsKey("ingestendpointparams")) {
+            try {
+                endpointParams = (ObjectNode)objectMapper.readTree(params.get("ingestendpointparams"));
+            } catch (IOException e) {
+                throw new RuntimeException("Unable to parse ingestendpointparams, cause: " + e.getMessage(), e);
+            }
+        } else {
+            endpointParams = objectMapper.createObjectNode();
+        }
+
+        boolean doesNotHaveApiPath = (!endpointParams.hasNonNull("apiPath") || StringUtils.isEmpty(endpointParams.get("apiPath").asText()));
+        boolean hasWorkUnitOrEndpointState = endpointParams.has("workUnit") || endpointParams.has("endpointState");
+        if (doesNotHaveApiPath && hasWorkUnitOrEndpointState) {
+            throw new RuntimeException("Cannot set workUnit or endpointState in ingestionendpointparams unless apiPath is defined as well.");
+        }
+
+        if (doesNotHaveApiPath) {
+            endpointParams.put("apiPath", "/data-hub/5/data-services/ingestion/bulkIngester.api");
+        }
+
+        // TODO : remove the below else block after java-client-api 5.3 release
+        if (!endpointParams.hasNonNull("endpointState")) {
+            endpointParams.set("endpointState", objectMapper.createObjectNode());
+        }
+
+        if (!endpointParams.hasNonNull("workUnit")) {
+            ObjectNode defaultWorkUnit = objectMapper.createObjectNode();
+            defaultWorkUnit.put("uriprefix", params.get("uriprefix")!=null ? params.get("uriprefix"):"");
+            endpointParams.set("workUnit", defaultWorkUnit);
+        }
+
+        return endpointParams;
+    }
 }
--- a/marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/sql/sources/v2/writer/HubDataWriterFactory.java
+++ b/marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/sql/sources/v2/writer/HubDataWriterFactory.java
@@ -50,10 +50,7 @@ public class HubDataWriterFactory extends LoggingObject implements DataWriterFac
 
     @Override
     public DataWriter<InternalRow> createDataWriter(int partitionId, long taskId, long epochId) {
-        if (logger.isDebugEnabled()) {
-            logger.debug("Creating DataWriter with taskId: " + taskId);
-        }
-        return new HubDataWriter(hubClient, taskId, schema, params);
+        return new HubDataWriter(hubClient, schema, params);
     }
 
     protected HubConfigImpl buildHubConfig(Map<String, String> params) {
@@ -63,7 +60,7 @@ public class HubDataWriterFactory extends LoggingObject implements DataWriterFac
         props.setProperty("hubssl", "true");
         // hubDHS and hubSsl passed in the params Map will be overridden.
         params.keySet().forEach(key -> props.setProperty(key, params.get(key)));
-        logger.info("Creating HubClient for host: " + props.getProperty("mlHost"));
+        logger.info("Creating HubClient for host: " + props.getProperty("mlhost"));
         HubConfigImpl hubConfig = new HubConfigImpl();
         hubConfig.registerLowerCasedPropertyConsumers();
         hubConfig.applyProperties(new SimplePropertySource(props));
--- a/marklogic-data-hub-spark-connector/src/test/java/com/marklogic/hub/spark/sql/sources/v2/WriteDataTest.java
+++ b/marklogic-data-hub-spark-connector/src/test/java/com/marklogic/hub/spark/sql/sources/v2/WriteDataTest.java
@@ -1,36 +1,20 @@
 package com.marklogic.hub.spark.sql.sources.v2;
 
+import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.marklogic.client.eval.EvalResultIterator;
-import org.apache.spark.sql.SaveMode;
 import org.apache.spark.sql.catalyst.InternalRow;
-import org.apache.spark.sql.catalyst.expressions.GenericInternalRow;
-import org.apache.spark.sql.sources.v2.DataSourceOptions;
-import org.apache.spark.sql.sources.v2.writer.DataSourceWriter;
 import org.apache.spark.sql.sources.v2.writer.DataWriter;
-import org.apache.spark.sql.sources.v2.writer.DataWriterFactory;
-import org.apache.spark.sql.types.DataTypes;
-import org.apache.spark.sql.types.Metadata;
-import org.apache.spark.sql.types.StructField;
-import org.apache.spark.sql.types.StructType;
-import org.apache.spark.unsafe.types.UTF8String;
 import org.junit.jupiter.api.Test;
 
 import java.io.IOException;
-import java.util.Map;
-import java.util.Optional;
 
 import static org.junit.jupiter.api.Assertions.*;
 
 public class WriteDataTest extends AbstractSparkConnectorTest {
 
-    private final static StructType SCHEMA = new StructType(new StructField[]{
-        new StructField("fruitName", DataTypes.StringType, true, Metadata.empty()),
-        new StructField("fruitColor", DataTypes.StringType, true, Metadata.empty()),
-    });
-
     @Test
     void ingestThreeFruitsWithBatchSizeOfTwo() throws IOException {
-        DataWriter<InternalRow> dataWriter = buildDataWriter("2", "/testFruit");
+        DataWriter<InternalRow> dataWriter = buildDataWriter(new Options(getHubPropertiesAsMap()).withBatchSize(2).withUriPrefix("/testFruit"));
 
         verifyFruitCount(0, "Shouldn't have any fruits ingested yet");
 
@@ -50,7 +34,7 @@ public class WriteDataTest extends AbstractSparkConnectorTest {
 
     @Test
     public void testBulkIngestWithoutUriPrefix() throws IOException {
-        DataWriter<InternalRow> dataWriter = buildDataWriter("1", "");
+        DataWriter<InternalRow> dataWriter = buildDataWriter(new Options(getHubPropertiesAsMap()).withBatchSize(1));
         dataWriter.write(buildRow("pineapple", "green"));
 
         String uriQuery = "cts.uris('', null, cts.andQuery([\n" +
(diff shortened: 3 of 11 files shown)


marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/sql/sources/v2/writer/HubDataWriter.java
```diff
@@ -45,24 +48,61 @@
...
+                loader.setEndpointState(new ByteArrayInputStream(("{\"next\":" + 0 + "}").getBytes()));
+            }
+
+            String uriPrefix = params.get("uriprefix") != null ? params.get("uriprefix") : "";
+
+            if (jsonObject.get("workUnit") != null && jsonObject.get("workUnit").toString().length()>0) {
+                JSONObject workUnitJson = new JSONObject(jsonObject.getString("workUnit"));
+                workUnitJson.putOpt("uriprefix", uriPrefix);
+                loader.setWorkUnit(new ByteArrayInputStream((workUnitJson.toString()).getBytes()));
+                System.out.println(workUnitJson.toString());
```
SameeraPriyathamTadikonda, line 96: Please remove Sysout

SameeraPriyathamTadikonda:
If you remove the taskID. we should fix the bulkIngesterTest. It is better if we have two different PR's/Commits.
1) For configuring customEndpointState and Workunits
2) Removing taskId and moving the URIPrefix to WorkUnit

SameeraPriyathamTadikonda (changes requested):

marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/sql/sources/v2/writer/HubDataWriter.java
```diff
@@ -45,24 +48,61 @@
...
-        loader.setWorkUnit(new ByteArrayInputStream(("{\"taskId\":" + taskId + "}").getBytes()));
-        loader.setEndpointState(new ByteArrayInputStream(("{\"next\":" + 0 + ", \"uriprefix\":\"" + params.get("uriprefix") + "\"}").getBytes()));
+            if(jsonObject.get("apiPath") == null || jsonObject.get("apiPath").toString().length()==0) {
+                if((jsonObject.get("endpointState") != null && jsonObject.get("endpointState").toString().length()>0) ||
+                    (jsonObject.get("workUnit")!=null && jsonObject.get("workUnit").toString().length()>0))
+                {
+                    throw new RuntimeException("Cannot override default endpointState and/or workUnit. Please provide custom endpoint.");
+                }
+            }
+            final String apiModulePath = (jsonObject.get("apiPath")!=null && jsonObject.get("apiPath").toString().length()>0) ?
```
rjrudin, line 70: There's a lot going on here, and generally, I like to use private/protected methods to break up long methods so that they're easier to read and maintain. I prefer private, though I'll use protected when I want to write unit tests against the logic.

I see the following steps in this code, where each can become a new private method:

1. Use the user params to determine the apiPath, the workUnit, and endpointState (and throw a validation error as needed)
2. Use the output of #1 to create a BulkInputCaller

I think the signature of the first private method would be "private JsonNode determineIngestionEndpointParams(Map)". If the user has defined "ingestendpointparams", then use ObjectMapper to parse that into a JsonNode, and then validate it. Else, use objectMapper.createObjectNode() and populate apiPath based on the default API path, and then populate workUnit based on uriPrefix (and then @SameeraPriyathamTadikonda  would add more logic to this for collections/permissions/etc). 

That first method would likely be protected so that you could easily write unit tests against it without having to involve everything else in this class. 

The second private method would have a signature of "private BulkInputCaller buildBulkInputCaller(JsonNode ingestionParams)". 

Each private method can then have a try/catch as needed to ensure that if an error is thrown, the appropriate context is provided in the error message.

marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/sql/sources/v2/writer/HubDataWriterFactory.java
```diff
@@ -53,7 +53,7 @@ public HubDataWriterFactory(Map<String, String> params, StructType schema) {
         if (logger.isDebugEnabled()) {
             logger.debug("Creating DataWriter with taskId: " + taskId);
         }
-        return new HubDataWriter(hubClient, taskId, schema, params);
+        return new HubDataWriter(hubClient, schema, params);
```
rjrudin, line 56: I think that since taskId doesn't matter, let's remove the debug logging above it. It would be meaningless to see it.

marklogic-data-hub-spark-connector/src/test/java/com/marklogic/hub/spark/sql/sources/v2/WriteDataTest.java
```diff
@@ -77,18 +96,16 @@ public void testBulkIngestWithoutUriPrefix() throws IOException {
      * @param uriPrefix
      * @return
      */
-    private DataWriter<InternalRow> buildDataWriter(String batchSize, String uriPrefix) {
+    private DataWriter<InternalRow> buildDataWriter(String batchSize, String uriPrefix, String endpointState, String workUnit, String apiPath) {
```
rjrudin, line 99: When there are a lot of params in a method, that's generally a sign that a class should be created. 

While we don't need this class in the application code, an "Options" class would be very helpful here for making it easier to write tests. It would have properties of batchSize, uriPrefix, apiPath, etc. And then it would have a "Map toMap()" method that builds a Map of params based on what's been configured. 

Then, each test will just build an Options class based on the scenario that it's testing. That will make it easier to write future tests - i.e. when @SameeraPriyathamTadikonda  needs to test for collections/permissions/etc, we don't have to add those to this method and fix every method that calls it - we just add new properties to our Options class, and the existing tests won't be affected.

rjrudin (changes requested):

anu3990:
@SameeraPriyathamTadikonda . for your below comments - 
If you remove the taskID. we should fix the bulkIngesterTest. It is better if we have two different PR's/Commits.
For configuring customEndpointState and Workunits
Removing taskId and moving the URIPrefix to WorkUnit

1. i have fixed bulkIngesterTest but the user can still have a field named "taskId" in their workUnit. I have renamed it for now to avoid confusion.
2. For your second point - "Removing taskId and moving the URIPrefix to WorkUnit", this is exactly what we do when we say "configure workUnit" so we can still combine it in the current story - https://project.marklogic.com/jira/browse/DHFPROD-5989 because we are configuring both endpointState and workUnit here. I have also added a bullet point in the acceptance criteria of the story.

marklogic-data-hub-spark-connector/src/test/java/com/marklogic/hub/spark/sql/sources/v2/WriteDataTest.java
```diff
@@ -30,7 +34,9 @@
 
     @Test
     void ingestThreeFruitsWithBatchSizeOfTwo() throws IOException {
-        DataWriter<InternalRow> dataWriter = buildDataWriter("2", "/testFruit");
+        Map<String, String> params = getHubPropertiesAsMap();
```
rjrudin, line 37: Pinging @SameeraPriyathamTadikonda  about this - for each workUnit input that the endpoint supports, we'll want to verify that it works correctly in the BulkIngestTest class, as that class has no knowledge of Spark.

The tests for our Spark connector should then focus on converting the params Map into the proper workUnit object. Those tests don't need to connect to ML - they can be very fast unit tests that live in a separate class that doesn't extend AbstractHubCoreTest. I'll put together a PR to demonstrate this and submit it against this branch. 

We'll still want one test in this project that writes data with every workUnit field populated, just to make sure that we're setting the correct workUnit fields.

marklogic-data-hub/src/test/java/com/marklogic/hub/dataservices/ingestion/BulkIngestTest.java
```diff
@@ -32,8 +32,8 @@ public void setupTest() {
     public void testBulkIngest() {
 
         String prefix = "/bulkIngesterTest";
-        String endpointState = "{\"next\":" + 0 + ", \"uriprefix\":\""+prefix+"\"}";
-        String workUnit      = "{\"taskId\":"+1+"}";
+        String endpointState =  "{\"next\":"+0+"}";
```
rjrudin, line 35: Can this just be "{}", since we don't care about "next"? It's supposed to go away once we shift to the newer Java client, but I think it'd be better to make it "{}" in the meantime.

rjrudin (commented):
I'm going to submit a separate PR with a test helper class.

rjrudin:
@anu3990  @SameeraPriyathamTadikonda See this PR - https://github.com/anu3990/marklogic-data-hub/pull/2 - for an Options class that should simplify writing tests. It also facilitates testing for null inputs, as  options in the map will only be populated if the test populates them.

marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/sql/sources/v2/writer/HubDataWriter.java
```diff
@@ -114,4 +113,68 @@ private String convertRowToJSONString(InternalRow record) {
...
+    private InputEndpoint.BulkInputCaller buildBulkInputCaller(HubClient hubClient, String apiPath, String endpointState,
+                                                               String workUnit) {
+        InputEndpoint.BulkInputCaller bulkInputCaller;
+        try {
+            bulkInputCaller= InputEndpoint.on(
+                hubClient.getStagingClient(),
+                hubClient.getModulesClient().newJSONDocumentManager().read(apiPath, new StringHandle())
+            ).bulkCaller();
+        } catch (ResourceNotFoundException ex) {
+            throw new RuntimeException("Endpoint not found.");
```
rjrudin, line 174: I am wondering if the ResourceNotFoundException needs to be caught here. If we do throw a new exception, we definitely need to include the original exception so that Ernie doesn't lose any context. 

I did a quick test, and this is what the exception shows:

```
com.marklogic.client.ResourceNotFoundException: Local message: Could not read non-existent document. Server Message: RESTAPI-NODOCUMENT: (err:FOER0000) Resource or document does not exist:  category: content message: /data-hub/5/data-services/ingestion/bulkIngesterrrr.api
```

That is likely helpful enough for Ernie - it shows the module he referenced, and the Java Client is giving a good indication of the error. So I think this try/catch can be removed. 

And because of that, we don't really need this method anymore, since it's just a few lines of code now. If we do need some special error handling here, I think it's worth the separate method. But that doesn't appear to be the case.

marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/sql/sources/v2/writer/HubDataWriter.java
```diff
@@ -114,4 +113,68 @@ private String convertRowToJSONString(InternalRow record) {
...
+        try {
+            endpointParamsJsonNode = objectMapper.readValue(ingestEndpointParams, JsonNode.class);
+        } catch (IOException e) {
+            throw new RuntimeException("Unable to parse ingestendpointparams, cause: " + e.getMessage(), e);
+        }
+
+        boolean doesNotHaveApiPath = !endpointParamsJsonNode.has("apiPath");
+        boolean hasWorkUnitOrEndpointState = endpointParamsJsonNode.has("workUnit") || endpointParamsJsonNode.has("endpointState");
+        if (doesNotHaveApiPath && hasWorkUnitOrEndpointState) {
+            throw new RuntimeException("Cannot override default endpointState and/or workUnit. Please provide custom endpoint.");
```
rjrudin, line 132: Exception messages are crucial here to explain to Ernie what to do if he made a mistake. We want to be more precise about how to fix the problem - e.g. 

Cannot set workUnit or endpointState in ingestionendpointparams unless apiPath is defined as well

marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/sql/sources/v2/writer/HubDataWriter.java
```diff
@@ -41,28 +44,24 @@
...
-     * @param taskId
      * @param schema
      * @param params contains all the params provided by Spark, which will include all connector-specific properties
      */
-    public HubDataWriter(HubClient hubClient, long taskId, StructType schema, Map<String, String> params) {
+    public HubDataWriter(HubClient hubClient, StructType schema, Map<String, String> params) {
         this.records = new ArrayList<>();
         this.schema = schema;
         this.batchSize = params.containsKey("batchsize") ? Integer.parseInt(params.get("batchsize")) : 100;
+        this.uriPrefix = params.get("uriprefix") != null ? params.get("uriprefix") : "";
```
rjrudin, line 59: No need to capture this here - batchSize needs to be captured, but not this. Just need to read this in the determineIngestionEndpointParams method.

marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/sql/sources/v2/writer/HubDataWriter.java
```diff
@@ -114,4 +113,68 @@ private String convertRowToJSONString(InternalRow record) {
...
+        if (endpointParamsJsonNode.hasNonNull("workUnit") && endpointParamsJsonNode.get("workUnit").asText().length() > 0) {
+
+            JsonNode workUnitNode;
+            try {
+                workUnitNode = objectMapper.readValue(endpointParamsJsonNode.get("workUnit").asText(), JsonNode.class);
+            } catch (JsonProcessingException e) {
+                throw new RuntimeException("Unable to parse workUnit, cause: " + e.getMessage(), e);
+            }
+            ((ObjectNode) workUnitNode).put("uriprefix", uriPrefix);
+            ((ObjectNode)endpointParamsJsonNode).put("workUnit", workUnitNode.toString());
```
rjrudin, line 158: Use set("workUnit", workUnitNode) instead. That returns the JSON structure instead of converting it into a string. You can then do setWorkUnit(new JacksonHandle(theNode)) and don't have to fiddle with converting things into strings. Same goes for endpointState.

rjrudin (changes requested):

marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/sql/sources/v2/writer/HubDataWriter.java
```diff
@@ -114,4 +116,52 @@ private String convertRowToJSONString(InternalRow record) {
...
+            JsonNode workUnitNode;
+            try {
+                workUnitNode = objectMapper.readValue(endpointParamsJsonNode.get("workUnit").asText(), JsonNode.class);
+            } catch (JsonProcessingException e) {
+                throw new RuntimeException("Unable to parse workUnit, cause: " + e.getMessage(), e);
+            }
+            ((ObjectNode) workUnitNode).put("uriprefix", params.get("uriprefix"));
+            ((ObjectNode)endpointParamsJsonNode).set("workUnit", workUnitNode);
+        } else {
+            ((ObjectNode)endpointParamsJsonNode).put("workUnit", "{\"uriprefix\":\"" + params.get("uriprefix") + "\"}");
```
rjrudin, line 163: I think this needs to be a JsonNode as well instead of a string for consistency. You can do mapper.createObjectNode() and then node.put("uriprefix", params.get("uriprefix")) . 

I think it's good to handle these as JsonNode's instead of as strings. When Ernie provides us with a custom workUnit and/or endpointState, we should convert those ASAP to JsonNode. That way, if Ernie's JSON is malformed, it'll fail as quickly as possible - i.e. in the connector code and not in ML. 

Also this is where @SameeraPriyathamTadikonda will add his code for collections/permissions/etc.

marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/sql/sources/v2/writer/HubDataWriter.java
```diff
@@ -45,24 +48,23 @@
...
+
         this.loader = InputEndpoint.on(
             hubClient.getStagingClient(),
-            hubClient.getModulesClient().newTextDocumentManager().read(apiModulePath, new StringHandle())
+            hubClient.getModulesClient().newJSONDocumentManager().read(endpointParamsJsonNode.get("apiPath").asText(), new StringHandle())
         ).bulkCaller();
 
-        loader.setWorkUnit(new ByteArrayInputStream(("{\"taskId\":" + taskId + "}").getBytes()));
-        loader.setEndpointState(new ByteArrayInputStream(("{\"next\":" + 0 + ", \"uriprefix\":\"" + params.get("uriprefix") + "\"}").getBytes()));
+        this.loader.setEndpointState(new ByteArrayInputStream((endpointParamsJsonNode.get("endpointState").asText()).getBytes()));
```
rjrudin, line 66: I'll test this locally, but I didn't think "asText" does what we want here - it's just getting the text of this node, it's not the same as toString. I think we want to use JacksonHandle here instead, but I'm not sure. 

Which makes me realize - we really need a test that uses a custom API path. That way, we know for sure that our custom workUnit and endpointState are passed to our custom endpoint correctly.

Let me handle that test - I want to think a bit about where the modules for this should go. I think they can go under ./marklogic-data-hub/src/test/ml-modules, I just need to confirm that. I'll try to submit that tonight or first thing tomorrow morning. 

In the meantime though, this can go forward so that @SameeraPriyathamTadikonda  can make use of it.

rjrudin (approved):

marklogic-data-hub/src/test/java/com/marklogic/hub/dataservices/ingestion/BulkIngestTest.java
```diff
@@ -32,8 +32,8 @@ public void setupTest() {
     public void testBulkIngest() {
 
         String prefix = "/bulkIngesterTest";
-        String endpointState = "{\"next\":" + 0 + ", \"uriprefix\":\""+prefix+"\"}";
-        String workUnit      = "{\"taskId\":"+1+"}";
+        String endpointState =  "{}";
+        String workUnit      = "{\"userDefinedValue\":" + 1 + ", \"uriprefix\":\""+prefix+"\"}";
```
SameeraPriyathamTadikonda, line 36: what's the purpose of userDefinedValue?

marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/sql/sources/v2/writer/HubDataWriter.java
SameeraPriyathamTadikonda, line 163: Do we need to add this even if its null?

marklogic-data-hub/src/test/java/com/marklogic/hub/dataservices/ingestion/BulkIngestTest.java
anu3990: "userDefinedValue" is just a variable added by the user. We can also have something like the below -
String workUnit      = "{\"testValue\":" + 1 + ", \"uriprefix\":\""+prefix+"\"}";

Adding the "userDefinedValue" confirms that no exception is thrown even when we have extra values defined by the user.

marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/sql/sources/v2/writer/HubDataWriter.java
anu3990: Yes we need this even when the user did not add URIPrefix because our sjs file looks for this value. Since, the uriPrefix value does not change, we moved it out from endpointState to workUnit

SameeraPriyathamTadikonda: Are we going to add null in that case?

marklogic-data-hub/src/test/java/com/marklogic/hub/dataservices/ingestion/BulkIngestTest.java
SameeraPriyathamTadikonda: How will the user pass this option from glue/spark? I'm still confused where are we using this?

marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/sql/sources/v2/writer/HubDataWriter.java
anu3990: No we will not be adding null to the uris. Our Glue script will be something like this -

val prefix = ""
val writer = dataframe.write.format("com.marklogic.client.spark.Writer.MarkLogicWriteDataSource")
                    .option("host", host)
                    .option("port", port)
                    .option("user", username)
                    .option("password",password)
                    .option("uriprefix", prefix).....(other options);
By default the value of variable prefix is an empty string. The user can add a custom value.

marklogic-data-hub/src/test/java/com/marklogic/hub/dataservices/ingestion/BulkIngestTest.java
anu3990: This test is not in the marklogic-data-hub-spark-connector folder. It is testing only the bulk caller in the java-client-api and uses the bulkIngester api and sjs. The uriPrefix appended in the workUnit is used by the sjs file here - https://github.com/anu3990/marklogic-data-hub/blob/660d87d7616098274a5b97ff3d2005b0a07461fb/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/ingestion/bulkIngester.sjs#L39 

In the Glue environment, uriprefix will be passed in as an option(i have defined in the next comment) and appended in the workUnit by the HubDataWriter class before calling bulkIngester.api.

marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/sql/sources/v2/writer/HubDataWriter.java
rjrudin: But I don't think Ernie has to specify '.option("uriprefix", prefix)', right? In which case the option will not exist in the options map? And so we'd only want to add it to the workUnit if it's not null (unless the endpoint is careful to not use it if it's null).

SameeraPriyathamTadikonda: Yes, that's exactly my point. I don't think that is handled in the endpoint.
Also the endpoint throws an exception if workUnit is empty when no options are set by the user.

marklogic-data-hub/src/test/java/com/marklogic/hub/dataservices/ingestion/BulkIngestTest.java
SameeraPriyathamTadikonda: I understand that. I'm confused on the userDefinedValue. How it is passed and what's the use of it?

rjrudin:
@anu3990  @SameeraPriyathamTadikonda  I created https://github.com/anu3990/marklogic-data-hub/pull/3 to add a test that uses a custom endpoint. I moved some methods from WriteDataTest to the parent class so that it's easier to write separate test classes, as we don't want every single scenario going into WriteDataTest.

I made a couple fixes to HubDataWriter so that workUnit and endpointState are always ObjectNodes, which means we can use JacksonHandle when passing them to the BulkInputCaller. All tests are passing.

rjrudin (changes requested):
Will review again after above PR is merged in.

marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/sql/sources/v2/writer/HubDataWriter.java
```diff
@@ -114,4 +121,41 @@ private String convertRowToJSONString(InternalRow record) {
...
+            try {
+                endpointParams = (ObjectNode)objectMapper.readTree(params.get("ingestendpointparams"));
+            } catch (IOException e) {
+                throw new RuntimeException("Unable to parse ingestendpointparams, cause: " + e.getMessage(), e);
+            }
+        } else {
+            endpointParams = objectMapper.createObjectNode();
+        }
+
+        boolean doesNotHaveApiPath = (!endpointParams.hasNonNull("apiPath") || StringUtils.isEmpty(endpointParams.get("apiPath").asText()));
```
SameeraPriyathamTadikonda, line 138: endpointParams.get("apiPath").asText() will throw s NPE

marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/sql/sources/v2/writer/HubDataWriter.java
```diff
@@ -114,4 +121,41 @@ private String convertRowToJSONString(InternalRow record) {
...
+        boolean hasWorkUnitOrEndpointState = endpointParams.has("workUnit") || endpointParams.has("endpointState");
+        if (doesNotHaveApiPath && hasWorkUnitOrEndpointState) {
+            throw new RuntimeException("Cannot set workUnit or endpointState in ingestionendpointparams unless apiPath is defined as well.");
+        }
+
+        if (!endpointParams.hasNonNull("apiPath") || StringUtils.isEmpty(endpointParams.get("apiPath").asText())) {
```
SameeraPriyathamTadikonda, line 144: we can use the variable "doesNotHaveApiPath" instead of writing the logic again

marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/sql/sources/v2/writer/HubDataWriter.java
```diff
@@ -114,4 +121,41 @@ private String convertRowToJSONString(InternalRow record) {
...
+            endpointParams.put("apiPath", "/data-hub/5/data-services/ingestion/bulkIngester.api");
+        }
+
+        // TODO : remove the below else block after java-client-api 5.3 release
```
SameeraPriyathamTadikonda, line 148: else block missing

marklogic-data-hub/src/test/java/com/marklogic/hub/dataservices/ingestion/BulkIngestTest.java
SameeraPriyathamTadikonda: We don't need the UserDefinedValue here.

marklogic-data-hub-spark-connector/src/test/java/com/marklogic/hub/spark/sql/sources/v2/WriteDataViaCustomEndpointTest.java
```diff
@@ -0,0 +1,67 @@
...
+import org.springframework.core.io.ClassPathResource;
+
+import java.io.IOException;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class WriteDataViaCustomEndpointTest extends AbstractSparkConnectorTest {
+
+    @Test
+    void test() throws IOException {
```
SameeraPriyathamTadikonda, line 22: provide a meaningful name for this test

SameeraPriyathamTadikonda:
Will the ingestionEndpointParams be sent as a JSON String from glue to the connector?

marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/sql/sources/v2/writer/HubDataWriter.java
rjrudin: An NPE won't occur because if apiPath were null or did not exist, then the first part of the conditional would be true and then Java won't evaluate the second part.

marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/sql/sources/v2/writer/HubDataWriter.java
rjrudin: There's no else needed - this is just saying "If Ernie didn't provide a non-null endpointState, then toss in an empty object node".

marklogic-data-hub-spark-connector/src/test/java/com/marklogic/hub/spark/sql/sources/v2/WriteDataViaCustomEndpointTest.java
rjrudin: This is actually a common pattern we have in existing tests - since the class name already identifies the scope of the test, the method is just "test". If a second test method is added, then this would need to be updated.

marklogic-data-hub-spark-connector/src/test/java/com/marklogic/hub/spark/sql/sources/v2/WriteDataTest.java
```diff
@@ -68,45 +52,68 @@ public void testBulkIngestWithoutUriPrefix() throws IOException {
...
+        RuntimeException ex = assertThrows(RuntimeException.class,
+            () -> buildDataWriter(new Options(getHubPropertiesAsMap()).withIngestApiPath("/incorrect.api")),
+            "Expected an error because a custom work unit was provided without a custom API path"
+        );
+        System.out.println(ex.getMessage());
+        assertTrue( ex.getMessage().contains("Could not read non-existent document."));
+    }
+
+    @Test
+    public void ingestWithCustomApiWithCustomWorkunit() throws IOException {
```
rjrudin, line 78: I think you can drop this test because WriteDataViaCustomEndpointTest already covers this with an actual custom endpoint, as opposed to using the default endpoint.

rjrudin (approved):
Looks good to me, though I think it's worth dropping ingestWithCustomApiWithCustomWorkunit in favor of the separate test class that tests a custom endpoint.

SameeraPriyathamTadikonda (dismissed):

anu3990:
@rjrudin and @SameeraPriyathamTadikonda. Need your approvals again, i have rebased and added the suggestions.

rjrudin (approved):

SameeraPriyathamTadikonda (approved):

PR #4631, opened 2020-09-28
DHFPROD-4809: Substitute tokens when loading user artifacts

akshaysonvane:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/LoadUserArtifactsCommand.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/LoadUserArtifactsCommand.java
@@ -21,7 +21,6 @@ import com.fasterxml.jackson.databind.node.ArrayNode;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.marklogic.appdeployer.command.AbstractCommand;
 import com.marklogic.appdeployer.command.CommandContext;
-import com.marklogic.client.DatabaseClient;
 import com.marklogic.client.document.DocumentWriteSet;
 import com.marklogic.client.document.JSONDocumentManager;
 import com.marklogic.client.ext.modulesloader.Modules;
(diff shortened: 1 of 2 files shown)


marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/LoadUserArtifactsCommand.java
```diff
@@ -155,9 +159,10 @@ private void loadModels(HubClient hubClient) throws IOException {
                 public FileVisitResult preVisitDirectory(Path dir, BasicFileAttributes attrs)  {
                     logger.info("Loading models from directory " + dir);
                     modulesFinder.findModules(dir.toString()).getAssets().forEach(r -> {
-                        try {
+                        try (InputStream inputStream = r.getInputStream()) {
```
rjrudin, line 162: We've got 5 instances of code all wanting to do the same thing - convert a File or Resource into a JsonNode, and now they all want to replace tokens at the same time. 

It's best to do this via a new private method - particularly as when we add a 6th instance of the code, we can just reuse that method and not worry about forgetting to replace tokens. 

Since we know that everything is a File here, the 2 places that use a Resource should just do "resource.getFile()". The reusable method can then have a signature of:

    private JsonNode readArtifact(File file);

And it should catch any exception that occurs and throwing a generic error message of:

    throw new RuntimeException("Unless to read file " + f.getFilename() " + as JSON; cause: " + e.getMessage(), e);

rjrudin (changes requested):
Test looks good, just one refactoring requested

akshaysonvane:
Added `readArtifact` method that reads the artifact file, replaces tokens, and then returns the content as a JsonNode.

marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/LoadUserArtifactsCommand.java
```diff
@@ -382,6 +378,25 @@ protected ObjectNode replaceLanguageWithLang(ObjectNode object) {
...
     }
 
+    /**
+     * Reads the artifact file, replaces tokens and then returns the content as a JsonNode.
+     *
+     * @param file
+     * @return
+     */
+    private JsonNode readArtifact(File file) {
+        JsonNode jsonNode;
```
rjrudin, line 388: This is fine to keep, but generally I think it's simpler to just do "return objectMapper.readTree" as opposed to declaring this variable here. Shortens the method too.

rjrudin (approved):

ryanjdew (approved):

PR #4633, opened 2020-09-28
DHFPROD-5944: Configure Document Ingestion via connector

SameeraPriyathamTadikonda:
### Description
Configure Document Ingestion, sets the permissions, collections and sources of the document configured through options from Spark
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/sql/sources/v2/writer/HubDataWriter.java
+++ b/marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/sql/sources/v2/writer/HubDataWriter.java
@@ -47,12 +47,12 @@ public class HubDataWriter extends LoggingObject implements DataWriter<InternalR
     private InputEndpoint.BulkInputCaller loader;
     private StructType schema;
     private int batchSize;
+    private ObjectNode defaultWorkUnit;
 
     /**
-     *
      * @param hubClient
      * @param schema
-     * @param params contains all the params provided by Spark, which will include all connector-specific properties
+     * @param params    contains all the params provided by Spark, which will include all connector-specific properties
      */
     public HubDataWriter(HubClient hubClient, StructType schema, Map<String, String> params) {
         this.records = new ArrayList<>();
@@ -127,7 +127,7 @@ public class HubDataWriter extends LoggingObject implements DataWriter<InternalR
         ObjectNode endpointParams;
         if (params.containsKey("ingestendpointparams")) {
             try {
-                endpointParams = (ObjectNode)objectMapper.readTree(params.get("ingestendpointparams"));
+                endpointParams = (ObjectNode) objectMapper.readTree(params.get("ingestendpointparams"));
             } catch (IOException e) {
                 throw new RuntimeException("Unable to parse ingestendpointparams, cause: " + e.getMessage(), e);
             }
@@ -151,11 +151,19 @@ public class HubDataWriter extends LoggingObject implements DataWriter<InternalR
         }
 
         if (!endpointParams.hasNonNull("workUnit")) {
-            ObjectNode defaultWorkUnit = objectMapper.createObjectNode();
-            defaultWorkUnit.put("uriprefix", params.get("uriprefix")!=null ? params.get("uriprefix"):"");
+            defaultWorkUnit = objectMapper.createObjectNode();
+            buildDefaultWorkUnit(params);
             endpointParams.set("workUnit", defaultWorkUnit);
         }
 
         return endpointParams;
+    }
+
+    protected void buildDefaultWorkUnit(Map<String, String> params) {
+        Stream.of("collections", "permissions", "sourcename", "sourcetype", "uriprefix").forEach(key -> {
+            if (params.containsKey(key)) {
+                defaultWorkUnit.put(key, params.get(key));
+            }
+        });
     }
 }
--- a/marklogic-data-hub-spark-connector/src/test/java/com/marklogic/hub/spark/sql/sources/v2/WriteDataWithOptionsTest.java
+++ b/marklogic-data-hub-spark-connector/src/test/java/com/marklogic/hub/spark/sql/sources/v2/WriteDataWithOptionsTest.java
@@ -1,0 +1,72 @@
+package com.marklogic.hub.spark.sql.sources.v2;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.marklogic.client.io.DocumentMetadataHandle;
+import com.marklogic.client.io.JacksonHandle;
+import org.apache.spark.sql.catalyst.InternalRow;
+import org.apache.spark.sql.sources.v2.writer.DataWriter;
+import org.junit.jupiter.api.Test;
+
+import java.io.IOException;
+import java.util.Set;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+public class WriteDataWithOptionsTest extends AbstractSparkConnectorTest {
+
+    @Test
+    void ingestDocsWithCollection() throws IOException {
+        String collections = "fruits,stuff";
+        DataWriter<InternalRow> dataWriter = buildDataWriter(newFruitOptions().withCollections(collections));
+        dataWriter.write(buildRow("pineapple", "green"));
+
+        DocumentMetadataHandle metadata = getFirstFruitMetadata();
+        assertEquals(2, metadata.getCollections().size());
+        assertTrue(metadata.getCollections().contains("fruits"));
+        assertTrue(metadata.getCollections().contains("stuff"));
+
+        // Verify default permissions
+        DocumentMetadataHandle.DocumentPermissions perms = getFirstFruitMetadata().getPermissions();
+        Set<DocumentMetadataHandle.Capability> capabilities = perms.get("data-hub-common");
+        assertTrue(capabilities.contains(DocumentMetadataHandle.Capability.READ));
+        assertTrue(capabilities.contains(DocumentMetadataHandle.Capability.UPDATE));
+    }
+
+    @Test
+    void ingestDocsWithSourceName() throws IOException {
+        String sourceName = "spark";
+        DataWriter<InternalRow> dataWriter = buildDataWriter(newFruitOptions().withSourceName(sourceName));
+        dataWriter.write(buildRow("pineapple", "green"));
+
+        JsonNode doc = getHubClient().getStagingClient().newJSONDocumentManager().read(getFruitUris()[0], new JacksonHandle()).get();
+        assertEquals(sourceName, doc.get("envelope").get("headers").get("sources").get(0).get("name").asText());
+    }
+
+    @Test
+    void ingestDocsWithSourceType() throws IOException {
+        String sourceType = "fruits";
+        DataWriter<InternalRow> dataWriter = buildDataWriter(newFruitOptions().withSourceType(sourceType));
+        dataWriter.write(buildRow("pineapple", "green"));
+
+        JsonNode doc = getHubClient().getStagingClient().newJSONDocumentManager().read(getFruitUris()[0], new JacksonHandle()).get();
+        assertEquals(sourceType, doc.get("envelope").get("headers").get("sources").get(0).get("datahubSourceType").asText());
+    }
+
+    @Test
+    void ingestDocsWithPermissions() throws IOException {
+        String permissions = "qconsole-user,read,manage-admin,update";
+        DataWriter<InternalRow> dataWriter = buildDataWriter(newFruitOptions().withPermissions(permissions));
+        dataWriter.write(buildRow("pineapple", "green"));
+
+        DocumentMetadataHandle.DocumentPermissions perms = getFirstFruitMetadata().getPermissions();
+        assertEquals(DocumentMetadataHandle.Capability.READ, perms.get("qconsole-user").iterator().next());
+        assertEquals(DocumentMetadataHandle.Capability.UPDATE, perms.get("manage-admin").iterator().next());
+    }
+
+    private DocumentMetadataHandle getFirstFruitMetadata() {
+        String[] uris = getFruitUris();
+        assertTrue(uris.length > 0, "Expected at least one fruit URI, found zero");
+        return getHubClient().getStagingClient().newDocumentManager().readMetadata(uris[0], new DocumentMetadataHandle());
+    }
+}
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/ingestion/bulkIngester.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/ingestion/bulkIngester.sjs
@@ -25,10 +25,31 @@ var input;         // jsonDocument*
 declareUpdate();
 
 const ingest = require("/data-hub/5/builtins/steps/ingestion/default/main.sjs");
-const consts = require('/data-hub/5/impl/consts.sjs');
+const consts = require("/data-hub/5/impl/consts.sjs");
+const HubUtils = require("/data-hub/5/impl/hub-utils.sjs");
 const state = fn.head(xdmp.fromJSON(endpointState));
 
 const work = fn.head(xdmp.fromJSON(workUnit));
+
+const uriPrefix = work.uriprefix != null ? work.uriprefix : "";
+
+const collections = work.collections != null ? work.collections.split(',') : [];
+
+const permissions = work.permissions != null ? work.permissions : 'data-hub-common,read,data-hub-common,update'
+const permissionsArray = new HubUtils().parsePermissions(permissions);
+
+const headers = {};
+headers.createdOn = consts.CURRENT_DATE_TIME
+headers.createdBy = consts.CURRENT_USER
+
+if(work.sourcename != null || work.sourcetype != null){
+  const sources = [];
+  const source = {};
+  source.name = work.sourcename
+  source.datahubSourceType = work.sourcetype
+  sources[0] = source;
+  headers.sources = sources
+}
 
 const inputs =
   (input instanceof Sequence) ? input.toArray().map(item => fn.head(xdmp.fromJSON(item))) :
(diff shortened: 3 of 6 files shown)


marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/sql/sources/v2/writer/HubDataWriter.java
```diff
@@ -61,7 +62,15 @@ public HubDataWriter(HubClient hubClient, long taskId, StructType schema, Map<St
             hubClient.getModulesClient().newTextDocumentManager().read(apiModulePath, new StringHandle())
         ).bulkCaller();
 
-        loader.setWorkUnit(new ByteArrayInputStream(("{\"taskId\":" + taskId + "}").getBytes()));
+        String workUnit = "{\"taskId\":" + taskId + "}";
+        ObjectMapper objectMapper = new ObjectMapper();
+        try {
+            workUnit = workUnit.concat(objectMapper.writeValueAsString(params));
```
rjrudin, line 68: I don't think it's safe to write all of the params as a JSON object, as that could contain the user password and other sensitive information. 

We should be much more precise about what goes into workUnit. So we should have a method of e.g. "ObjectNode buildWorkUnitFromParams(Map)" that builds an ObjectNode based on the params we know may exist in the Map, and nothing more than that.

marklogic-data-hub-spark-connector/src/test/java/com/marklogic/hub/spark/sql/sources/v2/WriteDataTest.java
```diff
@@ -86,9 +86,13 @@ public void testBulkIngestWithoutUriPrefix() throws IOException {
         Map<String, String> params = getHubPropertiesAsMap();
         params.put("batchsize", batchSize);
 
-        if(uriPrefix!=null && uriPrefix.length()!=0) {
+        if (uriPrefix != null && uriPrefix.length() != 0) {
             params.put("uriprefix", uriPrefix);
         }
+        params.put("collections", "fruits,gluefruits");
```
rjrudin, line 92: See my comment to @anu3990  about an Options class that will simplify writing tests, as opposed to building a Map. We will likely have many tests that populate different sets of options. The Options class will handle that for us.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/ingestion/bulkIngester.sjs
```diff
@@ -29,6 +29,29 @@ const consts = require('/data-hub/5/impl/consts.sjs');
...
+const permissionsArray = []
+
+if(!work.permissions){
+  work.permissions='data-hub-common,read,data-hub-common,update'
+}
+
+const permissions = work.permissions.split(',');
+
+if((permissions.length%2)!= 0){
+  xdmp.log("Error: Unable to parse Permissions")
```
rjrudin, line 46: No need for xdmp.log here, the stacktrace from the fn.error will be visible in the logs

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/ingestion/bulkIngester.sjs
```diff
@@ -29,6 +29,29 @@ const consts = require('/data-hub/5/impl/consts.sjs');
...
+  fn.error(null, "RESTAPI-SRVEXERR", Sequence.from([400, "Illegal Argument Exception", "Unable to parse permissions string, which must be a comma-separated list of role names and capabilities - i.e. role1,read,role2,update,role3,execute; string:" + permissions]));
```
rjrudin, line 47: Use ds-utils.sjs - that has helper functions in it to simplify this - e.g. throwBadRequest would be appropriate here.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/ingestion/bulkIngester.sjs
```diff
@@ -29,6 +29,29 @@ const consts = require('/data-hub/5/impl/consts.sjs');
...
+}
+
+for(let i=0;i<permissions.length;i+=2){
```
rjrudin, line 50: Use hub-utils.parsePermissions

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/ingestion/bulkIngester.sjs
```diff
@@ -38,16 +61,23 @@ inputs.forEach(record => {
   state.next = state.next + 1;
   const uri = (state.uriprefix) +  sem.uuidString() + '.json';
   record = ingest.main({uri: uri, value: record}, {
-    outputFormat: consts.JSON, headers: {createdOn: consts.CURRENT_DATE_TIME, createdBy: consts.CURRENT_USER}
+    outputFormat: consts.JSON,
+                  headers: {
+                      createdOn: consts.CURRENT_DATE_TIME,
+                      createdBy: consts.CURRENT_USER,
+                      sources: [
+                        { name: work.sourcename,
```
rjrudin, line 69: We only want to add a source object if at least one of these is set. So you should add a test that verifies that when neither sourcename nor sourcetype is provided, then the sources array does not exist.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/ingestion/bulkIngester.sjs
```diff
@@ -29,6 +29,29 @@ const consts = require('/data-hub/5/impl/consts.sjs');
...
+  collections=work.collections.split(',');
+}
+
+const permissionsArray = []
+
+if(!work.permissions){
+  work.permissions='data-hub-common,read,data-hub-common,update'
+}
+
+const permissions = work.permissions.split(',');
```
rjrudin, line 43: We need a test verifying that it's okay if permissions is not, then the documents will be added with no permissions. You'll need to use "runAsAdmin" after ingesting the data to verify that the documents exist.

We can't force Ernie to specify permissions, he's free to insert documents without any permissions.

rjrudin (changes requested):

rjrudin:
@anu3990  @SameeraPriyathamTadikonda I recommend we get @anu3990 's PR submitted first, which includes the Options class. This PR can then be updated to include that class and add fields to it for collections/permissions/etc.

SameeraPriyathamTadikonda:
Converting this PR to draft. Will start working on it once @anu3990's PR merges.

marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/sql/sources/v2/writer/HubDataWriter.java
```diff
@@ -151,11 +151,28 @@ protected JsonNode determineIngestionEndpointParams(Map<String, String> params)
         }
 
         if (!endpointParams.hasNonNull("workUnit")) {
-            ObjectNode defaultWorkUnit = objectMapper.createObjectNode();
-            defaultWorkUnit.put("uriprefix", params.get("uriprefix")!=null ? params.get("uriprefix"):"");
+            defaultWorkUnit = objectMapper.createObjectNode();
+            defaultWorkUnit.put("uriprefix", params.get("uriprefix") != null ? params.get("uriprefix") : "");
```
rjrudin, line 155: uriprefix should go into buildWorkUnitFromParams. Also, I'd just call that "buildDefaultWorkUnit" - it's clear that it's built from params because that's the sole argument.

marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/sql/sources/v2/writer/HubDataWriter.java
```diff
@@ -151,11 +151,28 @@ protected JsonNode determineIngestionEndpointParams(Map<String, String> params)
...
+            buildWorkUnitFromParams(params);
             endpointParams.set("workUnit", defaultWorkUnit);
         }
 
         return endpointParams;
     }
+
+    protected void buildWorkUnitFromParams(Map<String, String> params) {
+        if (params.get("collections") != null) {
```
rjrudin, line 164: A more succinct approach with Jackson is to use "has":

    if (params.has("collections"))

Also, all 4 conditionals are the same, so you could simplify this like so:

    Stream.of("collections", "permissions", "sourcename", "sourcetype", "uriprefix").forEach(key -> {
      if (params.containsKey(key)) {
        defaultWorkUnit.put(key, params.get(key));
      }
    }

The one downside is we have to use the lower-casing that Spark requires. I think we can live with that for now though.

marklogic-data-hub-spark-connector/src/test/java/com/marklogic/hub/spark/sql/sources/v2/WriteDataWithOptions.java
```diff
@@ -0,0 +1,89 @@
...
+import org.apache.spark.sql.catalyst.InternalRow;
+import org.apache.spark.sql.sources.v2.writer.DataWriter;
+import org.junit.Assert;
+import org.junit.jupiter.api.Test;
+
+import java.io.IOException;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class WriteDataWithOptions extends AbstractSparkConnectorTest {
```
rjrudin, line 19: Put "Test" as a suffix on this class name, that's our convention for every test class.

rjrudin (changes requested):
I'll try this out locally on Monday too - feeling too tired on a Friday afternoon to do it right now. :)

marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/sql/sources/v2/writer/HubDataWriter.java
SameeraPriyathamTadikonda: Added uriprefix to buildDefaultWorkUnit

marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/sql/sources/v2/writer/HubDataWriter.java
SameeraPriyathamTadikonda: Updated the changes

marklogic-data-hub-spark-connector/src/test/java/com/marklogic/hub/spark/sql/sources/v2/WriteDataWithOptions.java
SameeraPriyathamTadikonda: Renamed the class

marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/sql/sources/v2/writer/HubDataWriter.java
```diff
@@ -151,11 +151,22 @@ protected JsonNode determineIngestionEndpointParams(Map<String, String> params)
         }
 
         if (!endpointParams.hasNonNull("workUnit")) {
-            ObjectNode defaultWorkUnit = objectMapper.createObjectNode();
-            defaultWorkUnit.put("uriprefix", params.get("uriprefix")!=null ? params.get("uriprefix"):"");
+            defaultWorkUnit = objectMapper.createObjectNode();
+            if (params.get("uriprefix") == null) {
```
rjrudin, line 155: The endpoint should do this, not the connector. The endpoint should say - if uriprefix is in the workUnit, then I'll use that, even if it's null. It's up to the client to pass in the correct value for uriprefix - if the client passes in "uriprefix": null, then the endpoint says - Well I guess you want null as the prefix, so I'll use that.

marklogic-data-hub-spark-connector/src/test/java/com/marklogic/hub/spark/sql/sources/v2/WriteDataWithOptionsTest.java
```diff
@@ -0,0 +1,89 @@
...
+
+import java.io.IOException;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class WriteDataWithOptionsTest extends AbstractSparkConnectorTest {
+
+    @Test
+    void ingestDocsWithCollection() throws IOException {
+        String collections = "fruits";
```
rjrudin, line 23: This test should verify that two collections work, so e.g. "fruits,stuff".

marklogic-data-hub-spark-connector/src/test/java/com/marklogic/hub/spark/sql/sources/v2/WriteDataWithOptionsTest.java
```diff
@@ -0,0 +1,89 @@
...
+        DataWriter<InternalRow> dataWriter = buildDataWriter(new Options(getHubPropertiesAsMap()).withBatchSize(1).withUriPrefix("/testFruit").withCollections(collections));
```
rjrudin, line 24: Let's not specify a uriprefix here, as the scope of this test is to verify the collections. 

So instead of doing a uriMatch query, do a cts.uris(null, null, cts.collectionQuery('fruits')). And I recommend getting the value back as a string, and then splitting that into an array based on the newline symbol:

```
String[] uris = getHubClient().getStagingClient().newServerEval().javascript("cts.uris(null, null, cts.collectionQuery('fruits'))").evalAs(String.class).split("\n");
        assertEquals(1, uris.length);
        DocumentMetadataHandle metadata = getHubClient().getStagingClient().newDocumentManager().readMetadata(uris[0], new DocumentMetadataHandle());
        assertEquals(2, metadata.getCollections().size());
        assertTrue(metadata.getCollections().contains("fruits"));
        assertTrue(metadata.getCollections().contains("stuff"));
```

marklogic-data-hub-spark-connector/src/test/java/com/marklogic/hub/spark/sql/sources/v2/WriteDataWithOptionsTest.java
```diff
@@ -0,0 +1,89 @@
...
+
+        assertEquals(collections, docMetaData.getCollections().toArray()[0].toString());
+    }
+
+    @Test
+    void ingestDocsWithSourceName() throws IOException {
+        String sourceName = "spark";
+        DataWriter<InternalRow> dataWriter = buildDataWriter(new Options(getHubPropertiesAsMap()).withBatchSize(1).withUriPrefix("/testFruit").withSourceName(sourceName));
+        dataWriter.write(buildRow("pineapple", "green"));
+        String uriQuery = "cts.uriMatch('/testFruit**')";
```
rjrudin, line 42: There's a lot of duplication across these 4 tests - I'm going to submit a PR to yours to resolve that.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/ingestion/bulkIngester.sjs
```diff
@@ -25,11 +25,38 @@ var input;         // jsonDocument*
...
 
 const ingest = require("/data-hub/5/builtins/steps/ingestion/default/main.sjs");
-const consts = require('/data-hub/5/impl/consts.sjs');
+const consts = require("/data-hub/5/impl/consts.sjs");
+const HubUtils = require("/data-hub/5/impl/hub-utils.sjs");
 const state = fn.head(xdmp.fromJSON(endpointState));
 
 const work = fn.head(xdmp.fromJSON(workUnit));
 
+let collections = [];
```
rjrudin, line 34: "const" is preferable to "let", as it makes it clear that the value will not be changing. So do this instead:

    const collections = work.collections != null ? work.collections.split(',') : [];

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/ingestion/bulkIngester.sjs
```diff
@@ -25,11 +25,38 @@ var input;         // jsonDocument*
...
+if(work.collections != null){
+  collections = work.collections.split(',');
+}
+
+let permissionsArray = []
```
rjrudin, line 39: Same thing here - can adapt this to use "const" instead of "let".

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/ingestion/bulkIngester.sjs
```diff
@@ -25,11 +25,38 @@ var input;         // jsonDocument*
...
+
+const headers = {};
+headers.createdOn = consts.CURRENT_DATE_TIME
+headers.createdBy = consts.CURRENT_USER
+
+if(work.sourcename != null || work.sourcetype != null){
+  const sources = [];
+  const source = {};
+  source.name = work.sourcename
+  source.datahubSourceType = work.sourcetype
```
rjrudin, line 55: Pinging @rahulvudutala  and @fsnow  and @ryanjdew  about this - this is the property name we're proposing for capturing the source type. Note that this is going on the develop branch. Nothing in DHF will care about it yet though.

marklogic-data-hub-spark-connector/src/test/java/com/marklogic/hub/spark/sql/sources/v2/WriteDataWithOptionsTest.java
rjrudin: See https://github.com/SameeraPriyathamTadikonda/marklogic-data-hub/pull/25/files

rjrudin (changes requested):

marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/sql/sources/v2/writer/HubDataWriter.java
SameeraPriyathamTadikonda: Made the changes

marklogic-data-hub-spark-connector/src/test/java/com/marklogic/hub/spark/sql/sources/v2/WriteDataWithOptionsTest.java
SameeraPriyathamTadikonda: Addressed via https://github.com/SameeraPriyathamTadikonda/marklogic-data-hub/pull/25/files

marklogic-data-hub-spark-connector/src/test/java/com/marklogic/hub/spark/sql/sources/v2/WriteDataWithOptionsTest.java
SameeraPriyathamTadikonda: Same https://github.com/SameeraPriyathamTadikonda/marklogic-data-hub/pull/25/files

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/ingestion/bulkIngester.sjs
SameeraPriyathamTadikonda: Made the changes

rjrudin (approved):

anu3990 (approved):

PR #4652, opened 2020-09-30
DHFPROD-5891: Running a step collapses the expanded flow

briantang:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/src/pages/Run.test.tsx
+++ b/marklogic-data-hub-central/ui/src/pages/Run.test.tsx
@@ -725,13 +725,25 @@ describe('Verify Add Step function', () => {
 
     })
 
-    test('Verify a flow panel that is closed when a step is added to it is then opened', async () => {
+    test('Verify a flow panel that is closed reopens when a step is added to it', async () => {
         mocks.runAddStepAPI(axiosMock);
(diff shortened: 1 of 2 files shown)


wooldridge (changes requested):
@briantang , I think I'm seeing a remaining issue, can you see if you can reproduce?

The buggy scenario described in the ticket looks fixed:

https://drive.google.com/file/d/1tkkETOn2DbbkkUTnCqzz4Ea0EF_sjXbE/view

However, when I manually try the scenario described in Run.test.tsx: "Verify a flow panel that is closed reopens when a step is added to it"

When I try adding a step to test1, I see a different flow panel open, the panel where I most recently ran a step:

https://drive.google.com/file/d/1R2lIesLdbVc3bPccWt4_SnKuPGYx2ujI/view

Do we need to set panel state during the Add Step process?

briantang:
Thanks for the catch @wooldridge. It's fixed now.

wooldridge (approved):

Sanjeevani19 (approved):

timur-isangulov (approved):

PR #4654, opened 2020-10-01
DHFPROD-5865: Generate explorer search options w/o conflicting with entity-type name

akshaysonvane:
Also removed exp-default.xml staging and final options that aren't used anymore

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/hub-entities.xqy
```diff
@@ -201,25 +201,30 @@ declare %private function hent:fix-options-for-explorer(
...
-            else
-              element {fn:node-name($search-range-node)} {
-                $search-range-node/attribute()[not(name()='facet')],
-                attribute facet {"false"},
-                hent:fix-options-for-explorer($search-range-node, $sortable-properties, $entity-namespace-map)/node()
-              }
-        }
+        let $search-container-node := $n/search:container
+        return
+          if (fn:not(fn:empty($search-container-node))) then
```
rjrudin, line 206: Pinging @fsnow  to confirm, as I'm forgetting xquery each day, but I'm pretty certain that since you have a node or an empty sequence here, you can just do this:

    if ($search-container-node)

Though the more elegant approach would be this:

    let $container := $n/search:container
    where fn:not($container)
    return ...

I like let/where/return as a replacement for an if/else where either condition has empty sequence as a return value

rjrudin: And to make it more self-documenting, you could do this:

let $container-for-entity-property-generated-by-es := $n/search:container

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/impl/hub-entities/generateSearchOptions.sjs
```diff
@@ -359,6 +359,31 @@ function testHubCentralSupportedDatatypeMappingsForSort() {
...
+        },
+        "Address": {
+          "properties": {}
+        }
+      }
+    }];
+  const options = hent.dumpSearchOptions(input, true);
+
+  return [
+    test.assertNotExists(options.xpath("/*:constraint[@name = 'Book']/*:container")),
```
rjrudin, line 382: These assertions should have a message explaining why we don't want these here - i.e. touch on the Collection constraint and the fact that Explorer doesn't need these container constraints.

rjrudin (changes requested):
We also need a JUnit test for a project with an entity type named "Collection". That test will load the project and then verify that the exp-* options are generated, but without container constraints, and that the default options are not loaded, but also that an error is not thrown. The test should explain why it's fine that the default options are not loaded.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/hub-entities.xqy
akshaysonvane: Updated to let/where/return

akshaysonvane:
- Updated the code to use let/where/return instead of if/else
- Added a JUnit test with an entity-type named "Collection"
- Fixed a copy-paste typo in deploySearchOptionsAsDataHubDeveloper test

marklogic-data-hub/src/test/java/com/marklogic/hub/entity/EntityManagerTest.java
```diff
@@ -191,6 +194,33 @@ public void overrideExistingExplorerOptions() {
...
+        entityManager.deployQueryOptions();
+
+        runAsAdmin();
+        Stream.of(HubConfig.DEFAULT_STAGING_NAME, HubConfig.DEFAULT_FINAL_NAME)
+                .forEach(dbName -> assertNull(getModulesFile("/Default/" + dbName + "/rest-api/options/exp-default.xml"), "Did not expect exp-default.xml options file in " + dbName + "since we dont use it anymore."));
+
+        String stagingExplorerOptions = getModulesFile("/Default/" + HubConfig.DEFAULT_STAGING_NAME + "/rest-api/options/" + HubConfig.EXP_STAGING_ENTITY_QUERY_OPTIONS_FILE);
+        String finalExplorerOptions = getModulesFile("/Default/" + HubConfig.DEFAULT_FINAL_NAME + "/rest-api/options/" + HubConfig.EXP_FINAL_ENTITY_QUERY_OPTIONS_FILE);
+
+        Stream.of(stagingExplorerOptions, finalExplorerOptions).forEach(option -> {
```
rjrudin, line 216: This is good, but the test needs to assert two more things - that the "default" options were written to src/main/entity-config in the project directory, and that they are not in the modules database because of the conflict between Collection constraints.

rjrudin (changes requested):

marklogic-data-hub/src/test/java/com/marklogic/hub/entity/EntityManagerTest.java
akshaysonvane: Updated the tests.

rjrudin (approved):

ryanjdew (approved):

PR #4656, opened 2020-10-01
DHFPROD-6073: Middle-tier Endpoint For Default Mastering Collections

fsnow:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/controllers/steps/MergingStepController.java
+++ b/marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/controllers/steps/MergingStepController.java
@@ -5,6 +5,7 @@ import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.marklogic.hub.central.controllers.BaseController;
 import com.marklogic.hub.central.schemas.StepSchema;
 import com.marklogic.hub.dataservices.ArtifactService;
+import com.marklogic.hub.dataservices.MasteringService;
 import com.marklogic.hub.dataservices.StepService;
 import io.swagger.annotations.ApiImplicitParam;
 import io.swagger.annotations.ApiOperation;
@@ -51,6 +52,15 @@ public class MergingStepController extends BaseController {
     public ResponseEntity<Void> deleteStep(@PathVariable String stepName) {
         newService().deleteStep(STEP_DEFINITION_TYPE, stepName);
         return emptyOk();
+    }
+
+    @RequestMapping(value = "/defaultCollections/{entityType}", method = RequestMethod.GET)
+    @ResponseBody
+    @ApiOperation("Returns the default mastering collections for an entity type")
+    public JsonNode getDefaultCollections(@PathVariable String entityType) {
+        return MasteringService
+            .on(getHubClient().getStagingClient())
+            .getDefaultCollections(entityType);
     }
 
     private StepService newService() {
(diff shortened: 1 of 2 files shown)


ryanjdew (approved):

srinathgit (dismissed):

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/controllers/MasteringController.java
```diff
@@ -0,0 +1,40 @@
...
+import com.marklogic.hub.dataservices.MasteringService;
+import io.swagger.annotations.ApiOperation;
+import org.springframework.stereotype.Controller;
+import org.springframework.web.bind.annotation.PathVariable;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestMethod;
+import org.springframework.web.bind.annotation.ResponseBody;
+
+@Controller
+@RequestMapping(value = "/api/mastering")
```
rjrudin, line 29: @fsnow  @ryanjdew  Do you anticipate having other endpoints specific to the concept of mastering? I'm wondering if these collections are specific to matching or merging and thus should be in the Matching or Merging step controller.

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/controllers/MasteringController.java
```diff
@@ -0,0 +1,40 @@
...
+public class MasteringController extends BaseController {
+
+    @RequestMapping(value = "/defaultCollections/{entityType}", method = RequestMethod.GET)
+    @ResponseBody
+    @ApiOperation("Returns the default mastering collections for an entity type")
+    public JsonNode getDefaultCollection(@PathVariable String entityType) {
```
rjrudin, line 35: Small nit, should be getDefaultCollections

rjrudin (changes requested):

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/controllers/MasteringController.java
fsnow: Fixed

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/controllers/MasteringController.java
ryanjdew: This will be used for populating the defaults to the merge step, so it could go under merging. I can't think of anything at the moment that couldn't be split between matching or merging.

fsnow: We don't have a MergingController in that directory, so it would be a rename of this one.

rjrudin: @fsnow I think this should go into MergingStepController then, as it's specific to a merging step.

rjrudin (changes requested):

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/controllers/MasteringController.java
fsnow: done

rjrudin (approved):

ryanjdew (approved):

PR #4658, opened 2020-10-02
DHFPROD-5252: Create match ruleset form and match detail view

brucean52:
### Description
Mock matching step for Customer entity type only.

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


xnikhil08 (approved):
Looks good.

Sanjeevani19 (approved):
Reviewed looks good, the zero state step details page was out of scope for this story and it was part of DHFPROD-5230 which I have been working on.

For future reference let’s coordinate on what we will be working on so that we are on the same page.

PR #4663, opened 2020-10-02
DHFPROD-4543:Map empty or null value to null when using parseDate and parseDateTime

srinathgit:
### Description

Map empty or null value to null when using parseDate and parseDateTime
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/mapping-functions/core.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/mapping-functions/core.sjs
@@ -91,8 +91,12 @@ function parseDate(value, pattern) {
   if (pattern != null) {
     pattern = pattern.toString().replace(", ", ",");
   }
-  if (value != null) {
+
+  if(isNullOrEmpty(value)){
     value = value.toString().replace(", ", ",");
+  }
+  else{
+    return null;
   }
 
   if(standardFormats.includes(pattern.trim())){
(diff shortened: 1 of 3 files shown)


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/mapping-functions/core.sjs
```diff
@@ -91,8 +91,20 @@ function parseDate(value, pattern) {
   if (pattern != null) {
     pattern = pattern.toString().replace(", ", ",");
   }
-  if (value != null) {
+  //return in case 'value' is null or undefined
```
rjrudin, line 94: I recommend removing these comments. They don't add anything - the code is clear in that if the value is null or whitespace, then null will be returned.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/mapping-functions/core.sjs
```diff
@@ -151,6 +163,19 @@ function parseDateTime(value, pattern) {
   let supportedFormats = ["YYYYMMDDThhmmss", "DD/MM/YYYY-hh:mm:ss", "DD/MM/YYYY hh:mm:ss", "YYYY/MM/DD-hh:mm:ss" , "YYYY/MM/DD hh:mm:ss"];
   let response;
   let errorMsg = `The pattern '${pattern}' cannot be applied to the value '${value}'`;
+
+  if (value) {
```
rjrudin, line 167: This logic is almost identical. You can reduce the duplication and further remove the need for any comments by introducing a new function here. 

The 'replace(", ", ",");' bit should then be done on "value" after the new function eturns false. 

The code would then read e.g.:

    if (isNullOrEmpty(value)) {
      return null;
    }

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/mapping-functions/parseDate.sjs
```diff
@@ -22,7 +22,10 @@ function testMoreParseDate() {
     test.assertEqual(xs.date("1996-07-16"), core.parseDate("1996.07.16", "yyyy.MM.dd")),
 
     test.assertEqual(null, core.parseDate("07.16.1996", "Mon DD,YYYY")),
-    test.assertEqual(null, core.parseDate("notADate", "Mon DD,YYYY"))
+    test.assertEqual(null, core.parseDate("notADate", "Mon DD,YYYY")),
+    test.assertEqual(null, core.parseDate(null, "Mon DD,YYYY")),
```
rjrudin, line 26: Throw in a test case for "undefined" as well - shouldn't require any code changes.

rjrudin (changes requested):

rjrudin (approved):

fsnow (approved):

PR #4666, opened 2020-10-05
DHFPROD-6072: Background on Login screen is still old red, please update to match t…

AkshayBajajML:
…he new graphic

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- N/A Added Tests
  

- ##### Reviewer:

- N/A Reviewed Tests


wooldridge (changes requested):
Akshay, I see Jira ID isn't included in the commit and PR title, you will need to rebase/update those with the Jira info so it doesn't cause problems in the pipeline.

PR #4669, opened 2020-10-06
DHFPROD-6085:Allow removal of properties from step when loading from gradle

srinathgit:
### Description
Allow removal of properties from step when loading from gradle
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/dataservices/StepService.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/dataservices/StepService.java
@@ -83,17 +83,18 @@ public interface StepService {
             }
 
             @Override
-            public com.fasterxml.jackson.databind.JsonNode saveStep(String stepDefinitionType, com.fasterxml.jackson.databind.JsonNode stepProperties) {
+            public com.fasterxml.jackson.databind.JsonNode saveStep(String stepDefinitionType, com.fasterxml.jackson.databind.JsonNode stepProperties, Boolean overwrite) {
                 return saveStep(
-                    this.req_saveStep.on(this.dbClient), stepDefinitionType, stepProperties
+                    this.req_saveStep.on(this.dbClient), stepDefinitionType, stepProperties, overwrite
                     );
             }
-            private com.fasterxml.jackson.databind.JsonNode saveStep(BaseProxy.DBFunctionRequest request, String stepDefinitionType, com.fasterxml.jackson.databind.JsonNode stepProperties) {
+            private com.fasterxml.jackson.databind.JsonNode saveStep(BaseProxy.DBFunctionRequest request, String stepDefinitionType, com.fasterxml.jackson.databind.JsonNode stepProperties, Boolean overwrite) {
               return BaseProxy.JsonDocumentType.toJsonNode(
                 request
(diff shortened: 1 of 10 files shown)


marklogic-data-hub/src/main/java/com/marklogic/hub/dataservices/StepService.java
```diff
@@ -140,21 +147,30 @@ private void deleteStep(BaseProxy.DBFunctionRequest request, String stepDefiniti
...
     com.fasterxml.jackson.databind.JsonNode getStepsByType(Stream<String> propertiesToReturn);
 
+    /**
+     * Invokes the saveStep operation on the database server
+     *
+     * @param stepDefinitionType	provides input
+     * @param stepProperties	provides input
+     * @return	Return the created/updated step document
+     */
+    com.fasterxml.jackson.databind.JsonNode saveStep(String stepDefinitionType, com.fasterxml.jackson.databind.JsonNode stepProperties);
```
rjrudin, line 157: Shouldn't DS have removed this method? I think we only want one saveStep method, and I think it's good that the client has to explicitly choose whether to do an overwrite or not.

rjrudin: I just tried this locally - I added overwrite:boolean to the API, then ran "gradle generatestepinterface", and the method was updated to include the boolean. That's what we want, we don't want a saveStep method that doesn't require the boolean.

marklogic-data-hub/src/test/java/com/marklogic/hub/dataservices/customStep/CreateAndUpdateCustomStep.java
```diff
@@ -0,0 +1,44 @@
...
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import com.marklogic.hub.AbstractHubCoreTest;
+import com.marklogic.hub.dataservices.StepService;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+public class CreateAndUpdateCustomStep extends AbstractHubCoreTest {
```
rjrudin, line 12: Needs to have "Test" suffix so we know it's a test

marklogic-data-hub/src/test/java/com/marklogic/hub/dataservices/customStep/CreateAndUpdateCustomStep.java
```diff
@@ -0,0 +1,44 @@
...
+    @Test
+    void customStep() {
+        ObjectMapper mapper = new ObjectMapper();
```
rjrudin, line 15: Can use objectMapper in the parent class

rjrudin (changes requested):

rjrudin (approved):
This looks good

akshaysonvane (approved):

PR #4671, opened 2020-10-06
DHFPROD-5253: Add Zip, Exact, Reduce to Single Ruleset Matching Form

brucean52:
### Description
Saves new ruleset to context only.

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/src/components/entities/matching/matching-step-detail/matching-step-detail.tsx
+++ b/marklogic-data-hub-central/ui/src/components/entities/matching/matching-step-detail/matching-step-detail.tsx
@@ -9,8 +9,7 @@ import NumberIcon from '../../../number-icon/number-icon';
 
 import { CurationContext } from '../../../../util/curation-context';
 import {
-  MatchingStep,
-  MatchRule
+  MatchingStep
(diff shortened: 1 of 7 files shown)


xnikhil08 (commented):
I see that both the properties are 'Required' here but still the Save button is not disabled even when there is no value provided in those fields. Is it part of another story ?

image

ryanjdew (dismissed):

brucean52:
> I see that both the properties are 'Required' here but still the Save button is not disabled even when there is no value provided in those fields. Is it part of another story ?
> 
> image

As we discussed, the plan is to not disable save buttons and show error messages when the form has errors. I've updated this PR to show all errors when Save is clicked.

xnikhil08 (approved):

xnikhil08 (dismissed):

timur-isangulov (approved):

Sanjeevani19 (approved):

timur-isangulov (approved):

ryanjdew (approved):

PR #4672, opened 2020-10-06
DHFPROD-4732: Explore All Data

rahulvudutala:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/entities/search/EntitySearchManager.java
+++ b/marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/entities/search/EntitySearchManager.java
@@ -126,14 +126,7 @@ public class EntitySearchManager {
 
             return queryMgr.search(queryDefinition, resultHandle, searchQuery.getStart());
         } catch (MarkLogicServerException e) {
-            // If there are no entityModels to search, then we expect an error because no search options will exist
-            if (searchQuery.getQuery().getEntityTypeIds().isEmpty() || modelManager.getModels().size() == 0) {
-                logger.warn("No entityTypes present to perform search");
-                return new StringHandle("");
-            }
-
             logger.error(e.getLocalizedMessage());
-
             // Resorting to string contains check as there isn't any other discernible difference
             if (e.getLocalizedMessage().contains(QUERY_OPTIONS)) {
                 logger.error("If this is a configuration issue, fix the configuration issues as shown in"
@@ -145,10 +138,7 @@ public class EntitySearchManager {
                         + "time for the file to get generated. This file is required to enable "
                         + "various search features.");
             }
-
-            throw new DataHubException(e.getServerMessage(), e);
-        } catch (Exception e) { //other runtime exceptions
-            throw new DataHubException(e.getLocalizedMessage(), e);
+            throw e;
         }
     }
 
@@ -167,9 +157,7 @@ public class EntitySearchManager {
             } else { //FailedRequestException || ResourceNotResendableException
                 logger.error(e.getLocalizedMessage());
             }
-            throw new DataHubException(e.getServerMessage(), e);
-        } catch (Exception e) { //other runtime exceptions
-            throw new DataHubException(e.getLocalizedMessage(), e);
+            throw e;
         }
     }
 
@@ -193,9 +181,6 @@ public class EntitySearchManager {
                     .andNot(queryBuilder.collection(entityTypeCollections),
                             queryBuilder.collection(excludedCollections));
 
-            queries.add(finalCollQuery);
-        } else { // If entity-model collections are empty, don't return any documents
-            StructuredQueryDefinition finalCollQuery = queryBuilder.and(queryBuilder.collection());
             queries.add(finalCollQuery);
         }
 
@@ -352,7 +337,7 @@ public class EntitySearchManager {
                     .newQueryOptionsManager()
                     .readOptionsAs(queryOptionsName, Format.XML, String.class);
         } catch (ResourceNotFoundException e) {
-            throw new DataHubException(String.format("Could not find search options: %s", queryOptionsName), e);
+            throw new RuntimeException(String.format("Could not find search options: %s", queryOptionsName), e);
         }
         return queryOptions;
     }
--- a/marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/managers/ModelManager.java
+++ b/marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/managers/ModelManager.java
@@ -80,9 +80,9 @@ public class ModelManager {
             } else { //FailedRequestException || ResourceNotResendableException || other runtime exceptions
                 logger.error(e.getLocalizedMessage());
             }
-            throw new DataHubException(e.getServerMessage(), e);
+            throw e;
         } catch (Exception e) {
-            throw new DataHubException(e.getLocalizedMessage(), e);
+            throw new RuntimeException(e.getLocalizedMessage(), e);
         }
 
         return jsonRes;
--- a/marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/entities/search/EntitySearchManagerTest.java
+++ b/marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/entities/search/EntitySearchManagerTest.java
@@ -20,12 +20,13 @@ import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.node.ObjectNode;
+import com.marklogic.client.DatabaseClient;
+import com.marklogic.client.MarkLogicServerException;
 import com.marklogic.client.io.Format;
 import com.marklogic.client.io.StringHandle;
 import com.marklogic.hub.central.AbstractHubCentralTest;
 import com.marklogic.hub.central.entities.search.models.DocSearchQueryInfo;
 import com.marklogic.hub.central.entities.search.models.SearchQuery;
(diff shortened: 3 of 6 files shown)


marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/entities/search/EntitySearchManagerTest.java
```diff
@@ -70,37 +70,29 @@ void searchWithTransformInStagingDatabase() {
...
         runAsDataHubDeveloper();
-        EntitySearchManager.QUERY_OPTIONS = "non-existent-options";
         String collectionDeleteQuery = "declareUpdate(); xdmp.collectionDelete(\"http://marklogic.com/entity-services/models\")";
         getHubClient().getFinalClient().newServerEval().javascript(collectionDeleteQuery).evalAs(String.class);
 
         runAsHubCentralUser();
-        assertTrue(new EntitySearchManager(getHubClient()).search(new SearchQuery()).get().isEmpty());
+        StringHandle results = new EntitySearchManager(getHubClient()).search(new SearchQuery());
+        ObjectNode node = readJsonObject(results.get());
+        assertEquals(14, node.get("total").asInt());
```
rjrudin, line 91: Any time there's an assertion on a number, it's helpful to have an assertion message to explain why that number is expected. In this case, you're expecting 14 because that's the number of OOTB DH artifacts. That assertion is a little brittle - if we add a new artifact, this test would break, which would be surprising. 

Instead of assuming 14, do a quick newServerEval and run "cts.estimate(cts.trueQuery())" to get back the count of expected documents. That allows you to make a better assertion - i.e. that you expect the count to be that of all documents in the database (which the user has read access to).

marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/entities/search/EntitySearchManagerTest.java
```diff
@@ -70,37 +70,29 @@ void searchWithTransformInStagingDatabase() {
...
 
         SearchQuery query = new SearchQuery();
         query.getQuery().setEntityTypeIds(Arrays.asList("Some-entityType"));
-        assertTrue(new EntitySearchManager(getHubClient()).search(query).get().isEmpty());
+        assertThrows(DataHubException.class, () -> new EntitySearchManager(getHubClient()).search(query));
```
rjrudin, line 95: This needs an assertion message too - why do we expect this exception? And what's the point of a DataHubException - what value is that adding over the original exception? 

I noticed that class is used in 7 places, but there's no code that cares about DataHubException. For that reason, I recommend removing it. It's needlessly wrapping the original exception but not providing any value.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/hub-entities.xqy
```diff
@@ -198,7 +198,12 @@ declare %private function hent:fix-options-for-explorer(
             </search:range>
           </search:constraint>,
           hent:build-sort-operator($sortable-properties, $entity-namespace-map),
-          hent:fix-options-for-explorer($n/node(), $sortable-properties, $entity-namespace-map)
+          hent:fix-options-for-explorer($n/node(), $sortable-properties, $entity-namespace-map),
+          <search:transform-results apply="snippet">
```
rjrudin, line 202: To avoid errors in the future, I think we need to adjust the first case on search:options to exclude "search:transform-results" if it exists. So this line needs to be adjusted:

    $n/namespace::node(),

That should become (I think, need to test):

    $n/node()[fn:not(self::search:transform-results)]

There's not a test to add because we don't expect the search options generated by ES to return a search:transform-results (though we could add an ML unit test that sends in an options document with an existing transform-results, and then verify that it's discarded in favor of this one). This is just to prevent an error from happening in the future.

rjrudin (changes requested):

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/hub-entities.xqy
rahulvudutala: es.searchOptionsgenerate function generated the search:transform-results and it defaults to empty-snippet. Datahub overrides this by commenting out. Datahub is not excluding the transform results. It looks like the following
`case element(search:transform-results) return <!--<search:transform-results apply="empty-snippet"></search:transform-results>-->`

We should probably change to the following.
`case element(search:transform-results) return (
            <search:transform-results apply="snippet">
                <per-match-tokens>30</per-match-tokens>
                <max-matches>4</max-matches>
                <max-snippet-chars>200</max-snippet-chars>
            </search:transform-results>
      )`

rjrudin: The problem with the case on transform-results is that if ES stops adding a transform-results (not likely, but could happen), then our transform-results won't get added. It's safer then to return () for transform-results, and then add our own like you're doing.

rahulvudutala: I will leave with the existing implementation. If transform-results exists, then they are commented out. If they exist, the transform-results are overridden to datahub specific options, which is happening now.

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/entities/search/EntitySearchManager.java
```diff
@@ -146,9 +139,9 @@ public StringHandle search(SearchQuery searchQuery) {
                         + "various search features.");
             }
 
-            throw new DataHubException(e.getServerMessage(), e);
+            throw new RuntimeException(e.getServerMessage(), e);
```
rjrudin, line 142: Just do "throw e" here - MarkLogicServerException is a subclass of RuntimeException, so there's no need to wrap it (vs something like IOException).

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/entities/search/EntitySearchManager.java
```diff
@@ -146,9 +139,9 @@ public StringHandle search(SearchQuery searchQuery) {
...
         } catch (Exception e) { //other runtime exceptions
-            throw new DataHubException(e.getLocalizedMessage(), e);
+            throw new RuntimeException(e.getLocalizedMessage(), e);
```
rjrudin, line 144: And this catch block can be removed entirely. Same goes for the code below.

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/managers/ModelManager.java
```diff
@@ -80,9 +80,9 @@ public JsonNode getModels() {
             } else { //FailedRequestException || ResourceNotResendableException || other runtime exceptions
                 logger.error(e.getLocalizedMessage());
             }
-            throw new DataHubException(e.getServerMessage(), e);
+            throw new RuntimeException(e.getServerMessage(), e);
```
rjrudin, line 83: Same thing here - "throw e" and remove the catch block.

rjrudin (changes requested):

marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/entities/search/EntitySearchManagerTest.java
```diff
@@ -316,4 +312,22 @@ private void validateSearchWithTransform(String databaseType) {
                 "entityProperties so that the UI knows what structured values to show for each entity instance");
         assertTrue(node.get("results").get(1).has("entityProperties"));
     }
+
+    private void validateAllDataSearch(String databaseType) {
+        DatabaseClient client = databaseType.equalsIgnoreCase("staging") ? getHubClient().getStagingClient() : getHubClient().getFinalClient();
```
rjrudin, line 317: Given that this client is used after runAsHubCentralUser is called, I think you should move this line to after runAsHubCentralUser. Looking at the code, it's not clear what user is associated with this DatabaseClient, but it looks like it's supposed to be running as an HC user.

marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/entities/search/EntitySearchManagerTest.java
```diff
@@ -316,4 +312,22 @@ private void validateSearchWithTransform(String databaseType) {
...
+        project.createCustomerInstance(new Customer(1, "Jane"), databaseType);
+        project.createCustomerInstance(new Customer(2, "Sally"), databaseType);
+
+        runAsHubCentralUser();
+
+        String query = "cts.estimate(cts.trueQuery())";
+        int count = Integer.parseInt(client.newServerEval().javascript(query).evalAs(String.class));
+        StringHandle results = new EntitySearchManager(getHubClient(), databaseType).search(new SearchQuery());
+        ObjectNode node = readJsonObject(results.get());
+        assertEquals(count, node.get("total").asInt(), "OOTB datahub artifacts, Customer and Order entity models and " +
```
rjrudin, line 330: The assertion message shouldn't assume what's being returned. The benefit of doing an estimate on trueQuery is that you get a count of all the docs that an HC user can see. Then, you want to make sure that when you do an empty search, you get the same count back. You don't care what the count is, nor do you care what docs should be there - you just care that the two counts are the same. So the assertion message should be along the lines of:

"Expected " + count + " total documents; an empty search should result in a count equal to all the docs that the user can read in the database"

rjrudin (changes requested):

rjrudin (approved):

timur-isangulov (approved):

PR #4675, opened 2020-10-06
DHFPROD-6056: Structured Streaming in Spark

anu3990:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/sql/sources/v2/HubDataSource.java
+++ b/marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/sql/sources/v2/HubDataSource.java
@@ -20,31 +20,43 @@ import com.marklogic.hub.spark.sql.sources.v2.writer.HubDataWriterFactory;
 import org.apache.spark.sql.SaveMode;
 import org.apache.spark.sql.catalyst.InternalRow;
 import org.apache.spark.sql.sources.v2.DataSourceOptions;
+import org.apache.spark.sql.sources.v2.StreamWriteSupport;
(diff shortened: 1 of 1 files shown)


marklogic-data-hub-spark-connector/src/test/java/com/marklogic/hub/spark/sql/sources/v2/AbstractSparkConnectorTest.java
```diff
@@ -157,4 +159,24 @@ protected GenericInternalRow buildRow(String... values) {
         }
         return new GenericInternalRow(rowValues);
     }
+
+    /**
+     * @param options
+     * @return
+     * */
+    protected DataWriter<InternalRow> buildStreamWriter(Options options) {
```
rjrudin, line 167: We don't need this test, as it's not testing anything new. The DataWriter that's being returned is identical to the one being returned by buildDataWriter - it's being constructed in the same fashion, since OutputMode doesn't matter. . 

What we need tests for are the commit/abort methods that take an epoch and don't take an epoch. But, we don't have any logic for those yet. That's where we need some research to figure out when Glue/Spark is invoking these methods - we should add some logging and then test out the connector to see when these methods are invoked, and what's in the WriterCommitMessage objects that are passed to it.

rjrudin (changes requested):

rjrudin (approved):
@anu3990  @SameeraPriyathamTadikonda This works, and we'll revisit the abort/commit methods once we know when Spark invokes those. We'll get tests into place at that time.

SameeraPriyathamTadikonda (approved):

PR #4676, opened 2020-10-06
DHFPROD-6097: Disabling the Parallel Run

SameeraPriyathamTadikonda:
### Description
Disabling the parallel run as the next test picks up before assertions.

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [NA] Added Tests
  

- ##### Reviewer:

- [NA] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-spark-connector/src/test/resources/junit-platform.properties
+++ b/marklogic-data-hub-spark-connector/src/test/resources/junit-platform.properties
@@ -1,0 +1,2 @@
+junit.jupiter.execution.parallel.enabled = false
+#junit.jupiter.execution.parallel.mode.default = concurrent


SameeraPriyathamTadikonda:
> This should go against develop, and then it'll be rebased onto 5.4-develop, right? I think that's better to avoid the confusion of this being the only commit for spark against 5.4-develop.

"junit-platform.properties" file in other projects are added only in 5.4-develop. So, I did the same. Should I make changes to other projects as well?

rjrudin (approved):
That's a good point about junit-platform only being on 5.4-develop

srinathgit (dismissed):

ryanjdew (approved):

rjrudin (approved):

SameeraPriyathamTadikonda:
Merging the PR manually to unblock other PRs

PR #4687, opened 2020-10-08
DHFPROD-5999:Clear user artifacts from DHF instance

srinathgit:
### Description
Clear user artifacts from DHF instance
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/impl/DataHubImpl.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/impl/DataHubImpl.java
@@ -42,6 +42,7 @@ import com.marklogic.client.io.DocumentMetadataHandle;
 import com.marklogic.client.io.JacksonHandle;
 import com.marklogic.client.io.QueryOptionsListHandle;
 import com.marklogic.hub.*;
+import com.marklogic.hub.dataservices.ArtifactService;
 import com.marklogic.hub.deploy.HubAppDeployer;
 import com.marklogic.hub.deploy.commands.*;
 import com.marklogic.hub.deploy.util.HubDeployStatusListener;
@@ -403,6 +404,18 @@ public class DataHubImpl implements DataHub, InitializingBean {
             logger.error("Failed to clear user modules, cause: " + e.getMessage(), e);
         }
         logger.info("Finished clearing user modules");
+    }
+
+    public void clearUserArtifacts(){
+        final HubClient hubClientToUse = hubClient != null ? hubClient : hubConfig.newHubClient();
+
+        long start = System.currentTimeMillis();
+        logger.info("Clearing user artifacts as user: " + hubClientToUse.getUsername());
+
+        ArtifactService.on(hubClientToUse.getStagingClient()).clearUserArtifacts();
+        ArtifactService.on(hubClientToUse.getFinalClient()).clearUserArtifacts();
+
+        logger.info("Finished clearing user artifacts; time elapsed: " + (System.currentTimeMillis() - start));
     }
 
     public void deleteDocument(String uri, DatabaseKind databaseKind) {
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/artifacts/clearUserArtifacts.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/artifacts/clearUserArtifacts.sjs
@@ -1,0 +1,24 @@
+/**
+ Copyright (c) 2020 MarkLogic Corporation
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+ http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+// No privilege required: If the user doesn't have permission to update artifacts, their deletion will fail
+'use strict';
+declareUpdate();
+const consts = require('/data-hub/5/impl/consts.sjs');
+
+let urisToDelete = cts.uris(null, null, cts.andQuery([cts.collectionQuery(consts.USER_ARTIFACT_COLLECTIONS.concat('http://marklogic.com/data-hub/mappings'))
+  ,cts.notQuery(cts.collectionQuery(consts.HUB_ARTIFACT_COLLECTION))])).toArray();
+urisToDelete.forEach(uri => xdmp.documentDelete(uri));
+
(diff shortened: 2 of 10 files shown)


marklogic-data-hub/src/main/java/com/marklogic/hub/impl/DataHubImpl.java
```diff
@@ -405,6 +405,24 @@ public void clearUserModules(List<String> resourceNamesToNotDelete) {
         logger.info("Finished clearing user modules");
     }
 
+    public void clearUserArtifacts(){
+        final HubClient hubClientToUse = hubClient != null ? hubClient : hubConfig.newHubClient();
+
+        long start = System.currentTimeMillis();
+        logger.info("Clearing user artifacts as user: " + hubClientToUse.getUsername());
+
+        String query = "declareUpdate();\n" +
```
rjrudin, line 414: We want as much logic in ML as possible via data services, so I think an ArtifactService.clearUserArtifacts method is the right approach here. 

Pinging @fsnow and @ryanjdew  for thoughts - I was thinking perhaps we should start a "datahub" service directory, where we have "global" methods like this. We also have a "system" service which just has getVersion in it. That arguably should be part of a datahub service. 

But we can punt on reorganizing things right now, as I think ArtifactService.clearUserArtifacts is reasonable. 

I think the DS endpoint should only clear that database that it's associated with - thus, this method should still exist, and it'll call clearUserArtifacts using both a staging client and a final client.

marklogic-data-hub/src/test/java/com/marklogic/hub/impl/ClearUserArtifactsTest.java
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    @Test
+    void clearUserArtifacts() {
+        try{
+            new DataHubImpl(runAsDataHubOperator()).clearUserArtifacts();
+            fail("'data-hub-operator' should not be able delete user artifacts");
+        }
+        catch (Exception e){
+            logger.info("'data-hub-operator' cannot delete user artifacts");
```
rjrudin, line 31: A good practice here is to make an assertion on what the exception message contains - that way, we know we're catching the expected exception. Can also catch a more precise exception here - I assume that ML is throwing a FailedRequestException.

marklogic-data-hub/src/test/java/com/marklogic/hub/impl/ClearUserArtifactsTest.java
```diff
@@ -0,0 +1,40 @@
...
+        }
+        new DataHubImpl(runAsDataHubDeveloper()).clearUserArtifacts();
+
+        assertNotNull(getHubClient().getStagingClient().newServerEval().javascript("cts.doc('/customer1.json')").evalAs(JsonNode.class));
+        //The staging db will have one additional document  '/customer/1.json'
+        assertEquals(initialDbCount , Integer.parseInt(getHubClient().getStagingClient().newServerEval().javascript("cts.estimate(cts.trueQuery())").evalAs(String.class))-1);
```
rjrudin, line 37: This code is in 3 places here, and @rahulvudutala needed it in a PR, so let's add it to AbstractHubTest - e.g. "getDocumentCount(DatabaseClient client)". Try to work this into @rahulvudutala 's PR too.

ml-data-hub-plugin/src/main/groovy/com/marklogic/gradle/task/ClearUserArtifactsTask.groovy
```diff
@@ -0,0 +1,14 @@
+package com.marklogic.gradle.task
+
+import com.marklogic.hub.impl.DataHubImpl
+
+class ClearUserArtifactsTask extends AbstractConfirmableHubTask {
+
+    @Override
+    void executeIfConfirmed() {
+        println "Clearing user artifacts. This requires a user with at least 'data-hub-developer' role."
```
rjrudin, line 9: I don't think this is technically true - a user just needs update access to each artifact type. I think that's the better message to give - "Clearing user artifacts. This requires a user that has update permission on each artifact type."

rjrudin (changes requested):

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/DataHubImpl.java
ryanjdew: I agree that is would be nice to compile a list needs like this from the Java middle-tier and move as much of it as possible to be server-side.

ryanjdew (approved):

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/artifacts/clearUserArtifacts.sjs
```diff
@@ -0,0 +1,33 @@
...
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+'use strict';
+//No specific privilege available for writing step definition or legacy mappings
+xdmp.securityAssert("http://marklogic.com/data-hub/privileges/write-ingestion", "execute");
```
rjrudin, line 18: I actually don't think we need any privilege checks here. The delete calls will blow up if the user doesn't have update permission on any user artifacts.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/artifacts/clearUserArtifacts.sjs
```diff
@@ -0,0 +1,33 @@
...
+xdmp.securityAssert("http://marklogic.com/data-hub/privileges/write-entity-model", "execute");
+
+const consts = require('/data-hub/5/impl/consts.sjs');
+const DataHubSingleton = require('/data-hub/5/datahub-singleton.sjs');
+const dataHub = DataHubSingleton.instance();
+
+let urisToDelete = cts.uris(null, null, cts.andQuery([cts.collectionQuery(consts.USER_ARTIFACT_COLLECTIONS.concat('http://marklogic.com/data-hub/mappings'))
+  ,cts.notQuery(cts.collectionQuery(consts.HUB_ARTIFACT_COLLECTION))])).toArray();
+let database = xdmp.databaseName(xdmp.database());
+urisToDelete.forEach(uri => dataHub.hubUtils.deleteDocument(uri, database));
```
rjrudin, line 32: No need to do an xdmp.invoke here - just do xdmp.documentDelete. We know we're not trying to do this in some other database.

marklogic-data-hub/src/testFixtures/java/com/marklogic/hub/test/AbstractHubTest.java
```diff
@@ -207,6 +207,11 @@ protected void setTestUserRoles(String... roles) {
         user.save();
     }
 
+    protected int getDocumentCount(DatabaseClient client) {
```
rjrudin, line 210: Can you put this at the bottom of the file? Will help avoid merge conflicts (I'm making changes to this file too)

rjrudin (changes requested):

rjrudin (approved):

ryanjdew (approved):

rjrudin (approved):

akshaysonvane (approved):

PR #4688, opened 2020-10-08
DHFPROD-5988: Adding temporal collections

SameeraPriyathamTadikonda:
### Description
If the list of collections specified by the user has a temporal collection then the documents are ingested vi temporal.documentInsert().
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [ ] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/ingestion/bulkIngester.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/ingestion/bulkIngester.sjs
@@ -29,11 +29,21 @@ const consts = require("/data-hub/5/impl/consts.sjs");
 const HubUtils = require("/data-hub/5/impl/hub-utils.sjs");
 const state = fn.head(xdmp.fromJSON(endpointState));
 
+const temporal = require("/MarkLogic/temporal.xqy");
+const temporalLib = require("/data-hub/5/temporal/hub-temporal.sjs");
+
 const work = fn.head(xdmp.fromJSON(workUnit));
 
 const uriPrefix = work.uriprefix != null ? work.uriprefix : "";
 
 const collections = work.collections != null ? work.collections.split(',') : [];
+const temporalCollections = temporalLib.getTemporalCollections().toArray().reduce((acc, col) => {
+    acc[col] = true;
+    return acc;
+}, {});
+
+let temporalCollection = collections.concat(collections).find((col) => temporalCollections[col]);
+
 
 const permissions = work.permissions != null ? work.permissions : 'data-hub-common,read,data-hub-common,update'
 const permissionsArray = new HubUtils().parsePermissions(permissions);
@@ -67,14 +77,27 @@ inputArray.forEach(record => {
   record = ingest.main({uri: uri, value: record}, {
     outputFormat: consts.JSON, headers: headers
   }).value;
-  xdmp.documentInsert(
-    uri,
-    record,
-    {
-      permissions: permissionsArray,
-      collections: collections
-    }
-  );
+  if(temporalCollection) {
+    const collectionsReservedForTemporal = ['latest', uri];
+    temporal.documentInsert(
+            temporalCollection,
+            uri,
+            record,
+            {
+              permissions: permissionsArray,
+              collections: collections.filter((col) => !(temporalCollections[col] || collectionsReservedForTemporal.includes(col)))
+            }
+    );
+  } else {
+    xdmp.documentInsert(
+        uri,
+        record,
+        {
+          permissions: permissionsArray,
+          collections: collections
+        }
+      );
+  }
 });
 
 const returnValue = (fn.count(input) > 0) ? state : null;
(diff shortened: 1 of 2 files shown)


marklogic-data-hub-spark-connector/src/test/java/com/marklogic/hub/spark/sql/sources/v2/WriteTemporalDataTest.java
```diff
@@ -0,0 +1,125 @@
...
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+public class WriteTemporalDataTest extends AbstractSparkConnectorTest {
+
+    @BeforeEach
+    void setUp() {
+        runAsDataHubDeveloper();
+        addFieldAndIndexes();
```
srinathgit, line 25: The setup and tear down seems to duplicate code from TemporalWriteTest. Can they be reused ?

marklogic-data-hub-spark-connector/src/test/java/com/marklogic/hub/spark/sql/sources/v2/AbstractSparkConnectorTest.java
```diff
@@ -166,4 +171,13 @@ protected GenericInternalRow buildRow(String... values) {
         }
         return new GenericInternalRow(rowValues);
     }
+
+    protected ObjectNode getDatabaseProperties(String database) {
```
srinathgit, line 175: This method seems to be duplicated in HubTestbase

srinathgit, line 175: This method seems to be duplicated in HubTestbase

marklogic-data-hub-spark-connector/src/test/java/com/marklogic/hub/spark/sql/sources/v2/WriteTemporalDataTest.java
srinathgit, line 25: The setup and tear down seems to duplicate code from TemporalWriteTest. Can they be reused ?

srinathgit (changes requested):

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/ingestion/bulkIngester.sjs
```diff
@@ -29,11 +29,21 @@ const consts = require("/data-hub/5/impl/consts.sjs");
 const HubUtils = require("/data-hub/5/impl/hub-utils.sjs");
 const state = fn.head(xdmp.fromJSON(endpointState));
 
+const temporal = require("/MarkLogic/temporal.xqy");
+const temporalLib = require("/data-hub/5/temporal/hub-temporal.sjs");
```
srinathgit, line 33: FYI, this library is present only in Datahub 5.3.0. If the connector has to work with all 5.x DataHub, this will fail

marklogic-data-hub-spark-connector/src/test/java/com/marklogic/hub/spark/sql/sources/v2/AbstractSparkConnectorTest.java
rjrudin: Let's test this via a test in the core project instead, as there's nothing specific to the connector here. In fact, as @srinathgit points out, you can likely just add a new test method to TemporalWriteTest that hits the bulkIngester endpoint.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/ingestion/bulkIngester.sjs
rjrudin: @srinathgit Is this only needed if the user is not a data-hub-developer (which has the temporal-admin privilege), or does not already have the temporal-admin role? 

The connector can install a copy of hub-temporal.sjs if it doesn't exist - e.g. DHF is prior to 5.3. The amp won't exist, so the function won't do anything. That behavior is consistent with DHF prior to 5.3 - if you don't have the temporal-admin role/privilege, then temporal.collections returns an empty array.

rjrudin (changes requested):

marklogic-data-hub-spark-connector/src/test/java/com/marklogic/hub/spark/sql/sources/v2/AbstractSparkConnectorTest.java
SameeraPriyathamTadikonda: Moved the test to TemporalWriteTest

marklogic-data-hub/src/test/java/com/marklogic/hub/flow/TemporalWriteTest.java
```diff
@@ -68,15 +76,44 @@ void writeTemporalFile() {
...
+        );
+        input.forEach(bulkInputCaller::accept);
+        bulkInputCaller.awaitCompletion();
+
+        DocumentMetadataHandle metadata = getDocumentMetadata();
+        assertEquals(collections.split(",").length + 2, metadata.getCollections().size());
+        assertTrue(metadata.getCollections().contains("fruits"));
+        assertTrue(metadata.getCollections().contains("stuff"));
+        assertTrue(metadata.getCollections().contains("temporal/test"));
+        assertTrue(metadata.getCollections().contains("latest"));
```
rjrudin, line 113: This should have an assertion stating that because it's in the "latest" collection, that means that "temporal/test" was recognized as a temporal collection. Otherwise there's no indication here that the document was inserted via temporal.documentInsert.

rjrudin (changes requested):
Can you rebase this off 5.4-develop?

SameeraPriyathamTadikonda:
closing this as I created a new PR to 5.4-develop

PR #4691, opened 2020-10-09
DHFPROD-4659: Document metadata of the document to visualize in card view

rahulvudutala:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/controllers/EntitySearchController.java
+++ b/marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/controllers/EntitySearchController.java
@@ -212,4 +212,15 @@ public class EntitySearchController extends BaseController {
 
     public static class SavedQueries extends ArrayList<SavedQuery> {
     }
+
+    public static class HubMetadata {
+        public String lastProcessedByFlow;
+        public String lastProcessedByStep;
+        public String lastProcessedDateTime;
+        public ArrayList<DocumentSourceMetadata> sources;
+    }
+
+    public class DocumentSourceMetadata {
+        String name;
+    }
 }
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/entities/entity-search-lib.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/entities/entity-search-lib.sjs
@@ -482,8 +482,23 @@ function handleDuplicateSources (propToValidate, arrayWithDuplicates) {
   return deDupedArray;
 }
 
+function addDocumentMetadataToSearchResults(searchResponse) {
+  searchResponse.results.forEach(result => {
+    let hubMetadata = {};
+    let documentMetadata = xdmp.documentGetMetadata(result.uri);
+    if(documentMetadata) {
+      hubMetadata["lastProcessedByFlow"] = documentMetadata.datahubCreatedInFlow;
+      hubMetadata["lastProcessedByStep"] = documentMetadata.datahubCreatedByStep;
+      hubMetadata["lastProcessedDateTime"] = documentMetadata.datahubCreatedOn;
+      hubMetadata["sources"] = getEntitySources(result.uri);
+    }
+    result["hubMetadata"] = hubMetadata;
+  });
+}
+
 module.exports = {
+  addDocumentMetadataToSearchResults,
   addPropertiesToSearchResponse,
-  buildPropertyMetadata: buildPropertyMetadata,
-  getEntityInstance: getEntityInstance
+  buildPropertyMetadata,
+  getEntityInstance
 };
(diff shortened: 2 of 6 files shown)


ryanjdew (dismissed):

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/entities/entity-search-lib.sjs
```diff
@@ -482,8 +482,23 @@ function handleDuplicateSources (propToValidate, arrayWithDuplicates) {
   return deDupedArray;
 }
 
+function addDocumentMetadataToSearchResults(searchResponse) {
+  searchResponse.results.forEach(result => {
+    let metadata = {};
+    let documentMetadata = xdmp.documentGetMetadata(result.uri);
+    if(documentMetadata) {
+      metadata["Flow"] = documentMetadata.datahubCreatedInFlow ? documentMetadata.datahubCreatedInFlow : "";
```
rjrudin, line 490: While we don't want to change the labels yet in the UI, I think we should start using accurate names in the code. We can't change the names in the data, but for in-memory data and variables, we can use correct names. So:

- lastProcessedByFlow
- lastProcessedByStep
- lastProcessedDateTime

That should go into the Java class that the controller uses to generate Swagger too. The UI team will see them and at least understand what the values represent.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/entities/entity-search-lib.sjs
```diff
@@ -482,8 +482,23 @@ function handleDuplicateSources (propToValidate, arrayWithDuplicates) {
...
+      metadata["Step"] = documentMetadata.datahubCreatedByStep ? documentMetadata.datahubCreatedByStep : "";
+      metadata["createdOn"] = documentMetadata.datahubCreatedOn ? documentMetadata.datahubCreatedOn : "";
+      metadata["sources"] = getEntitySources(result.uri);
+    }
+    result["metadata"] = metadata;
```
rjrudin, line 495: To make it clear that this is a hub-specific addition to the search result, let's call this "hubMetadata". Otherwise, a developer could see this and think it's provided by the Search API, and specifically that it's the metadata keys on a document.

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/entities/search/addDocumentMetadataToSearchResults.sjs
```diff
@@ -0,0 +1,101 @@
...
+  xdmp.invokeFunction(function() {
+    declareUpdate();
+    xdmp.documentSetMetadata("/content/tim.json",
+        {
+          "datahubCreatedByStep": "map-step"
+        });
+  });
+  entitySearchLib.addDocumentMetadataToSearchResults(response);
+  return[
+    test.assertEqual("", response.results[0].metadata.Flow),
```
rjrudin, line 91: Why empty string instead of not setting the property at all? This is often a problematic approach for dates and dateTimes, where "" is not a valid value. I think it's best to just not include the property at all - that makes it crystal clear that it doesn't exist for this record.

rjrudin (changes requested):

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/controllers/EntitySearchController.java
```diff
@@ -212,4 +212,15 @@ private EntitySearchService getEntitySearchService(String database) {
 
     public static class SavedQueries extends ArrayList<SavedQuery> {
     }
+
+    public static class LastProcessedMetadata {
+        public String lastProcessedByFlow;
+        public String lastProcessedByStep;
+        public String lastProcessedDateTime;
+        public ArrayList<DocumentSourceMetadata> sources;
```
rjrudin, line 220: Let's call this "HubMetadata", as it's not all specific to when the document was last processed - for example, the sources are not specific to that concept.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/entities/entity-search-lib.sjs
```diff
@@ -482,8 +482,23 @@ function handleDuplicateSources (propToValidate, arrayWithDuplicates) {
...
+    result["hubMetadata"] = hubMetadata;
+  });
+}
+
 module.exports = {
   addPropertiesToSearchResponse,
   buildPropertyMetadata: buildPropertyMetadata,
-  getEntityInstance: getEntityInstance
+  getEntityInstance: getEntityInstance,
+  addDocumentMetadataToSearchResults
```
rjrudin, line 503: Can you put this in alphabetical order with the other functions? Makes life easier for future developers to find functions.

rjrudin (changes requested):

rjrudin (approved):

ryanjdew (approved):

PR #4692, opened 2020-10-09
DHFPROD-5254: Add Synonym Type to Single Ruleset Modal

brucean52:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/src/components/entities/matching/ruleset-single-modal/ruleset-single-modal.tsx
+++ b/marklogic-data-hub-central/ui/src/components/entities/matching/ruleset-single-modal/ruleset-single-modal.tsx
@@ -1,5 +1,5 @@
 import React, { useState, useEffect, useContext } from 'react';
-import { Modal, Form } from 'antd';
+import { Modal, Form, Input, Icon } from 'antd';
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import { faLayerGroup } from "@fortawesome/free-solid-svg-icons";
 import { MLButton, MLTooltip, MLSelect } from '@marklogic/design-system';
@@ -11,6 +11,7 @@ import EntityPropertyTreeSelect from '../../../entity-property-tree-select/entit
 import { CurationContext } from '../../../../util/curation-context';
 import { MatchingStep, MatchRule, MatchRuleset } from '../../../../types/curation-types';
 import { Definition } from '../../../../types/modeling-types';
+import { NewMatchTooltips } from '../../../../config/tooltips.config';
 
 type Props = {
   isVisible: boolean;
@@ -28,6 +29,7 @@ const layout = {
 
 const MATCH_TYPE_OPTIONS = [
   { name: 'Exact', value: 'exact' },
+  { name: 'Synonym', value: 'synonym' },
   { name: 'Reduce', value: 'reduce' },
   { name: 'Zip', value: 'zip' },
 ]
@@ -35,13 +37,18 @@ const MATCH_TYPE_OPTIONS = [
 const { MLOption } = MLSelect;
  
 const MatchRulesetModal: React.FC<Props> = (props) => {
-  const { curationOptions, updateActiveStepDefinition } = useContext(CurationContext);
+  const { curationOptions, updateActiveStepArtifact } = useContext(CurationContext);
 
-  const [entityTypeDefinition, setEntityTypeDefinition] = useState<Definition>(DEFAULT_ENTITY_DEFINITION)
-  const [selectedProperty, setSelectedProperty] = useState<string | undefined>(undefined)
+  const [entityTypeDefinition, setEntityTypeDefinition] = useState<Definition>(DEFAULT_ENTITY_DEFINITION);
+
+  const [selectedProperty, setSelectedProperty] = useState<string | undefined>(undefined);
   const [propertyTypeErrorMessage, setPropertyTypeErrorMessage] = useState('');
   const [matchType, setMatchType] = useState('')
   const [matchTypeErrorMessage, setMatchTypeErrorMessage] = useState('');
+
+  const [thesaurusValue, setThesaurusValue] = useState('');
+  const [thesaurusErrorMessage, setThesaurusErrorMessage] = useState('');
+  const [filterValue, setFilterValue] = useState('')
 
   useEffect(() => {
     if (props.isVisible && curationOptions.entityDefinitionsArray.length > 0 && curationOptions.activeStep.entityName !== '') {
@@ -49,6 +56,19 @@ const MatchRulesetModal: React.FC<Props> = (props) => {
       setEntityTypeDefinition(entityTypeDefinition);
     }
   }, [props.isVisible]);
+
+  const handleInputChange = (event) => {
+    if (event.target.id === 'thesaurus-uri-input') {
+      if (event.target.value === '') {
+        setThesaurusErrorMessage('A thesaurus URI is required')
+      } else {
+        setThesaurusErrorMessage('');
+      }
+      setThesaurusValue(event.target.value);
+    } else if (event.target.id === 'filter-input') {
+      setFilterValue(event.target.value);
+    }    
+  }
 
   const closeModal = () => {
     resetModal();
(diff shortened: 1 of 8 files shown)


Sanjeevani19 (approved):

marklogic-data-hub-central/ui/src/components/entities/matching/ruleset-single-modal/ruleset-single-modal.tsx
```diff
@@ -37,19 +39,37 @@ const { MLOption } = MLSelect;
...
     if (props.isVisible && curationOptions.entityDefinitionsArray.length > 0 && curationOptions.activeStep.entityName !== '') {
       let entityTypeDefinition: Definition = curationOptions.entityDefinitionsArray.find( entityDefinition => entityDefinition.name === curationOptions.activeStep.entityName) || DEFAULT_ENTITY_DEFINITION;
       setEntityTypeDefinition(entityTypeDefinition);
     }
   }, [props.isVisible]);
 
+  const handleInputChange = (event) => {
+    if (event.target.id === 'thesaurus-uri-input') {
+      if (event.target.value === '') {
+        setThesaurusErrorMessage('A thesaurus uri is required')
```
rjrudin, line 63: Should be "URI" instead of "uri".

marklogic-data-hub-central/ui/src/components/entities/matching/ruleset-single-modal/ruleset-single-modal.tsx
```diff
@@ -76,40 +100,78 @@ const MatchRulesetModal: React.FC<Props> = (props) => {
...
+  
+          let matchRuleset: MatchRuleset = {
+            name: propertyName,
+            weight: 0,
+            matchRules: [matchRule]
+          }
+  
+          //check for duplicate
+          let propertyRulesets = curationOptions.activeStep.stepDefinition['matchRulesets'].filter(matchRuleset => matchRuleset['name'] === propertyName)
+          if (propertyRulesets.length > 0) {
```
rjrudin, line 120: @ryanjdew @brucean52 @wooldridge I think the backend should validate this. But I also don't see any requirement in 5254 about prohibiting duplicates, nor what defines a duplicate.

marklogic-data-hub-central/ui/src/components/entities/matching/ruleset-single-modal/ruleset-single-modal.tsx
```diff
@@ -76,40 +100,78 @@ const MatchRulesetModal: React.FC<Props> = (props) => {
...
+          }
+  
+          let matchRuleset: MatchRuleset = {
+            name: propertyName,
+            weight: 0,
+            matchRules: [matchRule]
+          }
+  
+          //check for duplicate
+          let propertyRulesets = curationOptions.activeStep.stepDefinition['matchRulesets'].filter(matchRuleset => matchRuleset['name'] === propertyName)
```
rjrudin, line 119: Does "stepDefinition" here refer to the configuration of this step? If so, I highly recommend not using the term "step definition", which in DHF means "an artifact that identifies a name and type and module path for a step". I actually wouldn't expect anything in between "activeStep" and "matchRulesets" either, since matchRulesets is a top-level property of a step. If "activeStep" is actually a wrapper containing the step config and other properties, I would call it something besides a "step", because its name to me implies that matchRulesets should be a property of it.

marklogic-data-hub-central/ui/src/components/entities/matching/ruleset-single-modal/ruleset-single-modal.tsx
```diff
@@ -76,40 +100,78 @@ const MatchRulesetModal: React.FC<Props> = (props) => {
...
+            propertyRulesets.forEach( ruleset => {
+              if (ruleset['matchRules'].length === 1 && ruleset['matchRules'][0].matchType === matchType) {
+                propertyErrorMessage = 'A ruleset with these selections already exists';
+              }
+            });
+          }
+  
+          if (propertyErrorMessage === '' && matchErrorMessage === '') {
+            // TODO save step to backend
+            let newStepDefinition: MatchingStep = curationOptions.activeStep.stepDefinition;
```
rjrudin, line 130: Same thing - we should call this a "step", not a "stepDefinition".

marklogic-data-hub-central/ui/src/config/tooltips.config.js
```diff
@@ -62,6 +62,9 @@ const NewMatchTooltips = {
     'sourceDatabase': 'The database where the input data is read from. Must be the same as the Matching Step *Source Database*.',
     'description': 'The description of this matching configuration.',	    'targetDatabase': 'The database where to store the processed data. Must be the same as the Matching Step *Source Database*.',
     'sourceQuery' : 'The collection or CTS query that selects the source data to process in this configuration.',
+    'thesaurusUri' : 'The path to the thesaurus used to determine synonyms',
```
rjrudin, line 65: @ryanjdew Does it matter what database the thesaurus is in? As a user, that's the first question I have when reading this.

rjrudin (changes requested):

marklogic-data-hub-central/ui/src/components/entities/matching/ruleset-single-modal/ruleset-single-modal.tsx
brucean52: stepDefinition = stepArtifact from payload. I can rename it to that.

marklogic-data-hub-central/ui/src/components/entities/matching/ruleset-single-modal/ruleset-single-modal.tsx
brucean52: I confirmed with @ryanjdew that we shouldn't allow duplicate rulesets.

rjrudin: But what is the logic for defining "duplicate"? That's not in the JIRA ticket either. This logic only is applied on rulesets whose matchRules array has a length of 1. What if it's 2? 

My main point is - since the JIRA ticket says nothing about duplicates, there shouldn't be any code in this PR for duplicates. Otherwise, how does whomever does QA on this know if the duplicate logic is correct?

brucean52: If it's 2 then it isn't a duplicate. It's a ruleset for multiple properties and not a single one. 

Error handling isn't described in all tickets. Should it be a separate ticket?

rjrudin: We only want to write code in support of requirements written in JIRA tickets. I assume that this duplicate logic would be relevant regardless of how the step is being saved - i.e. it should be applied when saving a step via Gradle. So the logic should be in the backend. Since 5254 seems to be a UI-only ticket, then a separate ticket makes sense.

marklogic-data-hub-central/ui/src/components/entities/matching/ruleset-single-modal/ruleset-single-modal.tsx
```diff
@@ -76,40 +100,68 @@ const MatchRulesetModal: React.FC<Props> = (props) => {
...
+  
+          let matchRuleset: MatchRuleset = {
+            name: propertyName,
+            weight: 0,
+            matchRules: [matchRule]
+          }
+  
+          if (propertyErrorMessage === '' && matchErrorMessage === '') {
+            // TODO save step to backend
+            let newStepArtifact: MatchingStep = curationOptions.activeStep.stepArtifact;
```
rjrudin, line 120: These 5 lines are repeated below, but I assume that can be addressed when this is enhanced to talk to the backend.

rjrudin (approved):

Sanjeevani19 (approved):

PR #4699, opened 2020-10-09
DHFPROD-4659: Visualize metadata of the document in card view

timur-isangulov:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


Sanjeevani19 (changes requested):
@timur-isangulov Could you please add unit tests for Docs with no metadata, just like you have added for docs with all metadata

timur-isangulov:
> @timur-isangulov Could you please add unit tests for Docs with no metadata, just like you have added for docs with all metadata

Added, please review again.

Sanjeevani19:
> > @timur-isangulov Could you please add unit tests for Docs with no metadata, just like you have added for docs with all metadata
> 
> Added, please review again.

@timur-isangulov  will there be a scenario where there are docs with partial metadata ? In that case we might need a test for that scenario too. Also the info tooltip doesn't stick to the card where it has been opened, as the page scrolls position also changes.

Sanjeevani19 (changes requested):
@timur-isangulov will there be a scenario where there are docs with partial metadata ? In that case we might need a test for that scenario too. 
- Also the info tooltip doesn't stick to the card where it has been opened, as the page scrolls position also changes.

marklogic-data-hub-central/ui/src/components/raw-data/raw-data-card-view/raw-data-card-view.tsx
```diff
@@ -67,6 +65,44 @@ const RawDataCardView = (props) => {
     }
   }
 
+  const displayDocumentSources = (item) => {
+    let sources = item.hubMetadata?.sources.map((el) => {
+      return el.name;
+    }).join(", ");
```
rjrudin, line 71: @timur-isangulov  @rahulvudutala  @yasht2010  Should consider that a source name could have a comma in it. We don't expect that, but it's also not prohibited. That's a reason why I favor some sort of structured grouping any time we list sources, as opposed to relying on textual delimiters.

marklogic-data-hub-central/ui/src/components/raw-data/raw-data-card-view/raw-data-card-view.tsx
```diff
@@ -67,6 +65,44 @@ const RawDataCardView = (props) => {
...
+    return sources;
+  }
+
+  const emptyField = (
+    <p className={styles.textDisabled}>none</p>
+  )
+
+  const displayDocumentMetadata = (item) => {
```
rjrudin, line 79: Can you change this to "RecordMetadata"? We are trying to use the term "Record" instead of document, as Record is more familiar to a user. Additionally, "document metadata" is a specific thing in MarkLogic - https://docs.marklogic.com/xdmp.documentGetMetadata - and that differs from what we are defining as "metadata" here.

marklogic-data-hub-central/ui/src/components/raw-data/raw-data-card-view/raw-data-card-view.tsx
```diff
@@ -67,6 +65,44 @@ const RawDataCardView = (props) => {
...
+      <div className={styles.popover} data-testid={item.uri + '-popover'}>
+        <div className={styles.colKey}>
+          <p>Source:</p>
+          <p>Flow:</p>
+          <p>Step:</p>
+          <p>Created On:</p>
+        </div>
+        <div className={styles.colValue}>
+          {item.hubMetadata?.sources ? <span className={styles.valText} data-testid={item.uri + '-sources'}>
+            <OverflowTooltip title={displayDocumentSources(item)} placement={"bottom"} content={displayDocumentSources(item).substring(0, 28)} width={"200px"} />
```
rjrudin, line 90: Likewise, "RecordSources" would be more appropriate here. 

In fact, I'd rather this component be called "record-view" instead of "raw-data-card". This data is not necessarily "raw", which we don't have a clear definition of either. It may be highly structured data, just not an entity instance.

rjrudin (changes requested):

marklogic-data-hub-central/ui/src/components/raw-data/raw-data-card-view/raw-data-card-view.tsx
rahulvudutala: @rjrudin This is just an overview of source names limited to 23 characters. The accurate Record metadata can be seen in the Detail view. I assume it should be alright for one edge case. 
@yasht2010 can confirm.

timur-isangulov:
> @timur-isangulov will there be a scenario where there are docs with partial metadata ? In that case we might need a test for that scenario too.
> 
> * Also the info tooltip doesn't stick to the card where it has been opened, as the page scrolls position also changes.

The will be cases with particular metadata but since the two other cases are covered with existing and missing metadata probably it doesn't add much value. 
Fixed the issue with the popover. Thank you for noticing!

marklogic-data-hub-central/ui/src/components/raw-data/raw-data-card-view/raw-data-card-view.tsx
timur-isangulov: Changed to RecordSources and renamed component. Thanks.

Sanjeevani19 (approved):

rahulvudutala (dismissed):

xnikhil08 (changes requested):
Just few CSS changes as mentioned below.
1. Change the cursor to pointer when user hovers over the metadata icon. It should be consistent with other clickable icons in the application.
2. Reduce the padding on the right(like the one in design ) so the text looks evenly spread in the popover.
3. Add just one more space between 'CreatedOn:' and the Date part after it.

timur-isangulov:
Due to high number of file conflicts for several commits opened https://github.com/marklogic/marklogic-data-hub/pull/4716 with addressing Nikhil's comment.

PR #4700, opened 2020-10-10
DHFPROD-5534:Clear user modules from DHS instance

srinathgit:
### Description
Clear user modules from DHS instance
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


marklogic-data-hub/src/main/java/com/marklogic/hub/impl/DataHubImpl.java
```diff
@@ -402,8 +403,9 @@ public void clearUserModules(List<String> resourceNamesToNotDelete) {
             runInDatabase(query, hubConfig.getDbName(DatabaseKind.MODULES));
         } catch (Exception e) {
             logger.error("Failed to clear user modules, cause: " + e.getMessage(), e);
+            throw new RuntimeException(e);
```
rjrudin, line 406: Let's drop the logger command above instead include the message in the first arg to RuntimeException.

marklogic-data-hub/src/test/java/com/marklogic/hub/impl/ClearUserModulesTest.java
```diff
@@ -0,0 +1,83 @@
...
+import java.util.List;
+import java.util.Map;
+import org.apache.commons.io.FileUtils;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Assumptions;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+class ClearUserModulesTest extends AbstractHubCoreTest {
+    ClearUserModulesTest() {
```
rjrudin, line 22: No need for a constructor

ml-data-hub-plugin/src/main/groovy/com/marklogic/gradle/DataHubPlugin.groovy
```diff
@@ -139,6 +139,9 @@ class DataHubPlugin implements Plugin<Project> {
                 "Requires -Pconfirm=true to be set so this isn't accidentally executed.")
         project.task("hubClearUserArtifacts", type: ClearUserArtifactsTask, group: developGroup,
             description: "Clears all user artifacts in the staging, final databases. Requires -Pconfirm=true to be set so this isn't accidentally executed.")
+        project.task("hubClearUserModules", type: ClearUserModulesTask, group: developGroup,
+            description: "Clears user modules in the modules database, only leaving only the modules " +
```
rjrudin, line 143: Couple typos - do "only leaving the modules that are loaded when Data Hub is installed."

ml-data-hub-plugin/src/main/groovy/com/marklogic/gradle/task/ClearUserModulesTask.groovy
```diff
@@ -0,0 +1,14 @@
+package com.marklogic.gradle.task
+
+import com.marklogic.hub.impl.DataHubImpl
+
+class ClearUserModulesTask extends AbstractConfirmableHubTask {
+
+    @Override
+    void executeIfConfirmed() {
+        println "Clearing user modules. This requires a user with at least 'data-hub-developer' role."
```
rjrudin, line 9: I think technically, the user just needs the data-hub-module-writer role and the rest-writer privilege. But I think it's good to say data-hub-developer here because for modules, I don't think we want to advertise anything less than the data-hub-developer role.

marklogic-data-hub/src/test/java/com/marklogic/hub/impl/ClearUserModulesTest.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    @BeforeEach
+    void setUp() throws IOException {
+        Assumptions.assumeTrue(this.isVersionCompatibleWith520Roles());
+        ClassLoader classLoader = Thread.currentThread().getContextClassLoader();
```
rjrudin, line 28: It looks like this is setting up a project with a user module of each type. Just make a new test-project instead - e.g. test-project/all-module-types. And then add a file for each module type. That avoids all of this custom code, you can just use installProjectInFolder.

marklogic-data-hub/src/test/java/com/marklogic/hub/impl/ClearUserModulesTest.java
```diff
@@ -0,0 +1,83 @@
...
+        FileUtils.copyInputStreamToFile(classLoader.getResourceAsStream("ml-modules-staging/options/default.xml"), optionsFile.resolve("myOptions.xml").toFile());
+        Path transformsFile = this.getHubProject().getModulesDir().resolve("transforms");
+        transformsFile.toFile().mkdirs();
+        FileUtils.copyInputStreamToFile(classLoader.getResourceAsStream("ml-modules/transforms/hubEntitySearchTransform.sjs"), transformsFile.resolve("myTransform.sjs").toFile());
+        this.installProjectInFolder("flow-runner-test");
+    }
+
+    @Test
+    void testClearUserModules() {
+        ServerConfigurationManager configMgr = this.runAsDataHubDeveloper().newFinalClient().newServerConfigManager();
```
rjrudin, line 43: I think these assertions to verify that each user module type was actually loaded should be part of the setup. Though since there's only one test method, I wouldn't bother with a BeforeEach method. Just make a private method called "installProjectWithEachTypeOfModule" - that documents what's going on here too. That method will use installProjectInFolder and will then verify that each user module exists.

marklogic-data-hub/src/test/java/com/marklogic/hub/impl/ClearUserModulesTest.java
```diff
@@ -0,0 +1,83 @@
...
+        QueryOptionsManager finalOptionsManager = configMgr.newQueryOptionsManager();
+        Map<String, String> optionsMap = ((QueryOptionsListHandle)finalOptionsManager.optionsList(new QueryOptionsListHandle())).getValuesMap();
+        Assertions.assertEquals(2, optionsMap.keySet().size());
+        Assertions.assertTrue(optionsMap.containsKey("myOptions"));
+        String query = "cts.estimate(cts.collectionQuery('hub-core-module'))";
+        int hubCoreModuleCount = Integer.parseInt((String)this.getHubClient().getModulesClient().newServerEval().javascript(query).evalAs(String.class));
```
rjrudin, line 49: I think that before you install the project with user modules, you need to get counts of all docs in the modules database, all docs in hub-core-module, and then counts of each REST extension type that DHF installs - services, transforms, and options. 

Then you install the user modules project, and you verify that the counts are now increased by the appropriate amount.

Then you clear user modules, and you verify the counts are back to what they used to be.

To simplify all of this, I recommend creating a class - ModuleCounts - that has 5 fields, one for each count. Then make a private method that returns a ModuleCounts, doing all the queries necessary to get the correct amount.

I know that's a lot to test, but it's much better to automate this, this would be a real pain to test manually. And because of REST extensions resulting in multiple modules, we really need to verify individual counts of each extension type.

rjrudin (changes requested):

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/DataHubImpl.java
```diff
@@ -401,9 +402,9 @@ public void clearUserModules(List<String> resourceNamesToNotDelete) {
                     "] ! xdmp:document-delete(.)\n";
             runInDatabase(query, hubConfig.getDbName(DatabaseKind.MODULES));
         } catch (Exception e) {
-            logger.error("Failed to clear user modules, cause: " + e.getMessage(), e);
+            throw new RuntimeException(e.getMessage());
```
rjrudin, line 405: Let's keep the message though that was being logged, but just put it in the exception - e.g. throw new RuntimeException("Failed to clear user modules, cause: " + e.getMessage(), e).

marklogic-data-hub/src/test/java/com/marklogic/hub/impl/ClearUserModulesTest.java
```diff
@@ -0,0 +1,126 @@
...
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Assumptions;
+import org.junit.jupiter.api.Test;
+
+class ClearUserModulesTest extends AbstractHubCoreTest {
+
+    @Test
+    void testClearUserModules() {
+        new DataHubImpl(runAsDataHubDeveloper()).clearUserModules();
+        Assumptions.assumeTrue(isVersionCompatibleWith520Roles());
```
rjrudin, line 25: This is fine, but for future tests, check out the Intellij Tips Wiki page I just wrote that explains how to auto-import static functions. The test is easier to read if "Assertions." isn't repeated in many places.

marklogic-data-hub/src/test/java/com/marklogic/hub/impl/ClearUserModulesTest.java
```diff
@@ -0,0 +1,126 @@
...
+        ModuleCounts beforeClearingModules = new ModuleCounts();
+        beforeClearingModules.setModuleCounts();
+
+        Assertions.assertEquals(1, beforeClearingModules.optionsCount);
```
rjrudin, line 29: I don't think we want precise counts here, because these assertions will break whenever a REST extension is added/removed (most likely removed). I think all we care about is capturing the initial counts. No need to assert on the expected values.

marklogic-data-hub/src/test/java/com/marklogic/hub/impl/ClearUserModulesTest.java
```diff
@@ -0,0 +1,126 @@
...
+            Assertions.fail("'data-hub-operator' should not be able delete user modules");
+        } catch (RuntimeException e) {
+            Assertions.assertTrue(e.getMessage().contains("User is not allowed to delete /config/query"));
+            logger.info("'data-hub-operator' cannot delete user modules");
+        }
+
+        ModuleCounts afterClearingModulesAsOperator = new ModuleCounts();
+        afterClearingModulesAsOperator.setModuleCounts();
+
+        Assertions.assertEquals(2, afterClearingModulesAsOperator.optionsCount);
```
rjrudin, line 46: These assertions should then be relative to the before counts - e.g. optionsCount = before + 1.

marklogic-data-hub/src/test/java/com/marklogic/hub/impl/ClearUserModulesTest.java
```diff
@@ -0,0 +1,126 @@
...
+
+        ModuleCounts afterClearingModulesAsDeveloper = new ModuleCounts();
+        afterClearingModulesAsDeveloper.setModuleCounts();
+
+        Assertions.assertEquals(beforeClearingModules.optionsCount, afterClearingModulesAsDeveloper.optionsCount);
+        Assertions.assertEquals(beforeClearingModules.transformsCount, afterClearingModulesAsDeveloper.transformsCount);
+        Assertions.assertEquals(beforeClearingModules.resourceExtensionCount, afterClearingModulesAsDeveloper.resourceExtensionCount);
+        Assertions.assertEquals(beforeClearingModules.hubCoreModuleCount, afterClearingModulesAsDeveloper.hubCoreModuleCount);
+        //Before test starts , additional custom modules may be present
+        Assertions.assertTrue(afterClearingModulesAsDeveloper.totalModulesCount <= beforeClearingModules.totalModulesCount);
```
rjrudin, line 62: For assertTrue/assertFalse, it's often helpful to have an assertion message that shows the actual values; otherwise JUnit just says that we didn't get true back, which makes debugging more difficult. We can remove the comment in favor of an assertion message too - e.g. "Expected total module count to be same or less than the before count, as some custom modules may have existed from other tests; before count: " + beforeCount + "; after count: " + afterCount

Generally, assertion messages are better at explaining things than comments, and they're more likely to stay in sync with the code.

marklogic-data-hub/src/test/resources/test-projects/all-module-types/modules/options/myOptions.xml
```diff
@@ -0,0 +1,15 @@
+<options xmlns="http://marklogic.com/appservices/search">
```
rjrudin, line 1: Just make this an empty options element, we don't need anything further in it.

marklogic-data-hub/src/test/resources/test-projects/all-module-types/modules/services/myResourceExtension.sjs
```diff
@@ -0,0 +1,53 @@
...
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+*/
+'use strict';
+const DataHub = require("/data-hub/5/datahub.sjs");
+const datahub = new DataHub();
+
+
+function get(context, params) {
```
rjrudin, line 21: Same principle as above - we don't need anything real in here, just have a "get" function that does nothing. Otherwise, the reader may think that what's in here matters, but it doesn't.

marklogic-data-hub/src/test/resources/test-projects/all-module-types/modules/transforms/myTransform.sjs
```diff
@@ -0,0 +1,34 @@
...
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+'use strict';
+
+const ds = require("/data-hub/5/data-services/ds-utils.sjs");
+const entitySearchLib = require("/data-hub/5/entities/entity-search-lib.sjs");
+
+// Expects JSON content
+function transform(context, params, content) {
```
rjrudin, line 22: Same thing, just return content, nothing further

rjrudin (changes requested):

rjrudin (dismissed):

srinathgit:
Rebased against 5.4-develop

rjrudin (dismissed):

ryanjdew (approved):

srinathgit:
Had to make a few changes in the test as test compilation failed after PR #4695 got merged

rjrudin (approved):

akshaysonvane (approved):

PR #4705, opened 2020-10-12
DHFPROD-6062:Review Explore test data

Sanjeevani19:
### Description

- Removed unused files from mock-data folder
- Subcategorized mock data into 3 categories Curation, Explore, Modeling and kept the common files under assets/mock-data
- Removed enzyme tests for search-result and side-bar components, updated the existing RTL tests to have all those scenarios
- Deleted search-results.test.tsx file since it was only testing whether the component is being rendered or not ( Todo - Add RTL test for search-results component)

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [N/A] Added Tests
  

- ##### Reviewer:

- [N/A] Reviewed Tests


wooldridge (changes requested):
Great work! Looks like we use "modeling" (1 L) across the app so we should use that for the subfolder under mock-data.

Also, what do you think about using "curate" (same as page name) and "explore" for the other subfolders? Consistency and shortness is good.

I know we use Explorer in some places and Explore in others so maybe there's a reason for explorer...

Sanjeevani19:
> Great work! Looks like we use "modeling" (1 L) across the app so we should use that for the subfolder under mock-data.
> 
> Also, what do you think about using "curate" (same as page name) and "explore" for the other subfolders? Consistency and shortness is good.
> 
> I know we use Explorer in some places and Explore in others so maybe there's a reason for explorer...

@wooldridge made those changes, ready for review.

briantang (approved):

wooldridge (approved):

PR #4713, opened 2020-10-13
DHFPROD-5757:Ensure the flow panel scrolled upon adding a step to a flow in Run

chriskellyml:
### Description

DHFPROD-5757: Ensure the flow panel scrolled upon adding a step to a flow in Run

- When the user adds a step from any interface, the flow panel is opened and scrolled to the right.

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests: Checked affected tests are passing but did feel an additional automated test was required
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/package-lock.json
+++ b/marklogic-data-hub-central/ui/package-lock.json
@@ -1672,6 +1672,31 @@
         "use-debounce": "^3.4.2"
       }
     },
+    "@marklogic/react-compound-slider": {
+      "version": "3.0.0",
+      "resolved": "https://registry.npmjs.org/@marklogic/react-compound-slider/-/react-compound-slider-3.0.0.tgz",
+      "integrity": "sha512-fJywgC1edg3w34DK8Q/M64cNEulr1AmzapY9EK6/giWOATHznvflV0SVWcHvG4O08OBOoZ3GVsEQ0VYJBudedw==",
+      "requires": {
+        "@babel/runtime": "^7.10.4",
+        "d3-array": "^2.4.0",
+        "warning": "^4.0.3"
+      },
+      "dependencies": {
+        "@babel/runtime": {
+          "version": "7.11.2",
+          "resolved": "https://registry.npmjs.org/@babel/runtime/-/runtime-7.11.2.tgz",
+          "integrity": "sha512-TeWkU52so0mPtDcaCTxNBI/IHiz0pZgr8VEFqXFtZWpYD08ZB6FaSwVAS8MKRQAP3bYKiVjwysOJgMFY28o6Tw==",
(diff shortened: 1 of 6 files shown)


srinathgit (changes requested):
1. I rebased your branch locally and added a step to a flow that has horizontal scrollbar, I don't see the scrollbar scrolled to the far right and showing the newly added step.
https://drive.google.com/file/d/1BZJqhoXra2MXLv5b67aOpkyJTKpQDDSV/view
2. Once you fix the functionality, please rebase against feature/5.4-develop and push to your feature branch

chriskellyml:
> 1. I rebased your branch locally and added a step to a flow that has horizontal scrollbar, I don't see the scrollbar scrolled to the far right and showing the newly added step.

@srinathgit This is not what the fix does. When a user adds a new step it should scroll to the right if it does not fit on the page so the user can see the step they just added. The steps should not be scrolled to the right by default or when you load the page as you showed in your screen capture.

srinathgit:
> > 1. I rebased your branch locally and added a step to a flow that has horizontal scrollbar, I don't see the scrollbar scrolled to the far right and showing the newly added step.
> 
> @srinathgit This is not what the fix does. When a user adds a new step it should scroll to the right if it does not fit on the page so the user can see the step they just added. The steps should not be scrolled to the right by default or when you load the page as you showed in your screen capture.

When a user adds a step, he is automatically redirected to “Run” tile and the flow to which the step was added is opened. My understanding as per AC in the story is that when this happens, the scrollbar is automatically scrolled to far right to display the step which I don’t see happening

srinathgit:
I misunderstood the scope of the ticket. For DHFPROD-5757, we can restrict the scope to only adding steps from the Run tile(as described in the ticket) and consider scrolling to far right  when adding steps from Load, Curate tile in a separate ticket
I tried adding step from Run tile and see some scenarios where it’s not working correctly

https://drive.google.com/file/d/146f9ALmDXoNM8VWudGlc4dmA_B2JmNGt/view

srinathgit (changes requested):
1. Looks like it needs a rebase
2. Commit message should be of format,(DHFPROD-####:<msg>) e.g. DHFPROD-5757:Ensure the flow panel scrolled upon adding a step to a flow in Run

Sanjeevani19 (approved):

srinathgit:
Tests are handled in separate task.

srinathgit (approved):

Sanjeevani19:
> Tests are handled in separate task.

Tests will be added as a part of https://project.marklogic.com/jira/browse/DHFPROD-6157

PR #4725, opened 2020-10-15
DHFPROD-6117:Function picker dropdown should be sorted in alphabetical order

srinathgit:
### Description
Function picker dropdown should be sorted in alphabetical order
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/mapping/getMappingFunctions.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/mapping/getMappingFunctions.sjs
@@ -18,4 +18,15 @@ const esMappingLib = require("/data-hub/5/builtins/steps/mapping/entity-services
 
 let mlFunctions = esMappingLib.getMarkLogicMappingFunctions();
 let xpathFunctions = esMappingLib.getXpathMappingFunctions();
-Object.assign({}, mlFunctions, xpathFunctions);
+let mappingFunctions = mlFunctions.concat(xpathFunctions);
+mappingFunctions.sort(function(funcA, funcB) {
+  var nameA = String(funcA.functionName).toLowerCase();
+  var nameB = String(funcB.functionName).toLowerCase();
+  if (nameA < nameB) {
+    return -1;
+  }
+  if (nameA > nameB) {
+    return 1;
+  }
+  return 0;
+});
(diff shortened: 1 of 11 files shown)


ryanjdew (commented):
Should the sorting be done on the frontend? Have we tested if browsers honor the sorted order? Key order isn't something guaranteed by the JSON schema and I believe behavior is left to the implementation.

srinathgit:
> Should the sorting be done on the frontend? Have we tested if browsers honor the sorted order? Key order isn't something guaranteed by the JSON schema and I believe behavior is left to the implementation.

It works fine in the browser and believe we did the sorting in the backend for other artifacts as well.
functions

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/mapping/getMappingFunctions.sjs
```diff
@@ -18,4 +18,16 @@ const esMappingLib = require("/data-hub/5/builtins/steps/mapping/entity-services
 
 let mlFunctions = esMappingLib.getMarkLogicMappingFunctions();
 let xpathFunctions = esMappingLib.getXpathMappingFunctions();
-Object.assign({}, mlFunctions, xpathFunctions);
+let unOrderedMappingFunctions = Object.assign({}, mlFunctions, xpathFunctions);
+let orderedMappingFunctions = {};
```
rjrudin, line 22: @srinathgit @xnikhil08 How easily can this return an array instead of an object? An array seems expected here and also addresses @ryanjdew 's concern about the order of keys not being guaranteed.

ryanjdew (approved):

rjrudin (approved):

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/mapping/getMappingFunctions.sjs
srinathgit: The endpoint now returns an array.

ryanjdew (approved):

rjrudin (approved):

srinathgit:
@rjrudin , @ryanjdew 
Can one of you please check "Reviewed Tests" checkbox ?

xnikhil08 (approved):

PR #4736, opened 2020-10-19
DHFPROD-6121: Set the right target database in mapping step navigation

akshaysonvane:
### Description
- Added "targetDatabase" key with the value being the name of the database to RunStepResponse.
- The JobsController then replaces the targetDatabase name with the Database kind (staging/final) so the FE can set the correct database for data exploration.
- Also added the functionality to "explore curated data" by clicking on the status icon for the step.

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/controllers/JobsController.java
+++ b/marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/controllers/JobsController.java
@@ -19,6 +19,8 @@ package com.marklogic.hub.central.controllers;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.node.ArrayNode;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import com.marklogic.hub.DatabaseKind;
 import com.marklogic.hub.job.JobDocManager;
 import io.swagger.annotations.ApiModelProperty;
 import io.swagger.annotations.ApiOperation;
@@ -29,6 +31,7 @@ import org.springframework.web.bind.annotation.RequestMethod;
 import org.springframework.web.bind.annotation.ResponseBody;
 
 import java.util.Map;
+import java.util.Optional;
 
 @Controller
 @RequestMapping(value = "/api/jobs")
@@ -62,10 +65,36 @@ public class JobsController extends BaseController {
             });
             return array;
         } else if (jobJSON.has("job")) {
-            return jobJSON.get("job");
+            return processTargetDatabase(jobJSON.get("job"));
         } else {
             return jobJSON;
         }
+    }
+
+    /**
+     * Modifies the passed in JsonNode to replace the targetDatabase name with the Database kind
+     * so the FE can set the correct database for data exploration.
+     * @param jobNode
+     * @return - Modified JsonNode
+     */
+    protected JsonNode processTargetDatabase(JsonNode jobNode) {
+        Optional.of(jobNode)
+                .map(node -> node.get("stepResponses"))
+                .ifPresent(stepResponsesNode -> {
+                    stepResponsesNode.forEach(stepResponseNode -> {
+                        String targetDatabase = Optional.ofNullable(stepResponseNode.get("targetDatabase"))
+                                .map(JsonNode::asText)
+                                .orElse(null);
+                        if (getHubClient().getDbName(DatabaseKind.STAGING).equals(targetDatabase)) {
+                            ((ObjectNode) stepResponseNode).put("targetDatabase", "staging");
+                        }
+                        else if ((getHubClient().getDbName(DatabaseKind.FINAL).equals(targetDatabase))) {
+                            ((ObjectNode) stepResponseNode).put("targetDatabase", "final");
+                        }
+                    });
+                });
+
+        return jobNode;
     }
 
     public static class Job {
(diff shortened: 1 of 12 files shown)


ryanjdew (dismissed):

marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/controllers/FlowControllerTest.java
```diff
@@ -126,6 +126,7 @@ void test() throws Exception {
                 .andDo(result -> {
                     JsonNode response = parseJsonResponse(result);
                 assertEquals(mappingInfo.targetEntityType, response.path("stepResponses").path("1").path("targetEntityType").asText(), "Info for the step should specify the Target Entity Type; response: " + response);
+                assertEquals("final", response.path("stepResponses").path("1").path("targetDatabase").asText(), "Info for the step should specify the target database; response: " + response);
```
rjrudin, line 129: Is this always true? Or is it only true if targetDatabase is configured in the step options?

marklogic-data-hub/src/main/java/com/marklogic/hub/step/RunStepResponse.java
```diff
@@ -90,6 +92,7 @@ public RunStepResponse withStep(String stepNum) {
         if (targetEntityTextNode != null) {
             this.targetEntityType = targetEntityTextNode.asText();
         }
+        Optional.of(step.getOptions()).map(optionMap -> (TextNode) optionMap.get("targetDatabase")).ifPresent(node -> this.targetDatabase = node.asText());
```
rjrudin, line 95: It's a bit difficult to look at this and be sure there's no error. I think some unit tests are needed for the withStep method - none exist yet. Those tests would verify that targetDatabase is set if it exists in the options for the given step, but not set if targetDatabase doesn't exist in the given step's options but does exist in another step's options.

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/controllers/JobsController.java
```diff
@@ -62,12 +65,38 @@ private JsonNode flattenJobsJson(JsonNode jobJSON) {
...
             return array;
         } else if (jobJSON.has("job")) {
-            return jobJSON.get("job");
+            return processTargetDatabase(jobJSON.get("job"));
         } else {
             return jobJSON;
         }
     }
 
+    /**
```
rjrudin, line 74: I don't see any test coverage for this. I think it would be reasonable to do unit tests against this method with a fake jobNode - i.e. no need to run a job first.

rjrudin (changes requested):

marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/controllers/FlowControllerTest.java
akshaysonvane: It's only true if targetDatabase is configured in step options.

akshaysonvane:
Added tests for RunStepResponse and JobsController

timur-isangulov (approved):

marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/controllers/FlowControllerTest.java
rjrudin: Can you add that to the assertion message? The current assertion message is restating what the assertEquals is communicating - that we expect the stepResponse to have the targetDatabase. The assertion message needs to explain why that is - in this case, the reason is that if the step options contain targetDatabase, then it's expected to be added to the stepResponse.

marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/controllers/JobsControllerTest.java
```diff
@@ -0,0 +1,55 @@
...
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.marklogic.hub.DatabaseKind;
+import com.marklogic.hub.central.AbstractHubCentralTest;
+import org.junit.jupiter.api.Test;
+import org.springframework.beans.factory.annotation.Autowired;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class JobsControllerTest extends AbstractHubCentralTest {
```
rjrudin, line 27: Because this test has no dependency on ML or Spring, don't extend this base class. Just instantiate a new JobsController. That makes it clear that the method being tested has no dependencies too (and the test runs a bit faster too).

rjrudin (changes requested):

marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/controllers/JobsControllerTest.java
akshaysonvane: The processTargetDatabase function in JobsController depends on Spring since it needs access to the HubClient for getting the database name.

rjrudin: Ah - bummer

rjrudin (approved):

timur-isangulov (approved):

PR #4743, opened 2020-10-21
DHFPROD-6038:Ensure HC mapping tool works when step reads from final db

srinathgit:
### Description
Ensure HC mapping tool works when step reads from final db
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/mapping/getUris.api
```diff
@@ -0,0 +1,17 @@
+{
+    "functionName": "getUris",
+    "params": [
```
rjrudin, line 3: Same thing, add a "desc" here to explain what this is doing

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/mapping/getUris.api
```diff
@@ -0,0 +1,17 @@
...
+        {
+           "name": "stepName",
+           "datatype": "string"
+        },
+        {
+           "name": "limit",
```
rjrudin, line 9: I think this would benefit from a "desc" stating what the limit is if no value is provided

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/mapping/getUris.sjs
```diff
@@ -0,0 +1,21 @@
...
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+'use strict';
+const core = require('/data-hub/5/artifacts/core.sjs')
+
+xdmp.securityAssert("http://marklogic.com/data-hub/privileges/read-mapping", "execute");
+
+var stepName, limit;
```
rjrudin, line 18: Is "limit" required then? I think to make this safe, it should default to something like 10 if no value is provided.

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/data-services/mapping/getUrisAndDoc/suiteSetup.sjs
```diff
@@ -0,0 +1,32 @@
+declareUpdate();
```
rjrudin, line 1: I think you can reuse the test-data/content concept here - the docs are written to "raw-content" by default, so you can use that as your collection to query on instead of "finalCollection".

rjrudin (changes requested):

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/mapping/getDocument.api
```diff
@@ -0,0 +1,17 @@
...
+           "name": "stepName",
+           "datatype": "string"
+        },
+        {
+           "name": "uri",
+           "datatype": "string"
+        }
+    ],
+    "return": {
+        "datatype": "textDocument",
```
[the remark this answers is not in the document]
srinathgit: If I had string, I see the format being set to JSON when I generated the interfaces. Hence, I had to make it "textDocument".

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/mapping/getUris.sjs
srinathgit: "limit" is required parameter. Should we still have a default ?

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/mapping/getDocument.api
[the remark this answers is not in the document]
rjrudin: Yeah that's why I pinged @ehennum - curious if there's a better practice here for an endpoint that can return a JSON or XML document.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/mapping/getUris.sjs
rjrudin: It's defined as a string, which means null could be passed in. It should really be defined as an integer. DS will likely generate an Integer for it, in which case it can still be null. But yeah, if it's an Integer, then it seems fine for it to be required.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/mapping/getUris.api
rjrudin: And make this "int" or "integer", whichever DS prefers.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/mapping/getDocument.sjs
```diff
@@ -0,0 +1,21 @@
...
+ limitations under the License.
+ */
+'use strict';
+const core = require('/data-hub/5/artifacts/core.sjs')
+
+xdmp.securityAssert("http://marklogic.com/data-hub/privileges/read-mapping", "execute");
+
+var stepName, uri;
+let mappingStep = core.getArtifact("mapping", stepName);
+fn.head(xdmp.eval("String(cts.doc('" + uri + "'))", null, {database: xdmp.database(mappingStep.sourceDatabase)}));
```
rjrudin, line 20: Is it guaranteed that sourceDatabase will exist? I feel this needs to be a little more defensive - if sourceDatabase doesn't exist, I think it should just do a cts.doc(uri) on the current database (no eval).

rjrudin: Also, if the mappingStep is not found, I think this should throw an error via ds-util - probably throwBadRequest.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/mapping/getUris.sjs
```diff
@@ -0,0 +1,22 @@
...
+let mappingStep = core.getArtifact("mapping", stepName);
+xdmp.eval("cts.uris(null, ['limit=" + limit + "'], " + mappingStep.sourceQuery + ")", null, {database: xdmp.database(mappingStep.sourceDatabase)}).toArray();
+
+pm
```
rjrudin, line 22: Typo?

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/mapping/getUris.sjs
```diff
@@ -0,0 +1,22 @@
...
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+'use strict';
+const core = require('/data-hub/5/artifacts/core.sjs')
+
+xdmp.securityAssert("http://marklogic.com/data-hub/privileges/read-mapping", "execute");
+
+var stepName, limit;
+let mappingStep = core.getArtifact("mapping", stepName);
```
rjrudin, line 19: Same thing here, I think this needs to be more defensive. If sourceQuery doesn't exist, return an empty array. If the step doesn't exist, throw an error. If the sourceDatabase doesn't exist, don't do an eval.

rjrudin (changes requested):

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/mapping/getDocument.sjs
srinathgit: core.getArtifact throws HTTP 404 if step is not found

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/mapping/getUris.sjs
srinathgit: Yeah , will remove.

rjrudin (approved):

akshaysonvane (approved):

PR #4755, opened 2020-10-23
DHFPROD-6142:Add test program for streaming data into ML

anu3990:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


marklogic-data-hub-spark-connector/spark-test-project/src/main/java/test/SparkTest.java
```diff
@@ -52,7 +52,7 @@ private static void setConnectionProperties(String[] args) {
         SQLContext sqlContext = new SQLContext(javaSparkContext);
         final String filePath = getTestFilePath();
         logger.info("Loading from file: " + filePath);
-        Dataset<org.apache.spark.sql.Row> rows = sqlContext.load(filePath, "csv");
+        Dataset<org.apache.spark.sql.Row> rows = sqlContext.read().option("header", true).csv(getTestFilePath());
```
rjrudin, line 55: Nice - I was wondering how to do this!

marklogic-data-hub-spark-connector/spark-test-project/src/main/java/test/StreamTest.java
```diff
@@ -0,0 +1,104 @@
...
+
+    public static void main(String[] args) throws Exception {
+        setConnectionProperties(args);
+
+        logger.info("Creating JavaSparkContext");
+
+        SparkConf sparkConf = new SparkConf()
+            .setAppName("TestAppName")
+            .setMaster("local[1]")
+            .set("spark.driver.allowMultipleContexts", "true");
```
rjrudin, line 33: Do you know if this line is needed? I added it here when I was hacking around, but I don't know if it's needed. We want this to be as slim as possible, we want the bare minimum amount of code to enable streaming.

marklogic-data-hub-spark-connector/spark-test-project/src/main/java/test/StreamTest.java
```diff
@@ -0,0 +1,104 @@
...
+            rows.join(rows);
+            writeRowsToDataHub(rows);
+        } finally {
+            logger.info("Closing JavaSparkContext");
+            javaSparkContext.close();
+        }
+    }
+
+    private static void setConnectionProperties(String[] args) {
+        if (args.length != 3) {
```
rjrudin, line 49: Just a note, this duplication is totally fine right now. I'll submit a PR later to refactor it.

marklogic-data-hub-spark-connector/spark-test-project/src/main/java/test/StreamTest.java
```diff
@@ -0,0 +1,104 @@
...
+        }
+    }
+
+    private static Dataset<Row> loadRowsFromTestFile(JavaSparkContext javaSparkContext) {
+        SQLContext sqlContext = new SQLContext(javaSparkContext);
+        final String filePath = getTestFilePath();
+        logger.info("Loading from file: " + filePath);
+        StructType structType = sqlContext.read().option("header", true).csv(getTestFilePath()).schema();
+        logger.info("Schema of the input file is "+structType);
+        Dataset<org.apache.spark.sql.Row> rows = sqlContext.readStream().schema(structType).format("csv").option("header", true).csv("src/main/resources/data");
```
rjrudin, line 66: Nice - so the key is to read the header first and determine the schema from it, and then pass the schema in when calling readStream?

rjrudin (changes requested):
@anu3990 This is really good - thanks for figuring this out! Just wondering if that one line is required.

marklogic-data-hub-spark-connector/spark-test-project/src/main/java/test/StreamTest.java
anu3990: Not sure about that Rob. However i did notice that SQLContext(inside loadRowsFromTestFile) is deprecated, so we may want to shift towards SparkSession.

marklogic-data-hub-spark-connector/spark-test-project/src/main/java/test/StreamTest.java
anu3990: Yes i believe so.

marklogic-data-hub-spark-connector/spark-test-project/src/main/java/test/StreamTest.java
```diff
@@ -0,0 +1,104 @@
...
+import org.slf4j.LoggerFactory;
+
+import java.io.File;
+
+public class StreamTest {
+
+    private static Logger logger = LoggerFactory.getLogger(SparkTest.class);
+
+    private static String host = "localhost";
+    private static String username = "pari";
```
srinathgit, line 22: Is there any reason why  "test-data-hub-operator" isn't used here (like SparkTest) ?

marklogic-data-hub-spark-connector/spark-test-project/src/main/java/test/StreamTest.java
anu3990: Changed the code to use SparkSession.

marklogic-data-hub-spark-connector/spark-test-project/src/main/java/test/StreamTest.java
```diff
@@ -0,0 +1,96 @@
...
+            .option("mlHost", host)
+            .option("mlUsername", username)
+            .option("mlPassword", password)
+            .option("collections", "sparkTestOne,sparkTestTwo")
+            .option("hubDhs", "false")
+            .option("hubSsl", "false")
+            .option("batchSize", "3")
+            .option("checkpointLocation", "src/main/resources/logs")
+            .start()
+            .awaitTermination();
```
rjrudin, line 94: At least for this test program, I think we need to pass in e.g. 3000l here to force a timeout. Our connector doesn't otherwise know when to stop.

marklogic-data-hub-spark-connector/spark-test-project/src/main/java/test/StreamTest.java
```diff
@@ -0,0 +1,96 @@
...
+        logger.info("Creating SparkSession");
+
+        SparkSession sparkSession = SparkSession.builder()
+            .master("local")
+            .getOrCreate();
+
+
+        try {
+            Dataset<Row> rows = loadRowsFromTestFile(sparkSession);
+            rows.join(rows);
```
rjrudin, line 33: Can you remove this line from both test programs? I think I included it via copy/paste, but it doesn't seem necessary, and we want as few lines of code here as possible.

rjrudin (changes requested):
Can you also add the following to the log4j.properties file so we get info-logging for our test program:

log4j.logger.test=INFO, stdout
log4j.additivity.test=false

rjrudin:
Actually don't add the logging, I'm going to adjust that in a separate PR.

rjrudin:
@anu3990  After rebasing, can you update both this program and SparkTest to use SparkSession instead of JavaSparkContext? That'll get rid of the deprecation warning when SQLContext is constructed.

rjrudin (approved):

SameeraPriyathamTadikonda (approved):

PR #4757, opened 2020-10-23
DHFPROD-5664: Run Hub Central merge steps

ryanjdew:
### Description

Still working through new test cases, but current tests are passing with the new "compiled" merge options approach that allows normalizing the different formats and reducing the amount of redundant logic.

Also expanding the detail of merge trace events.

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mastering/default/merging.sjs
```diff
@@ -22,13 +22,24 @@ const processedURIs = [];
...
+  if (options.stepId) {
+    const stepDoc = fn.head(cts.search(cts.andQuery([
+      cts.collectionQuery("http://marklogic.com/data-hub/steps"),
+      cts.jsonPropertyValueQuery("stepId", options.stepId, "case-insensitive")
+    ])));
+    if (stepDoc) {
+      options = stepDoc.toObject();
+      isSeparateMatchStep = true;
+    } else {
+      fn.error(null, 'RESTAPI-SRVEXERR', Sequence.from([400, `Could not find step with stepId ${options.stepId}`]));
```
rjrudin, line 34: ds-utils.sjs can be used for this, though it needs a throwNotFound function for handling a 400.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mastering/default/merging.sjs
```diff
@@ -22,13 +22,24 @@ const processedURIs = [];
 
 function main(content, options) {
   // These index references can't be out this function scope or the jobReport will error, since they don't exist for the jobs DB
+  if (options.stepId) {
+    const stepDoc = fn.head(cts.search(cts.andQuery([
```
rjrudin, line 26: Is it known what type of step this is - i.e. is it known to be a merging step, since this is the merging module? If so, can the artifacts/core library be used here instead to find the step document?

marklogic-data-hub/src/main/resources/ml-modules/root/com.marklogic.smart-mastering/survivorship/merging/options.xqy
```diff
@@ -564,3 +583,300 @@ declare function merge-impl:propertyspec-to-json($property-spec as element()) as
...
+      if (fn:exists($name[. ne ""])) then
+        $name
+      else
+        "standard"
+    ),
+    $arity
+  )
+};
+
+declare variable $cached-merging-map as map:map := map:map();
```
rjrudin, line 797: I think a cache needs a comment describing why it helps. Based on the code below, my assumption would be that build-merging-map is called multiple times, and thus we're avoid a potentially expensive operation - the md5/describe call. But I only see build-merging-map called by compile-merge-options. Is that called enough times to warrant having a cache around?

marklogic-data-hub/src/main/resources/ml-modules/root/com.marklogic.smart-mastering/survivorship/merging/options.xqy
```diff
@@ -564,3 +583,300 @@ declare function merge-impl:propertyspec-to-json($property-spec as element()) as
...
+  )
+};
+
+(:
+ : Calculate merge information once per unique merge options in request to reduce repeat logic
+ : @param $merge-options  Options specifying how documents will be merged
+ : @param $only-warn-on-error  boolean indicating if errors should be returned rather than thrown
+ : @return map:map with compiled information about merge options
+ :)
+declare function merge-impl:compile-merge-options(
```
rjrudin, line 603: I'm wondering if this function can be broken up into several smaller functions. It's about 150 lines with many conditionals, and thus it's difficult to understand all the things that comprise compiling match options. Perhaps it's an exercise for later, I just find long methods like this to be difficult to read and understand. 

I know it can be difficult too to create lots of small functions in XQuery. My experience has been that in the absence of classes (which are very effective at carrying out related bits of data with self-documenting field names), we can use maps and arbitrary XML fragments to carry around related bits of data. So a typical problem I've run into with trying to decompose a large XQuery function into smaller functions is I've got a lot of state represented by variables, and I may have to pass in several arguments to a function in order for it to do what it needs to do. So I do find it more difficult to create lots of small functions with self-documenting names that do one thing and one thing only - at least it's more difficult in XQuery than in Java. Still, I think at some point, we need more of this in the mastering code to help more developers be able to read and understand it, and thus be able to maintain and enhance it.

rjrudin (commented):
I think it would be good to have docs somewhere - possibly code comments - explaining what "compiled merged options" means. For example, I understand what merge options are in a merge step - what is the result of compiling them, and what are the benefits we get?

rjrudin (approved):
I don't understand most of the changes here, but if the scope of the changes is just "Support the new format for match options", then I'm fine with the changes.

In the future, I think it'd be good to tackle large-scale changes like this via multiple JIRA tickets, with each ticket biting off a chunk of the work. For example, one ticket for strategy-based merge rules, and one ticket for custom merge rules, etc. And the changeset would be much smaller for each ticket, making it easier to understand (particularly when referring back to the changeset to look at what changed). I don't know if that would have been possible here though.

fsnow (dismissed):

fsnow (approved):

Sanjeevani19 (dismissed):

rjrudin (dismissed):

Sanjeevani19 (commented):
@ryanjdew https://jenkins.marklogic.com/blue/organizations/jenkins/Datahub_CI/detail/PR-4757/13/tests, looks like two core unit tests are failing, Please look into them

Sanjeevani19 (approved):

PR #4759, opened 2020-10-23
DHFPROD-6153: Loading default Initialize and Finalize endpoints

SameeraPriyathamTadikonda:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/sql/sources/v2/DefaultSource.java
+++ b/marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/sql/sources/v2/DefaultSource.java
@@ -19,8 +19,7 @@ import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.marklogic.client.DatabaseClient;
-import com.marklogic.client.document.DocumentWriteSet;
-import com.marklogic.client.document.TextDocumentManager;
+import com.marklogic.client.document.DocumentManager;
 import com.marklogic.client.ext.helper.LoggingObject;
 import com.marklogic.client.ext.util.DefaultDocumentPermissionsParser;
 import com.marklogic.client.io.DocumentMetadataHandle;
@@ -52,7 +51,6 @@ import java.util.Properties;
 import java.util.stream.Stream;
 
 public class DefaultSource extends LoggingObject implements WriteSupport, StreamWriteSupport {
-
     @Override
     public Optional<DataSourceWriter> createWriter(String writeUUID, StructType schema, SaveMode mode, DataSourceOptions options) {
         return Optional.of(new HubDataSourceWriter(options.asMap(), schema, false) {
@@ -76,6 +74,8 @@ public class DefaultSource extends LoggingObject implements WriteSupport, Stream
         hubClientConfig.applyProperties(props);
         return hubClientConfig;
     }
+
+
 }
 
 class HubDataSourceWriter extends LoggingObject implements StreamWriter {
@@ -169,14 +169,10 @@ class HubDataSourceWriter extends LoggingObject implements StreamWriter {
 
         if (doesNotHaveApiPath) {
             String apiPath = "/marklogic-data-hub-spark-connector/bulkIngester.api";
+            String scriptPath = "/marklogic-data-hub-spark-connector/bulkIngester.sjs";
+            loadModuleIfNotPresent(scriptPath, Format.TEXT);
+            loadModuleIfNotPresent(apiPath, Format.JSON);
 
-            if(hubClient.getModulesClient().newJSONDocumentManager().exists(apiPath) == null) {
-                try {
-                    loadEndpoints();
-                } catch (IOException e) {
-                    throw new RuntimeException("Unable to write default ingestion endpoint at path: " + apiPath + "; cause: " + e.getMessage(), e);
-                }
-            }
 
             endpointParams.put("apiPath", apiPath);
         }
@@ -204,16 +200,16 @@ class HubDataSourceWriter extends LoggingObject implements StreamWriter {
         });
     }
 
-    private void loadEndpoints() throws IOException {
-        String scriptPath = "marklogic-data-hub-spark-connector/bulkIngester.sjs";
-        String apiPath = "marklogic-data-hub-spark-connector/bulkIngester.api";
-
-        TextDocumentManager modMgr = hubClient.getModulesClient().newTextDocumentManager();
-        DocumentWriteSet writeSet = modMgr.newWriteSet();
-        DocumentMetadataHandle metadata = buildDocumentMetadata();
-        writeSet.add("/" + apiPath, metadata, new InputStreamHandle(new ClassPathResource(apiPath).getInputStream()).withFormat(Format.JSON));
-        writeSet.add("/" + scriptPath, metadata, new InputStreamHandle(new ClassPathResource(scriptPath).getInputStream()).withFormat(Format.TEXT));
-        modMgr.write(writeSet);
+    private void loadModuleIfNotPresent(String modulePath, Format format) {
+        if (!endpointExists(modulePath)) {
+            try {
+                DocumentManager modMgr = hubClient.getModulesClient().newDocumentManager();
+                DocumentMetadataHandle metadata = buildDocumentMetadata();
+                modMgr.write(modulePath, metadata, new InputStreamHandle(new ClassPathResource(modulePath).getInputStream()).withFormat(format));
+            } catch (IOException e) {
+                throw new RuntimeException("Unable to write endpoint at path: " + modulePath + "; cause: " + e.getMessage(), e);
+            }
+        }
     }
 
     private DocumentMetadataHandle buildDocumentMetadata() {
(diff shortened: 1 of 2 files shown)


marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/sql/sources/v2/DefaultSource.java
```diff
@@ -75,17 +64,28 @@ public static HubClientConfig buildHubClientConfig(Map<String, String> options)
         hubClientConfig.applyProperties(props);
         return hubClientConfig;
     }
+
+    @Override
```
rjrudin, line 68: Do these need to move as part of this PR? I'm planning on moving buildHubClientConfig out of here soon - the Reader will need it. In which case these two methods can stay put.

marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/sql/sources/v2/DefaultSource.java
```diff
@@ -216,6 +214,16 @@ private DocumentMetadataHandle buildDocumentMetadata() {
     }
 
     private String initializeJob(StructType schema) {
+        String apiPath = "/marklogic-data-hub-spark-connector/initializeJob.api";
+        String scriptPath = "/marklogic-data-hub-spark-connector/initializeJob.sjs";
+        try {
```
rjrudin, line 219: I think it's worth making another private method to avoid the duplication of the "if" check and the error handling - e.g. "loadEndpointsIfNotPresent".

marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/sql/sources/v2/DefaultSource.java
```diff
@@ -216,6 +214,16 @@ private DocumentMetadataHandle buildDocumentMetadata() {
...
+            if (endpointExists(apiPath)) {
+                loadEndpoint(scriptPath, apiPath);
+            }
+        } catch (IOException e) {
+            throw new IllegalArgumentException("Unable to load initializeJob Endpoint, cause: " + e.getMessage(), e);
```
rjrudin, line 224: Just throw a RuntimeException here - IllegalArgumentException is for when the user provided some argument that's not valid, which isn't the case here.

marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/sql/sources/v2/DefaultSource.java
```diff
@@ -155,10 +155,11 @@ private JsonNode determineIngestionEndpointParams(Map<String, String> options) {
 
         if (doesNotHaveApiPath) {
             String apiPath = "/marklogic-data-hub-spark-connector/bulkIngester.api";
+            String scriptPath = "/marklogic-data-hub-spark-connector/bulkIngester.sjs";
 
-            if(hubClient.getModulesClient().newJSONDocumentManager().exists(apiPath) == null) {
+            if (endpointExists(apiPath)) {
```
rjrudin, line 160: Need to check both the api and the sjs separately - it's conceivable that one exists but the other does not.

rjrudin (changes requested):

marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/sql/sources/v2/DefaultSource.java
SameeraPriyathamTadikonda: The changes are just formatting in this PR.

marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/sql/sources/v2/DefaultSource.java
SameeraPriyathamTadikonda: Done.

marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/sql/sources/v2/DefaultSource.java
SameeraPriyathamTadikonda: Updated the method to check and load.

marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/sql/sources/v2/DefaultSource.java
SameeraPriyathamTadikonda: Checking both the files.

marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/sql/sources/v2/DefaultSource.java
```diff
@@ -190,16 +185,19 @@ private void applyOptionsToEndpointConstants(ObjectNode endpointConstants, Map<S
...
-        String scriptPath = "marklogic-data-hub-spark-connector/bulkIngester.sjs";
-        String apiPath = "marklogic-data-hub-spark-connector/bulkIngester.api";
-
-        TextDocumentManager modMgr = hubClient.getModulesClient().newTextDocumentManager();
-        DocumentWriteSet writeSet = modMgr.newWriteSet();
-        DocumentMetadataHandle metadata = buildDocumentMetadata();
-        writeSet.add("/" + apiPath, metadata, new InputStreamHandle(new ClassPathResource(apiPath).getInputStream()).withFormat(Format.JSON));
-        writeSet.add("/" + scriptPath, metadata, new InputStreamHandle(new ClassPathResource(scriptPath).getInputStream()).withFormat(Format.TEXT));
-        modMgr.write(writeSet);
+    private void loadEndpointsIfNotPresent(String scriptPath, String apiPath) {
```
rjrudin, line 188: To be safe, we really need to do one document at a time. If someone how module exists but the other doesn't, this will throw an error. The efficiency of writing 2 at a time isn't important here; this is going to finish in a very small amount of time no matter what, and it's a one-time operation. 

So this method should just handle a single module. It just needs to know the path and the Format.

rjrudin (changes requested):

marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/sql/sources/v2/DefaultSource.java
SameeraPriyathamTadikonda: Done.

marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/sql/sources/v2/DefaultSource.java
```diff
@@ -204,16 +201,18 @@ private void applyOptionsToEndpointConstants(ObjectNode endpointConstants, Map<S
...
-        String scriptPath = "marklogic-data-hub-spark-connector/bulkIngester.sjs";
-        String apiPath = "marklogic-data-hub-spark-connector/bulkIngester.api";
-
-        TextDocumentManager modMgr = hubClient.getModulesClient().newTextDocumentManager();
-        DocumentWriteSet writeSet = modMgr.newWriteSet();
-        DocumentMetadataHandle metadata = buildDocumentMetadata();
-        writeSet.add("/" + apiPath, metadata, new InputStreamHandle(new ClassPathResource(apiPath).getInputStream()).withFormat(Format.JSON));
-        writeSet.add("/" + scriptPath, metadata, new InputStreamHandle(new ClassPathResource(scriptPath).getInputStream()).withFormat(Format.TEXT));
-        modMgr.write(writeSet);
+    private void loadEndpointsIfNotPresent(String scriptPath, Format format) {
```
rjrudin, line 204: Can you rename this to "loadModuleIfNotPresent" and rename "scriptPath" to "modulePath"?

marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/sql/sources/v2/DefaultSource.java
```diff
@@ -204,16 +201,18 @@ private void applyOptionsToEndpointConstants(ObjectNode endpointConstants, Map<S
...
+        if (!endpointExists(scriptPath)) {
+            try {
+                TextDocumentManager modMgr = hubClient.getModulesClient().newTextDocumentManager();
+                DocumentWriteSet writeSet = modMgr.newWriteSet();
```
rjrudin, line 208: No need for a DocumentWriteSet since we're only writing one doc. Also, do "newDocumentManager" so it doesn't look like we expect these to be text documents.

marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/sql/sources/v2/DefaultSource.java
```diff
@@ -246,13 +250,23 @@ private void addJobIdToEndpointConstants() {
...
             JsonNode endpointConstants = endpointParams.get("endpointConstants");
             if (endpointConstants instanceof ObjectNode) {
-                ((ObjectNode)endpointConstants).put("jobId", this.jobId);
+                ((ObjectNode) endpointConstants).put("jobId", this.jobId);
             }
         }
     }
 
     private void finalizeJob(String status) {
+        String apiPath = "/marklogic-data-hub-spark-connector/finalizeJob.api";
```
rjrudin, line 259: I think we need to initialize these modules before we create the job document. That ensures we fail fast if there's any problem - i.e. we don't want to load a bunch of data and then have an error if we can't write these modules for some reason.

rjrudin (changes requested):

rjrudin (approved):
Looks good!

anu3990 (approved):

PR #4764, opened 2020-10-26
DHFPROD-5858: Check ML version compatibility before running hubDeploy* tasks

akshaysonvane:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [ ] Added Tests
  

- ##### Reviewer:

- [ ] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/dhs/DhsDeployer.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/dhs/DhsDeployer.java
@@ -27,6 +27,8 @@ import com.marklogic.hub.deploy.commands.*;
 import com.marklogic.hub.dhs.installer.deploy.DeployHubAmpsCommand;
 import com.marklogic.hub.dhs.installer.deploy.DeployHubQueryRolesetsCommand;
 import com.marklogic.hub.impl.HubConfigImpl;
+import com.marklogic.hub.impl.VersionInfo;
+import com.marklogic.hub.impl.Versions;
 
 import java.io.File;
 import java.util.ArrayList;
@@ -53,6 +55,26 @@ public class DhsDeployer extends LoggingObject {
         HubAppDeployer dhsDeployer = new HubAppDeployer(hubConfig.getManageClient(), hubConfig.getAdminManager(), null, null);
         dhsDeployer.setCommands(buildCommandsForSecurityAdmin());
         dhsDeployer.deploy(hubConfig.getAppConfig());
+    }
+
+    /**
+     * Compares current ML version with the minimum required version and fails immediately
+     * if current ML version is lower than the minimum required version.
+     * @param hubConfig
+     */
+    public void checkCompatibility(HubConfig hubConfig) {
+        Versions versions = new Versions(hubConfig);
+        String mlVersionString = versions.getMarkLogicVersion();
+        Versions.MarkLogicVersion mlVersion = versions.getMLVersion(mlVersionString);
+        if (!((mlVersion.getMajor() > 10) || (mlVersion.getMajor() == 10 && mlVersion.getMinor() >= 201) || (mlVersion.getMajor() == 9 && mlVersion.getMinor() >= 1100))) {
+            logger.warn("Cannot proceed as this version of Data Hub does not support the detected version of MarkLogic:");
+            logger.warn("MarkLogic host: " + hubConfig.getHost());
+            logger.warn("MarkLogic version: " + mlVersionString);
+            logger.warn("Data Hub client version: " + VersionInfo.getBuildVersion());
+            logger.warn("Please see https://docs.marklogic.com/datahub/refs/version-compatibility.html for information on the minimum required MarkLogic version.");
+
+            throw new RuntimeException("Cannot deploy due to version mismatch.");
+        }
     }
 
     /**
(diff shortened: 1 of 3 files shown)


marklogic-data-hub/src/main/java/com/marklogic/hub/dhs/DhsDeployer.java
```diff
@@ -55,6 +57,26 @@ public void deployAsSecurityAdmin(HubConfigImpl hubConfig) {
...
+    /**
+     * Compares current ML version with the minimum required version and fails immediately
+     * if current ML version is lower than the minimum required version.
+     * @param hubConfig
+     */
+    public void checkCompatibility(HubConfig hubConfig) {
+        Versions versions = new Versions(hubConfig);
+        String mlVersionString = versions.getMarkLogicVersion();
+        Versions.MarkLogicVersion mlVersion = versions.getMLVersion(mlVersionString);
+        if (!((mlVersion.getMajor() > 10) || (mlVersion.getMajor() == 10 && mlVersion.getMinor() >= 201) || (mlVersion.getMajor() == 9 && mlVersion.getMinor() >= 1100))) {
```
rjrudin, line 69: We have this logic in AbstractInstallerCommand too, and it's different - because of the dependency on granular privileges, 10.0-3 or higher is needed. 

I think this question - "Given an ML version, can it support this version of DHF on DHS?" - and the question - "Given an ML version, can it support this version of DHF?" - should both be answered by the Versions.MarkLogicVersion class, which should become a real class instead of an embedded once. The problem is that the current MLV design encourages this logic to be duplicated. We should instead have methods on MLV itself that answer those questions - e.g.:

```
String mlVersion = ... // get from Manage API, or wherever, we don't care
return new MarkLogicVersion(mlVersion).supportsDataHubService();
// or .supportsDataHubFramework()
```

That class can then include all the stuff in Versions.getMLVersion. Arguably, it could go into the API project, but let's not do that yet. 

Making these changes allows you to easily test the MarkLogicVersion class too, as it's now decoupled from actually getting the ML version. So you can have a unit test that passes in all kinds of ML version strings, and MarkLogicVersion will parse it into major/minor/nightly/dateString (rename "isNightly" to "nightly" too, the Java getter will then naturally be isNightly). 

This refactoring will affect a few clients, but I think it's definitely worth doing - it gives us a much more OO design that is easier to test.

marklogic-data-hub/src/main/java/com/marklogic/hub/dhs/DhsDeployer.java
```diff
@@ -55,6 +57,26 @@ public void deployAsSecurityAdmin(HubConfigImpl hubConfig) {
         dhsDeployer.deploy(hubConfig.getAppConfig());
     }
 
+    /**
+     * Compares current ML version with the minimum required version and fails immediately
+     * if current ML version is lower than the minimum required version.
+     * @param hubConfig
+     */
+    public void checkCompatibility(HubConfig hubConfig) {
```
rjrudin, line 65: I think this should be built into the deploy methods and not be a separate public method that clients have to call (which means you don't need to change the Gradle tasks). 

Let's give it a more descriptive name too - "check" methods are often vague about what they actually do or what the output will be. Our purpose here is to throw an exception if the ML version is invalid. So let's call the method that - "throwExceptionIsMarkLogicVersionIsInvalid". That leaves no doubt as to what the method will do.

marklogic-data-hub/src/main/java/com/marklogic/hub/dhs/DhsDeployer.java
```diff
@@ -55,6 +57,26 @@ public void deployAsSecurityAdmin(HubConfigImpl hubConfig) {
...
+        Versions versions = new Versions(hubConfig);
+        String mlVersionString = versions.getMarkLogicVersion();
+        Versions.MarkLogicVersion mlVersion = versions.getMLVersion(mlVersionString);
+        if (!((mlVersion.getMajor() > 10) || (mlVersion.getMajor() == 10 && mlVersion.getMinor() >= 201) || (mlVersion.getMajor() == 9 && mlVersion.getMinor() >= 1100))) {
+            logger.warn("Cannot proceed as this version of Data Hub does not support the detected version of MarkLogic:");
```
rjrudin, line 70: Per the JIRA ticket, let's make this the actual error message. The problem with logging is that the user will see the error message below and may see a stacktrace as well, but this logging will be above all of that. So the user may miss it. Let's make it obvious and build up a string as described in the JIRA ticket, and then use that as the exception message.

rjrudin (changes requested):

marklogic-data-hub/src/main/java/com/marklogic/hub/dhs/DhsDeployer.java
akshaysonvane: Makes sense. I was thinking along the same lines so opened this PR as a draft.

marklogic-data-hub/src/main/java/com/marklogic/hub/dhs/DhsDeployer.java
akshaysonvane: I did not add it to deploy methods because these methods are being used by certain tests that might not have the same version requirements. For instance, the version check condition does not account for the nightlies.

akshaysonvane:
I'll refactor the MLV class, closing this PR for now.

PR #4769, opened 2020-10-26
DHFPROD-6117:Handle duplicate custom mapping functions

srinathgit:
### Description
Handle duplicate custom mapping functions. This failure https://jenkins.marklogic.com/blue/organizations/jenkins/Datahub_CI/detail/PR-4759/10/tests was because there were 2 custom mapping functions of name 'echo'.

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mapping/entity-services/lib.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mapping/entity-services/lib.sjs
@@ -379,7 +379,7 @@ function runMapping(mapping, uri, propMapping={"targetEntityType":mapping.target
         paths.pop();
       }
       else {
-         propMapping = addNode(propMapping, paths, mappedProperty,  false);
+        propMapping = addNode(propMapping, paths, mappedProperty,  false);
       }
     }
     if(mappedProperty && !mappedProperty.errorMessage && ! mappedProperty.hasOwnProperty("targetEntityType") && sourcedFrom.length > 0){
@@ -476,6 +476,7 @@ function getMarkLogicMappingFunctions() {
   return fn.head(datahub.hubUtils.queryLatest(function() {
     let fnMetadata = fn.collection("http://marklogic.com/entity-services/function-metadata")
     let ns = {"m":"http://marklogic.com/entity-services/mapping"};
+    const functionMap = new Map();
     let output = [];
 
     for (const metaData of fnMetadata){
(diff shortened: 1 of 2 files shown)


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mapping/entity-services/lib.sjs
```diff
@@ -476,21 +476,31 @@ function getMarkLogicMappingFunctions() {
...
-          let singleFunction ={};
-          singleFunction["functionName"] = funcName;
-          singleFunction["signature"] = funcName +"("+params+")";
-          singleFunction["category"] = (String(fnLocation).includes("/data-hub/5/mapping-functions")) ? "builtin" : "custom";
-          output.push(singleFunction)
+          if(functionsAdded.includes(funcName)){
+            let addedFunction = output.find(func => {
+              return func.functionName == funcName
+            });
+            addedFunction["signature"] = funcName +"("+params+")";
```
rjrudin, line 494: If there's a test failing for this, can the test be upgraded to verify that there's only one instance of the "echo" function? And this seems to be introducing logic that says - if there are duplicates, then whichever one is found last "wins" - i.e. the signature is for the last one. Is that the case?

srinathgit: The functionality is going to remain the same. Earlier the response was an object, so the last one always wins. 
```
{"echo":
"category":"custom"
"signature":"echo(value1, value2)"}
```
Now since it is an array, it can look like the below array, so I will have to include the logic.
```
[
{
"functionName": "echo,
"category":"custom",
"signature": "echo(value)"
},
{
"functionName": "echo,
"category":"custom",
"signature": "echo(value1, value2)"
}
]
```

 I will add a test that checks that echo (if present as the function was addd by previous tests) is present only once.

rjrudin (dismissed):

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mapping/entity-services/lib.sjs
```diff
@@ -476,21 +476,31 @@ function getMarkLogicMappingFunctions() {
...
           j++;
-          let singleFunction ={};
-          singleFunction["functionName"] = funcName;
-          singleFunction["signature"] = funcName +"("+params+")";
-          singleFunction["category"] = (String(fnLocation).includes("/data-hub/5/mapping-functions")) ? "builtin" : "custom";
-          output.push(singleFunction)
+          if(functionsAdded.includes(funcName)){
+            let addedFunction = output.find(func => {
+              return func.functionName == funcName
+            });
```
akshaysonvane, line 493: Instead of searching in both `functionsAdded` and `output` arrays, can we use a Map here with the key being the function name and the value being the function object?
Map also provides an iterator for values that can then be used to create the output array.
```
const map1 = new Map();
map1.set('a', {'a':'a'});
map1.set('b', {'b':'b'});
const ar = [...map1.values()];
```

rjrudin: That's a good call, that will be a lot faster than calling "includes" N times and then having to call "find" as well.

srinathgit: The spread operator doesn't work in ML 9. So, iterated through the values.

ryanjdew (approved):

rjrudin (approved):

akshaysonvane (approved):

PR #4773, opened 2020-10-27
DHFPROD-6153: Adding the buildJobPermissions to initializeEndpoint

SameeraPriyathamTadikonda:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/marklogic-data-hub-spark-connector/initializeJob.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/marklogic-data-hub-spark-connector/initializeJob.sjs
@@ -22,10 +22,21 @@ declareUpdate();
 var sparkMetadata = sparkMetadata != null ? fn.head(xdmp.fromJSON(sparkMetadata)) : {};
 
 const config = require("/com.marklogic.hub/config.sjs");
-const Jobs = require("/data-hub/5/impl/jobs.sjs");
-const jobs = new Jobs.Jobs(config);
 
 const id = sem.uuidString();
+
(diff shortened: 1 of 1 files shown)


marklogic-data-hub/src/main/resources/ml-modules/root/marklogic-data-hub-spark-connector/initializeJob.sjs
```diff
@@ -27,6 +27,18 @@ const jobs = new Jobs.Jobs(config);
 
 const id = sem.uuidString();
 
+function buildJobPermissions(config) {
```
rjrudin, line 30: This needs a comment as to why it's here too - otherwise a developer is going to see it and think - why is this duplicated here!?!? So explain that we've duplicated it here because we need this to work on 5.2.x, but the function is different between 5.2 and > 5.2.

rjrudin: Also, I think you can get rid of the "jobs" library dependency as a result?

rjrudin (changes requested):

marklogic-data-hub/src/main/resources/ml-modules/root/marklogic-data-hub-spark-connector/initializeJob.sjs
SameeraPriyathamTadikonda: Done.

rjrudin (approved):

anu3990 (approved):

ryanjdew (approved):

PR #4774, opened 2020-10-27
DHFPROD-6164: Null JSON node throws error in double metaphone match

fsnow:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/smart-mastering/matching/null-node-match.xqy
+++ b/marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/smart-mastering/matching/null-node-match.xqy
@@ -1,0 +1,19 @@
+xquery version "1.0-ml";
+
+import module namespace matcher = "http://marklogic.com/smart-mastering/matcher"
+  at "/com.marklogic.smart-mastering/matcher.xqy";
+import module namespace constants = "http://marklogic.com/smart-mastering/constants"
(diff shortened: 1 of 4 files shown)


ryanjdew (approved):

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/smart-mastering/matching/null-node-match.xqy
```diff
@@ -0,0 +1,19 @@
...
+  at "/com.marklogic.smart-mastering/constants.xqy";
+import module namespace lib = "http://marklogic.com/smart-mastering/test" at "lib/lib.xqy";
+
+import module namespace test = "http://marklogic.com/test" at "/test/test-helper.xqy";
+
+declare option xdmp:mapping "false";
+
+let $doc := fn:doc($lib:URI9)
+let $actual := matcher:find-document-matches-by-options-name($doc, $lib:MATCH-OPTIONS-NAME, fn:true(), cts:true-query())
+return
```
rjrudin, line 15: One style note - I think comments like this often are better expressed as assertion messages, since if the assertion does fail, the test report shows the very helpful message. No need to change, just a note.

rjrudin (approved):

PR #4778, opened 2020-10-27
DHFPROD-6163:Add the AWS Glue job info in the job document.

anu3990:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/dataservices/SparkService.java
+++ b/marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/dataservices/SparkService.java
@@ -49,34 +49,20 @@ public interface SparkService {
             private DatabaseClient dbClient;
             private BaseProxy baseProxy;
 
-            private BaseProxy.DBFunctionRequest req_finalizeJob;
             private BaseProxy.DBFunctionRequest req_bulkIngest;
+            private BaseProxy.DBFunctionRequest req_finalizeJob;
             private BaseProxy.DBFunctionRequest req_initializeJob;
 
             private SparkServiceImpl(DatabaseClient dbClient, JSONWriteHandle servDecl) {
                 this.dbClient  = dbClient;
                 this.baseProxy = new BaseProxy("/marklogic-data-hub-spark-connector/", servDecl);
 
-                this.req_finalizeJob = this.baseProxy.request(
-                    "finalizeJob.sjs", BaseProxy.ParameterValuesKind.MULTIPLE_ATOMICS);
                 this.req_bulkIngest = this.baseProxy.request(
                     "bulkIngester.sjs", BaseProxy.ParameterValuesKind.MULTIPLE_NODES);
+                this.req_finalizeJob = this.baseProxy.request(
+                    "finalizeJob.sjs", BaseProxy.ParameterValuesKind.MULTIPLE_ATOMICS);
                 this.req_initializeJob = this.baseProxy.request(
                     "initializeJob.sjs", BaseProxy.ParameterValuesKind.SINGLE_NODE);
-            }
-
-            @Override
-            public void finalizeJob(String jobId, String status) {
-                finalizeJob(
-                    this.req_finalizeJob.on(this.dbClient), jobId, status
-                    );
-            }
-            private void finalizeJob(BaseProxy.DBFunctionRequest request, String jobId, String status) {
-              request
-                      .withParams(
-                          BaseProxy.atomicParam("jobId", false, BaseProxy.StringType.fromString(jobId)),
-                          BaseProxy.atomicParam("status", false, BaseProxy.StringType.fromString(status))
-                          ).responseNone();
             }
 
             @Override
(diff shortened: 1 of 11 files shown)


marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/sql/sources/v2/DefaultSource.java
```diff
@@ -237,14 +237,17 @@ private DocumentMetadataHandle buildDocumentMetadata() {
     private String initializeJob(StructType schema) {
         loadJobEndpoints();
 
-        ObjectNode sparkMetadata = objectMapper.createObjectNode();
+        ObjectNode externalMetadata = objectMapper.createObjectNode();
         try {
-            sparkMetadata.set("schema", objectMapper.readTree(schema.json()));
+            externalMetadata.set("additionalExternalMetadata",
```
SameeraPriyathamTadikonda, line 242: why do we need to set this to null explicitly if user does not configure?

SameeraPriyathamTadikonda:
You need to update the Jobs.Schema.json file.

SameeraPriyathamTadikonda (changes requested):
Meant to press request changes

marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/dataservices/SparkService.java
```diff
@@ -103,7 +103,7 @@ private String initializeJob(BaseProxy.DBFunctionRequest request, com.fasterxml.
               return BaseProxy.StringType.toString(
                 request
                       .withParams(
-                          BaseProxy.documentParam("sparkMetadata", true, BaseProxy.JsonDocumentType.fromJsonNode(sparkMetadata))
```
SameeraPriyathamTadikonda, line 106: you might need to regenerate the service class again.

marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/sql/sources/v2/DefaultSource.java
rjrudin: I'll edit the JIRA ticket with my proposal - I don't think we should force an "additionalExternalMetadata" key to exist. Instead, each key from the additionalexternalmetadata object should be added to the externalMetadata object.

marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/sql/sources/v2/writer/HubDataWriter.java
```diff
@@ -80,6 +82,8 @@ public WriterCommitMessage commit() {
         bulkInputCaller.awaitCompletion();
         if (writeException != null) {
             logger.info("At least one write failed");
+            logger.info("Additional External Metadata - "+ options.get("additionalexternalmetadata"));
```
rjrudin, line 85: The story isn't calling for this, so I don't think we should add more logging here. I don't think this logging is solving any problem either. Ernie already knows what the Run ID of the Glue job is; he can use that to find the DHF job document.

marklogic-data-hub/src/main/resources/ml-modules/root/marklogic-data-hub-spark-connector/initializeJob.api
```diff
@@ -2,7 +2,7 @@
   "functionName" : "initializeJob",
   "desc" : "Creates a job document with the given sparkMetadata as a property",
   "params" : [ {
-    "name" : "sparkMetadata",
+    "name" : "externalMetadata",
```
rjrudin, line 5: Let's add a "desc" to this to make it clear what the purpose of this is. How about: "JSON object that, if not null, will be added to the job document with a key of 'externalMetadata'".

marklogic-data-hub/src/main/resources/ml-modules/root/marklogic-data-hub-spark-connector/initializeJob.api
```diff
@@ -2,7 +2,7 @@
   "functionName" : "initializeJob",
   "desc" : "Creates a job document with the given sparkMetadata as a property",
```
rjrudin, line 3: Need to rename to externalMetadata

rjrudin (changes requested):
This needs a new test method where the user configures additionalexternalmetadata, which should become a new property of our test Options class. The test should add a couple keys to that option and then verify that both of those keys are present under externalMetadata in the job document.

marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/sql/sources/v2/writer/HubDataWriter.java
anu3990: But if the option "additionalexternalmetadata" is not there in the script, there is no way Ernie will  know the Glue Run id. In that case we might want to add a logger saying what and where he needs to add/change his Glue script.

rjrudin: I'm not following - I can see logging the DHF job ID here, because if Ernie doesn't provide the Glue Run ID in the options, then the DHF job ID would be useful in the Glue logs as a way of finding the associated DHF job document - though it won't be that helpful, since it won't have any additional info in it. 

But if Ernie doesn't provide additionalexternalmetadata - well that's his choice. Our documentation can make it clear that that's the way he can add additional data to the job document. Also, we wouldn't Glue-specific logging in a Spark-generic connector.

rjrudin (dismissed):

SameeraPriyathamTadikonda:
> You need to update the Jobs.Schema.json file.

This is not addressed.

SameeraPriyathamTadikonda (changes requested):
Need to update jobs schema with externalMetaData

specs/models/Job.schema.json
```diff
@@ -35,7 +35,7 @@
       "type": "string",
       "description": "dateTime at which the job ended"
     },
-    "sparkMetadata": {
+    "externalMetadata": {
```
anu3990, line 38: @SameeraPriyathamTadikonda it is here.

SameeraPriyathamTadikonda: sorry, missed it somehow.

SameeraPriyathamTadikonda (approved):

Sanjeevani19 (commented):
@anu3990 There are few rebase conflicts, Please resolve them

anu3990:
@rjrudin and @SameeraPriyathamTadikonda i have rebased and resolved the conflicts. Need your approvals again.

SameeraPriyathamTadikonda (dismissed):

rjrudin (dismissed):

rjrudin (approved):

SameeraPriyathamTadikonda (approved):

rjrudin (approved):

PR #4779, opened 2020-10-28
DHFPROD-5997: Can now exclude items already processed by a step

rjrudin:
### Description

This introduces a new step option named "excludeAlreadyProcessed" that depends on some enhancements to Batch documents and a new range index in the jobs database. 

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/flow.sjs
```diff
@@ -139,6 +139,12 @@ class Flow {
     let uris = null;
     if (options.uris) {
       uris = this.datahub.hubUtils.normalizeToArray(options.uris);
+
+      if (true === options.excludeAlreadyProcessed || "true" === options.excludeAlreadyProcessed) {
```
ryanjdew, line 143: Another option here is `options.excludeAlreadyProcessed && xs.boolean(options.excludeAlreadyProcessed)`

rjrudin: Thanks, I'll make that change. I initially only had the boolean true value there, but felt like supporting "true" was worthwhile in case a user accidentally puts in "true" in their options - seems expected that that would still work.

rjrudin: Made the change, and also added a new trace event so that the impact of this feature can be logged.

ryanjdew (approved):

akshaysonvane (approved):

srinathgit (approved):

PR #4787, opened 2020-10-29
DHFPROD-5948: Add sourceName and sourceType parameters to OOTB load step

rahulvudutala:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/src/components/search-result/search-result.tsx
+++ b/marklogic-data-hub-central/ui/src/components/search-result/search-result.tsx
@@ -50,18 +50,9 @@ const SearchResult: React.FC<Props> = (props) => {
     }
 
     if ((props.item.format === 'json' || props.item.format === 'xml') && props.item.hasOwnProperty('sources')) {
-        if (!props.item.sources.length) {
-            // content data does not exist in payload
-            setAlertMessage('Error', 'No instance information in payload');
-        } else {
-            if (Array.isArray(props.item.sources)) {
-                sourcesVal = props.item.sources.map(src => {
-                    return src.name;
-                }).join(', ');
-            } else {
-                sourcesVal = props.item.sources.name;
-            }
-        }
+        sourcesVal = props.item.sources.map(src => {
+            return src.name;
+        }).join(', ');
     }
 
     function getSnippet() {
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/flow-utils.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/flow-utils.sjs
@@ -546,6 +546,23 @@ class FlowUtils {
     if(options.file) {
         headers["createdUsingFile"] = options.file;
     }
+
+    const sourceName = options.sourceName ? options.sourceName : null;
+    const sourceType = options.sourceType ? options.sourceType : null;
+
+    if(headers["sources"] && !Array.isArray(headers["sources"])) {
+      headers.sources = [headers.sources];
+    }
+
+    if(sourceName || sourceType) {
+      if(!headers["sources"] || headers["sources"].length == 0) {
+        headers["sources"] = [];
+      }
+      headers["sources"].push({
+        "name": sourceName === null ? undefined : sourceName,
+        "datahubSourceType": sourceType === null ? undefined : sourceType
+      });
+    }
     return headers;
   }
 
--- a/marklogic-data-hub/src/test/java/com/marklogic/hub/step/impl/IngestWithSourceNameAndSourceTypeTest.java
+++ b/marklogic-data-hub/src/test/java/com/marklogic/hub/step/impl/IngestWithSourceNameAndSourceTypeTest.java
@@ -1,0 +1,88 @@
+package com.marklogic.hub.step.impl;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import com.marklogic.hub.AbstractHubCoreTest;
+import com.marklogic.hub.dataservices.FlowService;
+import com.marklogic.hub.dataservices.StepService;
+import com.marklogic.hub.flow.FlowInputs;
+import com.marklogic.hub.flow.FlowRunner;
+import com.marklogic.hub.flow.RunFlowResponse;
+import com.marklogic.hub.flow.impl.FlowRunnerImpl;
+import com.marklogic.hub.job.JobStatus;
+import com.marklogic.hub.test.ReferenceModelProject;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+public class IngestWithSourceNameAndSourceTypeTest extends AbstractHubCoreTest {
+
+    ReferenceModelProject project;
+
+    @BeforeEach
+    void beforeEach() {
+        project = installReferenceModelProject();
+    }
+
+    @Test
+    void runDefaultIngestionWithSourceNameAndSourceType(){
+        String path = "entity-reference-model/data/customers";
+        ObjectNode info = new ObjectMapper().createObjectNode();
+        info.put("name", "sources-test");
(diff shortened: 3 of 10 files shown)


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/flow-utils.sjs
```diff
@@ -546,6 +546,22 @@ class FlowUtils {
...
+      let tempSourcesArray = [];
+      tempSourcesArray.push(headers["sources"]);
+      headers["sources"] = tempSourcesArray;
+    }
+
+    if(sourceName || sourceType) {
+      if(!headers["sources"] || headers["sources"].length == 0) {
+        headers["sources"] = [];
+      }
+      headers["sources"].push({"name": sourceName, "datahubSourceType": sourceType});
```
rjrudin, line 563: Let's only populate something if it's not null - this will result in "datahubSourceType": "" if it's not specified. So above, default sourceName and sourceType to null if they're not in options. And then only do any of this code if either sourceName or sourceType is not null. And only add them to the source object if they're not null.

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/data-services/step/convertIngestionStep.sjs
```diff
@@ -75,7 +75,7 @@ assertions.push(
   test.assertEqual("", flowStep.fileLocations.outputURIPrefix),
   test.assertEqual(",", flowStep.fileLocations.separator),
 
-  test.assertEqual("myIngester", options.headers.sources[0].name),
+  test.assertEqual("", options.headers.sources[0].name),
```
rjrudin, line 78: This should instead assert that "sources" doesn't even exist.

rjrudin (changes requested):

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/artifacts/core.sjs
```diff
@@ -162,6 +162,13 @@ function setArtifact(artifactType, artifactName, artifact) {
     } catch (e) {}
     if (fn.empty(existingArtifact) && artifactLibrary.defaultArtifact) {
         artifact = Object.assign({}, artifactLibrary.defaultArtifact(artifactName, artifact.targetEntityType), artifact);
+        if(artifact.sourceName || artifact.sourceType) {
```
rjrudin, line 165: I don't think this is needed. To find out,  there should be a JUnit test for this story that creates an ingestion step with sourceName and sourceType populated, and then the test runs the ingestion step to ensure that each ingested document (only need to ingest one, really) ends up with headers/source set correctly. 

To make that test pass - I don't think you need anything here. Instead, you just need to confirm that when the createHeaders is called by the default ingestion step module, sourceName and sourceType are both present in the options object (which I think will happen by default, I don't think you'll need to do anything for that). And then createHeaders will do all the work.

rjrudin (changes requested):

marklogic-data-hub-central/ui/src/components/search-result/search-result.tsx
```diff
@@ -50,17 +50,12 @@ const SearchResult: React.FC<Props> = (props) => {
     }
 
     if ((props.item.format === 'json' || props.item.format === 'xml') && props.item.hasOwnProperty('sources')) {
-        if (!props.item.sources.length) {
-            // content data does not exist in payload
-            setAlertMessage('Error', 'No instance information in payload');
+        if (Array.isArray(props.item.sources)) {
```
rjrudin, line 53: Do we know of examples where sources is an object and not an array? Pinging @srinathgit  and @ryanjdew  - that would violate the schema, and hopefully that's not something we've created. I think it's fine to have this code as a sanity check in case sources is an object somehow, but I want to make sure that we're not making sources an object somehow.

marklogic-data-hub/src/test/java/com/marklogic/hub/step/impl/IngestWithSourceNameAndSourceTypeTest.java
```diff
@@ -0,0 +1,68 @@
...
+    void runDefaultIngestionWithSourceNameAndSourceType(){
+        ServerTransform serverTransform = new ServerTransform("mlRunIngest");
+        Map<String, Object> options = new HashMap<>();
+        options.put("sourceName", "user-defined-source-name");
+        options.put("sourceType", "user-defined-source-type");
+        String optionString = jsonToString(options);
+        serverTransform.addParameter("options", optionString);
+
+        DatabaseClient stagingClient = getHubClient().getStagingClient();
+        WriteBatcher writeBatcher = stagingClient.newDataMovementManager().newWriteBatcher()
```
rjrudin, line 33: We still want to run the ingestion step here via FlowRunner. The reason that @srinathgit  uses WriteBatcher directly in RunDefaultIngestionTest is so that he can test the mlRunIngest transform in isolation, without running a step. 

But for this test, we need to run a step. You can use the all-steps-referenced test project - it has some test data in it along with an ingestion step. See AllStepsReferencedTest as an example.

rjrudin (changes requested):

marklogic-data-hub-central/ui/src/components/search-result/search-result.tsx
rahulvudutala: Yes. In createHeaders, even if it is an object, it is converted to an array.
In entity-search-lib.sjs as well, sources array is returned. The check can be removed.

marklogic-data-hub/src/test/java/com/marklogic/hub/step/impl/IngestWithSourceNameAndSourceTypeTest.java
```diff
@@ -0,0 +1,89 @@
...
+        // Add it to the existing echoFlow, which is assumed to have one step
+        FlowService.on(getHubClient().getStagingClient()).addStepToFlow("echoFlow", "sources-test", "ingestion");
+
+        // And run the flow
+        FlowInputs inputs = new FlowInputs("echoFlow", "2");
+        inputs.setInputFilePath(getClass().getClassLoader().getResource(path).getPath());
+
+        FlowRunner flowRunner = new FlowRunnerImpl(getHubClient());
+        RunFlowResponse response = flowRunner.runFlow(inputs);
+        flowRunner.awaitCompletion();
```
rjrudin, line 53: Use runFlow(FlowInputs) instead - that will call makeInputFilePathsAbsoluteInFlow for you, which means you don't need to call inputs.setInputFilePath above either.

marklogic-data-hub/src/test/java/com/marklogic/hub/step/impl/IngestWithSourceNameAndSourceTypeTest.java
```diff
@@ -0,0 +1,89 @@
...
+        assertEquals(JobStatus.FINISHED.toString(), response.getJobStatus());
+        JsonNode sourceDoc = getHubClient().getStagingClient().newJSONDocumentManager().read("customer1.json").next().getContent(new JacksonHandle()).get();
```
rjrudin, line 55: Can use getStagingDoc instead - less lines of code is almost always good in a test, as it makes it easier to read and understand.

rjrudin (changes requested):
Looks good, just a couple tweaks to shorten up the test.

marklogic-data-hub/src/test/java/com/marklogic/hub/step/impl/IngestWithSourceNameAndSourceTypeTest.java
rahulvudutala: makeInputFilePathsAbsoluteInFlow in runFlow method is reading the flow from the project directory. In the test, a new step is added which is saved in the database but not on the directory. So the step that the test is trying to add is overridden to what is in the project by makeInputFilePathsAbsoluteInFlow.

rjrudin (approved):

ryanjdew (approved):

PR #4804, opened 2020-11-02
DHFPROD-6098: Create endpoint for fetching source document for mapping UI

brent-hartwig:
### Description
Offer the `/api/steps/mapping/{stepName}/newDoc` endpoint where the intent is to enable the mapping UI to retrieve a source document associated to a step.

Supports XML and JSON formatted source documents but only ever returns a JSON representation.  The implementation relies on fast-xml-parser to convert XML to JSON.  Version v3.17.1 thereof was introduced to the backend.  The frontend is already set up to use version ^3.16.0.

Whether in this PR or a subsequent one, this endpoint is *replace* `/api/steps/mapping/{stepName}/doc`.  When the PR was originally submitted, both endpoints were present.

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [ ] Added Tests
  

- ##### Reviewer:

- [ ] Reviewed Tests


marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/controllers/steps/MappingStepController.java
```diff
@@ -83,6 +84,20 @@
         return new ResponseEntity<>(body, headers, HttpStatus.OK);
     }
 
+    @RequestMapping(value = "/{stepName}/newDoc", method = RequestMethod.GET)
+    @ApiOperation(value = "Returns an XML or JSON source document as a string of JSON")
+    @Secured("ROLE_readMapping")
+    public ResponseEntity<JsonNode> getNewDocument(@PathVariable String stepName, @PathVariable String docUri) {
+        // Step defines which database to pull the identified doc from.
+        final JsonNode stepNode = newService().getStep(STEP_DEFINITION_TYPE, stepName);
+        final String databaseName = stepNode.get("sourceDatabase").asText();
```
brent-hartwig, line 93: Is it okay to hard-code "sourceDatabase" (part of the step schema)?  I didn't find it used anywhere else within marklogic-data-hub-central.

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/controllers/steps/MappingStepController.java
```diff
@@ -83,6 +84,20 @@
...
+        // In the hub context, we're presuming staging or final.
+        final DatabaseClient databaseClient = databaseName.toLowerCase().equals(getHubClient().getDbName(DatabaseKind.STAGING).toLowerCase()) ?
+            getHubClient().getStagingClient() :
+            getHubClient().getFinalClient();
+        return ResponseEntity.ok(MappingService.on(databaseClient).getNewDocument(stepName, docUri));
```
brent-hartwig, line 98: The other endpoints herein use `StepService`.  Is it okay that this method uses `MappingService`?

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/controllers/steps/MappingStepController.java
```diff
@@ -83,6 +84,20 @@
         return new ResponseEntity<>(body, headers, HttpStatus.OK);
     }
 
+    @RequestMapping(value = "/{stepName}/newDoc", method = RequestMethod.GET)
+    @ApiOperation(value = "Returns an XML or JSON source document as a string of JSON")
+    @Secured("ROLE_readMapping")
+    public ResponseEntity<JsonNode> getNewDocument(@PathVariable String stepName, @PathVariable String docUri) {
+        // Step defines which database to pull the identified doc from.
+        final JsonNode stepNode = newService().getStep(STEP_DEFINITION_TYPE, stepName);
```
brent-hartwig, line 92: Should we determine the database from the step name here or on the backend, within `getNewDocument.sjs`?  The latter has the advantage of not presuming staging or final ...it's whatever the sourceDatabase property is.  From `getDocument.sjs`: `{database: xdmp.database(mappingStep.sourceDatabase)}`

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/mapping/xmlToJsonForMapping.sjs
```diff
@@ -0,0 +1,311 @@
+/**
```
brent-hartwig, line 1: Is this a good directory for this file?

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/data-services/mapping/getNewDocumentTest.sjs
```diff
@@ -0,0 +1,183 @@
+const test = require("/test/test-helper.xqy");
+const DataHubSingleton = require('/data-hub/5/datahub-singleton.sjs');
+const dataHub = DataHubSingleton.instance();
+const hubTest = require("/test/data-hub-test-helper.sjs");
+
+const jsonUri = '/test/sampleCustomerDoc.json';
```
brent-hartwig, line 6: Should all test data be moved out of this file?

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/controllers/steps/MappingStepController.java
brent-hartwig: No longer relevant given preceding answer from Ryan.

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/controllers/steps/MappingStepController.java
rjrudin: Yeah definitely have the backend figure it out.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/mapping/getNewDocument.api
```diff
@@ -0,0 +1,18 @@
+{
+    "functionName": "getNewDocument",
```
rjrudin, line 2: This endpoint is returning a JSON object with 3 things - the actual document, an optional namespaces object, and a format string. I think we need a noun other than "document" here, as "document" implies you're just getting back a document, but you're getting more than that. 

The use case is - I want to test my mapping against a document. How about "getDocumentForTesting"? I think that gives us the semantic flexibility to return a JSON object where "document" is one property (as opposed to "data", which is too vague).

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/mapping/getNewDocument.sjs
```diff
@@ -0,0 +1,42 @@
...
+
+var stepName, uri;
+
+const rtn = {
+  data: null,
+  namespaces: {},
+  format: null
+}
+
+const doc = cts.doc(uri);
```
rjrudin, line 28: If the doc isn't found, we should immediately throw an error. Use ds-util.sjs, and add a new function to it that we haven't needed yet, but we need now - ds.throwNotFound("Could not find a document with URI: " + uri);

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/mapping/xmlToJsonForMapping.sjs
```diff
@@ -0,0 +1,311 @@
...
+const textPropName = textStartsWith + 'text';
+const attrStartsWith = '@';
+const outputAttrStartsWith = '@'; // in case this ever needs to vary from the input.
+const nsPrefixDelim = ':';
+const nsAttrStartsWith = attrStartsWith + 'xmlns';
+const defaultNSAttrName = nsAttrStartsWith;
+const nonDefaultNSAttrStartsWith = nsAttrStartsWith + nsPrefixDelim
+
+// Set up for parser
+const parser = require('/fast-xml-parser/src/parser.js');
```
rjrudin, line 29: Is all of this code coming from our UI? There's a ton of logic in here and I'm wondering how much automated test coverage we have of it.

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/data-services/mapping/getNewDocumentTest.sjs
rjrudin: Here's what I recommend:

1. Use JUnit tests against the DS-generated interface for the scenarios of - URI exists, and URI doesn't exist. 
1. Use ml-unit-test for all of the actual logic, but invoke a function that does all the logic once the result of cts.doc has been obtained

That allows you to write ml-unit-tests that don't persist anything, which leads to much simpler tests.

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/data-services/mapping/getNewDocumentTest.sjs
```diff
@@ -0,0 +1,183 @@
...
+            'Sparrow'
+          ]
+        }
+      }
+    }
+  }
+};
+
+const xmlUri = '/test/sampleCustomerDoc.xml';
+const xmlInput = `
```
rjrudin, line 28: I think it'll be cleaner to have one test files for a JSON test, and one test file for an XML test. That's because the XML test is naturally going to be much more interesting. So e.g. getJsonDocumentForTestingTest.sjs and getXmlDocumentForTestingTest.sjs

rjrudin (changes requested):
This looks good, but I definitely want to avoid adding all the fast-xml-parser files to source control. I'm assuming we need most of the files though? (I'm guessing we don't need the LICENSE file, e.g.)

I'll submit a PR into your PR that takes a different approach here. This is a new thing for us, and I want to get a good pattern setup for how we should reuse 3rd party libraries without storing all their source files in version control.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/mapping/xmlToJsonForMapping.sjs
brent-hartwig: I wrote all of this code for this ticket.  I wasn't aware of existing code :/  Can you provide a pointer to that code?

srinathgit: The UI has some of the code (as the requirements for endpoint is slightly different) in mapping-card.tsx and source-to-entity-map.tsx

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/data-services/mapping/getNewDocumentTest.sjs
```diff
@@ -0,0 +1,183 @@
...
+<Order xmlns="https://www.w3schools.com/OrderNS">
```
rjrudin, line 29: I recommend slimming this down to the bare minimum needed to verify that the XML parsing is working correctly. For example, you likely don't need all of the top-level properties that do not have a namespace - those are just adding noise.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/mapping/xmlToJsonForMapping.sjs
rjrudin: That's fine that you wrote it, I was just curious if it came from the UI first. 

The challenge here is there's a lot of code / conditionals / logic here, and I'm wondering how to ensure it's all covered correctly. There's one ml-unit-test for it, and while the XML input document has a lot of complexity to it, there's no way to know what the total set of scenarios is that we need to support. 

The tests can identify that though. Instead of having one large test that involves many scenarios, I think a better approach is to have a lot of small tests (they could all be separate functions in one file) that each test a single scenario. That would better identify what the total set of known scenarios that we support is.

brent-hartwig: Thanks, gents.  I'll check out the existing code (as it is proven) and will work over the tests as suggested.

rjrudin:
Created https://github.com/brent-hartwig/marklogic-data-hub/pull/2 to avoid storing the fast-xml-parser files in source control. 

@brent-hartwig If that looks good, please modify the zip file to include the version number in the filename.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/mapping/xmlToJsonForMapping.sjs
brent-hartwig: Initial assessment of referenced code...

### marklogic-data-hub-central/ui/src/components/entities/mapping/mapping-card.tsx:

* When source format is XML, fast-xml-parser's output is passed into generateNestedDataSource()
* When source format is JSON, that JSON is passed into generateNestedDataSource() as is.
* generateNestedDataSource() traverses the given JSON and creates a new structure that's quite a bit different than 6098's spec.
* Seems like only the traversal logic could be compared with DHFPROD-6098's implementation.

### /marklogic-data-hub-central/ui/src/components/entities/mapping/source-entity-map/source-to-entity-map.tsx

* Receives sourceData (output from generateNestedDataSource()) and fetchSrcDocFromUri (function) from mapping-card.tsx.
* fetchSrcDocFromUri appears only used when changing source document (via navigation).
* sourceData given to flattenSourceDoc()
* flattenSourceDoc() is the UI's implementation of DHFPROD-6101.

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/controllers/steps/MappingStepController.java
srinathgit: That's finr

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/mapping/xmlToJsonForMapping.sjs
srinathgit: DHFPROD-6101 covers entirety of flattenSourceDoc() and some more. We will have to handle scenarios where properties are not Qnames in DHFPROD-6101.

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/controllers/steps/MappingStepController.java
brent-hartwig: Addressed in fork; will be part of next or updated PR.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/mapping/getNewDocument.api
brent-hartwig: Changed to getDocumentForTesting.  Endpoint for front end also changed to `testingDoc`.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/mapping/getNewDocument.sjs
brent-hartwig: Addressed.  Will be present in updated/next PR.

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/data-services/mapping/getNewDocumentTest.sjs
brent-hartwig: Done.  In the updated or next PR there will be a getDocumentForTesting sub-directory in the suites/data-hub/5/data-services/mapping directory.  At the time of this writing, it includes three files defining tests: nonExistentDocTest.sjs, startWithJson.sjs, and startWithXml.sjs.

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/data-services/mapping/getNewDocumentTest.sjs
brent-hartwig: I was able to remove some elements that were not being tested, and added some tests for other unique portions that were not otherwise tested.

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/data-services/mapping/getNewDocumentTest.sjs
brent-hartwig: I added the JUnit tests as recommended, which surfaced a couple errors with that HC endpoint 👍.

brent-hartwig:
This PR is ready for its second review.  Thank you in advance.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/mapping/getDocumentForTesting.api
```diff
@@ -0,0 +1,18 @@
+{
+    "functionName": "getDocumentForTesting",
```
rjrudin, line 2: I know I proposed this, but seeing it here, I think it makes sense.

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/schemas/ExternalMetadata.java
```diff
@@ -0,0 +1,71 @@
+
```
rjrudin, line 1: Can you actually exclude this and any other schema-generated classes? I submitted a separate PR to update them.

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/data-services/lib/mappingService.sjs
```diff
@@ -0,0 +1,28 @@
+'use strict';
+
+// Utils for getDocumentForTesting's tests.
+const DocumentForTestingUtils = {
+  STEP_NAME: 'prospect2CustomerMappingStep',
+  DATABASE_NAME: 'data-hub-FINAL',
+  // Load data not provided via test-data directory.
+  loadTestData: function () {
+    // Found it difficult to figure out how to load a mapping step via test-data directory and get it into the collection
```
rjrudin, line 9: @srinathgit  Can you figure out the problem here? I know we have lots of ML tests that load mappings

rjrudin (changes requested):
Looks good to me, just one comment about dropping ExternalMetadata, and hoping @srinathgit  can figure out the problem with loading mappings in a test. We want as little friction as possible with writing tests.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/ds-utils.sjs
```diff
@@ -29,12 +29,17 @@ function throwForbidden(message) {
   fn.error(null, 'RESTAPI-SRVEXERR', Sequence.from([403, message]));
 }
 
+function throwNotFound(message) {
```
rjrudin, line 32: Pinging @loriwilhelm  for awareness - this is the library where we're defining convenience functions for throwing errors from ML. As more of DHF is converted to use this library, we'll have an easy way of seeing where legit server error are being thrown.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/mapping/getDocumentForTesting.api
brent-hartwig: Does or does not make sense?

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/schemas/ExternalMetadata.java
brent-hartwig: Removed!

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/mapping/getDocumentForTesting.api
rjrudin: Does, I'm just patting myself on the back :)

rjrudin (approved):
Approving, and you can remove the "draft" status from it

srinathgit (changes requested):
Can you squash all the commits to one and ensure all the tasks in checklist are done ? If they are all done, please go ahead and check them

srinathgit:
@xnikhil08 
Can you review this PR and ensure that the response is in line with what we want for the UI ?

brent-hartwig:
PR https://github.com/marklogic/marklogic-data-hub/pull/4835 is to replace this one.  It has all of this ticket's commits squashed into one.

PR #4810, opened 2020-11-02
DHFPROD-6168:Ensure we can get latest job details if a step is added more than once to a flow

srinathgit:
### Description
Ensure we can get latest job details if a step is added more than once to a flow
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/flow/getFlowWithLatestJobInfo.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/flow/getFlowWithLatestJobInfo.sjs
@@ -30,13 +30,19 @@ fn.head(datahub.hubUtils.queryLatest(function() {
     jobQueries.push(cts.collectionQuery('Job'));
     jobQueries.push(cts.jsonPropertyValueQuery("flow",flowWithStepDetails.name));
     jobQueries.push(cts.jsonPropertyValueQuery("stepName",step.stepName));
+    //A flow may contain same step more then once. 'status' in step response always contains step number
+    jobQueries.push(cts.jsonPropertyWordQuery("status",step.stepNumber));
+
     let latestJob = fn.head(fn.subsequence(cts.search(cts.andQuery(jobQueries),[cts.indexOrder(cts.jsonPropertyReference("timeStarted"), "descending")]), 1, 1));
     if(latestJob) {
       latestJob = latestJob.toObject();
-      step.jobId = latestJob.job.jobId;
-      let stepRunResponse = latestJob.job.stepResponses[step.stepNumber];
-      step.lastRunStatus = stepRunResponse.status;
-      step.stepEndTime = stepRunResponse.stepEndTime;
+      let stepRunResponses = latestJob.job.stepResponses;
+      if(stepRunResponses && stepRunResponses[step.stepNumber]){
+        let stepRunResponse = stepRunResponses[step.stepNumber];
+        step.jobId = latestJob.job.jobId;
+        step.lastRunStatus = stepRunResponse.status;
+        step.stepEndTime = stepRunResponse.stepEndTime;
+      }
     }
   });
 }, datahub.config.JOBDATABASE));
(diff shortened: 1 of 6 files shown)


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/flow/getFlowWithLatestJobInfo.sjs
```diff
@@ -30,13 +30,19 @@ fn.head(datahub.hubUtils.queryLatest(function() {
     jobQueries.push(cts.collectionQuery('Job'));
     jobQueries.push(cts.jsonPropertyValueQuery("flow",flowWithStepDetails.name));
     jobQueries.push(cts.jsonPropertyValueQuery("stepName",step.stepName));
+    //A flow may contain same step more then once. 'status' in step response always contains step number
```
rjrudin, line 33: Oof this is brittle - a word search on an unstructured string. Also, the step number can change over time - e.g. a user can move steps around in the flow. I'm really reluctant to base anything on step number for this reason. 

In fact, this seems like it could break for steps that aren't in a flow twice. For example, let's say my ABC mapping step is step 2, and I run it. Then I change my flow and make it step 3. Then I run this query - I no longer get anything back for my ABC step. Right? 

Because of the rarity of a step being in a flow twice (we shouldn't prohibit it, because we have no reason to do that, and there may be a valid reason to do it), I think we should just take the latest document and then say - see if there's a stepRunResponse with the same stepNumber and stepName as the given step. If so, use it. If not, then check for a stepRunResponse with the same stepName. If there's 1 or more, use the first one.

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/data-services/flow/test-data/jobs/job3.json
```diff
@@ -0,0 +1,32 @@
+{
```
rjrudin, line 1: For testing, don't you need a job document that has at least two step responses - one for each occurrence of the step in the flow?

rjrudin (changes requested):

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/flow/getFlowWithLatestJobInfo.sjs
srinathgit: So, in HC where we can run only one step at a time, it essentially means that if a step occurs twice in a flow, the latest job details for both the steps will be the same ?

rjrudin: This is a tough call, as stepNumber is effectively the unique ID for a step within a flow - except that the stepNumber can change, meaning that it's not a very good ID. 

Given that the previous code was already depending on stepNumber to find the given step, the simplest thing is to keep doing that. The scenario I mentioned above is already a problem with the current approach - i.e. that if the stepNumber changes for a step, then the latest job info will not be found. So constraining the query on the stepNumber being in status isn't any different. 

We still need tests for at least these scenarios that I can think of:

- A single job document that has stepResponses for both occurrences of a step in a flow
- Two job documents, one for each occurrence of the step in a flow (which would be produced by HC)

srinathgit: @rjrudin 
Just to be clear, the current implementation(changes in getFlowWithLatestJobInfo.sjs) in the PR is good. You want a couple of tests to be added. Is that correct ?

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/data-services/flow/test-data/jobs/job3.json
rjrudin: Right, I think we need those two scenarios above addressed. But depending on the stepNumber - while brittle - seems reasonable for now, since it's what DHF has been doing.

srinathgit: job3.json and job4.json addresses scenario 1 and job5.json scenario2.

rjrudin (approved):

briantang (approved):
Tested from the UI and it looks good.

PR #4820, opened 2020-11-03
DHFPROD-6138: Set columns for join table to nullable to avoid error w…

ryanjdew:
…ith PII fields

### Description

If the join property is PII then there will be an error on the mapping step when applying the TDE index if the column isn't nullable.

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [ ] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/impl/hub-entities/generateTDE.sjs
+++ b/marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/impl/hub-entities/generateTDE.sjs
@@ -1,0 +1,107 @@
+const hent = require("/data-hub/5/impl/hub-entities.xqy");
+const test = require("/test/test-helper.xqy");
+
+function generateTdeWithViewJoin() {
+    const input =
+        [{
+            "info": {
+                "title": "Customer",
+                "version": "0.0.1",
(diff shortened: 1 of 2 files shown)


srinathgit (approved):

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/impl/hub-entities/generateTDE.sjs
```diff
@@ -0,0 +1,92 @@
+const hent = require("/data-hub/5/impl/hub-entities.xqy");
+const test = require("/test/test-helper.xqy");
+
+function generateTdeWithViewJoin() {
+    const input =
```
rjrudin, line 5: Can you slim this down to the minimum needed to reproduce the problem? I'm guessing just the customerId and billing, and then you can drop Zip from Address? Or is Zip needed?

ryanjdew: I adjusted the test to also check the single value join of address to zip.

rjrudin (approved):

PR #4828, opened 2020-11-04
DHFPROD-6209: Convert reference project match/merge steps to HubCentr…

ryanjdew:
…al format

### Description

Ran the hubConvertForHubCentral task to update the match/merge steps.

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- N/A Added Tests
  

- ##### Reviewer:

- N/A Reviewed Tests

### DIFF
--- a/examples/reference-entity-model/steps/matching/match-customers.step.json
+++ b/examples/reference-entity-model/steps/matching/match-customers.step.json
@@ -7,32 +7,21 @@
   "targetEntity" : "Customer",
   "sourceDatabase" : "data-hub-FINAL",
   "collections" : [ "matched-customers", "match-customers" ],
-  "matchOptions" : {
-    "dataFormat" : "json",
-    "propertyDefs" : {
-      "property" : [ {
-        "localname" : "customerId",
-        "name" : "customerId"
-      } ]
-    },
-    "scoring" : {
-      "add" : [ {
-        "propertyName" : "customerId",
-        "weight" : "10"
-      } ]
-    },
-    "thresholds" : {
-      "threshold" : [ {
-        "above" : "10",
-        "label" : "Definitive Match",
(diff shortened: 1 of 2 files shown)


Sanjeevani19 (approved):

xnikhil08 (approved):

examples/reference-entity-model/steps/matching/match-customers.step.json
```diff
@@ -7,32 +7,21 @@
...
-      } ]
-    },
-    "tuning" : {
-      "maxScan" : 200
-    }
-  },
   "targetDatabase" : "data-hub-FINAL",
   "targetFormat" : "json",
-  "stepId" : "match-customers-matching"
+  "stepId" : "match-customers-matching",
```
rjrudin, line 12: @ryanjdew  @fsnow  Two thoughts on seeing this side-by-side with the "before" config:

1. This is much easier to read
1. I feel like there should be a top-level property that contains "matchRulesets" and "thresholds" - e.g. "matchConfig" . That makes it easy to grab all of the match config in a step, as opposed to having to know what properties constitute the match config and which ones do not.

Thoughts on #2?

ryanjdew: @rjrudin We don't group feature specific properties for mapping steps, granted it is only "namespaces" and "properties".  Also, having the properties at the top-level can have the added benefit of the front-end being able to only push updates to thresholds, or matchRulesets, etc. I'm not sure if the front-end was planning on using that approach or not.

It may be worth drafting a ticket and getting feedback. At this point, it would probably be best to make such a change after the current work is completed and it can be changed across the board to avoid disruption.

rjrudin (approved):

PR #4830, opened 2020-11-06
DHFPROD-6217:Catch and log error if updating job doc fails

srinathgit:
### Description
Catch and log error if updating job doc fails
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-spark-connector/src/test/java/com/marklogic/hub/spark/sql/sources/v2/writer/HandleAbortedWriterTest.java
+++ b/marklogic-data-hub-spark-connector/src/test/java/com/marklogic/hub/spark/sql/sources/v2/writer/HandleAbortedWriterTest.java
@@ -31,8 +31,14 @@ public class HandleAbortedWriterTest extends AbstractSparkConnectorTest {
         assertEquals(2, getFruitCount(), "Just verifying the two fruits were written");
 
         dataSourceWriter.abort(null);
-        assertEquals("canceled", getJobDocumentStatus(), "If the DataSourceWriter is aborted for any reason, the job " +
-            "document should have a status of 'canceled'. Note that this does not imply whether any writes failed. " +
-            "But that is a limitation of the JobStatus class in DHF.");
+        if(canUpdateJobDoc()){
+            assertEquals("canceled", getJobDocumentStatus(), "If the DataSourceWriter is aborted for any reason, the job " +
+                "document should have a status of 'canceled'. Note that this does not imply whether any writes failed. " +
+                "But that is a limitation of the JobStatus class in DHF.");
+        }
+        else{
+            verifyJobDocumentWasNotUpdated(getJobDocumentStatus());
+        }
+
     }
 }
--- a/marklogic-data-hub-spark-connector/src/test/java/com/marklogic/hub/spark/sql/sources/v2/writer/WriteJobWithCustomEndpointsTest.java
+++ b/marklogic-data-hub-spark-connector/src/test/java/com/marklogic/hub/spark/sql/sources/v2/writer/WriteJobWithCustomEndpointsTest.java
@@ -38,7 +38,7 @@ public class WriteJobWithCustomEndpointsTest extends AbstractSparkConnectorTest 
 
         verifyCustomInitializeEndpointIsUsed();
         dataSourceWriter.commit(null);
-        verifyCustomFinalizeEndpointIsUsed();
+        verifyStatusIsStopOnError();
     }
 
     @Test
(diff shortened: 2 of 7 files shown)


marklogic-data-hub-spark-connector/src/test/java/com/marklogic/hub/spark/sql/sources/v2/writer/HandleAbortedWriterTest.java
```diff
@@ -31,8 +32,10 @@ void dataSourceWriterIsAborted() {
         assertEquals(2, getFruitCount(), "Just verifying the two fruits were written");
 
         dataSourceWriter.abort(null);
-        assertEquals("canceled", getJobDocumentStatus(), "If the DataSourceWriter is aborted for any reason, the job " +
+        assertTrue(getJobDocumentStatus().equals("canceled") ||
```
rjrudin, line 35: Is the "or" check here only because it will be "started" on ML 9? If so, this should first do a check of the ML version to see if it's ML 9, and use that to determine what the expected status should be. Same goes for the other "or" checks below.

rjrudin (changes requested):

marklogic-data-hub-spark-connector/src/test/java/com/marklogic/hub/spark/sql/sources/v2/writer/HandleAbortedWriterTest.java
srinathgit: 1. The failures are not consistent in 9.0-x. In WriteJobWithCustomEndpointsTest , whenever custom endpoint is used to initialize write , we are able to update the job document (regardless of whether custom endpoint is used to finalize write or not). 
2. If default endpoints are used, we can create job doc but are not able to update it.

marklogic-data-hub-spark-connector/src/test/java/com/marklogic/hub/spark/sql/sources/v2/writer/HandleAbortedWriterTest.java
```diff
@@ -31,8 +31,14 @@ void dataSourceWriterIsAborted() {
...
-        assertEquals("canceled", getJobDocumentStatus(), "If the DataSourceWriter is aborted for any reason, the job " +
-            "document should have a status of 'canceled'. Note that this does not imply whether any writes failed. " +
-            "But that is a limitation of the JobStatus class in DHF.");
+        if(canUpdateJobDoc()){
+            assertEquals("canceled", getJobDocumentStatus(), "If the DataSourceWriter is aborted for any reason, the job " +
+                "document should have a status of 'canceled'. Note that this does not imply whether any writes failed. " +
+                "But that is a limitation of the JobStatus class in DHF.");
+        }
+        else{
+            assertEquals("started", getJobDocumentStatus(), "In case of 9.0-x server, status will remain 'started'");
```
rjrudin, line 40: Do we know why this is not working on 9.0-x? Is it something about amps and SJS functions? Whatever the reason is, it should be part of the assertion message.

marklogic-data-hub-spark-connector/src/test/java/com/marklogic/hub/spark/sql/sources/v2/writer/WriteJobWithCustomEndpointsTest.java
```diff
@@ -38,7 +38,10 @@ void bothEndpointsAreCustom() {
 
         verifyCustomInitializeEndpointIsUsed();
         dataSourceWriter.commit(null);
-        verifyCustomFinalizeEndpointIsUsed();
+        assertEquals("stop-on-error", getJobDocumentStatus(),
```
rjrudin, line 41: Just to avoid duplication, toss this into a new private method - e.g verifyStatusIsStopOnError.

marklogic-data-hub-spark-connector/src/test/java/com/marklogic/hub/spark/sql/sources/v2/writer/HandleAbortedWriterTest.java
rjrudin: Actually, this can become a new protected method in AbstractSparkConnectorTest - verifyJobDocumentWasNotUpdated(). Then you can reuse it across tests.

marklogic-data-hub-spark-connector/src/test/resources/custom-job-endpoints/finalizeWrite.sjs
```diff
@@ -26,4 +26,11 @@ const Jobs = require("/data-hub/5/impl/jobs.sjs");
 
 const alwaysUseThisStatus = "stop-on-error";
 
-Jobs.updateJob(datahub, jobId, alwaysUseThisStatus, null, null, null, null);
+try{
+  Jobs.updateJob(datahub, jobId, alwaysUseThisStatus, null, null, null, null);
+}
+catch(ex){
+  console.log("Failed to update job document; cause: " + ex);
+}
```
rjrudin, line 34: Just to confirm - the try/catch is needed for the scenario where the custom finalize endpoint is used, but not the custom init endpoint?

rjrudin (changes requested):

marklogic-data-hub-spark-connector/src/test/resources/custom-job-endpoints/finalizeWrite.sjs
srinathgit: That is correct.

rjrudin (approved):

SameeraPriyathamTadikonda (approved):

PR #4832, opened 2020-11-06
DHFPROD-6146: Setup initial JMeter performance project

devinnicholson:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


performance-tests/src/test/jmeter/README.txt
```diff
@@ -0,0 +1,4 @@
+To run JMeter Test:
```
rjrudin, line 1: Let's put this in the README.md file in the root of this project - that's where folks expect to find the instructions for using a project.

rjrudin (changes requested):
Looks good, just a couple small tweaks

srinathgit (changes requested):
I think this needs a rebase with latest develop. Also, there should be just one commit or you will have to squash multiple commits to one

PR #4843, opened 2020-11-10
DHFPROD-5984: Adding indexes to filter on sourceName and sourceType as hub facets

rahulvudutala:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/AbstractHubCentralTest.java
+++ b/marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/AbstractHubCentralTest.java
@@ -27,6 +27,7 @@ import java.nio.file.Paths;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
+import java.util.Properties;
 
 /**
  * Intended base class for any test that wishes to connect to MarkLogic. Tests that do not have any need to connect to
@@ -76,8 +77,8 @@ public abstract class AbstractHubCentralTest extends AbstractHubTest {
 
         testHubProject = new HubProjectImpl();
         testHubConfig = new HubConfigImpl(testHubProject);
-
         resetHubProject();
+        testHubConfig.applyProperties(new Properties());
 
         // Run as the least-privileged HC user
         runAsHubCentralUser();
@@ -126,9 +127,9 @@ public abstract class AbstractHubCentralTest extends AbstractHubTest {
      */
     @Override
     protected HubClient runAsUser(String username, String password) {
-        // Initialize a new HubConfigImpl
-        testHubConfig = hubCentral.newHubConfig(username, password);
-        testHubConfig.setHubProject(testHubProject);
+        // Need to create the project directory before applying properties
+        testHubConfig.createProject(testProjectDirectory);
+        testHubConfig.applyProperties(hubCentral.buildPropertySource(username, password));
 
         // Update the provider with a new HubClient
         hubClientProvider.setHubClientDelegate(testHubConfig.newHubClient());
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/DeployDatabaseFieldCommand.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/DeployDatabaseFieldCommand.java
@@ -78,6 +78,7 @@ public class DeployDatabaseFieldCommand extends AbstractResourceCommand {
         addExistingFields(newProps, existingProps);
         addExistingRangeFieldIndexes(newProps, existingProps);
         addExistingRangePathIndexes(newProps, existingProps);
+        addExistingPathNamespaces(newProps, existingProps);
         return newProps.getPrettyXml();
     }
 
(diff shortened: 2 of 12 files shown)


ryanjdew (dismissed):

marklogic-data-hub/src/main/resources/ml-config/databases/final-database.json
```diff
@@ -1,5 +1,9 @@
 {
   "database-name": "%%mlFinalDbName%%",
+  "path-namespace" : [ {
```
rjrudin, line 3: I'm now remembering that if we modify this file in particular, we have to update HubProjectImpl:upgradeProject so that this addition is made programmatically. That's because we can't just overwrite this file, as the user may have made changes in it. 

For that reason, I'm thinking it's better to add the es path-namespace to the XML fields files and then enhance DeployDatabaseFieldsCommand to ensure that the es path-namespace is added if it doesn't exist yet. And we'll need to add tests to verify that it properly adds it or doesn't add it.

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/impl/hub-entities/generateSearchOptions.sjs
```diff
@@ -409,6 +409,39 @@ function verifySnippetOptions() {
   ]
 }
 
+function verifySourceNameAndSourceTypeOptions() {
```
rjrudin, line 412: I don't think this is the right way to test this, because we're not programmatically generating any options - we're just hardcoding two new options in.

Instead, I think we need an ml-test that loads a couple documents - one with sourceName/sourceType, the other without - and then use search:search with the installed options. Then verify that the search:response has the facets for sourceName/sourceType. That verifies both that the options were generated correctly and that they work.

rjrudin (changes requested):

ryanjdew (approved):

marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/DeployDatabaseFieldCommand.java
```diff
@@ -141,6 +142,30 @@ protected void addExistingRangePathIndexes(Fragment newProps, Fragment existingP
         }
     }
 
+    protected void addExistingPathNamespaces(Fragment newProps, Fragment existingProps) {
+        Element newNamespaces = newProps.getInternalDoc().getRootElement().getChild("path-namespaces", MANAGE_NS);
+        if(newNamespaces == null) {
```
rjrudin, line 147: We know path-namespaces exists in newProps - I don't see the point of this null check here?

marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/DeployDatabaseFieldCommand.java
```diff
@@ -141,6 +142,30 @@ protected void addExistingRangePathIndexes(Fragment newProps, Fragment existingP
...
+            String esNamespace = "<path-namespaces><path-namespace><prefix>es</prefix><namespace-uri>http://marklogic.com/entity-services</namespace-uri>" +
+                    "</path-namespace></path-namespaces>";
+            newNamespaces = newProps.getInternalDoc().getRootElement().addContent(esNamespace);
+        }
+
+        List<String> newNamespacePrefixes = new ArrayList<>();
+        newProps.getElements("/m:database-properties/m:path-namespaces/m:path-namespace").forEach(namespace -> {
+            String prefix = namespace.getChildText("prefix", MANAGE_NS);
+            if (StringUtils.isNotBlank(prefix)) {
```
rjrudin, line 156: You don't need this check here - prefix is a required field, it's guaranteed to exist.

marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/DeployDatabaseFieldCommand.java
```diff
@@ -141,6 +142,30 @@ protected void addExistingRangePathIndexes(Fragment newProps, Fragment existingP
...
+                newNamespacePrefixes.add(prefix);
+            }
+        });
+
+        for (Element namespace : existingProps.getElements("/m:database-properties/m:path-namespaces/m:path-namespace")) {
+            String prefix = namespace.getChildText("prefix", MANAGE_NS);
+            if (StringUtils.isNotBlank(prefix) && !newNamespacePrefixes.contains(prefix)) {
```
rjrudin, line 163: Same thing, don't need the isNotBlank check - if a path-namespace exists, it's guaranteed to have a prefix.

marklogic-data-hub/src/test/java/com/marklogic/hub/deploy/commands/DeployExistingNamespacesTest.java
```diff
@@ -0,0 +1,122 @@
...
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+public class DeployExistingNamespacesTest extends AbstractHubCoreTest {
```
rjrudin, line 21: No need for a separate test here - just add the path-namespace stuff to the existing DeployDatabaseFieldCommandTest .

rjrudin (changes requested):

timur-isangulov (dismissed):

ryanjdew (dismissed):

rjrudin (approved):

marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/AbstractHubCentralTest.java
```diff
@@ -127,8 +128,8 @@ protected File getTestProjectDirectory() {
     @Override
     protected HubClient runAsUser(String username, String password) {
         // Initialize a new HubConfigImpl
-        testHubConfig = hubCentral.newHubConfig(username, password);
-        testHubConfig.setHubProject(testHubProject);
+        testHubConfig.createProject(testProjectDirectory);
```
rjrudin, line 131: The comment above this should be removed now - and really, we ought to explain why the project is being created here. So I think a comment of "Need to create the project directory before applying properties" (assuming that is the correct reason).

rjrudin (changes requested):

rjrudin (approved):

ryanjdew (approved):

xnikhil08 (approved):

PR #4851, opened 2020-11-11
DHFPROD-6227: Explorer failing cypress tests.

timur-isangulov:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/e2e/cypress/integration/explore/scenarios.queries.tsx
+++ b/marklogic-data-hub-central/ui/e2e/cypress/integration/explore/scenarios.queries.tsx
@@ -374,6 +374,7 @@ describe('save/manage queries scenarios, developer role', () => {
         browsePage.selectQuery('reset-query');
         browsePage.clickColumnTitle(2);
         browsePage.getResetQueryButton().click();
+        cy.waitForAsyncRequest();
         browsePage.getResetConfirmationYesClick();
         browsePage.getEditSaveChangesButton().click();
         browsePage.getExploreButton().should('be.visible');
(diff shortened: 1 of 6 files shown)


marklogic-data-hub-central/ui/e2e/cypress/integration/explore/scenarios.queries.tsx
```diff
@@ -382,6 +383,7 @@ describe('save/manage queries scenarios, developer role', () => {
         browsePage.selectQuery('reset-query');
         cy.waitForAsyncRequest();
         browsePage.getAppliedFacets('Adams Cole').should('exist');
+        cy.wait(500)
```
xnikhil08, line 386: Did we try adding waitUntil on the next line? Since browsePage.getSortIndicatorDesc() might be returning a cy.get() output so we can add a waitUntil() there, forcing the browser to wait until it fetches the element.

Basically the goal is to exploit every possible scenario before adding explicit waits in our test cases since there would always be 1 out of 100 or 1000 cases when it might fail because of timeout delay more than the wait time limit.

marklogic-data-hub-central/ui/e2e/cypress/support/pages/browse.tsx
```diff
@@ -293,6 +293,7 @@ class BrowsePage {
   }
 
   clickColumnTitle(index: number) {
+    cy.wait(500);
```
xnikhil08, line 296: Same comment as earlier. Let's try to add waitUntil on the next get statement and see if it works, so that the explicit waits can be avoided.

xnikhil08 (changes requested):
Looks good, except few changes regarding explicit waits. I just want to ensure we have explored all possible options before using waits.

marklogic-data-hub-central/ui/e2e/cypress/integration/explore/scenarios.queries.tsx
timur-isangulov: Yes I did try to add waitUntil but it didn't help reliably while running tests on DHS.

marklogic-data-hub-central/ui/e2e/cypress/support/pages/browse.tsx
timur-isangulov: I had to add explicit wait here also to fix random failure. I did try to add waitUntil but it didn't help to fix random failure issue on DHS.

briantang (approved):

Sanjeevani19 (approved):

xnikhil08 (approved):

PR #4853, opened 2020-11-12
DHFPROD-6234:Fix failing tests in 9.0-x server

srinathgit:
### Description
Fix failing tests in 9.0-x server. This PR fixes 3 failing tests in https://jenkins.marklogic.com/blue/organizations/jenkins/Datahub_CI/detail/develop/1611/tests/.  

deployExplorerOptionsWithoutContainerConstraint() – com.marklogic.hub.entity.EntityManagerTest test is failing in 9.0-x.
@rjrudin, @akshaysonvane , @ryanjdew ,
It looks like we do not generate entity constraints in 9.0-x server. 
Not sure what should be done here, should the test be excluded in 9.0-x ?
 
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/test/ml-modules/root/test/suites/marklogic-data-hub-spark-connector/writeRecords/ingestionWithNullParams.sjs
+++ b/marklogic-data-hub/src/test/ml-modules/root/test/suites/marklogic-data-hub-spark-connector/writeRecords/ingestionWithNullParams.sjs
@@ -33,6 +33,7 @@ const envelope = record.document.envelope;
 
   //should add default metadata
   test.assertExists(record.metadata.datahubCreatedOn),
-  test.assertEqual("test-data-hub-developer",record.metadata.datahubCreatedBy),
+  test.assertExists(record.metadata.datahubCreatedBy),
+  test.assertNotEqual('', record.metadata.datahubCreatedBy),
   test.assertExists(record.metadata.datahubCreatedByJob)
 ];
--- a/marklogic-data-hub/src/test/java/com/marklogic/hub/entity/EntityManagerTest.java
+++ b/marklogic-data-hub/src/test/java/com/marklogic/hub/entity/EntityManagerTest.java
@@ -17,6 +17,7 @@ package com.marklogic.hub.entity;
 
 import com.marklogic.hub.AbstractHubCoreTest;
 import com.marklogic.hub.HubConfig;
+import com.marklogic.hub.MarkLogicVersion;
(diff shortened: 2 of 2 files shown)


ryanjdew:
We've excluded other tests for 9.0 when they've tested entity constraints.

marklogic-data-hub/src/test/ml-modules/root/test/suites/marklogic-data-hub-spark-connector/writeRecords/ingestionWithNullParams.sjs
```diff
@@ -33,6 +33,7 @@ const envelope = record.document.envelope;
 
   //should add default metadata
   test.assertExists(record.metadata.datahubCreatedOn),
+  String(xdmp.version()).startsWith("9") ? test.assertEqual("flow-developer",record.metadata.datahubCreatedBy):
```
rjrudin, line 36: Let's change this to asserting that datahubCreatedBy exists and is not "". The tests can be run by different users. We don't care about the identity of the user, we just care that the metadata is set to something.

marklogic-data-hub/src/test/java/com/marklogic/hub/scaffolding/CreateStepTest.java
```diff
@@ -52,8 +52,11 @@ void ingestionStep() throws IOException {
         verifyArtifactsAreWrittenToDbAndProject("myIngest2", "ingestion", null);
     }
 
+    // runAsDataHubDeveloper() runs as "flow-developer-role" in 9.0-x who doesn't have "hub-central-match-merge-writer, so explicitly
```
rjrudin, line 55: It's not the role that's needed, it's the http://marklogic.com/data-hub/privileges/write-match-merge privilege, right? I think it's totally fine to add those privileges to flow-developer-role, I think that's the right fix here. flow-developer-role is already over-powered since it has the manage-admin role, there's no harm in giving it more DHF privileges.

rjrudin (changes requested):
@srinathgit  @akshaysonvane  The supportsRangeIndexConstraints method in MarkLogicVersion may be the thing to check for the failing EntityManager test.

ryanjdew (approved):

akshaysonvane (approved):

rjrudin (approved):

PR #4854, opened 2020-11-12
DHFPROD-5278: Improve SMT merging step, so it is easier to track prog…

fsnow:
…ress and to restart it

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [ ] Added Tests
  

- ##### Reviewer:

- [ ] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mastering/default/merging.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mastering/default/merging.sjs
@@ -21,7 +21,6 @@ const httpUtils = require("/data-hub/5/impl/http-utils.sjs");
 const quickStartRequiredOptionProperty = 'mergeOptions';
 const hubCentralRequiredOptionProperty = 'mergeRules';
 const requiredOptionProperties = [[quickStartRequiredOptionProperty, hubCentralRequiredOptionProperty]];
-const processedURIs = [];
 
 function main(content, options) {
   let isSeparateMergeStep = false;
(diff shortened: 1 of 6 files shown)


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mastering/default/merging.sjs
```diff
@@ -40,84 +40,105 @@ function main(content, options) {
...
+      cts.pathRangeQuery("/matchSummary/URIsToProcess", "=" , uriToProcess));
+    if (lastSummaryWithURI != thisMatchSummaryURI) {
+      continue;
     }
-    otherMergesForURI.forEach((actionDetails) => actionDetails.uris.forEach(
-      (uri) => {
-        if (!uriActionDetails.uris.includes(uri)) {
-          uriActionDetails.uris.push(uri);
+
+    const uriRangeQuery = cts.rangeQuery(urisPathReference, '=', uriToProcess);
```
rjrudin, line 65: My main recommendation so far is to try to break this 70-line stretch of code up into several functions. There are several chunks of processing to work through here - moving each chunk into its own function would allow the function name to document what the chunk is doing, and also what the inputs and outputs are.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mastering/default/merging.sjs
```diff
@@ -40,84 +40,105 @@ function main(content, options) {
...
-        // filter out false positives that don't have merge actions for the given URI
-        .filter((actionNode) => !!actionNode)
-        // convert node to JSON object
-        .map((actionNode) => actionNode.toObject());
-    if (otherMergesForURI.some((otherMerge) => otherMerge.uris.length > uriActionDetails.uris.length)) {
-      return Sequence.from([]);
+
+  let returnArray = [];
+
+  let thisMatchSummary = fn.head(cts.search(cts.documentQuery(thisMatchSummaryURI), ['unfiltered']));
```
ryanjdew, line 50: You should be able to retrieve this from `content.value`.  Also in a case like this where we want a single document from a URI we can use `cts.doc("uri")`.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mastering/default/merging.sjs
```diff
@@ -40,84 +40,105 @@ function main(content, options) {
...
+    }
+
+    let results = mastering.buildContentObjectsFromMatchSummary(
+      uriToProcess,
+      matchSummaryJson,
+      options,
+      datahub.prov.granularityLevel() === datahub.prov.FINE_LEVEL
+    );
+    processedURIs.push(uriToProcess);
+    for (let matchSummary of relatedMatchSummaries) {
```
ryanjdew, line 119: This for loop can go away. We can change it so that we set `content["$delete"] = true;` and then push content into the results.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mastering/default/merging.sjs
```diff
@@ -40,84 +40,105 @@ function main(content, options) {
...
+    if (lastSummaryWithURI != thisMatchSummaryURI) {
+      continue;
     }
-    otherMergesForURI.forEach((actionDetails) => actionDetails.uris.forEach(
-      (uri) => {
-        if (!uriActionDetails.uris.includes(uri)) {
-          uriActionDetails.uris.push(uri);
+
+    const uriRangeQuery = cts.rangeQuery(urisPathReference, '=', uriToProcess);
+    let relatedMatchSummaries = cts.search(
```
ryanjdew, line 66: Due to not needing the for loop below, we can change this to only get the single most recent match summary.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mastering/default/merging.sjs
```diff
@@ -40,84 +40,105 @@ function main(content, options) {
...
+        (uri) => {
+          if (!uriActionDetails.uris.includes(uri)) {
+            uriActionDetails.uris.push(uri);
+          }
         }
+      ));
+    }
+
+    let results = mastering.buildContentObjectsFromMatchSummary(
+      uriToProcess,
```
ryanjdew, line 113: This can accept a Sequence of URI strings, we could potentially change this to be a single call for all the URIs that don't have a mostRecentMatchSummary with a different URI from the `content.uri`.

ryanjdew (changes requested):
`const processedURIs = [];` can be removed since we no longer need to track them since we are handling an entire match summary at a time.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mastering/default/merging.sjs
```diff
@@ -39,86 +38,85 @@ function main(content, options) {
...
+          .map(
+            (result) => result.xpath(`/matchSummary/actionDetails/*[action = 'merge']`).toArray()
+              .filter((actionNode) => cts.contains(actionNode, urisQuery))[0]
+          )
+          // filter out false positives that don't have merge actions for the given URI
+          .filter((actionNode) => !!actionNode)
+          // convert node to JSON object
+          .map((actionNode) => actionNode.toObject());
+      if (otherMergesForURI.some((otherMerge) => otherMerge.uris.length > uriActionDetails.uris.length)) {
+        return Sequence.from([]);
```
ryanjdew, line 94: I believe this could cause us to miss out on other URIs in the match summary that doesn't have more recent updates in other match summaries. 

It would probably be better to filter `theseURIsToProcess` with this logic and then make a single call to `buildContentObjectsFromMatchSummary` with the sequence of URIs in that filtered list.

ryanjdew (changes requested):

ryanjdew:
Closing in favor of PR 4922, which has been rebased and has fixed ML unit tests.

PR #4857, opened 2020-11-12
DHFPROD-6244:Create match and merge step in the new format by default

srinathgit:
### Description
Create match and merge step in the new format by default
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/artifacts/matching.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/artifacts/matching.sjs
@@ -16,7 +16,6 @@
 'use strict';
 
 const DataHubSingleton = require('/data-hub/5/datahub-singleton.sjs');
-const Security = require("/data-hub/5/impl/security.sjs");
 
 // define constants for caching expensive operations
 const dataHub = DataHubSingleton.instance();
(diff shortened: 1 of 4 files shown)


ryanjdew (approved):

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/artifacts/matching.sjs
```diff
@@ -78,40 +77,8 @@ function defaultArtifact(artifactName) {
...
-      },
-      thresholds: {
-        threshold: []
-      },
-      tuning: {
-        maxScan: 200
-      }
-    }
-  }
-
```
fsnow, line 114: This change is backing out the changes that I just made for DHFPROD-6178 and -6179, so that the match and merge options are created in either the old or new format depending on the roles of the user. Do we not want to support the creation of QuickStart-compatible steps?

fsnow (changes requested):

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/artifacts/matching.sjs
srinathgit: While working on test failures in 9.0-x, we realized that when a step is created in QS, it doesn't follow this codepath. This codepath is meant for creating steps either in HC or using hubCreateStep (which creates step in HC format);  so we might as well create the steps in the new format

fsnow (approved):

PR #4859, opened 2020-11-13
DHFPROD-6143:Run OOTB ml-gradle DMSDK tasks against Final db in DHS.

srinathgit:
### Description
Run OOTB ml-gradle DMSDK tasks against Final db in DHS.
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/test/java/com/marklogic/hub/impl/HubConfigImplTest.java
+++ b/marklogic-data-hub/src/test/java/com/marklogic/hub/impl/HubConfigImplTest.java
@@ -1,5 +1,6 @@
 package com.marklogic.hub.impl;
 
+import com.marklogic.client.DatabaseClient;
 import com.marklogic.client.ext.SecurityContextType;
 import com.marklogic.hub.DatabaseKind;
 import com.marklogic.hub.HubConfig;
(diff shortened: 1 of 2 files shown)


marklogic-data-hub/src/test/java/com/marklogic/hub/impl/HubConfigImplTest.java
```diff
@@ -129,6 +131,15 @@ void hostOnAppConfigShouldBeUpdated() {
             "LoadSchemasCommand requires that the schemas path be absolute; path: " + config.getAppConfig().getSchemaPaths());
     }
 
+    @Test
+    void connectionTypeIsGatewayInProvisionedEnvironment(){
+        HubConfigImpl hubConfig = new HubConfigImpl("localhost", "admin", "admin");
+        Assertions.assertNull(hubConfig.getAppConfig().getRestConnectionType());
+        hubConfig.setIsProvisionedEnvironment(true);
+        hubConfig.applyProperties(new Properties());
+        Assertions.assertEquals(DatabaseClient.ConnectionType.GATEWAY, hubConfig.getAppConfig().getRestConnectionType());
```
rjrudin, line 140: Looks like this should assert on appServicesConnectionType as well? Though is that actually required for the ml-gradle tasks to work? Those tasks will only use the appServices port if the user specifies a "-Pdatabase=". And that path doesn't always work well due to a bug in the REST API. The ml-gradle docs tell the user to override databaseClient directly as a better approach.

rjrudin (changes requested):

rjrudin (approved):

akshaysonvane (approved):

PR #4862, opened 2020-11-13
DHFPROD-6101: Enhance the /api/steps/mapping/{stepName}/testingDoc endpoint to return "sourceProperties"

brent-hartwig:
### Description
This is an extension of DHFPROD-6098, "Create endpoint for fetching source document for mapping UI".

It adds the sourceProperties property to the endpoint's return.

Full endpoint path: /api/steps/mapping/{stepName}/testingDoc

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [ ] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/mapping/getDocumentForTesting.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/mapping/getDocumentForTesting.sjs
@@ -19,13 +19,83 @@ xdmp.securityAssert('http://marklogic.com/data-hub/privileges/read-mapping', 'ex
 
 const core = require('/data-hub/5/artifacts/core.sjs')
 const ds = require("/data-hub/5/data-services/ds-utils.sjs");
+const xmlToJson = require('./xmlToJsonForMapping.sjs');
 
 var stepName, uri;
+
+function _isSourceJson(format) {
+  return format.toUpperCase() === 'JSON'
+}
+
+function _isValidQName(name) {
+  try {
+    fn.QName('', name);
+    return true;
+  } catch (e) {
+    return false;
+  }
+}
+
+function _isObject(value) {
+  // Added key criteria as typeof(value) returned 'object' for some scalar values.
+  return (value && typeof(value) === 'object' && Object.keys(value).length > 0) === true;
+}
+
+function _isAtomic(value) {
+  return !_isObject(value);
+}
+
+function _isArray(value) {
+  // False negatives from Array.isArray(value)
+  return value && value.hasOwnProperty('0');
+}
+
+function _getXPath(leadingPath, nextPart, value, format, isArray) {
+  // Account for invalid QNames, which is possible when the source format is JSON.
+  let funcStart = '';
+  let funcEnd = '';
+  if (_isSourceJson(format) && !_isValidQName(nextPart)) {
+    // Array of atomic values
+    if (isArray && value.length > 0 && _isAtomic(value[0])) {
+      funcStart = "array-node('";
+      funcEnd = "')/node()";
+    } else {
+      // Either not an array, an empty array, or an array of object values.
+      funcStart = "node('";
+      funcEnd = "')";
+    }
+  }
+  return `${leadingPath}/${funcStart}${nextPart}${funcEnd}`;
+}
+
+// Recursive function used to populate the sourceProperties portion/array of the return.
+function _flatten(sourceData, sourceFormat, flatArray, flatArrayKey = '', level = 0) {
+  let value, isObject, isArray, xpath;
+  for (let key of Object.keys(sourceData)) {
+    // sourceProperties is not to receive the #text properties.
+    if (key === xmlToJson.PROP_NAME_FOR_TEXT) { continue }
+
+    value = sourceData[key];
+    isObject = _isObject(value);
+    isArray = _isArray(value);
+    xpath = _getXPath(flatArrayKey, key, value, sourceFormat, isArray);
+    flatArray.push({
+      name: key,
+      xpath: xpath,
+      struct: isObject,
+      level: level
+    })
+    if (isObject && !isArray) {
+      _flatten(value, sourceFormat, flatArray, `${flatArrayKey}/${key}`, level + 1);
+    }
+  }
+}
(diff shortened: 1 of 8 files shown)


srinathgit (approved):
Looks good ..

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/mapping/getDocumentForTesting.sjs
```diff
@@ -19,13 +19,83 @@ xdmp.securityAssert('http://marklogic.com/data-hub/privileges/read-mapping', 'ex
...
+      // Either not an array, an empty array, or an array of object values.
+      funcStart = "node('";
+      funcEnd = "')";
+    }
+  }
+  return `${leadingPath}/${funcStart}${nextPart}${funcEnd}`;
+}
+
+// Recursive function used to populate the sourceProperties portion/array of the return.
+function _flatten(sourceData, sourceFormat, flatArray, flatArrayKey = '', level = 0) {
```
rjrudin, line 72: Assuming all of the functions added here are needed for this "flatten" function, let's move them into a separate library (in the same directory). That allows us to export the one public method that's needed here, and to avoid using "_", which is not a convention that's used anywhere else in DHF. 

"flatten" doesn't seem like the right function name either, as that's understood in JS to mean what "Array.flat()" means. And the comment indicates that it does something more interesting too - to populate the sourceProperties object. 

The user goal appears to be - given a source document and a source format, I want back an array with the source properties that I find in the source document. Right? So I think that public function in the library should be:

    buildSourceProperties(sourceDocument, sourceRecord)

Which would return an array. And then that library can have a recursive function that has a signature that includes the array, the array key, and the level.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/mapping/getDocumentForTesting.sjs
```diff
@@ -19,13 +19,83 @@ xdmp.securityAssert('http://marklogic.com/data-hub/privileges/read-mapping', 'ex
...
+function _isAtomic(value) {
+  return !_isObject(value);
+}
+
+function _isArray(value) {
+  // False negatives from Array.isArray(value)
+  return value && value.hasOwnProperty('0');
+}
+
+function _getXPath(leadingPath, nextPart, value, format, isArray) {
```
rjrudin, line 53: This function could use some docs on it to explain what each of these args means. And then we typically use "build" and "make" for functions that produce new things, whereas "get" means "get me something from this data structure". Also, I believe the goal of this function is to make a "safe" XPath expression, right? So how about "makeSafeXPathExpression" as the function name, and the docs should clarify what "safe" actually means. 
So I think this should be "makeXPathExpression"

rjrudin (changes requested):

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/mapping/xmlToJsonForMapping.sjs
```diff
@@ -309,3 +309,4 @@ function transform(xmlNode) {
   }
 }
 exports.transform = transform;
+exports.PROP_NAME_FOR_TEXT = textPropName;
```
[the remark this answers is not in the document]
brent-hartwig: True.  Background: `textNodeName` is a Fast XML Parser option name, I established a naming convention for that and other constants within xmlToJsonForMapping while developing DHFPROD-6098, found a need to export that value while developing DHFPROD-6101, elected to use the all-caps naming convention, yet only did so for the exported one.  How about sticking with `PROP_NAME_FOR_TEXT` and changing the others (less Fast XML parser's) within xmlToJsonForMapping?

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/data-services/mapping/getDocumentForTesting/nonExistentDocTest.sjs
```diff
@@ -3,16 +3,15 @@
...
 const uri = '/uri/to/non-existent/doc.json';
 
 // Using try/catch as an alternative to test.assertThrowsError*().
 try {
   utils.invokeService(utils.STEP_NAME, uri);
   assertions.push(test.fail('Exception not thrown when attempting to process a non-existent document.'));
 } catch (e) {
-  xdmp.log(JSON.stringify(e));
-  assertions.concat([
+  assertions = assertions.concat([
```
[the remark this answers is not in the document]
brent-hartwig: Addressed, but not yet pushed here.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/mapping/getDocumentForTesting.sjs
brent-hartwig: Went with `makeSafeXPathExpression()` and included function-level documentation.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/mapping/getDocumentForTesting.sjs
brent-hartwig: Implemented as requested.  The relocated code is within `testingSourceProperties.sjs` (open to a different name).

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/mapping/xmlToJsonForMapping.sjs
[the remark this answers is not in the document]
brent-hartwig: > How about sticking with PROP_NAME_FOR_TEXT and changing the others (less Fast XML parser's) within xmlToJsonForMapping?

Implemented.

brent-hartwig:
Feedback addressed in https://github.com/marklogic/marklogic-data-hub/pull/4867.  Closing this PR.

PR #4863, opened 2020-11-16
DHFPROD-6246:Enable logging in dhs installer, client jars

srinathgit:
### Description
Enable logging in dhs installer, client jars
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [n/a] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/build.gradle
+++ b/marklogic-data-hub/build.gradle
@@ -40,6 +40,12 @@ ext {
     thirdPartySourcePath = "src/main/resources/ml-modules/root/data-hub/third-party"
 }
 
+// This enables including a logging jar in the application artifacts produced by this project - the installer and the client jars
+// without that jar ending up in the library artifact.
+configurations {
+    datahubLogging
+}
+
 dependencies {
     // Forcing use of 1.3.71 - which is what OkHttp 4.7.2 wants, which is what Java Client 5.3 uses - so that 1.2.x
     // versions don't come onto the classpath
@@ -62,7 +68,7 @@ dependencies {
 
     compile 'commons-io:commons-io:2.4'
     compile 'org.apache.commons:commons-text:1.1'
-
+    datahubLogging 'ch.qos.logback:logback-classic:1.2.3'
     // For installer program
     compile "com.beust:jcommander:1.72"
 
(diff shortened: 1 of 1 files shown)


akshaysonvane (approved):

ryanjdew (approved):

marklogic-data-hub/build.gradle
```diff
@@ -62,7 +62,7 @@ dependencies {
 
     compile 'commons-io:commons-io:2.4'
     compile 'org.apache.commons:commons-text:1.1'
-
+    compile 'ch.qos.logback:logback-classic:1.2.3'
```
rjrudin, line 65: We don't want to include logback as a compile dependency, as that can cause issues for applications that depend on the DHF library as a dependency (it's not considered good form for libraries to include logging implementations either). 

I found this link for Spring Boot 1.x that shows a mechanism for creating a custom Gradle configuration that excludes a logging dependency - https://docs.spring.io/spring-boot/docs/1.2.0.M1/reference/html/build-tool-plugins-gradle-plugin.html#build-tool-plugins-gradle-repackage-custom-configuration 

I didn't find something similar in the Spring Boot 2.1.x docs though - https://docs.spring.io/spring-boot/docs/2.1.18.RELEASE/reference/html/build-tool-plugins-gradle-plugin.html 

Check for something like this though. Also test out the client JAR, which expects logging to work and in theory would have the same issue.

rjrudin (changes requested):

marklogic-data-hub/build.gradle
rjrudin: I tried this locally, it seems to do the trick - be sure to test out both the client JAR and installer JAR:

```
configurations {
    datahubLogging
}

dependencies {
  ...
  datahubLogging 'ch.qos.logback:logback-classic:1.2.3'
}

// add this to both clientJar and bootJar
from(configurations.datahubLogging.collect { it.isDirectory() ? it : zipTree(it) }) {
        exclude "META-INF/*.SF"
        exclude "META-INF/*.DSA"
        exclude "META-INF/*.RSA"
    }
```

srinathgit: It works fine for both client and installer jars. Thanks for the info on configuration. I will check it out.

marklogic-data-hub/build.gradle
```diff
@@ -40,6 +40,10 @@ ext {
     thirdPartySourcePath = "src/main/resources/ml-modules/root/data-hub/third-party"
 }
 
+configurations {
+    datahubLogging
```
rjrudin, line 44: Sorry one more thing - can you add a comment on this explaining why it exists? We want to document that this is used so we can include a logging implementation jar in the application artifacts produced by this project - the installer and the client jars - without that ending up in the library artifact.

rjrudin (changes requested):

rjrudin (approved):

akshaysonvane (approved):

PR #4868, opened 2020-11-16
DHFPROD-6218:Throw 404 or 400 instead of 500 when backend operation cannot find something

srinathgit:
### Description
Throw 404 or 400 instead of 500 when backend operation cannot find something. Not sure if it requires any new tests. The existing tests should run fine.
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [ n/a] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/entities/search/EntitySearchManagerTest.java
+++ b/marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/entities/search/EntitySearchManagerTest.java
@@ -96,7 +96,7 @@ public class EntitySearchManagerTest extends AbstractHubCentralTest {
 
         SearchQuery searchQuery = new SearchQuery();
         searchQuery.getQuery().setEntityTypeIds(Arrays.asList("Some-entityType"));
-        assertThrows(MarkLogicServerException.class, () -> new EntitySearchManager(getHubClient()).search(searchQuery), "Entity Model with name Some-entityType doesn't exist ");
+        assertNull(new EntitySearchManager(getHubClient()).search(searchQuery), "Entity Model with name Some-entityType doesn't exist ");
     }
 
     @Test
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/4/transforms/mlcp-flow-transform.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/4/transforms/mlcp-flow-transform.sjs
@@ -19,6 +19,7 @@ const config = require("/com.marklogic.hub/config.sjs");
 const consts = require("/data-hub/4/impl/consts.sjs");
 const flowlib = require("/data-hub/4/impl/flow-lib.sjs");
 const tracelib = require("/data-hub/4/impl/trace-lib.sjs");
+const httpUtils = require("/data-hub/5/impl/http-utils.sjs");
 
 function transform(content, context) {
   let uri = content.uri;
@@ -35,7 +36,7 @@ function transform(content, context) {
     optionsString = match[1];
     parsedTransformParam = transformString.replace(optionsString, '');
   }
-  
+
   let splits = parsedTransformParam.split(',');
   for (let i in splits) {
     let pair = splits[i];
@@ -49,7 +50,7 @@ function transform(content, context) {
   let flow = flowlib.getFlow(entityName, flowName, consts.INPUT_FLOW);
 
   if (!flow) {
-    fn.error(null, "RESTAPI-SRVEXERR", "The specified flow " + params.flow + " is missing.");
+    httpUtils.throwNotFoundWithArray(["Not Found", "The specified flow " + params.flow + " is missing."]);
   }
   let options = {};
   if (optionsString) {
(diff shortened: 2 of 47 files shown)


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/4/transforms/mlcp-flow-transform.sjs
```diff
@@ -49,7 +49,7 @@ function transform(content, context) {
   let flow = flowlib.getFlow(entityName, flowName, consts.INPUT_FLOW);
 
   if (!flow) {
-    fn.error(null, "RESTAPI-SRVEXERR", "The specified flow " + params.flow + " is missing.");
+    fn.error(null, "RESTAPI-SRVEXERR", Sequence.from([404, "Not Found", "The specified flow " + params.flow + " is missing."]));
```
rjrudin, line 52: Let's convert this to use ds-util.sjs too. That library of course is a little mis-named now - we may want to rename it to "http-util.sjs", since the functions are generic to HTTP. But the more error handling that's routed through those functions, the better, as it makes it easier for us to track where we're throwing errors. And of course, it reduces duplication of the usage of RESTAPI-SRVEXERR.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/transforms/mlcp-flow-transform.sjs
```diff
@@ -62,7 +62,7 @@ function transform(content, context = {}) {
 
     if (!flow) {
       datahub.debug.log({message: params, type: 'error'});
-      fn.error(null, "RESTAPI-SRVEXERR", Sequence.from(["DH-FLOWMISSING", "The specified flow " + flowName + " is missing.", content.uri]));
+      fn.error(null, "RESTAPI-SRVEXERR", Sequence.from([404, "DH-FLOWMISSING", "The specified flow " + flowName + " is missing.", content.uri]));
```
rjrudin, line 65: For these, let's add throwNotFoundWithArray(args) and throwBadRequestWithArray(args) so that when a client wants to return multiple things, it's easy to do so.

rjrudin (changes requested):

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/4/transforms/mlcp-flow-transform.sjs
srinathgit: The uri of http-utils.sjs will be /data-hub/5/data-services/http-utils.sjs . Should that be changed to /data-hub/5/impl/http-utils.sjs or /data-hub/5/http-utils.sjs ?

marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/entities/search/EntitySearchManagerTest.java
```diff
@@ -96,7 +96,7 @@ public void testSearchResultsOnNonExistentEntityModel() {
 
         SearchQuery searchQuery = new SearchQuery();
         searchQuery.getQuery().setEntityTypeIds(Arrays.asList("Some-entityType"));
-        assertThrows(MarkLogicServerException.class, () -> new EntitySearchManager(getHubClient()).search(searchQuery), "Entity Model with name Some-entityType doesn't exist ");
+        assertNull(new EntitySearchManager(getHubClient()).search(searchQuery), "Entity Model with name Some-entityType doesn't exist ");
```
srinathgit, line 99: @rahulvudutala ,
In entity-search-lib.sjs line 48, I changed the HTTP error code from 500 to 404. It causes this endpoint to return null instead of exception it used to throw earlier ? let me know if that's ok.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/4/transforms/mlcp-flow-transform.sjs
rjrudin: Oh I was saying - don't rename it now, let's take care of that in a separate PR. Sorry for the confusion.

srinathgit: Since we are using it outside the library outside of ds, is it ok to rename in this PR itself ?

rjrudin: Sure, that's fine. I'm not a fan of the "impl" directory (since "impl" is meaningless), but putting it in there is consistent with "hub-utils" and "flow-utils", so that'll work for now.

srinathgit: I think the PR is ready for review then

rjrudin (changes requested):
Looks good just one more change - DataServiceGenerator needs to be updated to reference http-utils.

marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/entities/search/EntitySearchManagerTest.java
rahulvudutala: Yeah it should be ok.

rjrudin (approved):

srinathgit:
Fixed the merge conflict

rjrudin (approved):

ryanjdew (approved):

PR #4884, opened 2020-11-19
DHFPROD-5984: Adding null check for failing deployment in DHS

rahulvudutala:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

### DIFF
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/DeployDatabaseFieldCommand.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/DeployDatabaseFieldCommand.java
@@ -144,6 +144,9 @@ public class DeployDatabaseFieldCommand extends AbstractResourceCommand {
 
     protected void addExistingPathNamespaces(Fragment newProps, Fragment existingProps) {
         Element newNamespaces = newProps.getInternalDoc().getRootElement().getChild("path-namespaces", MANAGE_NS);
+        if(newNamespaces == null) {
+            return;
+        }
         List<String> newNamespacePrefixes = new ArrayList<>();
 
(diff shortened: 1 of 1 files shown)


rjrudin (approved):

marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/DeployDatabaseFieldCommand.java
```diff
@@ -144,6 +144,9 @@ protected void addExistingRangePathIndexes(Fragment newProps, Fragment existingP
 
     protected void addExistingPathNamespaces(Fragment newProps, Fragment existingProps) {
         Element newNamespaces = newProps.getInternalDoc().getRootElement().getChild("path-namespaces", MANAGE_NS);
+        if(newNamespaces == null) {
```
rjrudin, line 147: Can you add a comment here as to why newNamespaces may be null? I just reproduced this, am trying to find out why it happens.

rjrudin (changes requested):

marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/DeployDatabaseFieldCommand.java
rjrudin: I'm trying this on dh-5-example project, and it appears that the new final-database.xml file is not being written to the project. Not sure what's up yet.

rjrudin: Okay I see the problem now. We need a different fix, this PR can be closed.

PR #4897, opened 2020-11-20
DHFPROD-6201: Visualize history of a record in the document metadata view

rahulvudutala:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


ryanjdew (approved):

timur-isangulov (changes requested):
Hi @rahulvudutala, can you please format the history time stamp on the Detail page according to UX design https://wiki.marklogic.com/display/ENGINEERING/Explore+Source+Data+Design#ExploreSourceDataDesign-DetailView
Screen Shot 2020-11-20 at 10 21 56 AM

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/entities/entity-search-lib.sjs
```diff
@@ -535,11 +540,58 @@ function isHubEntityInstance(docUri) {
...
+      const currentObject = {};
+      currentObject.updatedTime = metadata.datahubCreatedOn ? metadata.datahubCreatedOn : undefined;
+      currentObject.flow = metadata.datahubCreatedInFlow ? metadata.datahubCreatedInFlow : undefined;
+      currentObject.step = metadata.datahubCreatedByStep ? metadata.datahubCreatedByStep : undefined;
+      currentObject.user = metadata.datahubCreatedBy ? metadata.datahubCreatedBy : undefined;
+      history.push(currentObject);
+    }
+  } else {
+    const flowNames = fn.collection("http://marklogic.com/data-hub/flow").toArray().map(flow => JSON.stringify(flow.toObject().name));
+    const stepNames = fn.collection(["http://marklogic.com/data-hub/steps", "http://marklogic.com/data-hub/step-definition"]).toArray().map(step => JSON.stringify(step.toObject().name));
```
akshaysonvane, line 571: - Should we be ignoring "hub-core-artifacts" for both `flowNames` and `stepNames` here?
- We can use a Set for `flowNames` and `stepNames` instead of an array.
- No need to `JSON.stringify` here since `name` is already a string.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/entities/entity-search-lib.sjs
```diff
@@ -535,11 +540,58 @@ function isHubEntityInstance(docUri) {
...
+    for(let i=0; i<provResults.length; i+=3) {
+      const currentObject = {};
+      currentObject.updatedTime = provResults[i].dateTime ? provResults[i].dateTime : undefined;
+      for(let j=i; j<i+3; j++) {
+        if(flowNames.includes(JSON.stringify(provResults[j].associatedWith))) {
+          currentObject.flow = provResults[j].associatedWith;
+        } else if(stepNames.includes(JSON.stringify(provResults[j].associatedWith))) {
+          currentObject.step = provResults[j].associatedWith;
```
akshaysonvane, line 579: - Once `flowNames` and `stepNames` become a Set we can use `has` instead of `includes`
- Since `provResults[j].associatedWith` is an object, we can use `String()` or `.toString()` here.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/entities/entity-search-lib.sjs
```diff
@@ -535,11 +540,58 @@ function isHubEntityInstance(docUri) {
   return isHubEntityInstance;
 }
 
+function getRecordHistory(docUri) {
+  const history = [];
+  const provResults = xdmp.invokeFunction(function () {
```
rjrudin, line 545: Only this part needs to be amped. And we have a natural home for it - in prov.sjs. But let's not put it in the Provenance class (I don't think we can amp it either if we do that). Just make a new exported function in there. 

To keep it flexible, make "out" (if that is really "output", let's call it "output" - abbreviations rarely buy us anything) an argument to the function. So how about - findProvenance(uri, relations). 

Then, make the above amp "findProvenance.json" - no need for "amp" or "dhf" in the filename.

I think it's worth putting the invokeFunction in the amped function too, so that the client doesn't have to worry about doing that part.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/entities/entity-search-lib.sjs
```diff
@@ -535,11 +540,58 @@ function isHubEntityInstance(docUri) {
...
+    const out = {
+      dateTime: '?',
+      relations: {'associatedWith': '?', 'attributedTo': '?'}
+    };
+    const kvPattern = ps.opTriplePattern(match, out);
+    return op.fromTriples(kvPattern).result();
+  }, { 'database' : xdmp.database(datahub.config.JOBDATABASE)}).toArray();
+
+  if(provResults.length == 0) {
+    const metadata = xdmp.documentGetMetadata(docUri);
```
rjrudin, line 560: To shorten this function, put all of this logic into a separate function - e.g. "getRecordMetadata".

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/entities/entity-search-lib.sjs
```diff
@@ -535,11 +540,58 @@ function isHubEntityInstance(docUri) {
...
+    if(metadata) {
+      const currentObject = {};
+      currentObject.updatedTime = metadata.datahubCreatedOn ? metadata.datahubCreatedOn : undefined;
+      currentObject.flow = metadata.datahubCreatedInFlow ? metadata.datahubCreatedInFlow : undefined;
+      currentObject.step = metadata.datahubCreatedByStep ? metadata.datahubCreatedByStep : undefined;
+      currentObject.user = metadata.datahubCreatedBy ? metadata.datahubCreatedBy : undefined;
+      history.push(currentObject);
+    }
+  } else {
+    const flowNames = fn.collection("http://marklogic.com/data-hub/flow").toArray().map(flow => JSON.stringify(flow.toObject().name));
```
rjrudin, line 570: Getting all objects in a collection is generally not a safe practice. While we don't expect there to be hundreds of flows or steps, we can't prevent that. 

Instead, try the following approach:

1. Iterate through provResults first, building up the history array and also building arrays of flow and step names. 
1. Take all the flow names and use those to query for only the flow objects you need. 
1. Take all the step names and use those to query for only the steps / step definitions that you need. 
1. Iterate back over the provResults, replacing flow/step if it doesn't exist.

Though - what's the requirement behind populating flow/step only if the flow/step still exists? If it's something for the UI, I think it would be better to add boolean properties of flowExists and stepExists. We shouldn't hide the flow/step name just because it no longer exists.

rjrudin (changes requested):

marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/controllers/EntitySearchControllerTest.java
```diff
@@ -108,9 +108,17 @@ void testCRUDOnSavedQuery() throws Exception {
         verifyRequestIsForbidden(MockMvcRequestBuilders.delete(SAVED_QUERIES_PATH + "/query").param("id", "some-id"));
     }
 
+    // This test is done at the Java layer as multiple scenarios cannot be tested in ml-unit-tests due to the optic calls occurring in the same request.
+    // There appears to be a bug with Optic caching values in the same request that is being investigated.
     @Test
     void testGetDocumentByURI() throws Exception {
         ReferenceModelProject project = installOnlyReferenceModelEntities(true);
+        project.setCustomerDocumentMetadata(new DocumentMetadataHandle()
```
rjrudin, line 116: It's difficult to read this test and see that the 3 scenarios called out in the story - no prov records, 1 prov record, and 2 prov records - are being tested. 

Also, I'd much rather see this tests in core DHF as opposed to HC. In HC, a single smoke test will do the trick. 

So move the tests to core DHF, write the test against the EntitySearchService - call it GetRecordTest, since you're testing the getRecord endpoint - and I recommend a separate test method for each of the 3 scenarios. The steps don't matter. So for the first one, you can just ingest a record. For the second one, ingest it and map it. For the 3rd, ingest it, map it, and maybe use a custom step (or 2 custom steps instead of map). All values in each row in the history object need to be verified.

rjrudin (changes requested):

marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/controllers/EntitySearchControllerTest.java
rahulvudutala: There is a getRecordHistoryTest.sjs for that covers all the scenarios. getRecord.sjs test is failing as multiple scenarios cannot be tested in ml-unit-tests due to the optic calls occurring in the same request. There appears to be a bug with Optic caching values in the same request that is being investigated.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/entities/entity-search-lib.sjs
```diff
@@ -535,11 +541,138 @@ function isHubEntityInstance(docUri) {
...
+          attributes: {
+            location: docUri
+          }
+        };
+        const output = {
+          dateTime: '?',
+          relations: relations
+        };
+        const kvPattern = ps.opTriplePattern(match, output);
+        return op.fromTriples(kvPattern).result();
```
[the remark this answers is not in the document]
rjrudin: Though I'm not 100% sure - check with Erik about it. If groupBy doesn't work, custom code is fine.

rjrudin (changes requested):

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/entities/entity-search-lib.sjs
```diff
@@ -535,11 +537,94 @@ function isHubEntityInstance(docUri) {
...
+    }
+  });
+  return artifactNames;
+}
+
 module.exports = {
   addDocumentMetadataToSearchResults,
   addPropertiesToSearchResponse,
   buildPropertyMetadata,
+  getArtifactNamesFromUris,
```
rjrudin, line 625: I didn't see anywhere that this needed to be exported, I think it can be private to this library.

marklogic-data-hub/src/test/java/com/marklogic/hub/dataservices/entitySearch/GetRecordTest.java
```diff
@@ -0,0 +1,181 @@
...
+    private ReferenceModelProject project;
+
+    @BeforeEach
+    void beforeEach() {
+        service = EntitySearchService.on(getHubClient().getFinalClient());
+        project = installReferenceModelProject();
+    }
+
+    @Test
+    void testGetRecord() {
```
rjrudin, line 36: To make it clear what this scenario is testing, name this "recordHasMetadataButNoProvenance".

marklogic-data-hub/src/test/java/com/marklogic/hub/dataservices/entitySearch/GetRecordTest.java
```diff
@@ -0,0 +1,181 @@
...
+import com.marklogic.hub.test.Customer;
+import com.marklogic.hub.test.ReferenceModelProject;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import java.util.Arrays;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+//This test is done at the Java layer as multiple scenarios cannot be tested in ml-unit-tests due to the optic calls occurring in the same request.
```
rjrudin, line 22: I think JUnit is fine here, but was the problem that you had multiple scenarios in one test module? Or was there a problem with multiple test modules, each testing one scenario? In ml-unit-test, the setup.xqy module can be used for doing setup before each test module.

marklogic-data-hub/src/test/resources/entity-reference-model/flows/ingestToFinal.flow.json
```diff
@@ -37,6 +37,12 @@
         "targetDatabase": "data-hub-STAGING",
         "permissions": "data-hub-common,read,data-hub-common,update"
       }
+    },
```
rjrudin, line 40: Why not use the existing ingestion / mapping steps in this flow? I ask because this project gets loaded by a lot of tests, so I'd like to only add a new artifact to it if we really need it.

rjrudin (changes requested):

marklogic-data-hub/src/test/java/com/marklogic/hub/dataservices/entitySearch/GetRecordTest.java
rahulvudutala: This is testing all other properties in the response from getRecord endpoint except for the history. There are 3 methods for 3 different scenarios for record history

marklogic-data-hub/src/test/resources/entity-reference-model/flows/ingestToFinal.flow.json
rahulvudutala: Added new steps to verify if the step name is returned or not. Otherwise step-defintion name is returned when the existing steps are used.

marklogic-data-hub/src/test/java/com/marklogic/hub/dataservices/entitySearch/GetRecordTest.java
rahulvudutala: This was the error.
OPTIC-INVALARGS: fn.error(null, 'OPTIC-INVALARGS', errMsg+arg); -- Invalid arguments: cannot call pattern() with invalid predicates : [object Object]

I had multiple multiple scenarios in one test module

ryanjdew (dismissed):

timur-isangulov (changes requested):
Please provide unique key for each record for the MLTable component history table.
Screen Shot 2020-11-24 at 9 27 25 AM

rjrudin (changes requested):
Putting a Request changes on this, as I know the tests are being reworked to address some new scenarios we realized.

timur-isangulov (dismissed):
Looks good.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/entities/entity-search-lib.sjs
```diff
@@ -535,11 +537,112 @@ function isHubEntityInstance(docUri) {
...
 }
 
+function getRecordHistory(docUri) {
+  const history = [];
+  const relations = {'associatedWith': '?', 'attributedTo': '?'};
+  const provenanceRecords = Provenance.findProvenance(docUri, relations);
+
+  if(provenanceRecords.length) {
+    const flows = findFlowsAsMap();
+    const flowNames = getArtifactNamesFromUris(consts.FLOW_COLLECTION, "/flows/", ".flow.json");
```
rjrudin, line 547: You don't need this anymore, as flowNames = Object.keys(flows).

Also, I recommend renaming "flows" to "flowsMap". That makes it self-documenting that it's a map and not an array or sequence, which is what "flows" implies.

marklogic-data-hub/src/test/resources/test-projects/provenance-test/entities/Customer.entity.json
```diff
@@ -0,0 +1,17 @@
...
+    "title": "Customer",
+    "version": "0.0.1",
+    "baseUri": "http://example.org/"
+  },
+  "definitions": {
+    "Customer": {
+      "properties": {
+        "customerId": {
+          "datatype": "integer",
+          "sortable": true
```
rjrudin, line 12: Recommend removing this so there's no expectation that a range index is needed.

rjrudin (changes requested):
Looks good, just a couple tweaks.

rjrudin (approved):

ryanjdew (approved):

timur-isangulov (approved):

PR #4890, opened 2020-11-19
DHFPROD-6064: First attempt at running tests against DHS

rjrudin:
### Description

You'll need to add the following properties to gradle-local.properties, as the test plumbing will read properties in that file and in gradle-local.properties:

mlHost=changeme
mlUsername=changeme
mlPassword=changeme
mlSecurityUsername=same as mlUsername
mlSecurityPassword=same as mlUsername
hubDhs=true
hubSsl=true

Also deleted gradle-dhs.properties, which has no bearing on anything. 

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [ ] JIRA_ID included in all the commit messages
- [ ] PR title is in the format JIRA_ID:Title
- [ ] Rebase the branch with upstream
- [ ] Squashed all commits into a single commit
- [ ] Added Tests
  

- ##### Reviewer:

- [ ] Reviewed Tests

PR #4900, opened 2020-11-24
DHFPROD-6290:Original gradle properties get reset if runAsUser() is called

srinathgit:
### Description
Original gradle properties get reset if runAsUser() is called. Once PR #4890 and this gets merged, most of the java tests should pass in DHS. No new tests needed, existing ones should pass.

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [n/a ] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/test/java/com/marklogic/hub/HubTestBase.java
+++ b/marklogic-data-hub/src/test/java/com/marklogic/hub/HubTestBase.java
@@ -59,6 +59,7 @@ import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.util.Arrays;
+import java.util.Properties;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.fail;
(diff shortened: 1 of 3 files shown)


marklogic-data-hub/src/test/java/com/marklogic/hub/HubTestBase.java
```diff
@@ -183,8 +184,10 @@ protected HubClient runAsDataHubOperator() {
     @Override
     protected HubClient runAsUser(String mlUsername, String mlPassword) {
         hubClient = null;
-
-        applyMlUsernameAndMlPassword(mlUsername, mlPassword);
+        Properties props = hubConfigInterceptor.getHubConfigObjectFactory().getGradleProperties();
+        props.setProperty("mlUsername", mlUsername);
```
rjrudin, line 188: I think it would be better to construct a new Properties object and copy everything from "props" into it, then override mlUsername/mlPassword. That ensures we don't modify the set of Properties that were actually read from disk. 

Good catch though figuring this out.

rjrudin (changes requested):

rjrudin (approved):

ryanjdew (approved):

ryanjdew (approved):

srinathgit:
If all gradle properties are set, it looks like we will have to reset mlManageUsername/password, mlSecurityUsername/password as well. Hence, I am setting the property that only matters in case of dhs tests.

rjrudin (approved):

PR #4912, opened 2020-11-25
DHFPROD-5623: added focus to switch view buttons in load tile

jyangML:
### Description
made radio button group that switches view in Load tile focusable and added outline when focused
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/src/components/load/switch-view.scss
+++ b/marklogic-data-hub-central/ui/src/components/load/switch-view.scss
@@ -13,3 +13,8 @@
 #switch-view .ant-radio-group-large .ant-radio-button-wrapper:hover {
     color: #7F86B5 ;
   }
+
+.radioGroupView:focus-within {
+    box-shadow: 0 0 1px 1px #7F86b5;
+    background-color: #7f86b5;
+}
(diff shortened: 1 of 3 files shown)


wooldridge (changes requested):
@jyangML, nice work. We also want a focused switch view component to be actionable when the user presses return. I've updated the AC on the ticket to clarify this, can we add that as well?

FYI, screencast of current functionality: https://drive.google.com/file/d/1UZy7PUEhRpNPBl6DkXIhg57Cqzmlrt90/view

jyangML:
> @jyangML, nice work. We also want a focused switch view component to be actionable when the user presses return. I've updated the AC on the ticket to clarify this, can we add that as well?
> 
> FYI, screencast of current functionality: https://drive.google.com/file/d/1UZy7PUEhRpNPBl6DkXIhg57Cqzmlrt90/view

added the functionality you requested, also added toggle with arrow keys.
added an extra unit test in this pr as well, because I realized I forgot.

wooldridge (approved):
@jyangML, looks great! 

I can now tab to the switch element to put it in focus.
When in focus, I can use the left and right keys to switch to the left and right buttons, respectively.
When in focus, I can use the return key toggle the switch.
Unit tests test the control of the switch via the keys when element is in focus.

briantang (approved):
Looks great, and tests pass!

PR #4916, opened 2020-11-30
DHFPROD-6172: Customize Spark connector endpoints for reading rows

anu3990:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/sql/sources/v2/reader/HubDataSourceReader.java
+++ b/marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/sql/sources/v2/reader/HubDataSourceReader.java
@@ -5,9 +5,12 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.marklogic.client.FailedRequestException;
 import com.marklogic.client.ext.helper.LoggingObject;
+import com.marklogic.client.io.InputStreamHandle;
 import com.marklogic.hub.HubClient;
+import com.marklogic.hub.HubClientConfig;
 import com.marklogic.hub.spark.dataservices.SparkService;
 import com.marklogic.hub.spark.sql.sources.v2.Util;
+import org.apache.commons.lang.StringUtils;
 import org.apache.spark.sql.SparkSession;
 import org.apache.spark.sql.catalyst.InternalRow;
 import org.apache.spark.sql.sources.v2.DataSourceOptions;
@@ -15,6 +18,7 @@ import org.apache.spark.sql.sources.v2.reader.DataSourceReader;
 import org.apache.spark.sql.sources.v2.reader.InputPartition;
 import org.apache.spark.sql.types.StructType;
 
+import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
@@ -29,6 +33,8 @@ public class HubDataSourceReader extends LoggingObject implements DataSourceRead
     private final Map<String, String> options;
     private final JsonNode initializationResponse;
     private final StructType sparkSchema;
+    private JsonNode endpointParams;
+    private boolean hasCustomApiPath;
 
     /**
      * The current default for partition count is based on the active SparkSession. The PartitionCountProvider is used
@@ -50,6 +56,7 @@ public class HubDataSourceReader extends LoggingObject implements DataSourceRead
      * @param options
      */
     private void validateOptions(Map<String, String> options) {
+        this.endpointParams = determineReadRowsEndpointParams(options);
         if (options.containsKey("serializedplan")) {
             Stream.of("view", "schema", "sqlcondition", "selectedcolumns").forEach(key -> {
                 if (options.containsKey(key)) {
@@ -59,7 +66,7 @@ public class HubDataSourceReader extends LoggingObject implements DataSourceRead
             return;
         }
 
-        if (!options.containsKey("view")) {
+        if (!hasCustomApiPath && !options.containsKey("view")) {
             throw new RuntimeException("The 'view' option must define a TDE view name from which to retrieve rows");
         }
     }
@@ -70,14 +77,24 @@ public class HubDataSourceReader extends LoggingObject implements DataSourceRead
      */
     private JsonNode initializeRead(Map<String, String> options) {
         ObjectNode inputs = buildInitializeReadInputs(options);
+        HubClientConfig hubClientConfig = Util.buildHubClientConfig(options);
+        HubClient hubClient = HubClient.withHubClientConfig(hubClientConfig);
+        InputStreamHandle initializeDefinition;
+
+        initializeDefinition = readCustomInitializeApiDefinition(options, hubClient);
+
         try {
+            if(options.containsKey("initializereadapipath")) {
+                return SparkService.on(
+                    hubClient.getFinalClient(), initializeDefinition
+                ).initializeRead(inputs);
+            }
             return SparkService.on(
-                HubClient.withHubClientConfig(Util.buildHubClientConfig(options)).getFinalClient()
+                hubClient.getFinalClient()
             ).initializeRead(inputs);
         } catch (FailedRequestException ex) {
             throw new RuntimeException("Unable to initialize read, cause: " + ex.getMessage(), ex);
         }
-
     }
 
     /**
@@ -126,13 +143,27 @@ public class HubDataSourceReader extends LoggingObject implements DataSourceRead
         return numPartitions;
     }
 
+    private InputStreamHandle readCustomInitializeApiDefinition(Map<String, String> options, HubClient hubClient) {
+        InputStreamHandle initializeDefinition = null;
+        String key = "initializereadapipath";
+        if (options.containsKey(key)) {
+            try {
+                initializeDefinition = hubClient.getModulesClient().newDocumentManager().read(options.get(key), new InputStreamHandle());
+            } catch (Exception ex) {
+                throw new RuntimeException("Unable to read custom API module for initializing Read job: " + options.get(key) + "; cause: " + ex.getMessage(), ex);
+            }
+        }
+
+        return initializeDefinition;
+    }
+
     @Override
     public List<InputPartition<InternalRow>> planInputPartitions() {
         List<InputPartition<InternalRow>> inputPartitions = new ArrayList<>();
         int partitionCount = initializationResponse.get("partitions").size();
         logger.info("Input partition count: " + partitionCount);
         for (int i = 0; i < partitionCount; i++) {
-            inputPartitions.add(new HubInputPartition(this.options, initializationResponse, i));
+            inputPartitions.add(new HubInputPartition(this.options, initializationResponse, i, this.endpointParams));
         }
         return inputPartitions;
     }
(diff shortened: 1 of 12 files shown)


marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/sql/sources/v2/reader/HubDataSourceReader.java
```diff
@@ -29,6 +33,11 @@
     private final Map<String, String> options;
     private final JsonNode initializationResponse;
     private final StructType sparkSchema;
+    private final HubClient hubClient;
```
rjrudin, line 36: I don't think most of these should be class fields, which complicates the class. See next note below.

marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/sql/sources/v2/reader/HubDataSourceReader.java
```diff
@@ -41,7 +50,12 @@ public HubDataSourceReader(DataSourceOptions dataSourceOptions) {
 
         this.options = dataSourceOptions.asMap();
         validateOptions(this.options);
+        this.hubClientConfig = Util.buildHubClientConfig(options);
+        this.hubClient = HubClient.withHubClientConfig(hubClientConfig);
+        this.objectMapper = new ObjectMapper();
+        this.customReadApiDefinitions = readCustomJobApiDefinitions(options);
 
+        this.endpointParams = determineReadRecordsEndpointParams(options);
```
rjrudin, line 58: I think this is all work to be done in initializeRead, which is already instantiating a HubClient. That can become a local variable in initializeRead, which can then say - if you've specified an initializereadapipath option, then I'll read that using the HubClient and pass it in as the second arg to the SparkService.on call.

marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/sql/sources/v2/reader/HubDataSourceReader.java
```diff
@@ -126,6 +145,20 @@ private int determineNumPartitions(Map<String, String> options) {
         return numPartitions;
     }
 
+    private CustomReadApiDefinitions readCustomJobApiDefinitions(Map<String, String> options) {
```
rjrudin, line 148: Because there's only one API endpoint, there's no need for a CustomReadApiDefinition class. CustomWriteApiDefinitions was created as a holder for two API definitions. We don't have that need here, so this method should just return a single InputStreamHandle. 

Also, should name this method "readCustomInitializeApiDefinition" - "Job" is not an appropriate name here.

marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/sql/sources/v2/reader/HubDataSourceReader.java
```diff
@@ -141,5 +174,49 @@ private int determineNumPartitions(Map<String, String> options) {
     public StructType readSchema() {
         return sparkSchema;
     }
+
+    private JsonNode determineReadRecordsEndpointParams(Map<String, String> options) {
```
rjrudin, line 178: Should name this "ReadRows" instead of "ReadRecords" for consistency.

marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/sql/sources/v2/reader/HubDataSourceReader.java
```diff
@@ -141,5 +174,49 @@ private int determineNumPartitions(Map<String, String> options) {
...
+        }
+
+        boolean doesNotHaveApiPath = (!endpointParams.hasNonNull("apiPath") || StringUtils.isEmpty(endpointParams.get("apiPath").asText()));
+        boolean hasEndpointConstants = endpointParams.hasNonNull("endpointConstants") && !StringUtils.isEmpty(endpointParams.get("endpointConstants").asText());
+        boolean hasEndpointState = endpointParams.hasNonNull("endpointState") && !StringUtils.isEmpty(endpointParams.get("endpointState").asText());
+        if (doesNotHaveApiPath && hasEndpointState) {
+            throw new IllegalArgumentException("Cannot set endpointState in readrowsendpointparams unless apiPath is defined as well.");
+        }
+        if(hasEndpointConstants)
+            throw new IllegalArgumentException("Cannot override endpointConstants.");
```
rjrudin, line 197: Let's give the same context as the error message above this, and I think some guidance is useful here - e.g. "Cannot set endpointConstants in readrowsendpointparams option; can only set apiPath and endpointState".

marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/sql/sources/v2/reader/HubDataSourceReader.java
```diff
@@ -141,5 +174,49 @@ private int determineNumPartitions(Map<String, String> options) {
...
+            }
+        } else {
+            endpointParams = objectMapper.createObjectNode();
+        }
+
+        boolean doesNotHaveApiPath = (!endpointParams.hasNonNull("apiPath") || StringUtils.isEmpty(endpointParams.get("apiPath").asText()));
+        boolean hasEndpointConstants = endpointParams.hasNonNull("endpointConstants") && !StringUtils.isEmpty(endpointParams.get("endpointConstants").asText());
+        boolean hasEndpointState = endpointParams.hasNonNull("endpointState") && !StringUtils.isEmpty(endpointParams.get("endpointState").asText());
+        if (doesNotHaveApiPath && hasEndpointState) {
+            throw new IllegalArgumentException("Cannot set endpointState in readrowsendpointparams unless apiPath is defined as well.");
```
rjrudin, line 194: Good error here, but add "option" after "readrowsendpointparams" to give it a little more context

marklogic-data-hub-spark-connector/src/test/java/com/marklogic/hub/spark/sql/sources/v2/Options.java
```diff
@@ -27,6 +27,7 @@
     private String initializeWriteApiPath;
     private String finalizeWriteApiPath;
     private JsonNode additionalExternalMetadata;
+    private String initializeReadApiPath;
```
rjrudin, line 30: "readRowsEndpointParams" should be here as well, and it needs to be tested too. A good way to test that, along with the custom initialize endpoint, is the following:

- Don't load any customers, nor the TDE
- Have the initialize endpoint return a single static partition, regardless of what the other inputs are
- Have the initialize endpoint return a static Spark schema as well, regardless of what the other inputs are
- For readrowsendpointparams/endpointState, define a single JSON row that can be returned by the readRows endpoint
- Then have the readRows endpoint return whatever value is in endpointState (it'll also need to use something in the endpointState object to know whether it's been called, since you only want to call it once)

That will allow you to verify that both your custom initialize and read endpoints are being used.

marklogic-data-hub-spark-connector/src/test/java/com/marklogic/hub/spark/sql/sources/v2/reader/ReadDataWithCustomEndpointsTest.java
```diff
@@ -0,0 +1,46 @@
...
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+
+public class ReadDataWithCustomEndpointsTest extends AbstractSparkReadTest {
+
+    @BeforeEach
+    void installCustomJobEndpoints() {
```
rjrudin, line 18: Replace "Job" with "Read"

marklogic-data-hub-spark-connector/src/test/java/com/marklogic/hub/spark/sql/sources/v2/reader/ReadDataWithCustomEndpointsTest.java
```diff
@@ -0,0 +1,46 @@
...
+        GenericDocumentManager mgr = getHubClient().getModulesClient().newDocumentManager();
+        DocumentMetadataHandle metadata = new DocumentMetadataHandle()
+            .withPermission("data-hub-operator", DocumentMetadataHandle.Capability.READ, DocumentMetadataHandle.Capability.UPDATE, DocumentMetadataHandle.Capability.EXECUTE);
+
+        String apiPath = "/custom-job-endpoints/initializeRead.api";
```
rjrudin, line 23: Let's put these in a folder called "/custom-read-endpoints" so it's clear that they're for testing the read endpoint.

marklogic-data-hub-spark-connector/src/test/resources/custom-job-endpoints/initializeRead.sjs
```diff
@@ -0,0 +1,121 @@
+/**
```
rjrudin, line 1: See my note above for testing - this should return a static schema and partition object. I believe parameterizedPlan can be "{}" because your custom read endpoint won't use it.

rjrudin (changes requested):

rjrudin:
@anu3990  I'm working on a test case for this now, I'll submit a PR into your branch.

rjrudin:
Submitted https://github.com/anu3990/marklogic-data-hub/pull/7 to provide a test that uses custom endpoints for both initializeRead and readRows.

marklogic-data-hub-spark-connector/src/test/java/com/marklogic/hub/spark/sql/sources/v2/reader/ReadDataWithCustomEndpointsTest.java
```diff
@@ -0,0 +1,78 @@
...
+        readrowsendpointparams.put("endpointConstants", "testValue");
+        Options options = newOptions().withView("Customer").withReadRowsEndpointparams(readrowsendpointparams);
+        IllegalArgumentException ex = assertThrows(IllegalArgumentException.class,
+            () -> new HubDataSourceReader(options.toDataSourceOptions()));
+        assertEquals("Cannot set endpointConstants in readrowsendpointparams option; can only set apiPath and endpointState.",
+            ex.getMessage());
+    }
+
+    @Test
+    public void testReadRowsEndpointparams() {
```
rjrudin, line 62: This test isn't verifying that the customReadRows.api endpoint is being used, because it's asserting that rows is empty. Rows is empty because no customers have been loaded. 

The test class I added - ReadWithCustomEndpointsTest - is failing because HubInputPartitionReader needs to look at the options to determine if it should use a custom API path or not. That test should suffice too for the happy path (you'll still want test methods for when endpointState is provided by not apiPath, and also for when endpointConstants is erroneously provided; I recommend moving both of those into ReadWithCustomEndpointsTest).

marklogic-data-hub-spark-connector/src/test/java/com/marklogic/hub/spark/sql/sources/v2/reader/ReadWithCustomEndpointsTest.java
```diff
@@ -0,0 +1,41 @@
...
+        ObjectNode testRow = endpointState.putObject("testRow");
+        testRow.put("myId", "12345");
+        testRow.put("myName", "This is a test row");
+
+        Options options = newOptions().withView("doesnt-matter")
+            .withInitializeReadApiPath("/custom-read-endpoints/initializeRead.api")
+            .withReadRowsApiPath("/custom-read-endpoints/customReadRows.api")
+            .withReadRowsEndpointState(endpointState);
+
+        List<InternalRow> rows = readRows(new HubDataSourceReader(options.toDataSourceOptions()));
```
rjrudin, line 33: Once HubInputPartitionReader is updated to use the custom endpoint, this test should pass.

marklogic-data-hub/src/main/resources/ml-modules/root/marklogic-data-hub-spark-connector/readRows.sjs
```diff
@@ -27,4 +42,6 @@ if (endpointState.batchNumber <= partition.batchCount) {
   endpointState.batchNumber = endpointState.batchNumber + 1;
 }
 
+console.log("results", xdmp.toJsonString(results));
```
rjrudin, line 45: Can you remove this line of logging? I may have accidentally included it in the PR I submitted.

marklogic-data-hub/src/main/resources/ml-modules/root/marklogic-data-hub-spark-connector/initializeRead.sjs
```diff
@@ -20,7 +20,7 @@ const readLib = require("readLib.sjs");
 
 /**
  * Build an Optic plan based on the user's inputs.
- * 
+ *
```
rjrudin, line 23: Undo this change just to avoid making any changes to this file (which complicates the version history of the file).

rjrudin (changes requested):

marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/sql/sources/v2/reader/HubDataSourceReader.java
```diff
@@ -29,6 +33,9 @@
     private final Map<String, String> options;
     private final JsonNode initializationResponse;
     private final StructType sparkSchema;
+    private InputStreamHandle initializeDefinition;
```
rjrudin, line 36: I like to minimize the number of class fields as much as possible to simplify a class; this one can be a local variable in the initializeRead method, it doesn't need to be a class field.

marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/sql/sources/v2/reader/HubInputPartition.java
```diff
@@ -16,15 +16,17 @@
     private Map<String, String> options;
     private JsonNode initializeResponse;
     private int partitionNumber;
+    private JsonNode endpointParams;
 
-    public HubInputPartition(Map<String, String> options, JsonNode initializeResponse, int partitionNumber) {
+    public HubInputPartition(JsonNode endpointParams, Map<String, String> options, JsonNode initializeResponse, int partitionNumber) {
```
rjrudin, line 21: Let's make this consistent with the HubInputPartitionReader constructor, where endpointParams is the last argument out of 4.

rjrudin (changes requested):
Looks good! Just a couple small tweaks to make

srinathgit (approved):

rjrudin (dismissed):

rjrudin (approved):

srinathgit (dismissed):

rjrudin (approved):

srinathgit (approved):

PR #4923, opened 2020-12-01
DHFPROD-6015:Study effect of server bugfix 54003

srinathgit:
### Description
This PR shouldn't be merged. It is created just to see the effects of server bugfix 54003 in datahub. This PR will be closed after all the tests run successfully.
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [ ] Added Tests
  

- ##### Reviewer:

- [ ] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mapping/default/lib.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mapping/default/lib.sjs
@@ -15,13 +15,13 @@ function getModel(targetEntity, version = '0.0.1') {
 }
 
 function getMapping(mappingName) {
-  return fn.head(cts.search(cts.andQuery([cts.collectionQuery('http://marklogic.com/data-hub/mappings'), cts.jsonPropertyValueQuery('name', mappingName, ['case-insensitive'])]), ["unfiltered", cts.indexOrder(cts.uriReference(), "descending")]));
+  return fn.head(cts.search(cts.andQuery([cts.collectionQuery('http://marklogic.com/data-hub/mappings'), cts.jsonPropertyValueQuery('name', mappingName, ['unstemmed', 'case-insensitive'])]), ["unfiltered", cts.indexOrder(cts.uriReference(), "descending")]));
 }
 
 function getMappingWithVersion(mappingName, version) {
   let cacheKey = `${mappingName}:${version}`;
   if (!cachedMappingByNameAndVersion[cacheKey]) {
-    cachedMappingByNameAndVersion[cacheKey] = fn.head(cts.search(cts.andQuery([cts.collectionQuery('http://marklogic.com/data-hub/mappings'), cts.jsonPropertyValueQuery('name', mappingName, ['case-insensitive']), cts.jsonPropertyValueQuery('version', version)])));
+    cachedMappingByNameAndVersion[cacheKey] = fn.head(cts.search(cts.andQuery([cts.collectionQuery('http://marklogic.com/data-hub/mappings'), cts.jsonPropertyValueQuery('name', mappingName, ['unstemmed','case-insensitive']), cts.jsonPropertyValueQuery('version', version, ['unstemmed'])])));
   }
   return cachedMappingByNameAndVersion[cacheKey];
 }
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/step.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/step.sjs
@@ -41,7 +41,7 @@ class Step {
   getStepNames(){
     let names = {};
     for(let type of this.getStepTypes()){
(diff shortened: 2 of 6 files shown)


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mapping/default/lib.sjs
```diff
@@ -15,13 +15,13 @@ function getModel(targetEntity, version = '0.0.1') {
 }
 
 function getMapping(mappingName) {
-  return fn.head(cts.search(cts.andQuery([cts.collectionQuery('http://marklogic.com/data-hub/mappings'), cts.jsonPropertyValueQuery('name', mappingName, ['case-insensitive'])]), ["unfiltered", cts.indexOrder(cts.uriReference(), "descending")]));
+  return fn.head(cts.search(cts.andQuery([cts.collectionQuery('http://marklogic.com/data-hub/mappings'), cts.jsonPropertyValueQuery('name', mappingName, ['unstemmed', 'case-insensitive'])]), ["unfiltered", cts.indexOrder(cts.uriReference(), "descending")]));
```
rjrudin, line 18: Related to this is that we have a lot of cts.search functions that perform filtered searches when they don't need to. I am thinking a convenience function would be handy here that defaults to unfiltered searches. And to handle the common use case of "Find me an artifact with this name in this collection", it would have a signature of:

    findArtifact(artifactName, artifactCollection)

That would likely be in artifacts/core.sjs since it's specific to artifacts.

hub-utils could have a more general search function - e.g. 

    hubUtils.findDocuments(myQuery, optionalIndexOrder)

That function would default to an unfiltered search.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/step.sjs
```diff
@@ -77,20 +77,20 @@ class Step {
   }
 
   getStepsByType(type = 'custom') {
-    let query = [cts.collectionQuery('http://marklogic.com/data-hub/step-definition'), cts.jsonPropertyValueQuery('type', type, 'case-insensitive')];
+    let query = [cts.collectionQuery('http://marklogic.com/data-hub/step-definition'), cts.jsonPropertyValueQuery('type', type, ["unstemmed","case-insensitive"])];
```
rjrudin, line 80: Any idea why some of these artifact queries are case-insensitive and others are not? This makes it seem that DHF is not sure if case sensitivity matters or not when finding a step. 

I am guessing that all artifact queries should be case-insensitive on "name" because the name is used in the filename. And it also doesn't seem beneficial to the user to allow for separate steps of "myIngester" and "myingester".

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/step.sjs
```diff
@@ -77,20 +77,20 @@ class Step {
...
     return cts.search(cts.andQuery(query)).toArray();
   }
 
   getStepByNameAndType(name, type = 'custom') {
-    let query = [cts.collectionQuery('http://marklogic.com/data-hub/step-definition'), cts.jsonPropertyValueQuery('name', name, 'case-insensitive'), cts.jsonPropertyValueQuery('type', type, 'case-insensitive')];
+    let query = [cts.collectionQuery('http://marklogic.com/data-hub/step-definition'), cts.jsonPropertyValueQuery('name', name, ["unstemmed","case-insensitive"]), cts.jsonPropertyValueQuery('type', type, ["unstemmed","case-insensitive"])];
```
rjrudin, line 85: To support a query like this that goes beyond just artifact name, I think we should have another helper function in artifacts/core:

    artifactPropertyValueQuery(name, value)

And that can default to unstemmed / case-insensitive. 

The goal is to limit the number of places where we're coding "unstemmed" as a query option. This query would then become:

    hubUtils.findDocuments(cts.andQuery([artifactPropertyValueQuery("name", name), artifactPropertyValueQuery("type", type)]));

marklogic-data-hub/src/test/java/com/marklogic/bootstrap/TestAppInstaller.java
```diff
@@ -203,6 +204,9 @@ public void run() {
 
             HubTestBase.applyDatabasePropertiesForTests(hubConfig);
 
+            Group group = new Group(api, "Default");
```
rjrudin, line 207: This would benefit from a comment to explain why this is being enabled.

srinathgit: I will add a comment

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/step.sjs
srinathgit: I am not sure why, also this module step.sjs deals mostly with crud operation of step definition and not steps.

srinathgit:
Closing out this PR. All tests ran successfully with the server bugfix applied. All these changes will be part of a story in the future.

PR #4925, opened 2020-12-01
DHFPROD-6237: Can now read arrays and maps with Spark connector

rjrudin:
### Description

Reworked HubInputPartitionReader to use Scala's JacksonParser to read a string of JSON into an InternalRow; resulted in a lot of code being deleted from HubInputPartitionReader. 

Made a couple tweaks to ReadAllDataTypesTest too:

- The timestamp is being returned as a long by JacksonParser, so verifying the correct long is returned
- JacksonParser is not able to convert a hexBinary into a byte array, so readLib was updated to associate hexBinary with string

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


rjrudin:
@ehennum  Including you as a reviewer in case you wanted to verify the use of JacksonParser; thanks for pointing that out.

I believe we're now in the state you had recommended, which is that our Java code - specifically HubInputPartitionReader - has very little knowledge of how to convert a string of JSON into an InternalRow; JacksonParser is handling that. The logic of what Spark schema to build based on a TDE is in readLib.sjs.

srinathgit (approved):

anu3990 (approved):

PR #4936, opened 2020-12-03
DHFPROD-6311: Fix for intermittently failing GetRecordTest

rahulvudutala:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/entities/entity-search-lib.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/entities/entity-search-lib.sjs
@@ -537,12 +537,10 @@ function isHubEntityInstance(docUri) {
   return isHubEntityInstance;
 }
 
-// Adding console logs to debug the issue DHFPROD-6311. Console logs will be removed after the issue is resolved
 function getRecordHistory(docUri) {
   const history = [];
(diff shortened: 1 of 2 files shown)


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/entities/entity-search-lib.sjs
```diff
@@ -586,20 +582,13 @@ function findFlowAndStepNameFromProvenanceRecords(provenanceRecord, stepNames, s
       associatedWith[associatedWith.length-1] = artifactTobeRemoved;
       break;
     }
-    associatedWith.pop();
   }
-  console.log("associatedWith array after finding flowName: " + associatedWith);
+  associatedWith.pop();
```
rjrudin, line 586: Is this the fix then - that the pop() call was occurring within the for loop?

rahulvudutala: If the associatedWith was something like this ["inline", "map", "entity-services-mapping"], then the test runs fine.
The test fails when the associatedWith is something like this ["entity-services-mapping", "inline", "map"].
The pop should happen after breaking out of the for loop.

rjrudin (approved):

ryanjdew (approved):

PR #4938, opened 2020-12-04
DHFPROD-5243: Merge rule strategies select dropdown

ryanjdew:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/src/components/entities/merging/add-merge-rule/add-merge-rule-dialog.tsx
+++ b/marklogic-data-hub-central/ui/src/components/entities/merging/add-merge-rule/add-merge-rule-dialog.tsx
@@ -41,6 +41,7 @@ const AddMergeRuleDialog: React.FC<Props> = (props) => {
         handleError
     } = useContext(UserContext);
     const [entityTypeDefinition, setEntityTypeDefinition] = useState<Definition>(DEFAULT_ENTITY_DEFINITION);
+    const [mergeStrategyNames, setMergeStrategyNames] = useState<string[]>([]);
     const [property, setProperty] = useState<string | undefined>(undefined);
     const [propertyTouched, setPropertyTouched] = useState(false);
     const [propertyErrorMessage, setPropertyErrorMessage] = useState('');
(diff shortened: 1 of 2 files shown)


Sanjeevani19 (commented):
@ryanjdew Can we add two merge rules with same property name and strategy name ?
https://drive.google.com/file/d/1Ncmyx_Pq5eDtMnihBgcMnFIYQt4DM4eb/view

marklogic-data-hub-central/ui/src/components/entities/merging/add-merge-rule/add-merge-rule-dialog.tsx
```diff
@@ -418,16 +432,17 @@ const AddMergeRuleDialog: React.FC<Props> = (props) => {
                                 validateStatus={(strategyValue || !strategyValueTouched) ? '' : 'error'}
                                 help={(strategyValue || !strategyValueTouched) ? '' : 'Strategy Name is required'}
                             >
-                                <MLInput
+                                <MLSelect
                                     id="strategyName"
-                                    placeholder="Enter strategy name"
+                                    placeholder="Select strategy name"
                                     size="default"
                                     value={strategyValue}
```
xnikhil08, line 439: This placeholder will never be visible for this field because the default value here is empty or null. It should be **undefined** instead of empty string.

xnikhil08 (changes requested):
The placeholder "select "

Sanjeevani19 (changes requested):
@ryanjdew looks like even before I save the form, strategy name error message is being shown when strategy merge type is selected, instead it should show up only when it is submitted with an empty option.Please look into it.

https://drive.google.com/file/d/1QEOHO9Pg9MujBBqj43TMnBMUjddulP-j/view

Sanjeevani19 (approved):
Looks good, works as expected.
@ryanjdew Please update the story with all these scenarios when you get a chance.

xnikhil08 (approved):

PR #4948, opened 2020-12-07
DHFPROD-6323: Fix zip match function for HC step

ryanjdew:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/smart-mastering/entity-driven/test-data/steps/matching/matchCustomers.step.json
+++ b/marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/smart-mastering/entity-driven/test-data/steps/matching/matchCustomers.step.json
@@ -62,7 +62,7 @@
         },
         {
           "entityPropertyPath": "billing.zip.fiveDigit",
-          "matchType": "exact"
+          "matchType": "zip"
         }
       ]
     },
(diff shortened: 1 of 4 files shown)


marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/smart-mastering/entity-driven/test-data/steps/matching/matchCustomers.step.json
```diff
@@ -76,7 +76,7 @@
         },
         {
           "entityPropertyPath": "shipping.zip.fiveDigit",
-          "matchType": "exact"
+          "matchType": "zip"
```
rjrudin, line 79: I verified this works locally, but the challenge for me is understanding how I know that zip match works from this test. Because there are a lot of rulesets and entities in play, it takes a lot of work to understand why something matches and why something doesn't match. 

I'm approving this as the bug is fixed, but I'm going to work on a test that isolates the problem as much as possible - I'm thinking something like 2 persisted documents, and then an input doc that matches one but not the other. The entity model will have a single property, and the step config will have a single ruleset using the zip match. I'll see what I come up with.

rjrudin (approved):

fsnow (approved):

PR #4950, opened 2020-12-07
DHFPROD-6303: Download a record or artifact from All Data View

rahulvudutala:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/controllers/EntitySearchController.java
```diff
@@ -54,7 +57,22 @@ public String search(@RequestBody SearchQuery searchQuery, @RequestParam(default
...
+
+    @RequestMapping(value = "/downloadRecord", method = RequestMethod.GET)
+    public void downloadRecord(@RequestParam String docUri, @RequestParam(defaultValue = "final") String database, HttpServletResponse response) {
+        String[] docUriArray = docUri.split("/");
+        String fileName = docUriArray[docUriArray.length-1];
+        response.setContentType(MediaType.APPLICATION_OCTET_STREAM_VALUE);
+        response.addHeader("Content-Disposition", String.format("attachment; filename=%s", fileName));
+
+        try (OutputStream out = response.getOutputStream()) {
+            newEntitySearchManager(database).downloadRecord(docUri, out);
```
rjrudin, line 71: I think this can be simpler and wholly implemented in the controller by using a DocumentManager from a DatabaseClient. Just read the URI as an InputStreamHandle, then use Spring's FileCopyUtils class to copy that class's InputStream to the HTTP response's OutputStream. You can remove all the code from the Manager (which doesn't really add any value here), and only have one try/catch too.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/entitySearch/getRecord.api
```diff
@@ -1,10 +1,17 @@
 {
   "functionName" : "getRecord",
-  "params" : [ {
+  "params" : [
```
rjrudin, line 3: Can undo this, assuming just using DatabaseClient.newDocumentManager works.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/entities/entity-search-lib.sjs
```diff
@@ -637,12 +638,30 @@ function findFlowsAsMap() {
   return flows;
 }
 
+function getRecordSize(docUri) {
```
rjrudin, line 641: Should be "getDocumentSize", since "record" refers to the abstraction containing the URI, the document, metadata, etc - we're only interested in the document size.

marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/entities/search/EntitySearchManagerTest.java
```diff
@@ -274,6 +279,39 @@ void testGetQueryOptions() {
         assertThrows(RuntimeException.class, () -> new EntitySearchManager(getHubClient()).getQueryOptions("non-existent-options"), "Search options doesn't exist");
     }
 
+    @Test
+    void testDownloadRecord() {
```
rjrudin, line 283: Just test the controller instead via an AbstractMvcTest - that gives us coverage over the controller as well, which is missing right now.

Should test downloading a binary as well - that could be as simple as capturing the bytes of the string "hello world" and storing that as a document. Can assert on its size too.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/entities/entity-search-lib.sjs
```diff
@@ -637,12 +638,30 @@ function findFlowsAsMap() {
...
+
+  if(nodeKind === 'binary') {
+    bytes = xdmp.binarySize(fn.head(doc).root);
+  } else {
+    bytes = xdmp.binarySize(fn.head(xdmp.unquote(xdmp.quote(doc), null, "format-binary")).root)
+  }
+
+  let power = Math.floor(Math.log(bytes) / Math.log(1024));
+  power = power > 2 ? 2 : power;
+  return (bytes / Math.pow(1024, power)).toFixed(2) * 1 + ' ' + sizes[power];
```
rjrudin, line 655: I didn't see any tests for record size? I recommend doing some unit tests here to verify you're getting the correct result. And instead of passing in a URI, pass in the document itself - that makes it easier to test, as you don't have to persist a document first. 

Should test documents that result in each of the 3 sizes being used too.

marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/entities/search/EntitySearchManagerTest.java
rjrudin: Also, no need to install the reference project and create customer entities. The "download" feature doesn't care about whether the doc is an entity or not - it just takes a URI as an input and returns the document in the HTTP response output stream. 

For testing then, I'd have 5 separate tests - one a JSON doc, one for XML, one for Text, one for Binary, and one for a missing URI (verify you get an appropriate error message).

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/controllers/EntitySearchController.java
```diff
@@ -54,7 +57,22 @@ public String search(@RequestBody SearchQuery searchQuery, @RequestParam(default
...
+            response.flushBuffer();
+        } catch (IOException e) {
+            throw new RuntimeException(String.format("Unable to download record with docUri: %s with cause: %s", docUri, e.getMessage()));
```
rjrudin, line 74: Use "URI" in the error message ; "docUri" is a variable name that is not as meaningful to a user like Pari as "URI" is.

rjrudin (changes requested):

timur-isangulov (changes requested):
Hi @rahulvudutala, the hubMetadata object is empty for the *.step.json files. Can you please look into it.

rjrudin:
@timur-isangulov  @rahulvudutala  That's correct, as *.step.json documents won't have any metadata keys. The user may ingest all kinds of documents outside of DHF as well that do not have metadata keys. So we should support hubMetadata being empty.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/entities/entity-search-lib.sjs
```diff
@@ -494,6 +494,7 @@ function addDocumentMetadataToSearchResults(searchResponse) {
       hubMetadata["lastProcessedByStep"] = documentMetadata.datahubCreatedByStep;
       hubMetadata["lastProcessedDateTime"] = documentMetadata.datahubCreatedOn;
       hubMetadata["sources"] = getEntitySources(docUri);
+      hubMetadata["recordSize"] = getRecordSize(docUri);
```
timur-isangulov, line 497: Can you return recordSize as a two parameters, recordSize and recordSizeUnit ?

timur-isangulov:
> @timur-isangulov @rahulvudutala That's correct, as *.step.json documents won't have any metadata keys. The user may ingest all kinds of documents outside of DHF as well that do not have metadata keys. So we should support hubMetadata being empty.

There will be no file size returned for the step documents ? If that a case we will allow to download these files but won't show their file sizes ?

rjrudin (changes requested):
I'm trying this out locally - I'd really like to see if it's possible to avoid reading the data into memory in the middle tier, which can be a big memory hit for a large document. So seeing if there's a way to go from InputStream to OutputStream and also pretty-print the JSON/XML.

rjrudin:
Created https://github.com/rahulvudutala/marklogic-data-hub/pull/3 - note that I changed the controller to start the move from this being entity-search-specific to record-generic. The middle tier endpoint is then /api/record/download . 

I didn't test this in HC yet, but everything worked via curl - i.e. JSON and XML docs are both being pretty-printed. And this avoids ever reading the document into memory in the middle tier.

timur-isangulov (changes requested):
@rahulvudutala can you please add documentSize object to the response of "api/entitySearch?docUri=" api. According to the UX design we need to display the size of the file on the Detail view, as well.

ryanjdew (approved):

rjrudin (dismissed):

timur-isangulov (approved):

timur-isangulov (approved):

ryanjdew (approved):

rjrudin (approved):

PR #4956, opened 2020-12-08
DHFPROD-6319: Use targetEntityType in new match/merge steps

ryanjdew:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/com.marklogic.smart-mastering/survivorship/merging/options.xqy
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/com.marklogic.smart-mastering/survivorship/merging/options.xqy
@@ -634,9 +634,10 @@ declare function merge-impl:compile-merge-options(
         map:map()
       else ()
     (: Gather target Entity Type information :)
-    let $target-entity := $merge-options/(*:target-entity|targetEntity|targetEntityType) ! fn:string(.)
-    let $target-entity-type-def := es-helper:get-entity-def($target-entity)
-    let $target-entity-type := $target-entity-type-def/entityIRI ! fn:string(.)
+    let $target-entity-type-info := util-impl:get-entity-type-information($merge-options)
+    let $target-entity := $target-entity-type-info => map:get("targetEntityType")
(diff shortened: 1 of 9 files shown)


marklogic-data-hub/src/main/resources/ml-modules/root/com.marklogic.smart-mastering/survivorship/merging/options.xqy
```diff
@@ -632,7 +632,7 @@ declare function merge-impl:compile-merge-options(
         map:map()
       else ()
     (: Gather target Entity Type information :)
-    let $target-entity := $merge-options/(*:target-entity|targetEntity|targetEntityType) ! fn:string(.)
+    let $target-entity := fn:head(($merge-options/targetEntityType,$merge-options/(*:target-entity|targetEntity))) ! fn:string(.)
```
rjrudin, line 635: Is there a common library module that would make sense for this function to live in so that both options-impl and options.xqy can use it? Like the util.xqy module in com.marklogic.smart-mastering?

rjrudin (changes requested):
I think this needs a test to verify that an unconverted matching step that has "targetEntity" still works, as that's the bug I ran into here. That can be a super simple test - i.e. just two documents, an entity model with one property, and a single exact match option on that property. The match options would use "targetEntity" instead of "targetEntityType" to verify backwards compatibility with existing 5.x apps that use "targetEntity".

marklogic-data-hub/src/main/resources/ml-modules/root/com.marklogic.smart-mastering/survivorship/merging/options.xqy
ryanjdew: This has been moved into the util.xqy and a test has been added for that function

srinathgit (commented):
Will inline match/merge steps that use "targetEntity" continue to run fine ?

rjrudin (approved):

ryanjdew:
@srinathgit Yes, inline match/merge steps that use "targetEntity" continue to run as they did before.

srinathgit (approved):

PR #4962, opened 2020-12-08
DHFPROD-4947:Provide info when main.sjs is not found or user doesn't required permissions

srinathgit:
### Description
Provide info when main.sjs is not found or user doesn't required permissions
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/step.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/step.sjs
@@ -116,7 +116,10 @@ class Step {
     try {
       stepModule = this.hubUtils.retrieveModuleLibrary(moduleUri);
     } catch (e) {
-      httpUtils.throwBadRequest(`Unable to access module: ${moduleUri}. Verify that this module is in your modules database and that your user account has a role that grants read permission to this module.`);
+      if(e.stack && e.stack.includes("XDMP-MODNOTFOUND")){
+        httpUtils.throwBadRequest(`Unable to access module: ${moduleUri}. Verify that this module is in your modules database and that your user account has a role that grants read and execute permission to this module`);
+      }
+      httpUtils.throwBadRequest(`Unable to run module: ${moduleUri}; cause: ${e.stack}`);
     }
     if (this.performance.performanceMetricsOn())  {
       return this.performance.instrumentStep(stepModule, stepModule[funcName], flow.globalContext.jobId, flow.globalContext.batchId, flow.globalContext.flow.name, moduleUri, flow.globalContext.uri);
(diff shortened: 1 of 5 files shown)


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/step.sjs
```diff
@@ -116,7 +116,7 @@ class Step {
     try {
       stepModule = this.hubUtils.retrieveModuleLibrary(moduleUri);
     } catch (e) {
-      httpUtils.throwBadRequest(`Unable to access module: ${moduleUri}. Verify that this module is in your modules database and that your user account has a role that grants read permission to this module.`);
+      httpUtils.throwBadRequest(`Unable to access module: ${moduleUri}. Verify that this module is in your modules database and that your user account has a role that grants read and execute permission to this module`);
```
rjrudin, line 119: @srinathgit  Can you test one more thing - see what happens if the step module is present but has a compilation error in it. I think we may get the same error, in which case we really ought to include "e.stack" at the end of the message - e.g. "; cause: ${e.stack}".

rjrudin (changes requested):

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/step.sjs
srinathgit: We cannot have the same message if there is compilation error right ? I have changed the message to 
```
`Cannot run module: ${moduleUri}; cause: ${e.stack}`
```
Let me know if that's ok.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/step.sjs
```diff
@@ -116,7 +116,10 @@ class Step {
     try {
       stepModule = this.hubUtils.retrieveModuleLibrary(moduleUri);
     } catch (e) {
-      httpUtils.throwBadRequest(`Unable to access module: ${moduleUri}. Verify that this module is in your modules database and that your user account has a role that grants read permission to this module.`);
+      if(e.stack && e.stack.includes("XDMP-MODNOTFOUND")){
```
rjrudin, line 119: That looks good to me.

rjrudin (dismissed):

ryanjdew:
@srinathgit It looks like the expected error messages for a couple of unit tests elsewhere have failed. 

com.marklogic.hub.flow.impl.FlowRunnerTest#customStepReferencesModulePathThatDoesntExist
marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/impl/step/makeFunction.sjs

srinathgit:
> @srinathgit It looks like the expected error messages for a couple of unit tests elsewhere have failed.
> 
> com.marklogic.hub.flow.impl.FlowRunnerTest#customStepReferencesModulePathThatDoesntExist
> marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/impl/step/makeFunction.sjs

Thanks Ryan. I think latest commit should fix them.

akshaysonvane (approved):

rjrudin (approved):

PR #4963, opened 2020-12-08
DHFPROD-6032: Incorrect icon for JSON source

saarimrahman:
### Description
- no tests as this deals with updating existing icons.
- also removes some unused variables to get eslint to pass.

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [NA] Added Tests
  

- ##### Reviewer:

- [NA] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/src/components/entities/merging/add-merge-rule/add-merge-rule-dialog.tsx
+++ b/marklogic-data-hub-central/ui/src/components/entities/merging/add-merge-rule/add-merge-rule-dialog.tsx
@@ -14,7 +14,6 @@ import {CurationContext} from "../../../../util/curation-context";
 import arrayIcon from "../../../../assets/icon_array.png";
 import {MergeRuleTooltips} from "../../../../config/tooltips.config";
 import MultiSlider from "../../matching/multi-slider/multi-slider";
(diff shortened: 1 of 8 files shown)


wooldridge (changes requested):
I think we might have missed one more place to update the icon. In Explore, I select the "Staging" database and "All Data" from the drop down. In the card results view, I click through to details for a JSON doc. I see a Record tab with the XML-like icon. Should this tab have a JSON-like icon for JSON docs? Also, should the "Record" label be "XML"/"JSON" instead? Or are we leaving it ambiguous on purpose?

My screencast shows this at the end: https://drive.google.com/file/d/11yGHI0ODI6PqGSjra9UeENwwyiOBhJWg/view

FYI: @rahulvudutala @yasht2010

wooldridge:
@rahulvudutala said the following so let's address the "Record" tab in a separate issue:

The endpoint returns the datatype of the record being returned to be rendered in the detail view.
We can do couple of things:
1. We can have a generic icon of record/document in the detail view.
2. We also have text and binary types. So we need have different icons when rendering different type of records/documents

wooldridge (approved):

Sanjeevani19 (approved):

PR #4964, opened 2020-12-08
DHFPROD-6216: Optimize match summary range index

ryanjdew:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- N/A Added Tests
  

- ##### Reviewer:

-  N/A Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/impl/HubProjectImpl.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/impl/HubProjectImpl.java
@@ -507,7 +507,7 @@ public class HubProjectImpl extends LoggingObject implements HubProject {
             XPathFactory xPathFactory = XPathFactory.newInstance();
             XPath xPath = xPathFactory.newXPath();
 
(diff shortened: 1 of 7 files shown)


fsnow (approved):

rjrudin (changes requested):
I need to research this further, as we have DHFPROD-6271 open to undo a temporary change I made so that final-database.xml is updated during hubUpdate. I think this ticket will require a little more surgery - i.e. "//actionDetails" needs to be removed if it exists, and then "/matchSummary/actionDetails" needs to be added if it doesn't exist. But I need to dig a bit deeper to remember what's going on with these files under ml-config.

ryanjdew:
@rjrudin After doing a bootstrap on my instance I only have the "/matchSummary/actionDetails" index, but I'm fine with parking this in draft state for a bit.

rjrudin:
@ryanjdew  That's due to the hack I put in to fix a bug on the develop branch (the hack is that final-database.xml is overwritten if it exists); I opened up 6271 to undo the hack and surgically update final-database.xml. 

I think we may need a meeting though to consider the pros and cons of having DHF basically say - I own the definition of the files under ml-config/database-fields. That would make this issue go away.

rjrudin (approved):

PR #4967, opened 2020-12-09
DHFPROD-6231:Mapping step fails when provenance is set to "fine"

srinathgit:
### Description
Mapping step fails when provenance is set to "fine"
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/test/resources/test-projects/provenance-test/flows/inline.flow.json
+++ b/marklogic-data-hub/src/test/resources/test-projects/provenance-test/flows/inline.flow.json
@@ -29,6 +29,7 @@
           "name": "SimpleCustomerMapping",
           "version": 1
         },
+        "provenanceGranularityLevel": "fine",
         "sourceDatabase": "data-hub-FINAL",
         "collections": [
           "mastering-flow-mapping-json"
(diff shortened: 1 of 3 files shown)


marklogic-data-hub/src/test/resources/test-projects/provenance-test/flows/inline.flow.json
```diff
@@ -29,6 +29,7 @@
           "name": "SimpleCustomerMapping",
           "version": 1
         },
+        "provenanceGranularityLevel": "fine",
```
rjrudin, line 32: I think it's worth adding an assertion to GetRecordTest (it's tangential to the test, but I think it's fine to stuff it in there) verifying that there's no error in stepOutput, and the test should verify that prov is set to "fine" in the step config as well. The assertion messages can state that we expect the step to not fail, even though fine is not supported for a mapping step.

rjrudin (changes requested):

rjrudin (approved):

akshaysonvane (approved):

ryanjdew (dismissed):

rjrudin (approved):

akshaysonvane (approved):

PR #4968, opened 2020-12-09
DHFPROD-6279: Adjust placement of Match/Merge settings tooltips

wooldridge:
Apply same "right" placement as used in other step settings forms. Also update wrapper CSS class name to match component name.

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] JIRA_ID included in all the commit messages
- [x] PR title is in the format JIRA_ID:Title
- [x] Rebase the branch with upstream
- [x] Squashed all commits into a single commit
- N/A Added Tests
  

- ##### Reviewer:

- N/A Reviewed Tests

PR #4970, opened 2020-12-09
DHFPROD-6309: Add a tooltip for a disabled name field on edit step and edit flow

devinnicholson:
…d edit flow

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/src/components/entities/create-edit-step/create-edit-step.tsx
+++ b/marklogic-data-hub-central/ui/src/components/entities/create-edit-step/create-edit-step.tsx
@@ -418,17 +418,28 @@ const CreateEditStep: React.FC<Props>  = (props) => {
         validateStatus={(stepName || !isStepNameTouched) ? "" : "error"}
         help={(stepName || !isStepNameTouched) ? "" : "Name is required"}
         >
-          <Input
+          { tobeDisabled?<MLTooltip title={NewMatchTooltips.nameField} placement={'bottom'}> <Input
             id="name"
             placeholder="Enter name"
             value={stepName}
             onChange={handleChange}
             disabled={tobeDisabled}
             className={styles.input}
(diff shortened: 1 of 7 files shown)


wooldridge (changes requested):
Nice work, @devinnicholson! There's a minor adjustment to be made to the name and description "?" icon tooltips. Because of idiosyncrasies with the Ant Design and tile components, we need to use placement "right" for the tooltip alignment to work. That's what I did in the following PR:

https://github.com/marklogic/marklogic-data-hub/pull/4968

Can you do the same for the match/merge name and description icons? Looks like those are the only ones that still have a problem, e.g.:

Screen Shot 2020-12-09 at 2 52 15 PM

wooldridge (changes requested):
Test needed a little more work. Need to test with create-edit-load dialog in editing state. (First test has the dialog in new state due to an empty stepData property in the test data.) I've created a PR to your branch with a working test here:

https://github.com/devinnicholson/marklogic-data-hub/pull/1

wooldridge (approved):
Looks good, tests pass!

briantang (changes requested):
Looks good, but I think the three commits may need to be squashed into one to keep consistent with PR protocols. Can you look into squashing those commits? @devinnicholson

briantang (dismissed):

wooldridge (changes requested):
@devinnicholson, looks like we've reverted to the older failing test? Updating the test file so that it has the changes here should fix:

https://github.com/devinnicholson/marklogic-data-hub/pull/1/files

Then squash and force-push.

wooldridge (changes requested):

wooldridge (approved):

briantang (approved):

PR #4978, opened 2020-12-10
DHFPROD-6322:Wrong artifact returned when artifact name appears in another artifact

srinathgit:
### Description
Wrong artifact returned when artifact name appears in another artifact
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


srinathgit:
@rjrudin 
In order to get the uri of step definition, we need the artifact (see getDirectory() method in stepDefinition.sjs). So, I haven't included the fix for step defs. Let me know if that's ok.

rjrudin (changes requested):
@srinathgit  For step definitions, let's take the slightly less efficient approach - but guarantees correctness - of not calling fn.head on the results of cts.search, but rather iterating over each result and taking the first one that has a root-level "name" property with the given value. That ensures we never returned the wrong one, which could happen if a step definition has "name" in its options.

srinathgit:
@rjrudin 
For running flows, we obtain step definition from impl/step.sjs. That code also has the same issue. I made the change there as well and renamed step.sjs to stepDefinition.sjs (as the module is concerned only with step defs). Also removed couple of unused methods from it .

rjrudin (approved):
Good call on renaming step.sjs

ryanjdew (approved):

PR #4989, opened 2020-12-11
DHFPROD-6219:Remove '$ref' from empty entity instance

srinathgit:
### Description
Remove '$ref' from empty entity instance
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mapping/entity-services/xquery-lib.xqy
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mapping/entity-services/xquery-lib.xqy
@@ -19,6 +19,7 @@ xquery version "1.0-ml";
 module namespace xquery-lib = "http://marklogic.com/mapping/es/xquery";
 
 import module namespace inst="http://marklogic.com/entity-services-instance" at "/MarkLogic/entity-services/entity-services-instance.xqy";
+import module namespace json = "http://marklogic.com/xdmp/json" at "/MarkLogic/json/json.xqy";
 
 declare function document-with-nodes($nodes as node()*) {
   document {
@@ -64,5 +65,28 @@ declare function data-hub-map-to-canonical(
   return
     if ($format="xml")
     then inst:canonical-xml($results)
-    else inst:canonical-json($results)
+    else(
+      let $element := $results/element()
+      return
+      (: This change is necessitated by DHFPROD-6219. In case the entity doesn't have strucutred properties, it
+         returns empty json object. If it has structured properties, the empty structured properties are removed.
+         inst:canonical-json() method introduces the '$ref'.  :)
+      if (empty($element/*)) then
+        document{json:object()=>map:with(string(fn:node-name($element)), json:object())}
+      else
+        inst:canonical-json(xquery-lib:remove-empty-structured-properties($element))
+    )
+};
+
+
+declare function remove-empty-structured-properties($element as item()*) as document-node() {
+  document{
+    element {fn:node-name($element)} {
+      $element/namespace::node(),
+      $element/@*,
+      for $child in $element/element()
+      where not($child/element() instance of element() and empty($child/element()/element()))
+      return $child
+    }
+  }
 };
--- a/marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/builtins/steps/mapping/entity-services/runEmptyMapping.sjs
+++ b/marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/builtins/steps/mapping/entity-services/runEmptyMapping.sjs
@@ -1,0 +1,37 @@
+const mapping = require("/data-hub/5/builtins/steps/mapping/entity-services/main.sjs");
+const test = require("/test/test-helper.xqy");
+
+let assertions = [];
+
+//Only 'id' is mapped. Other properties shouldn't be present in entity instance
+function runStructuredPropertyMapping () {
+  let doc = fn.head(xdmp.unquote(`{"id":1,"customer":{"firstName":"Oralia"},"items":[{"name":"Voltsillam"},{"name":"Latlux"}]}`));
+  let result = mapping.main({uri: '/test.json', value: doc}, {
+    mapping: {name: 'OrdersMapping', version: 4},
+    outputFormat: 'json'
+  }).value.root;
+  let instance = result.envelope.instance;
+  assertions = assertions.concat([
+    test.assertEqual(null, instance.OrderType.customer),
+    test.assertEqual(null, instance.OrderType.items),
+    test.assertEqual("1", fn.string(instance.OrderType.id))
+  ]);
+}
+
+//None of the entity properties are mapped, so 'CustomerType' shouldn't be present in entity instance
+function runSimplePropertyMapping() {
+  let doc = fn.head(xdmp.unquote(`{"id":1}`));
+  let result = mapping.main({uri: '/test.json', value: doc}, {
+    mapping: {name: 'CustomerMapping', version: 1},
+    outputFormat: 'json'
+  }).value.root;
+  let instance = result.envelope.instance;
+  assertions = assertions.concat([
+    test.assertEqual(null, instance.CustomerType)
+  ]);
+}
+
+runStructuredPropertyMapping();
+runSimplePropertyMapping();
+
+assertions;
--- a/marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/builtins/steps/mapping/entity-services/test-data/mappings/CustomerMapping/CustomerMapping-1.mapping.json
+++ b/marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/builtins/steps/mapping/entity-services/test-data/mappings/CustomerMapping/CustomerMapping-1.mapping.json
@@ -1,0 +1,15 @@
+{
+  "lang": "zxx",
+  "name": "CustomerMapping",
+  "description": "This is intentionally mapping with invalid source properties",
+  "version": 1,
+  "targetEntityType": "http://marklogic.com/data-hub/example/CustomerType-0.0.1/CustomerType",
+  "properties": {
+    "id": {
+      "sourcedFrom": "missingProp1"
+    },
+    "firstname": {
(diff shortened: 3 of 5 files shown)


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mapping/entity-services/xquery-lib.xqy
```diff
@@ -61,8 +62,34 @@ declare function data-hub-map-to-canonical(
...
+};
+
+
+declare function clean-xml($element as item()*) as document-node() {
+  document{
+    element {fn:node-name($element)} {
+      $element/namespace::node(),
+      $element/@*,
+      for $child in $element/element()
+      return
```
ryanjdew, line 89: Since the if condition returns an empty sequence, this could be changed to a where clause with a not().

srinathgit: Added a where clause with a not()

ryanjdew (approved):

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mapping/entity-services/xquery-lib.xqy
```diff
@@ -64,5 +65,28 @@ declare function data-hub-map-to-canonical(
...
+         inst:canonical-json() method introduces the '$ref' .  :)
+      if (empty($element/*)) then
+        document{json:object()=>map:with(string(fn:node-name($element)), json:object())}
+      else
+        inst:canonical-json(xquery-lib:clean-xml($element))
+    )
+};
+
+
+declare function clean-xml($element as item()*) as document-node() {
```
rjrudin, line 82: Let's call this something more descriptive than "clean", which doesn't describe what it's doing. I think your comment indicates what the function name should be - perhaps "remove-empty-structured-properties"?

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/builtins/steps/mapping/entity-services/test-data/mappings/CustomerMapping/CustomerMapping-1.mapping.json
```diff
@@ -0,0 +1,15 @@
+{
+  "lang": "zxx",
+  "name": "CustomerMapping",
+  "description": "This is intentionally mapping with invalid source properties",
+  "version": 1,
+  "targetEntityType": "http://marklogic.com/data-hub/example/CustomerType-0.0.1/CustomerType",
+  "properties": {
+    "id": {
+      "sourcedFrom": "invalidProp1"
```
rjrudin, line 9: These aren't really invalid, they're just missing, right? "invalid" implies to me that some sort of error should be thrown.

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/builtins/steps/mapping/entity-services/test-data/mappings/OrdersMapping/OrdersMapping-4.mapping.json
```diff
@@ -0,0 +1,37 @@
...
+  "name": "OrdersMapping",
+  "description": "The structured properties are mapped to invalid properties",
+  "version": 4,
+  "targetEntityType": "http://marklogic.com/data-hub/example/OrderType-0.0.1/OrderType",
+  "sourceContext": "/",
+  "properties": {
+    "id": {
+      "sourcedFrom": "id"
+    },
+    "purchaseDate": {
```
rjrudin, line 12: Do we need anything more than id and customer here? If not, let's remove these other mapping expressions, as they complicate the test. We want the bare minimum of everything to isolate the problem and verify it's fixed.

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/builtins/steps/mapping/entity-services/test-data/mappings/OrdersMapping/OrdersMapping-4.mapping.json
```diff
@@ -0,0 +1,37 @@
...
+    },
+    "customer": {
+      "targetEntityType": "http://marklogic.com/data-hub/example/CustomerType-0.0.1/CustomerType",
+      "sourcedFrom": "customer",
+      "properties": {
+        "firstname": {
+          "sourcedFrom": "invalidFirstName"
+        }
+      },
+      "items": {
```
rjrudin, line 26: This doesn't look right - shouldn't "items" either be under "customer/properties" or at the same level as "customer"?

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/builtins/steps/mapping/entity-services/runEmptyMapping.sjs
```diff
@@ -0,0 +1,35 @@
...
+  assertions = assertions.concat([
+    test.assertEqual(null, instance.OrderType.customer),
+    test.assertEqual(null, instance.OrderType.items),
+    test.assertEqual("1", fn.string(instance.OrderType.id)),
+    test.assertEqual('2019-12-07', fn.string(instance.OrderType.purchaseDate))
+  ]);
+}
+
+function runSimplePropertyMapping() {
+  let doc = fn.head(xdmp.unquote(`{"id":1,"transactionDate":"12/7/2019","customer":{"firstName":"Oralia","lastName":"Dinesen","gender":"F"},"items":[{"name":"Voltsillam","price":2.0,"quantity":7},{"name":"Latlux","price":7.17,"quantity":10},{"name":"Biodex","price":5.01,"quantity":2},{"name":"Fixflex","price":8.77,"quantity":6},{"name":"Keylex","price":5.57,"quantity":3}]}`));
```
rjrudin, line 21: Same as above, use the simplest input doc possible to verify the problem is fixed.

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/builtins/steps/mapping/entity-services/runEmptyMapping.sjs
```diff
@@ -0,0 +1,35 @@
+const mapping = require("/data-hub/5/builtins/steps/mapping/entity-services/main.sjs");
+const test = require("/test/test-helper.xqy");
+
+let assertions = [];
+function runStructuredPropertyMapping () {
+  let doc = fn.head(xdmp.unquote(`{"id":1,"transactionDate":"12/7/2019","customer":{"firstName":"Oralia","lastName":"Dinesen","gender":"F"},"items":[{"name":"Voltsillam","price":2.0,"quantity":7},{"name":"Latlux","price":7.17,"quantity":10},{"name":"Biodex","price":5.01,"quantity":2},{"name":"Fixflex","price":8.77,"quantity":6},{"name":"Keylex","price":5.57,"quantity":3}]}`));
```
rjrudin, line 6: Let's simplify the input doc to the bare minimum here - no transactionDate needed, and only need one item.

rjrudin (changes requested):

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/builtins/steps/mapping/entity-services/test-data/mappings/OrdersMapping/OrdersMapping-4.mapping.json
srinathgit: "items" is of type 'array'. I want to make sure it works fine as well.

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/builtins/steps/mapping/entity-services/runEmptyMapping.sjs
srinathgit: Removed transactionDate, added 2 items so as to ensure all instances of it get removes from '$results' obtained by calling xdmp:xslt-invoke in line 64.
```
<?xml version="1.0" encoding="UTF-8"?>
<OrderType xmlns:xs="http://www.w3.org/2001/XMLSchema" xmlns:map="http://marklogic.com/xdmp/map" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
  <id xsi:type="xs:string">1</id>
  <customer>
    <CustomerType/>
  </customer>
  <items datatype="array">
    <ItemType/>
  </items>
  <items datatype="array">
    <ItemType/>
  </items>
</OrderType>
```

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/builtins/steps/mapping/entity-services/runEmptyMapping.sjs
```diff
@@ -0,0 +1,37 @@
...
+  }).value.root;
+  let instance = result.envelope.instance;
+  assertions = assertions.concat([
+    test.assertEqual(null, instance.OrderType.customer),
+    test.assertEqual(null, instance.OrderType.items),
+    test.assertEqual("1", fn.string(instance.OrderType.id))
+  ]);
+}
+
+//None of the entity properties are mapped, so 'CustomerType' shouldn't be present in entity instance
```
rjrudin, line 21: This is fine - no need to change - but a comment like this is best expressed as the assertion message - e.g. 

    test.assertEqual(null, instance.CustomerType, "Since none of the Customer properties are mapped, it should not exist in the entity instance");

There are two benefits to this:

1. If the test fails, we get a nice message as opposed to just that something isn't null
1. If the logic changes, or if the code is reorganized, the assertion message is almost guaranteed to stay in sync with the code, while the comment may not get updated or may not get moved

rjrudin (approved):

ryanjdew (approved):

PR #4992, opened 2020-12-12
DHFPROD-5983: Run new OOTB step to add sourceName and sourceType to user-specified documents

rahulvudutala:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/custom/addSourceNameAndType/main.sjs
```diff
@@ -0,0 +1,65 @@
+function main(content, options) {
+  if(options["sourceNameToAdd"] || options["sourceTypeToAdd"]) {
+    // See https://docs.marklogic.com/Node.nodeType
+    if(content.value.root && 1 === content.value.root.nodeType) {
+      addSourcesToXmlDocuments(content, options["sourceNameToAdd"], options["sourceTypeToAdd"]);
```
rjrudin, line 5: Call both of these functions "addSourceTo...", since only one source is being added (even though that may require adding a "sources" property).

marklogic-data-hub/src/test/java/com/marklogic/hub/flow/AddSourceNameAndTypeStepDefTest.java
```diff
@@ -0,0 +1,124 @@
...
+
+import static org.junit.jupiter.api.Assertions.*;
+
+public class AddSourceNameAndTypeStepDefTest extends AbstractHubCoreTest {
+    @BeforeEach
+    void beforeEach() {
+        installProjectInFolder("test-projects/addSources-step-definition-test");
+        String path = "test-projects/addSources-step-definition-test/data";
+        FlowInputs inputs = new FlowInputs("addSourcesFlow", "1");
+        inputs.setInputFilePath(getClass().getClassLoader().getResource(path).getPath());
```
rjrudin, line 23: You should be able to use "makeInputFilePathsAbsoluteInFlow" here. 

Pinging @akshaysonvane  about this - I am wondering though if the best way to override this in tests is via setInputFilePath. But it should be a lot easier than this when there's only one ingestion step (the above makeInputfilePaths function is useful when there are multiple ingestion steps, and each needs it's one absolute path). How about:

    addAbsoluteInputFilePath(flowInputs, "data");

That would use the current HubProject to resolve the "data" directory and then add the absolute file path of that directory to flowInputs.

marklogic-data-hub/src/test/resources/test-projects/addSources-step-definition-test/data/customer1.json
```diff
@@ -0,0 +1,3 @@
+{
```
rjrudin, line 1: Nice - simplest document possible!

marklogic-data-hub/src/test/java/com/marklogic/hub/flow/AddSourceNameAndTypeStepDefTest.java
```diff
@@ -0,0 +1,124 @@
...
+        runFlow(inputs);
+
+        inputs.setSteps(Arrays.asList("2"));
+        runFlow(inputs);
```
rjrudin, line 27: Can't both steps be run at the same time since they use the same input path?

marklogic-data-hub/src/test/java/com/marklogic/hub/flow/AddSourceNameAndTypeStepDefTest.java
```diff
@@ -0,0 +1,124 @@
...
+import org.w3c.dom.Document;
+import org.w3c.dom.NodeList;
+
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+public class AddSourceNameAndTypeStepDefTest extends AbstractHubCoreTest {
```
rjrudin, line 17: These are really good tests, but I think we need a couple more things - need to verify that if "sources" already exists, then a new source is added and the existing ones aren't modified. You can simply add one more JSON doc and one more XML doc to your test data and verify them as part of the first test scenario (I don't think you need to verify them in every other test).

rjrudin (changes requested):
Looks good, just a few tweaks.

marklogic-data-hub/src/test/java/com/marklogic/hub/flow/AddSourceNameAndTypeStepDefTest.java
rahulvudutala: I have run the same flow with different options on the same files without adding new files and verified if the previously added sources are not updated.

marklogic-data-hub/src/test/java/com/marklogic/hub/flow/AddSourceNameAndTypeStepDefTest.java
rahulvudutala: Added the method addAbsoluteInputFilePath in AbstractHubTest class which reads flowInputs and input filePath and adds the absolute path of the directory to flowInputs

rjrudin (approved):

akshaysonvane (approved):

PR #5002, opened 2020-12-14
DHFPROD-6359: Fix ML9 cluster regression in MasterTest

ryanjdew:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- N/A Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/com.marklogic.smart-mastering/impl/process.xqy
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/com.marklogic.smart-mastering/impl/process.xqy
@@ -325,7 +325,6 @@ declare function proc-impl:consolidate-notifies($all-matches as map:map, $merged
     let $updated-notification-uris :=
       for $key-notification in $key-notifications[@threshold = $key-threshold]
       let $key-uri as xs:string := $key-notification/@uri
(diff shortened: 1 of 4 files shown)


srinathgit (approved):

marklogic-data-hub/src/main/resources/ml-modules/root/com.marklogic.smart-mastering/impl/process.xqy
```diff
@@ -325,7 +325,6 @@ declare function proc-impl:consolidate-notifies($all-matches as map:map, $merged
     let $updated-notification-uris :=
       for $key-notification in $key-notifications[@threshold = $key-threshold]
       let $key-uri as xs:string := $key-notification/@uri
-      let $_lock-on-uri := if (fn:not(map:contains($all-matches, $key))) then merge-impl:lock-for-update($key-uri) else ()
```
rjrudin, line 328: Is this removed because it was not having any impact? I looked at the surrounding code - since "$key" comes from map:keys($all-matches), it looks like this would never resolve to true and thus never call lock-for-update - right?

ryanjdew: Yes, if it was going to do anything it should have been $key-uri, but it doesn't really make sense to lock for a URI in the notification document anyway since it doesn't write to those URIs.

rjrudin (approved):

PR #5003, opened 2020-12-15
DHFPROD-6241:'createdOn' and 'createdBy' in mapped documents are incorrect

srinathgit:
### Description
'createdOn' and 'createdBy' in mapped documents are incorrect
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/flow-utils.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/flow-utils.sjs
@@ -597,6 +597,36 @@ class FlowUtils {
     return headers;
   }
 
+  //If the document header has 'createdBy' or 'createdOn' properties, this method updates it with current values.
+  updateHeaders(headers, outputFormat){
+    const currentUser = xdmp.getCurrentUser();
+    const currentDateTime = fn.currentDateTime();
+    if (outputFormat === this.consts.XML) {
+      var response = [];
+      for (const headerElement of headers) {
+        if(fn.localName(headerElement) == "createdBy"){
+          response.push(this.normalizeValuesInNode(xdmp.unquote(`<createdBy xmlns="">${currentUser}</createdBy>`)));
+        }
+        else if(fn.localName(headerElement) == "createdOn"){
+          response.push(this.normalizeValuesInNode(xdmp.unquote(`<createdOn xmlns="">${currentDateTime}</createdOn>`)));
+        }
+        else {
+          response.push(headerElement);
+        }
+      }
+      return Sequence.from(response);
+    }
+    else{
+      if(headers.createdBy){
+        headers["createdBy"] = currentUser;
+      }
+      if(headers.createdOn){
+        headers["createdOn"] = currentDateTime;
+      }
+      return headers;
+    }
+  }
+
   createMetadata(metaData = {}, flowName, stepName, jobId) {
     if (!metaData) {
       metaData = {};
--- a/marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/controllers/FlowControllerTest.java
+++ b/marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/controllers/FlowControllerTest.java
@@ -2,9 +2,11 @@ package com.marklogic.hub.central.controllers;
 
 import com.fasterxml.jackson.databind.JsonNode;
 import com.marklogic.hub.central.AbstractMvcTest;
+import com.marklogic.hub.central.controllers.steps.IngestionStepControllerTest;
 import com.marklogic.hub.central.controllers.steps.MappingStepControllerTest;
 import com.marklogic.hub.dataservices.StepService;
 import com.marklogic.hub.flow.FlowRunner;
+import com.marklogic.hub.job.JobStatus;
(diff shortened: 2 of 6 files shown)


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/flow-utils.sjs
```diff
@@ -592,7 +592,7 @@ class FlowUtils {
         }
       }
       docHeaders = docHeadersArray.reduce((acc, cur) => Object.assign(acc,cur), {});
-      headers = Object.assign({}, headers, docHeaders);
+      headers = Object.assign({}, docHeaders, headers);
```
rjrudin, line 595: This definitely works when creating a new document. If a user is overwriting an existing document (perhaps the source document itself) - this still seems correct because we're really creating a brand new document from the perspective of the mapping step - i.e. we're overwriting a source document with an entity instance, and those are two totally different things. So overwriting createdOn/createdBy makes sense. 

Couple thoughts to make this easier to understand:

- Rename "headers" to "headersFromStepOptions" to make it more clear what they're describing (they could be interpreted as headers from a source document)
- I think we should swap the order of the first two args in the function so that it is consistent with the order in the "assign" call. And the function needs comments on it to make it clear that the output will be the result of merging "headersFromStepOptions" into "docHeaders", where the former will overwrite the latter where there are duplicate keys. 

Finally, we need a unit test for this function. The HC test is fine, but that's a very high level test for a low level detail like this. The unit test won't care about steps - it'll just pass in in-memory args. Should include a scenario for XML as well so we have test coverage there.

rjrudin (changes requested):

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/artifacts/mapping.sjs
```diff
@@ -86,6 +86,11 @@ function defaultArtifact(artifactName, entityTypeId) {
     defaultCollections.push(hubEs.findEntityServiceTitle(entityTypeId) || entityTypeId);
   }
   return {
+    headers: {
+      sources: [{name: artifactName}],
```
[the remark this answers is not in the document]
srinathgit: We currently have it for ingestion step as well. Should that be removed as well ?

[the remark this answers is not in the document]
rjrudin: We agreed to keep it there for backwards compatibility purposes.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/flow-utils.sjs
```diff
@@ -568,12 +568,19 @@ class FlowUtils {
...
     if (outputFormat === this.consts.XML) {
-      headers = this.cleanData(Sequence.from([
-        docHeaders,
-        this.jsonToXml(headers)
-      ]), "headers", outputFormat);
+      if(docHeaders instanceof Element || docHeaders instanceof Sequence){
+        headersFromStepOptions = this.mergeXmlHeaders(Sequence.from([docHeaders, this.jsonToXml(headersFromStepOptions)]));
+      }
+      else{
+        headersFromStepOptions = this.mergeXmlHeaders(Sequence.from([this.jsonToXml(docHeaders), this.jsonToXml(headersFromStepOptions)]));
```
srinathgit, line 581: @ryanjdew ,
Should we call this.cleanData(this.mergeXmlHeaders(Sequence.from([this.jsonToXml(docHeaders), this.jsonToXml(headersFromStepOptions)]))) here or is this sufficient ?

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/artifacts/mapping.sjs
```diff
@@ -86,6 +86,10 @@ function defaultArtifact(artifactName, entityTypeId) {
     defaultCollections.push(hubEs.findEntityServiceTitle(entityTypeId) || entityTypeId);
   }
   return {
+    headers: {
```
rjrudin, line 89: I'm realizing now that while this will fix the problem for steps created starting in 5.4.0, it's still a problem for everyone's mapping steps prior to 5.4.0. In that sense, we're adding a new feature, but not fixing the bug. 

The general problem here is that when a DHF step is creating a new document based on an input document, it shouldn't blindly copy createdOn/createdBy from the input document. The value of createdOn is almost certainly incorrect, and createdBy may be incorrect too. 

I don't think this should be fixed by step config, this should be fixed by code. 

This appears to only be a problem with mapping steps, as those are the only ones calling the "mergeHeaders" function. I think we need a code fix in that function that doesn't depend on step config. 

That function should say - after I merge the headers, then if createdOn/createdBy are present, I'll be sure to update those to the actual values - fn.currentDateTime() and xdmp.currentUser().

rjrudin (changes requested):

ryanjdew (approved):

rjrudin (approved):
Looks good, and I can do QA on it too

PR #5016, opened 2020-12-17
DHFPROD-6357: Return all user-defined steps

akshaysonvane:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/step/getStepsByType.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/step/getStepsByType.sjs
@@ -19,38 +19,49 @@ xdmp.securityAssert("http://marklogic.com/data-hub/privileges/read-ingestion", "
 xdmp.securityAssert("http://marklogic.com/data-hub/privileges/read-mapping", "execute");
 xdmp.securityAssert("http://marklogic.com/data-hub/privileges/read-match-merge", "execute");
 
-const Artifacts = require('/data-hub/5/artifacts/core.sjs');
-
 var propertiesToReturn;
 
-const response = {};
-const stepTypes = ['ingestion', 'mapping'];
+const response = {
+  'ingestionSteps': [],
+  'mappingSteps': [],
+  'matchingSteps': [],
+  'mergingSteps': [],
+  'masteringSteps': [],
+  'customSteps': []
+};
 
 function removeAllPropertiesExcept(step) {
-    if (fn.exists(propertiesToReturn)) {
-        const stepCopy = {};
-        for (const prop of propertiesToReturn) {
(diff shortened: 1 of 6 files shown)


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/step/getStepsByType.sjs
```diff
@@ -24,7 +24,7 @@ const Artifacts = require('/data-hub/5/artifacts/core.sjs');
 var propertiesToReturn;
 
 const response = {};
-const stepTypes = ['ingestion', 'mapping'];
+const stepTypes = ['ingestion', 'mapping', 'matching', 'merging', 'mastering'];
```
srinathgit, line 27: Should 'mastering' step be present as well ? The users cannot create / edit one in HC but they can add it to flow ?

rjrudin (dismissed):
The test is fine, but generally, we want to test ML functionality via ml-unit-test, which runs a lot faster than JUnit. Over time, the difference adds up.

akshaysonvane:
Updated the PR to now include all available user-defined steps.

briantang (approved):
Looks good, has everything needed for the UI story. Seeing all user-defined steps in the response now.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/step/getStepsByType.sjs
```diff
@@ -24,7 +24,7 @@ const Artifacts = require('/data-hub/5/artifacts/core.sjs');
 var propertiesToReturn;
 
 const response = {};
-const stepTypes = ['ingestion', 'mapping'];
+const stepTypes = ['ingestion', 'mapping', 'matching', 'merging', 'mastering', 'custom'];
```
rjrudin, line 27: Given that we want all steps, it'll be simpler and faster to just query for all steps (they all share a collection) and then group them into a sorted map structure. That's 1 query instead of 6 queries, and also ensures we don't need to update this again (stepTypes I believe only existed before as a way of preventing anything besides ingestion/mapping from showing up).

rjrudin (changes requested):

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/step/getStepsByType.sjs
akshaysonvane: Done

rjrudin (approved):

briantang (approved):

ryanjdew (approved):

PR #5020, opened 2020-12-17
DHFPROD-6374: Use target database for autocomplete endpoint in create…

brucean52:
… edit step

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- N/A Added Tests
  

- ##### Reviewer:

- N/A Reviewed Tests


Sanjeevani19 (approved):

briantang (changes requested):
Seeing the endpoint database change properly when the source database is edited inside a step (for example from staging to final). But following that, creating a new mapping step continues to use final database for the endpoint when I think it should be staging by default.

brucean52:
> Seeing the endpoint database change properly when the source database is edited inside a step (for example from staging to final). But following that, creating a new mapping step continues to use final database for the endpoint when I think it should be staging by default.

Mapping step uses the create-edit-mapping component and is out of scope of this bug.

briantang (approved):

PR #5021, opened 2020-12-17
DHFPROD-6306: Handle all tabbed content on step save

wooldridge:
### Description

The step (Load, Map, Match, Merge) tabbed dialog has been updated so that clicking Save under either tab saves changes under all tabs. The user is no longer prompted to save changes when switching tabs.

- Step saving tests have been moved up to the Steps parent component since that's where the saving is now handled rather than in the forms children.
- Similar for tests of the "Discard changes?" dialog on cancel, as that is also now handled exclusively in the Steps parent.

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


marklogic-data-hub-central/ui/e2e/cypress/integration/curation/load/defaultIngestion.spec.tsx
```diff
@@ -56,6 +56,7 @@ describe("Default ingestion ", () => {
     loadPage.stepNameInput().should("be.disabled");
     loadPage.stepDescriptionInput().clear().type("UPDATE");
     loadPage.saveButton().click();
+    cy.waitForAsyncRequest();
```
brucean52, line 59: I still don't think this function does anything. Navigating through the UI can be done using selectors that are based off of UI state changes. The cypress-wait-until library is a better option.

wooldridge: Thanks, I removed the instances that I added and didn't see any negative impact.

wooldridge:
Fixed a step-saving issue that was keeping saves from happening from the Advanced tab. Force-pushed and this should fix the failures that involve not finding JSON property names in mapped docs.

wooldridge:
e2e failures, reverting to version with the waitForAsyncRequest() calls after step saves.

marklogic-data-hub-central/ui/e2e/cypress/integration/curation/load/defaultIngestion.spec.tsx
wooldridge: I reverted on account of e2e failures on the CI/CD side. e2e tests now pass.

xnikhil08 (approved):

briantang (approved):
Looks great! Tests pass.

PR #5025, opened 2020-12-17
DHFPROD-6205: Customize Optic optimization level via Spark connector

anu3990:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/sql/sources/v2/reader/HubInputPartitionReader.java
+++ b/marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/sql/sources/v2/reader/HubInputPartitionReader.java
@@ -9,6 +9,7 @@ import com.marklogic.client.io.InputStreamHandle;
 import com.marklogic.client.io.JacksonHandle;
 import com.marklogic.hub.HubClient;
 import com.marklogic.hub.spark.sql.sources.v2.Util;
+import org.apache.commons.lang.StringUtils;
 import org.apache.spark.sql.catalyst.InternalRow;
 import org.apache.spark.sql.sources.v2.reader.InputPartitionReader;
 import org.apache.spark.sql.types.StructType;
@@ -43,7 +44,7 @@ public class HubInputPartitionReader extends LoggingObject implements InputParti
 
         StructType sparkSchema = (StructType) StructType.fromJson(initializationResponse.get("sparkSchema").toString());
         this.jsonRowParser = new JsonRowParser(sparkSchema);
-        ObjectNode endpointConstants = buildEndpointConstants(initializationResponse, partitionNumber);
+        ObjectNode endpointConstants = buildEndpointConstants(options, initializationResponse, partitionNumber);
         this.bulkOutputCaller = buildOutputCaller(endpointConstants, endpointParams);
     }
 
@@ -88,10 +89,11 @@ public class HubInputPartitionReader extends LoggingObject implements InputParti
         logger.debug("Closing");
     }
 
-    private ObjectNode buildEndpointConstants(JsonNode initializationResponse, int partitionNumber) {
+    private ObjectNode buildEndpointConstants(Map<String, String> options, JsonNode initializationResponse, int partitionNumber) {
         ObjectNode endpointConstants = objectMapper.createObjectNode();
         endpointConstants.set("initializationResponse", initializationResponse);
         endpointConstants.put("partitionNumber", partitionNumber);
+        addOptimizationLevel(options,endpointConstants);
         return endpointConstants;
     }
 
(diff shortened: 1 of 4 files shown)


marklogic-data-hub/src/main/resources/ml-modules/root/marklogic-data-hub-spark-connector/readRows.sjs
```diff
@@ -25,19 +25,32 @@ const results = [endpointState];
...
 
   // Run the parameterized plan, constraining it to the min and max row ID of the current batch
-  op.import(endpointConstants.initializationResponse.parameterizedPlan)
-    .result(null, {
-      "MIN_ROW_ID": batch.min,
-      "MAX_ROW_ID": batch.max
-    })
-    .toArray()
-    .forEach(row => results.push(row));
+  if(optimizationlevel >= 0) {
```
rjrudin, line 35: op.import returns a plan, so to avoid duplication here, try this:

```
let thePlan = op.import(...);
if (optimization level is set) thePlan = thePlan.prepare(...);
thePlan.result(...).toArray().forEach(...)
```

marklogic-data-hub-spark-connector/src/test/java/com/marklogic/hub/spark/sql/sources/v2/reader/ReadWithCustomOptimizationlevel.java
```diff
@@ -0,0 +1,62 @@
...
+    public void testDecimalOptimizationlevel() {
+        setUp();
+        Options options = newOptions().withView("Customer").withStringOptimizationlevel("1.5");
+        IllegalArgumentException ex = assertThrows(IllegalArgumentException.class,
+            () -> readRows(new HubDataSourceReader(options.toDataSourceOptions())));
+        assertEquals("optimizationlevel needs to be 0,1 or 2",
+            ex.getMessage());
+    }
+
+    private void setUp() {
```
rjrudin, line 55: Since this is needed for every test, put it in a BeforeEach-annotated method at the top of the test. For readability, it's helpful to see BeforeEach methods at the top.

marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/sql/sources/v2/reader/HubInputPartitionReader.java
```diff
@@ -88,10 +88,21 @@ public void close() {
         logger.debug("Closing");
     }
 
-    private ObjectNode buildEndpointConstants(JsonNode initializationResponse, int partitionNumber) {
+    private ObjectNode buildEndpointConstants(Map<String, String> options, JsonNode initializationResponse, int partitionNumber) {
         ObjectNode endpointConstants = objectMapper.createObjectNode();
         endpointConstants.set("initializationResponse", initializationResponse);
         endpointConstants.put("partitionNumber", partitionNumber);
+        String optimizationlevel = options.get("optimizationlevel");
```
rjrudin, line 95: Given that we have a couple levels of nesting going on here, that's a good indicator that we could use a new private method for readability here - e.g. 

```
addOptimizationLevel(options, endpointConstants);
```

That method will then do the error handling and validation of the value that the user provides.

rjrudin (changes requested):

marklogic-data-hub-spark-connector/src/main/java/com/marklogic/hub/spark/sql/sources/v2/reader/HubInputPartitionReader.java
```diff
@@ -117,4 +122,19 @@ private void readNextBatchOfRows() {
             rowIndex = 0;
         }
     }
+
+    private ObjectNode addOptimizationLevel(Map<String, String> options, ObjectNode endpointConstants){
+        String optimizationlevel = options.get("optimizationlevel");
+        if(optimizationlevel!=null && optimizationlevel.length()>0) {
```
rjrudin, line 128: It's best to use StringUtils.hasText or StringUtils.isNotEmpty here, as that will do a trim().length(), which rules out whitespace.

marklogic-data-hub/src/main/resources/ml-modules/root/marklogic-data-hub-spark-connector/readRows.sjs
rjrudin: There's still duplication here of the "result" call. You can avoid that by doing what I showed above - call thePlan = thePlan.prepare if there's an optimization level.

rjrudin (changes requested):

marklogic-data-hub/src/main/resources/ml-modules/root/marklogic-data-hub-spark-connector/readRows.sjs
anu3990: Done. Sorry about the confusion.!

rjrudin (approved):

srinathgit (approved):

PR #5033, opened 2020-12-21
DHFPROD-5935:Download project zip from HC

srinathgit:
### Description
Download project zip from HC
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/controllers/environment/DownloadProjectFilesMvcTest.java
+++ b/marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/controllers/environment/DownloadProjectFilesMvcTest.java
@@ -5,8 +5,6 @@ import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.Test;
 import org.springframework.http.MediaType;
 import org.springframework.mock.web.MockHttpServletResponse;
-import org.springframework.security.access.AccessDeniedException;
-import org.springframework.test.web.servlet.request.MockHttpServletRequestBuilder;
 
 import java.io.ByteArrayInputStream;
 import java.util.ArrayList;
@@ -17,11 +15,11 @@ import java.util.zip.ZipInputStream;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.get;
-import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;
 
 public class DownloadProjectFilesMvcTest extends AbstractMvcTest {
 
-    private final static String PATH = "/api/environment/downloadProjectFiles";
+    private final static String HUB_CENTRAL_FILES_PATH = "/api/environment/downloadHubCentralFiles";
+    private final static String PROJECT_FILES_PATH = "/api/environment/downloadProjectFiles";
 
     @AfterEach
     void afterEach() {
@@ -34,11 +32,11 @@ public class DownloadProjectFilesMvcTest extends AbstractMvcTest {
 
         loginAsTestUserWithRoles("hub-central-downloader");
 
-        mockMvc.perform(get(PATH).session(mockHttpSession))
+        mockMvc.perform(get(HUB_CENTRAL_FILES_PATH).session(mockHttpSession))
             .andDo(result -> {
                 MockHttpServletResponse response = result.getResponse();
                 assertEquals(MediaType.APPLICATION_OCTET_STREAM_VALUE, response.getContentType());
-                assertEquals("attachment; filename=hub-central-project.zip", response.getHeader("Content-Disposition"));
+                assertEquals("attachment; filename=datahub-hub-central-files.zip", response.getHeader("Content-Disposition"));
 
                 // We trust the zip to be constructed correctly based on DownloadConfigurationFilesTest, so just
                 // doing a quick sanity check here
@@ -53,11 +51,32 @@ public class DownloadProjectFilesMvcTest extends AbstractMvcTest {
                 assertTrue(entryNames.contains("entities/Order.entity.json"));
                 assertTrue(entryNames.contains("entities/Customer.entity.json"));
             });
+        mockMvc.perform(get(PROJECT_FILES_PATH).session(mockHttpSession))
+            .andDo(result -> {
+                MockHttpServletResponse response = result.getResponse();
+                assertEquals(MediaType.APPLICATION_OCTET_STREAM_VALUE, response.getContentType());
+                assertEquals("attachment; filename=datahub-project-files.zip", response.getHeader("Content-Disposition"));
+
+                List<String> entryNames = new ArrayList<>();
+                try (ZipInputStream zipStream = new ZipInputStream(new ByteArrayInputStream(response.getContentAsByteArray()))) {
+                    ZipEntry entry;
+                    while ((entry = zipStream.getNextEntry()) != null) {
+                        entryNames.add(entry.getName());
+                        zipStream.closeEntry();
+                    }
+                }
+                assertTrue(entryNames.contains("entities/Order.entity.json"));
+                assertTrue(entryNames.contains("entities/Customer.entity.json"));
+
+                assertTrue(entryNames.contains("build.gradle"));
+                assertTrue(entryNames.contains("src/main/hub-internal-config/security/roles/data-hub-developer.json"));
+            });
     }
 
     @Test
     void forbiddenUser() throws Exception {
         loginAsTestUserWithRoles("hub-central-user");
-        verifyRequestIsForbidden(get(PATH));
+        verifyRequestIsForbidden(get(HUB_CENTRAL_FILES_PATH));
+        verifyRequestIsForbidden(get(PROJECT_FILES_PATH));
     }
 }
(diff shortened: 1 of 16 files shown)


marklogic-data-hub/src/main/java/com/marklogic/hub/hubcentral/HubCentralManager.java
```diff
@@ -31,33 +33,93 @@
...
         final ZipOutputStream zipOutputStream = new ZipOutputStream(outputStream);
-        final ObjectWriter prettyWriter = buildPrettyWriter();
+        try {
+            writeArtifactsIntoZipOutputStream(hubClient, zipOutputStream);
+        } finally {
+           IOUtils.closeQuietly(zipOutputStream);
+        }
+    }
 
+    public void writeProjectAsZip(HubClient hubClient, OutputStream outputStream){
```
rjrudin, line 44: I think this needs less knowledge of what's actually in the project, as this is duplicating knowledge that initProject has in HubProject. How about the following approach instead:

1. Create a temp directory using java.io.tmpdir
1. Use HubProject to initialize a project in that temp dir
1. Get all of the project update files and write them over top of the initialized project dir
1. Zip up the entire project directory 

Now, if we add a new project file in initProject in HubProject, we don't have to remember to account for it here too.

rjrudin (changes requested):

marklogic-data-hub/src/main/java/com/marklogic/hub/hubcentral/HubCentralManager.java
srinathgit: I have used hubProject.exportProject() for step 4.

marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/controllers/environment/DownloadProjectFilesMvcTest.java
```diff
@@ -21,7 +21,8 @@
 
 public class DownloadProjectFilesMvcTest extends AbstractMvcTest {
 
-    private final static String PATH = "/api/environment/downloadProjectFiles";
+    private final static String ARTIFACTS_PATH = "/api/environment/downloadProjectUpdates";
```
rjrudin, line 24: I'll test this out locally tomorrow morning, but I think we need some more renaming here so that we consistently use "project updates" and "project files" - i.e. "artifacts" is no longer meaningful in this context, because it's not clear what artifacts we're getting.

rjrudin: Correction, trying it out now, will try to get feedback soon.

marklogic-data-hub/src/test/java/com/marklogic/hub/hubcentral/AllArtifactsProject.java
```diff
@@ -22,9 +22,9 @@
  */
 public class AllArtifactsProject extends TestObject {
 
-    private HubClient hubClient;
-    private Map<String, JsonNode> zipEntries;
-    private File zipFile;
+    protected HubClient hubClient;
+    protected Map<String, JsonNode> zipArtifactEntries;
```
rjrudin, line 26: Let's do "projectUpdateZipEntries" to make it clear this is the zip of project updates, not the full project zip.

marklogic-data-hub/src/test/java/com/marklogic/hub/hubcentral/AllArtifactsProject.java
```diff
@@ -22,9 +22,9 @@
...
+    private File zipArtifactsFile;
```
rjrudin, line 27: And then projectUpdatesZipFile

marklogic-data-hub/src/test/java/com/marklogic/hub/hubcentral/AllArtifactsProject.java
```diff
@@ -34,20 +34,20 @@ public AllArtifactsProject(HubClient hubClient) {
      * Writes the zip to a file and reads all the entries into memory.
      */
     public void writeProjectArtifactsToZipFile() {
-        try {
-            zipFile = new File("build/allArtifactsProject.zip");
-            FileOutputStream fos = new FileOutputStream(zipFile);
+        try{
```
rjrudin, line 37: And "writeProjectUpdatesToZipFile", and so on for other uses of the term "project artifacts" in this class.

marklogic-data-hub/src/test/java/com/marklogic/hub/hubcentral/InitializedDataHubProject.java
```diff
@@ -0,0 +1,96 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.stream.Stream;
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipFile;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+public class InitializedDataHubProject extends AllArtifactsProject{
```
rjrudin, line 17: I don't think this class is needed - the methods in here can just be private to the one test class that uses it. The name is a little confusing because it implies that it's a newly initialized project, but it's more than that - it's all the stuff from the "all-artifacts" project too.

marklogic-data-hub/src/test/java/com/marklogic/hub/hubcentral/InitializedDataHubProject.java
```diff
@@ -0,0 +1,96 @@
...
+    public void verifyZipProject(){
+        assertTrue(zipProjectEntries.containsKey("src/main/ml-modules/root/custom-modules/custom/"));
+        assertTrue(zipProjectEntries.containsKey("src/main/ml-modules/root/custom-modules/ingestion/"));
+        assertTrue(zipProjectEntries.containsKey("src/main/ml-modules/root/custom-modules/mapping/"));
+        assertTrue(zipProjectEntries.containsKey("src/main/ml-modules/root/custom-modules/mapping-functions/"));
+        assertTrue(zipProjectEntries.containsKey("src/main/ml-modules/root/custom-modules/mastering/"));
+        assertTrue(zipProjectEntries.containsKey("src/main/ml-modules/root/custom-modules/matching/"));
+        assertTrue(zipProjectEntries.containsKey("src/main/ml-modules/root/custom-modules/merging/"));
+        assertEquals(5, zipProjectEntries.get("rootDir"));
+        assertEquals(2, zipProjectEntries.get("src/main/hub-internal-config/database-fields"));
```
rjrudin, line 80: These assertions on exact counts create a maintenance burden, as any time someone adds a new file to hub-internal-config, they have to remember to update this test - which is very unlikely that someone would think to do that, which means this test will fail. And the developer will have to figure out what's going on in this test and how to fix it. 

Instead of exact counts that are brittle, I think we should just verify that one file in each directory exists. That should suffice for verifying that hub-internal-config was written out correctly. 

Also, by not depending on brittle counts, I don't think you need the additional logic above for making counts of files in directories.

marklogic-data-hub/src/test/java/com/marklogic/hub/hubcentral/InitializedDataHubProject.java
```diff
@@ -0,0 +1,96 @@
...
+
+    public void verifyZipProject(){
+        assertTrue(zipProjectEntries.containsKey("src/main/ml-modules/root/custom-modules/custom/"));
+        assertTrue(zipProjectEntries.containsKey("src/main/ml-modules/root/custom-modules/ingestion/"));
+        assertTrue(zipProjectEntries.containsKey("src/main/ml-modules/root/custom-modules/mapping/"));
+        assertTrue(zipProjectEntries.containsKey("src/main/ml-modules/root/custom-modules/mapping-functions/"));
+        assertTrue(zipProjectEntries.containsKey("src/main/ml-modules/root/custom-modules/mastering/"));
+        assertTrue(zipProjectEntries.containsKey("src/main/ml-modules/root/custom-modules/matching/"));
+        assertTrue(zipProjectEntries.containsKey("src/main/ml-modules/root/custom-modules/merging/"));
+        assertEquals(5, zipProjectEntries.get("rootDir"));
```
rjrudin, line 79: Given that there's special logic for ensuring that gradle-local.properties is written, there should be an explicit assertion that it exists. Also, we should spell out exactly which 5 things are expected here - just knowing that 5 exist doesn't really tell us anything.

rjrudin (changes requested):

briantang (changes requested):
Changes on the UI side look good. Only thing is there are cypress failures in 'authorization.spec.tsx'   caused by some lines that are still referencing the old Download text (refer to 'projectInfo.tsx'). Tests should pass once those are updated.

rjrudin (commented):
DownloadProjectFilesMvcTest should replace "artifacts" with "project updates" as well, but I think wait for the 0930 meeting to confirm that "project updates" is the right term here.

rjrudin (changes requested):
I think this is a bug - the non exp search options don't have entity-specific constrains in them, while the "normal" search options do - e.g.

<search:constraint name="Fruit">
    <search:container>
      <search:json-property>Fruit</search:json-property>
    </search:container>
  </search:constraint>

srinathgit:
> I think this is a bug - the non exp search options don't have entity-specific constrains in them, while the "normal" search options do - e.g.
> 
> <search:constraint name="Fruit">
> search:container
> search:json-propertyFruit</search:json-property>
> </search:container>
> </search:constraint>

Is this introduced by this PR or has been present already ?

rjrudin:
@srinathgit  I think you're right - actually I think this is what we intentionally removed. So ignore my comment.

rjrudin (approved):
You can assign me to do the QA for this.

briantang (approved):
Looks good on the UI side. E2e and unit tests passing.

srinathgit:
> You can assign me to do the QA for this.
@rjrudin ,
Can you please check "Reviewed Tests" ?

PR #5034, opened 2020-12-21
DHFPROD-6180: Converting SMC project for 5.4 usage

rjrudin:
### Description

Changes that had to be made:

1. Changed rest-reader/rest-writer permissions to data-hub-common (this is not specific to mastering)
1. Changed custom step to use new ES-based path range index instead of element index
1. Fixed dob-match.xqy custom action based on changes in 5.4 (will need docs to explain this to users)

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


Sanjeevani19 (commented):
@rjrudin Can we also add users and gradle.properties files as we do in reference project so that it’s easier to deploy and test using the project

rjrudin:
@Sanjeevani19 That's outside the scope of this ticket. smart-mastering-complete has had Pari as a target audience, such that she can see how to configure matching/merging and also do some custom coding where it provides extension points. 

A separate ticket could define the users/roles/etc that we want in this project, though I'd want to avoid just making a copy of everything under reference-entity-model.

Sanjeevani19 (approved):

ryanjdew (approved):

PR #5038, opened 2020-12-21
DHFPROD-6328: Failing test for unconverted zip match option

rjrudin:


rjrudin:
@ryanjdew  I created this for testing 6328, as I was not able to get an unconverted zip match to work in the SMC project. I was able to reproduce the problem in this test. I thought I was on the right path with having the max score calculation account for "weight" at any level, but I then realized that within options-impl, the "expand" rules were not being evaluated - i.e. the "fn:abs" call on a match ruleset was only happening on the "add" matcher on "city". Figuring you know how to fix this much more easily. 

Also, if I add "weight":"10" to the root of the zip matcher, it all works. But that's not how folks will have their match options from 5.2.x.

rjrudin:
Closing this, it's linked to in the relevant JIRA ticket.

PR #5041, opened 2020-12-21
DHFPROD-6149:Disable generation of TDE for entity models

srinathgit:
### Description
Disable generation of TDE for entity models
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/4/triggers/entity-model-trigger.xqy
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/4/triggers/entity-model-trigger.xqy
@@ -133,7 +133,7 @@ return (
     ),
 
   (: Attempt to generate TDE :)
-  if (local:should-write-tde($tde-uri)) then
+  if (local:should-write-tde($tde-uri) and hent:is-tde-generation-enabled($entity-def))then
     try {
       tde:template-insert(
         $tde-uri,
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/hub-entities.xqy
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/hub-entities.xqy
@@ -83,6 +83,19 @@ declare function hent:uber-model($models as object-node()*) as map:map
     $uber-model
 };
 
+declare function hent:is-tde-generation-enabled($entity-def as object-node()) as xs:boolean
+{
+  let $entity-def-map as map:map := $entity-def
+  let $definitions := $entity-def-map => map:get("definitions")
+  let $entity-title := $entity-def/info/title
+  (: Check if TDE generation is enabled. If the property is not set or has any value other than true/'true', it enables TDE generation. :)
+  let $primary-type-def := ($definitions => map:get($entity-title), $definitions => map:get(map:keys($definitions)[1]))[1]
+  let $is-property-set := map:contains($primary-type-def, "tdeGenerationDisabled")
+  let $property-value := xs:string(map:get($primary-type-def, "tdeGenerationDisabled"))
+  let $tdes-enabled := not($is-property-set) or $property-value != "true"
+  return $tdes-enabled
+};
+
 declare function hent:wrap-duplicates(
   $duplicate-map as map:map,
   $property-name as xs:string,
--- a/marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/entities/testTdeGenerationDisabled.sjs
+++ b/marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/entities/testTdeGenerationDisabled.sjs
@@ -1,0 +1,135 @@
+/**
+ Copyright (c) 2020 MarkLogic Corporation
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+ http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+'use strict';
+
+const test = require("/test/test-helper.xqy");
+const hent = require("/data-hub/5/impl/hub-entities.xqy");
+
+function checkForTde(uri){
+  return fn.head(xdmp.eval(
+    `fn.docAvailable('${uri}') `,
+    {uri:uri}, {database: xdmp.schemaDatabase()}
+  ));
+}
+
+function verifyTdeGeneration() {
+  return [
+    test.assertTrue(checkForTde("/tde/Entity1-0.0.1.tdex"), "TDE should be generated when 'tdeGenerationDisabled' is set to false"),
+    test.assertFalse(checkForTde("/tde/Entity2-0.0.1.tdex"), "TDE should not be generated when 'tdeGenerationDisabled' is set to true"),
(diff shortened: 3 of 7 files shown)


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/4/triggers/entity-model-trigger.xqy
```diff
@@ -132,8 +132,15 @@ return (
       }, map:entry("database", xdmp:schema-database())
     ),
 
+  (: Check if TDE generation is disabled. Any value other than false or "false" disables TDE generation. :)
```
rjrudin, line 135: For readability, let's toss this into a function in the module - e.g. local:is-tde-generation-enabled($definitions, $entity-title) . 

Better yet - let's put that function into hub-entities.xqy. You can write in-memory unit tests against that function, which makes it easier to test the scenario where there's not an entity definition whose name matches that of info/title (I don't think that's covered below?). 

It's good to have a test against the trigger too, but I'd opt for one trivial test there, and then exhaustive tests against this function, as the latter will be faster and easier to understand.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/4/triggers/entity-model-trigger.xqy
```diff
@@ -132,8 +132,15 @@ return (
...
+  let $primary-type-def := $definitions => map:get($entity-title)
+  let $primary-type-def := if (fn:exists($primary-type-def)) then $primary-type-def
```
rjrudin, line 137: An XQuery trick that I think works here is:

    let $primary-type-def := ($definitions => map:get($entity-title), $definitions => map:get(map:keys($definitions)[1]))[1]

i.e. return the first non-empty thing

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/entities/testTdeGenerationDisabled.sjs
```diff
@@ -0,0 +1,35 @@
...
+function checkForTde(uri){
+  return fn.head(xdmp.eval(
+    `fn.docAvailable('${uri}') `,
+    {uri:uri}, {database: xdmp.schemaDatabase()}
+  ));
+}
+
+function verifyTdeGeneration() {
+  return [
+    test.assertTrue(checkForTde("/tde/Customer-0.0.1.tdex")),
```
rjrudin, line 29: Really need assertion messages here to explain why it's true or false for each one.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/4/triggers/entity-model-trigger.xqy
srinathgit: @rjrudin ,
So, is this response correct for the following 2 entities
hent.isTdeGenerationDisabled(entity1) - false
hent.isTdeGenerationDisabled(entity2) - true
```
  let entity1 = {
    "info": {
      "title": "Person",
      "version": "0.0.1",
      "baseUri": "http://marklogic.com/",
      "description": "A person entity"
    },
    "definitions": {
      "Customer": {
        "tdeGenerationDisabled": "true"
      }
    }
  }
  let entity2 = {
    "info": {
      "title": "Person",
      "version": "0.0.1",
      "baseUri": "http://marklogic.com/",
      "description": "A person entity"
    },
    "definitions": {
      "Person": {
        "tdeGenerationDisabled": "randomValue"
      }
    }
  }
```

rjrudin: First, call the function "isTdeGenerationEnabled", as the typical use will be "if TDE generation is enabled, then do the following" (as opposed to "If TDE generation is disabled, then do nothing" or "If TDE generation is not disabled, then do the following", which is a double negative). 

The property should still be tdeGenerationDisabled, since for backwards compatibility reasons, we need to default to enabling it if the property is not there.

For entity1, TDE generation would not be enabled (because there's no entity with a name matching info/title, so we use the first one) and for entity2 it would be enabled because the value is not the boolean value true. (Let's only support the boolean true and not "true" - the user expects to put a boolean here, not a string).

rjrudin: Effectively, the rule is - "If the definition whose name matches info/title, or the first definition in the absence of such a match, has tdeGenerationDisabled = true, then do not generate a TDE".

srinathgit: Just to confirm, for entity1, TDE generation should be enabled right ? There's no entity with a name matching info/title, so we use the first one("Customer"), it has "tdeGenerationDisabled" set to "true" (which we don't support) and not true.

rjrudin: Oh right - yeah, it should be boolean true not "true", so yes, it would be enabled. 

All good explanations to put into assertion messages. :)

ryanjdew (dismissed):

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/hub-entities.xqy
```diff
@@ -83,6 +83,21 @@ declare function hent:uber-model($models as object-node()*) as map:map
...
+declare function hent:is-tde-generation-enabled($entity-def as object-node()) as xs:boolean
+{
+  let $entity-def-map as map:map := $entity-def
+  let $definitions := $entity-def-map => map:get("definitions")
+  let $entity-title := $entity-def/info/title
+  (: Check if TDE generation is enabled. Any value other than true enables TDE generation. :)
+  let $primary-type-def := ($definitions => map:get($entity-title), $definitions => map:get(map:keys($definitions)[1]))[1]
+  let $is-property-set := map:contains($primary-type-def, "tdeGenerationDisabled")
+  let $property-val := map:get($primary-type-def, "tdeGenerationDisabled")
+  let $is-property-type-boolean := $is-property-set and string(xdmp:type($property-val)) = "boolean"
```
rjrudin, line 95: Hmm - if the code is simpler in XQuery to just support "true" and true, let's do that (I don't think I've ever used xdmp:type before, so this seems more complex than it should be).

rjrudin (changes requested):

rjrudin (dismissed):

ryanjdew (dismissed):

srinathgit:
Fixed a test failing due to setup issues

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/entities/test-data/entities/Customer.entity.json
```diff
@@ -0,0 +1,86 @@
+{
+  "info": {
+    "title": "Customer",
+    "version": "0.0.1",
+    "baseUri": "http://example.org/"
+  },
+  "definitions": {
+    "Customer": {
```
rjrudin, line 8: Sorry to be picky here, but until we need all these properties, let's make these entity defs as minimal as possible. We always want to do that in tests, as it makes the focus of test that much more clear.

rjrudin (changes requested):

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/entities/test-data/entities/Customer.entity.json
srinathgit: I have renamed the entity models and included minimal properties.

rjrudin (approved):
Looks good. Thanks for simplifying the entity models, that's just something we want to do across all of our tests so that the tests are both easy to understand and easy to fix when something fails. The fewer moving parts, the better.

akshaysonvane (approved):

PR #5052, opened 2020-12-29
Trying to define content schema

rjrudin:


rjrudin:
@ryanjdew  @srinathgit  @fsnow Can you take a look at this and add comments for stuff that's incorrect or missing? I'd like to get this defined as part of enhancing the docs for editing custom steps.

specs/models/ContentObject.schema.json
```diff
@@ -0,0 +1,86 @@
+{
+  "$schema": "http://json-schema.org/schema#",
+  "$id": "http://marklogic.com/data-hub/ContentObject.schema.json",
+  "title": "ContentObject",
+  "description": "Defines the object that is passed, either by itself or within an array, to a step module for processing and is then used for writing documents",
+  "type": "object",
+  "properties": {
+    "uri": {
```
ryanjdew, line 8: There is also an optional `previousUri` property that can be a string or array of strings. It is used for the derivedFrom coarse-grained document-level provenance. The mastering/merging steps are the only ones that use it at the moment.

rjrudin: I noticed that one, but I'm thinking that isn't worth including yet until we can explain to a user what the result of populating that would be - i.e. why would a user populate this? 

I also noticed "$delete", I need to add that, as that's an easy one to explain.

ryanjdew: It may be worth changing previousUri to derivedFrom at some point. That would be easier to explain the provenance connection.

PR #5053, opened 2020-12-29
DHFPROD-6385: Optimize top-level value retrieval for matching

ryanjdew:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- N/A Code passes ESLint tests
- N/A Added Tests
  

- ##### Reviewer:

- N/A Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/test/java/com/marklogic/hub/master/MasterTest.java
+++ b/marklogic-data-hub/src/test/java/com/marklogic/hub/master/MasterTest.java
@@ -62,7 +62,12 @@ public class MasterTest extends AbstractHubCoreTest {
         makeInputFilePathsAbsoluteInFlow("myNewFlow");
 
         runAsDataHubOperator();
-        runFlow(new FlowInputs("myNewFlow", "1", "2"));
+        RunFlowResponse flowResponse = runFlow(new FlowInputs("myNewFlow", "1", "2"));
+        for(RunStepResponse stepResponse:flowResponse.getStepResponses().values()) {
+            if (stepResponse != null) {
+                assertTrue(stepResponse.isSuccess(), stepResponse.getStepName() + " job failed with " + stepResponse.getStepOutput());
+            }
+        }
         JsonNode matchResp = masteringManager.match("/person-1.json", "myNewFlow", "3", Boolean.TRUE, new ObjectMapper().createObjectNode()).get("results");
         assertEquals(7, matchResp.get("total").asInt(), "There should 7 match results");
         assertEquals(7, matchResp.get("result").size(), "There should 7 match results");
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/com.marklogic.smart-mastering/impl/util.xqy
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/com.marklogic.smart-mastering/impl/util.xqy
@@ -30,6 +30,8 @@ import module namespace es-helper = "http://marklogic.com/smart-mastering/entity
 import module namespace httputils="http://marklogic.com/data-hub/http-utils"
 at "/data-hub/5/impl/http-utils.xqy";
 
+declare namespace es = "http://marklogic.com/entity-services";
+
 declare variable $write-objects-by-uri as map:map := map:map();
 
 declare function util-impl:add-all-write-objects(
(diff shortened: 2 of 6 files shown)


marklogic-data-hub/src/test/java/com/marklogic/hub/master/MasterTest.java
```diff
@@ -62,7 +62,12 @@ public void testMatchEndpoint() {
         makeInputFilePathsAbsoluteInFlow("myNewFlow");
 
         runAsDataHubOperator();
-        runFlow(new FlowInputs("myNewFlow", "1", "2"));
+        RunFlowResponse flowResponse = runFlow(new FlowInputs("myNewFlow", "1", "2"));
```
rjrudin, line 65: This test, and testMasterStep, are failing locally for me, and also in Jenkins - https://jenkins.marklogic.com/blue/organizations/jenkins/Datahub_CI/detail/PR-5053/1/tests . 

It's very difficult to debug these tests. The assertion messages below only restate what the assertions state. Before the mastering step is run, there are 239 docs in the "mapping" collection. Why are 7 match results expected?

I recommend trying to rework this test as part of this PR. The scope appears to be just verifying that the match endpoint works. If so, I think all we need is a test that inserts two documents (one that will match and one that won't), one flow with a mastering step, and then uses a test document to verify that one match is returned. The assertion messages should explain why the match is expected, and why the test doc doesn't match the other document. That way, when the test fails in the future, it'll be much easier to debug. Right now, it's nearly impossible to debug because a developer is faced with a ton of analysis work to figure out why 7 matches are expected, and which ones those are. Instead of trying to document all of that, I think it'd be easier and better in the long term to just replace this test. 

I recommend replacing testMasterStep for the same reason - it's failing with a message of "At least 10 merges are expected", which doesn't inspire much confidence in how accurate merging is (i.e. is it okay if there are 10 merges, or 50 merges? If so, why?).

rjrudin (commented):
There are 12 failing ml unit tests in Jenkins too. I tried to debug the first one - entityDrivenMatches.sjs - but I ran into a problem similar to the one I encountered with MasterTest - the assertion messages are just restating what the assertions themselves state. Why is a merge action expected? Reading further down, I can see that CustMatchMerge1 and CustMatchMerge2 are expected to be in the merge action - but why, and why only those 2 - why not any of the other 12? In order to figure that out, I really need to bring up 3 documents in a side-by-side view - those 2 docs, plus the matching config - and then figure out what set of properties match such that I hit the "merge" threshold.

rjrudin:
I did find SM-MATCH tracing to be helpful - when I run entityDrivenMatches.sjs (and I commented out the XML function so as to isolate on the JSON test), I can see this for the expected matching doc:

```
Checking cts.doc('/content/CustMatchMerge1.json') for match against ruleset 'billingAddress with weight 5: false
Checking cts.doc('/content/CustMatchMerge1.json') for match against ruleset 'customerId with weight 4.5: false
Checking cts.doc('/content/CustMatchMerge1.json') for match against ruleset 'name with weight 3.5: false
Checking cts.doc('/content/CustMatchMerge1.json') for match against ruleset 'lastName with weight 2.5: true
Checking cts.doc('/content/CustMatchMerge1.json') for match against ruleset 'shippingAddress with weight 2: true
Checking cts.doc('/content/CustMatchMerge1.json') for match against ruleset 'reduce for household with weight -1: true
cts.doc('/content/CustMatchMerge1.json') score: 3.5 minimum-threshold: 4.5
```

That's very helpful - I can see that the score of 3.5 derives from the 3 "true" checks. 

Bringing up the 2 docs in a side by side - I can see that the customerId's are equal, so the "false" is clearly wrong. 

That indicates to me that something in this PR is affecting the "exact" algorithm. But that would be very easy to tell in a test that is limited strictly to an exact match on one property. I checked the other unit tests, but didn't see that sort of an isolated test. 

If we had a set of unit tests under /data-hub/5/smart-mastering/algorithms/exact, then those would likely all be failing. Since those tests would be limited to only exact matches, that would make the problem very clear right away. 

I think "entityDrivenMatches.sjs" has value as a way of verifying that multiple rulesets are working properly together, but if that is indeed the goal, the filename and assertion messages should reflect that.

rjrudin (approved):

srinathgit (approved):

SameeraPriyathamTadikonda:
Tests are failing
https://jenkins.marklogic.com/blue/organizations/jenkins/Datahub_CI/detail/PR-5053/7/tests

srinathgit (approved):

fsnow (approved):

PR #5062, opened 2021-01-05
DHFPROD-6423: building UI for the build task

SameeraPriyathamTadikonda:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [NA] Added Tests
  

- ##### Reviewer:

- [NA] Reviewed Tests

### DIFF
--- a/web/build.gradle
+++ b/web/build.gradle
@@ -217,7 +217,7 @@ task stopBootrun {
 e2eUI.finalizedBy stopBootrun
 
 // Need to ensure the UI files are built and included in the war file
-if (gradle.startParameter.taskNames*.toLowerCase().contains("bootwar"))
+if ((gradle.startParameter.taskNames*.toLowerCase().contains("bootwar")) || (gradle.startParameter.taskNames*.toLowerCase().contains("build")))
 {
   processResources.dependsOn copyIndexHtml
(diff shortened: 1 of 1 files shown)


rjrudin (changes requested):
I don't think this fixes the problem - you need to change these lines instead:

if (gradle.startParameter.taskNames*.toLowerCase().contains("bootwar"))
{
  processResources.dependsOn copyIndexHtml
}

The error is because copyIndexHtml is not being invoked. So you have to modify that if statement to see if taskNames contains "bootwar" or "build".

SameeraPriyathamTadikonda:
> I don't think this fixes the problem - you need to change these lines instead:
> 
> if (gradle.startParameter.taskNames*.toLowerCase().contains("bootwar"))
> {
> processResources.dependsOn copyIndexHtml
> }
> 
> The error is because copyIndexHtml is not being invoked. So you have to modify that if statement to see if taskNames contains "bootwar" or "build".

It worked as build depended on bootWar which will copyUI files. But we don't need to run bootWar in case of build. So I updated the condition.

srinathgit (approved):

rjrudin (approved):

PR #5086, opened 2021-01-11
DHFPROD-6385: Match query optimizations

ryanjdew:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/smart-mastering/helper/property-name-to-query.xqy
+++ b/marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/smart-mastering/helper/property-name-to-query.xqy
@@ -1,0 +1,99 @@
+xquery version "1.0-ml";
+
+import module namespace helper = "http://marklogic.com/smart-mastering/helper-impl"
+  at "/com.marklogic.smart-mastering/matcher-impl/helper-impl.xqy";
+
+
+import module namespace test = "http://marklogic.com/test" at "/test/test-helper.xqy";
+
+
+declare variable $match-options := xdmp:unquote('{
+  "dataFormat": "json",
+  "targetEntityType": "MasteringTestEntity",
+  "propertyDefs": [
+    {
+      "property": {
+        "name": "nicknames",
+        "indexReferences": [
+          {
+            "pathReference" : {
+              "pathExpression" : "/(es:envelope|envelope)/(es:instance|instance)/MasteringTestEntity/nicknames",
+              "scalarType" : "string",
+              "collation" : "http://marklogic.com/collation//S2",
+              "nullable" : false
+            }
+          }
+        ]
+      }
+    }
+  ],
+  "matchRulesets": [
+    {
+      "name": "name",
+      "matchRules": [
+        {
+          "entityPropertyPath" : "name",
+          "matchType": "exact"
+        }
+      ]
+    },
+    {
+      "name": "nicknames",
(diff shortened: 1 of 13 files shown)


marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/smart-mastering/helper/property-name-to-query.xqy
```diff
@@ -0,0 +1,99 @@
...
+
+
+declare variable $match-options := xdmp:unquote('{
+  "dataFormat": "json",
+  "targetEntityType": "Customer",
+  "propertyDefs": [
+    {
+      "property": {
+        "name": "nicknames",
+        "indexReferences": [
```
rjrudin, line 17: Is this something that Pari would provide - i.e. the entity model wouldn't specify an index on nicknames? Also, does the GUI support this, or is this something Pari would provide via editing the JSON file? If the GUI doesn't support it (which seems fine), is there a GUI test to ensure it won't drop this config?

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/smart-mastering/helper/suite-setup.xqy
```diff
@@ -0,0 +1,42 @@
+xquery version "1.0-ml";
+import module namespace admin = "http://marklogic.com/xdmp/admin"
+  at "/MarkLogic/admin.xqy";
+
+(: add indexes for the test :)
+
+let $config := admin:get-configuration()
```
rjrudin, line 7: If you'd like, you can add these to the test-config//final-database.json file so they don't have to be coded in here. I would change the entity name though so that it doesn't conflict with other tests that use a Customer entity - e.g. "MasteringTestEntity" or something like that.

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/smart-mastering/helper/property-name-to-query.xqy
ryanjdew: This is something that Pari would provide. I'll add a Hub Central test to ensure it isn't dropped.

ryanjdew: @rjrudin Reviewing it further, I believe the test here should be sufficient: https://github.com/marklogic/marklogic-data-hub/blob/1135085a67d780142ae4234003776fd17857f591/marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/controllers/steps/AbstractStepControllerTest.java#L47

It verifies that properties that haven't been explicitly sent in an update are retained.

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/smart-mastering/helper/suite-setup.xqy
ryanjdew: Done. Thanks for the suggestion.

rjrudin (approved):

srinathgit (dismissed):

rjrudin (approved):

srinathgit (approved):

PR #5088, opened 2021-01-12
DHFPROD-6476: Step processor failure now reports correct job counts

rjrudin:
### Description

Had to remove the throwing of an exception from stopJob - I couldn't find any reason why that would be good. With it there, the test was failing because 4 failed items were being reported instead of 2. Specifically, a stepItemFailureListener was throwing this exception, after the 2 failed items had already been captured. So it seems unnecessary to throw an exception when stopJob is called but the jobId cannot be found. Also ran all of the flow tests and none of them failed with this change. 

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/flow.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/flow.sjs
@@ -395,22 +395,27 @@ class Flow {
     }
     flowInstance.globalContext.uri = null;
 
+    let stepProcessorFailed = false;
     try {
       this.applyProcessorsBeforeContentPersisted(flowStep, contentArray, combinedOptions);
     } catch (e) {
       // If a processor throws an error, we don't know if it was specific to a particular item or not. So we assume that
       // all items failed; this is analogous to the behavior of acceptsBatch=true
+      flowInstance.globalContext.completedItems = [];
+      stepProcessorFailed = true;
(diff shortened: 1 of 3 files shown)


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/flow.sjs
```diff
@@ -395,22 +395,27 @@ class Flow {
...
+      contentArray.forEach(contentObject => {
+        if (contentObject.uri) {
+          this.writeQueue.push(contentObject);
+        } else {
+          this.datahub.debug.log({ type: 'error', message: `Couldn't add '${xdmp.toJsonString(contentObject)}' to the write queue due to missing uri.`});
+        }
+      });
+    }
 
     if (hook && !hook.runBefore) {
```
srinathgit, line 420: If a step has both processor and custom hook, should post commit hook run if processor fails ?

rjrudin: I went back and forth on this. Because a custom hook is already able to run when items in a batch fail, I think it should be able to run when a processor fails too.

ryanjdew (approved):

srinathgit (approved):

akshaysonvane (approved):

PR #5102, opened 2021-01-13
DHFPROD-6446: fixed missing entity link in Run tile

jyangML:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/src/api/__mocks__/mocks.data.ts
+++ b/marklogic-data-hub-central/ui/src/api/__mocks__/mocks.data.ts
@@ -280,7 +280,9 @@ const advancedAPI = (axiosMock) => {
   axiosMock.post["mockImplementationOnce"](jest.fn(() =>
     Promise.resolve({status: 200, data: {}})));
   return axiosMock.get["mockImplementation"]((url) => {
-    const defaultCollectionsURL = `/api/steps/merging/defaultCollections/${encodeURI(stepsData.stepMerging.targetEntityType)}`;
+    const targetEntityType = String(stepsData.stepMerging.targetEntityType);
+    const targetEntityTitle = targetEntityType.substring(targetEntityType.lastIndexOf("/") + 1);
+    const defaultCollectionsURL = `/api/steps/merging/defaultCollections/${encodeURIComponent(targetEntityTitle)}`;
     switch (url) {
     case "/api/steps/ingestion/AdvancedLoad":
       return Promise.resolve({status: 200, data: stepsData.stepLoad});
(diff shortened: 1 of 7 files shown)


ngodugu-marklogic (commented):
@jyangML When we click the view step on Match/Merge flow step, we are getting navigated to the curate page but not to the actual step. For instance, when we click Map Step we are getting navigated to the actual step(entity/Map tab). Please refer: https://drive.google.com/file/d/1eW-vRXWfrBGdlV0qyUyoj3EMQ4pmU7lZ/view

Sanjeevani19 (changes requested):
@jyangML Looks like there is a cypress test failure

jyangML:
@ngodugu-marklogic What you described seems to only affect existing flows/steps from our project setups, such as reference-entity-model or hc-qa-project in e2e.  If you create a new match/merge step, then add that step to any flow, the link should go to the correct entity.

There is a slight problem as of now.  If you click the link to go to the Curate tile, the right entity and the right tab in the entity is open, but the bar at the top shows the wrong tab.  Functional-wise, it's working.
Screenshot 2021-01-14 114943

wooldridge:
I'm seeing the following difference between the existing steps in example project setups and what gets saved for new steps currently.

Existing flows:
```
"targetEntity": "Customer",
"targetEntityType": *** UNDEFINED ***,
```

New flows:
```
"targetEntity": "Change this to a valid entity type name; e.g. Customer",
"targetEntityType": "http://example.org/Customer-0.0.1/Customer",
```

Ergo:
1. New flows need to be saved with a target entity name ("Customer").
2. If existing flows don't have "targetEntityType" properties, do we need to make the UI functionality work with targetEntity instead of targetEntityType?

@ryanjdew, what are your thoughts?

ryanjdew:
@wooldridge As part of the convert for Hub Central process, I believe we are moving the targetEntity value to the targetEntityType property. That value could either be the entity title or the full URI. If you'd like to file a ticket we can convert it to the full URI as part of the conversion. 

We shouldn't be setting targetEntity with the new steps. That is a bug.

wooldridge:
@ryanjdew, thanks for the reply. I updated my comment. It looks like the existing curation steps in example projects are different, mapping steps have targetEntityType while matching and merging have targetEntity. 

Having everything with a targetEntityType with the URL value should work with how the UI is currently working.

It looks like the backend is filling in that info string for targetEntity for new steps.

jyangML:
Incorporated all changes from DHFPROD-6514 (https://github.com/marklogic/marklogic-data-hub/pull/5112) into this PR as well, since changes are related.

Sanjeevani19 (approved):

briantang (approved):

PR #5108, opened 2021-01-14
DHFPROD-6385: Fix runaway XPath expression in compile match options

ryanjdew:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- N/A Added Tests
  

- ##### Reviewer:

- N/A Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/com.marklogic.smart-mastering/matcher-impl/options-impl.xqy
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/com.marklogic.smart-mastering/matcher-impl/options-impl.xqy
@@ -448,12 +448,7 @@ declare function opt-impl:compile-match-options(
               fn:string($match-rule/matchType)
             else
               $local-name
-          let $full-property-node :=
-              if ($is-complex-rule) then
-                $match-rule/(entityPropertyPath|documentXPath)
-              else
-                $match-rule/(@property-name|propertyName|((*:all-match|allMatch)/*:property))
-          let $full-property-name := fn:normalize-space(fn:string-join($full-property-node, ", "))
+          let $full-property-name := opt-impl:full-property-name-from-rule($match-rule, $is-complex-rule)
           let $algorithm-ref :=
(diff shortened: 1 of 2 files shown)


fsnow (dismissed):

marklogic-data-hub/src/main/resources/ml-modules/root/com.marklogic.smart-mastering/matcher-impl/options-impl.xqy
```diff
@@ -452,7 +452,10 @@ declare function opt-impl:compile-match-options(
               if ($is-complex-rule) then
                 $match-rule/(entityPropertyPath|documentXPath)
               else
-                $match-rule/(@property-name|propertyName|((*:all-match|allMatch)/*:property))
+                (
```
rjrudin, line 455: Looking at the surrounding code, I think it's worthwhile to extract a small function here and add a few unit tests, as this is part of a very large function that has dozens of variable assignments, making it difficult to understand.

This assigns a variable - full-property-node (which possibly should have a plural name since it looks like the assignment can return a sequence?) - that is only used on the next line to get full-property-name. So I think there's a function of:

    get-full-property-name-for-match-rule(match-rule, is-complex)

Some small unit tests could be written against that to verify that it works correctly and, more importantly, to demonstrate some of the possible inputs. For example, the code begs the question - When does all-match/allMatch exist, and does it also exist when property-name/propertyName exists? The test case would answer that question by showing the expected inputs.

rjrudin (changes requested):

marklogic-data-hub/src/main/resources/ml-modules/root/com.marklogic.smart-mastering/matcher-impl/options-impl.xqy
ryanjdew: Added the separate function with unit tests

fsnow (approved):

rjrudin (approved):

PR #5109, opened 2021-01-14
DHFPROD-6490:Add validation to step, stepdef and flow names in gradle task

srinathgit:
### Description
Add validation to step, stepdef and flow names in gradle task
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/artifacts/core.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/artifacts/core.sjs
@@ -176,12 +176,21 @@ function setArtifact(artifactType, artifactName, artifact) {
 }
 
 function validateArtifact(artifactType, artifactName, artifact) {
+  if(!validateArtifactName(artifactName)){
+    let message = `Invalid name: '${artifactName}'; it must start with a letter and can contain letters, numbers, hyphens and underscores only.`
+    httpUtils.throwBadRequest(message);
+  }
   const artifactLibrary = getArtifactTypeLibrary(artifactType);
   const validatedArtifact = artifactLibrary.validateArtifact(artifact, artifactName);
   if (validatedArtifact instanceof Error) {
     httpUtils.throwBadRequest(validatedArtifact.message);
   }
-    return validatedArtifact;
+  return validatedArtifact;
+}
+
+function validateArtifactName(artifactName){
+  const pattern = /^[a-zA-Z][a-zA-Z0-9\-_]*$/;
+  return pattern.test(artifactName);
 }
 
 function getArtifactNode(artifactType, artifactName, artifactVersion = 'latest') {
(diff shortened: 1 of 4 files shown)


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/artifacts/core.sjs
```diff
@@ -176,12 +176,25 @@ function setArtifact(artifactType, artifactName, artifact) {
 }
 
 function validateArtifact(artifactType, artifactName, artifact) {
+  if(!validateArtifactName(artifactName)){
+    let message = `Invalid ${artifactType}`;
+    if(!(artifactType == "flow" || artifactType == "stepDefinition")){
```
rjrudin, line 181: This is a leaky abstraction, where the generic artifact library has knowledge of particulars about different artifact types. I think we should remove this, particularly because how much value are we getting from including the artifact type in the message? I think "Invalid name: etc..." is sufficient and it avoids this leak.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/artifacts/core.sjs
```diff
@@ -176,12 +176,25 @@ function setArtifact(artifactType, artifactName, artifact) {
...
+      message += " step";
+    }
+    message += ` name: '${artifactName}'; it must start with a letter and can contain letters, numbers, hyphens and underscores only.`
+    httpUtils.throwBadRequest(message);
```
rjrudin, line 185: This is preventing the concrete artifact from doing its own validation. That can result in the user first getting this name validation error, and then fixing it, only to get another error next. Ideally, the user gets all errors at once so they only have to do one set of fixes. Is the UI able to handle many errors, or are we restricted for now to a single error message? That's fine if so, but likely something to improve in the near future.

rjrudin (commented):
Looks good, just one tweak.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/artifacts/core.sjs
srinathgit: It looks like HC UI will do client side validation based on comments in https://project.marklogic.com/jira/browse/DHFPROD-6454

rjrudin: Argh. The way to get rapid feedback isn't by duplicating knowledge across tiers, it's by immediately asking the backend to validate something after the user changes it.

rjrudin (approved):

ryanjdew (approved):

PR #5110, opened 2021-01-14
DHFPROD-6516: Fixing bug in flow-utils, and some cleanup too

rjrudin:
### Description

Deleting some dead code, moved a little too:

- datahub.getConfig() is not used
- FlowUtils doesn't need a config object
- Removed xquerySanitizer (unused) and its test
- Moved retrieveModuleLibrary from HubUtils to stepDefinition since it's only used there
- HubUtils doesn't need a config object

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/flow-utils.sjs
```diff
@@ -711,18 +705,8 @@ class FlowUtils {
   }
 
   parseText(text, outputFormat){
-    try {
```
rjrudin, line 714: This is a good example of what happens when there are no tests - this function has multiple problems with it (datahub isn't defined, and 'instance' is used instead of 'text'), and it's used by custom ingestion step modules.

ryanjdew (approved):

srinathgit (changes requested):
There's a test failure
https://jenkins.marklogic.com/blue/organizations/jenkins/Datahub_CI/detail/PR-5110/6/tests

rjrudin:
@srinathgit  @rahulvudutala The test failure is from options always being generated. It's failing for me on develop too. I'll bump up the assertion count on my branch though. The cypress failure is unrelated too.

srinathgit (approved):

ryanjdew (approved):

PR #5121, opened 2021-01-15
DHFPROD-6369: Fix error handling in Run when a flow is missing a step

wooldridge:
- Display modal with message from error object when calling api/flows fails
- Add test of api/flows error case

To test manually:

1. Create a flow and add a step to that flow.
2. Remove that added step via some external manner (e.g., remove the step file in QConsole). 
3. Go to the Run tile. An error modal should be displayed and the browser should no longer go into an infinite flows-retrieving loop.

Note that the resulting page displays no flows whatsoever (since the api/flows endpoint sends back and error and no flows).

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/src/api/__mocks__/mocks.data.ts
+++ b/marklogic-data-hub-central/ui/src/api/__mocks__/mocks.data.ts
@@ -196,6 +196,27 @@ const runAddStepAPI = (axiosMock) => {
   });
 };
 
+// For testing display of a flow missing a step (DHFPROD-6369)
+const runMissingStep = (axiosMock) => {
+  axiosMock.get["mockImplementation"]((url) => {
+    switch (url) {
+    case "/api/flows":
+      return Promise.reject({
+        "response": {
+          "data": {
+            "code": 400,
+            "message": "Error message"
+          }
+        }
+      });
+    case "/api/steps":
+      return Promise.resolve(curateData.steps);
+    default:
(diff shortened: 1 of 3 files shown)


briantang (approved):
Verified manually that infinite loop issue no longer occurs, error message is displayed. Tests look good

devinnicholson (approved):
Confirmed that error modal pops up, and no infinite loop. Tests pass, lint passes.

Sanjeevani19 (commented):
@wooldridge I see that the modal popup reappears for around 3 times, even though the page is refreshed the run tile no longer shows previous user steps and flows, Is this expected ?

https://drive.google.com/file/d/1ME6Bzp5nVnSCMD8AAGL75azRtGe615jZ/view

wooldridge:
> @wooldridge I see that the modal popup reappears for around 3 times, even though the page is refreshed the run tile no longer shows previous user steps and flows, Is this expected ?
> 
> https://drive.google.com/file/d/1ME6Bzp5nVnSCMD8AAGL75azRtGe615jZ/view

@Sanjeevani19, yes, sometimes this happens due to multiple api/flows calls going on. I hope we can live with this for the beta (this situation is an edge case, after all) and fix in a separate bug.

Sanjeevani19 (approved):

PR #5125, opened 2021-01-18
DHFPROD-6531: Mapping properties from first mapping pre-populated on all other mappings

xnikhil08:
### Description
We will revisit any test scenarios for this bug in a separate mini-hardening, in the interest of time. Just the fix for the bug is included in this PR.

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [N/A] Added Tests
  

- ##### Reviewer:

- [ N/A] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/src/components/entities/mapping/mapping-card.tsx
+++ b/marklogic-data-hub-central/ui/src/components/entities/mapping/mapping-card.tsx
@@ -227,7 +227,7 @@ const MappingCard: React.FC<Props> = (props) => {
         setSourceData([]);
         setSourceData([...sDta]);
         if (typeof(srcDocResp.data) === "string") {
-          let mData = await props.getMappingArtifactByMapName(props.entityModel.entityTypeId, props.data[index].name);
+          let mData = await props.getMappingArtifactByMapName(props.entityModel.entityTypeId, stepName);
           updateMappingWithNamespaces(mData);
         }
       }
--- a/marklogic-data-hub-central/ui/src/components/steps/steps.tsx
+++ b/marklogic-data-hub-central/ui/src/components/steps/steps.tsx
@@ -97,14 +97,14 @@ const Steps: React.FC<Props> = (props) => {
     setCurrentTab(key);
   };
 
-  const getStepPayload = (payload) => {
+  const getStepPayload = (payload, newStepFlag?: boolean) => {
     // Combine current payload from saved payloads from both tabs, ensure name prop exists
     let name = basicPayload["name"] ? basicPayload["name"] : props.stepData.name;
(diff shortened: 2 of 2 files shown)


Sanjeevani19 (commented):
Works as expected, @xnikhil08 Do you think it's better we add a e2e scenario for this, Given the time constraint if you think it would take some time, let's create a separate task to cover these scenarios

Sanjeevani19 (dismissed):

xnikhil08:
> Works as expected, @xnikhil08 Do you think it's better we add a e2e scenario for this, Given the time constraint if you think it would take some time, let's create a separate task to cover these scenarios

@Sanjeevani19 - We should definitely include an e2e scenario for this, but given the timing constraint i was thinking if we can take that up as a part of separate ticket, unless this bug is not a blocker anymore.

srinathgit (changes requested):
~~After applying the fix, can you try this, create a new map for entity1 and save it, open entity2 pane , UI breaks~~
The above issue is already present in develop

Sanjeevani19:
> > Works as expected, @xnikhil08 Do you think it's better we add a e2e scenario for this, Given the time constraint if you think it would take some time, let's create a separate task to cover these scenarios
> 
> @Sanjeevani19 - We should definitely include an e2e scenario for this, but given the timing constraint i was thinking if we can take that up as a part of separate ticket, unless this bug is not a blocker anymore.

https://project.marklogic.com/jira/browse/DHFPROD-6521, created this task to handle these test scenarios

srinathgit (changes requested):
I create a step 'map1', add a property, close it. I create another step 'map2', when I add a property, 'map1' gets overwritten. see
https://drive.google.com/file/d/1FfT4BhPT3pCx3L9vbIpoAaiFP-6doWwz/view

xnikhil08:
> I create a step 'map1', add a property, close it. I create another step 'map2', when I add a property, 'map1' gets overwritten. see
> https://drive.google.com/file/d/1FfT4BhPT3pCx3L9vbIpoAaiFP-6doWwz/view

@srinathgit - Fixed it. Another review ?

srinathgit (approved):
Looks good

Sanjeevani19 (approved):

srinathgit (changes requested):
https://jenkins.marklogic.com/blue/organizations/jenkins/Datahub_CI/detail/PR-5125/3/tests 
e2e test failure

Sanjeevani19:
> https://jenkins.marklogic.com/blue/organizations/jenkins/Datahub_CI/detail/PR-5125/3/tests
> e2e test failure

@srinathgit This might be a intermittent test failure, not related to this PR, Could you please remove Changes requested state so that the tests will run again

srinathgit (approved):

PR #5126, opened 2021-01-18
DHFPROD-6454: Add validation to step and flow names in Hub Central

briantang:
### Description

- Follow-up PR to address some issues caught in QA from @srinathgit 
- On special characters error in name field, save button should not submit but remain enabled
- Spaces should not be allowed in name field

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


srinathgit (approved):
It works fine. Spaces aren't allowed in artifact names and the request doesn't go through in case of invalid artifact name

timur-isangulov (approved):

PR #5145, opened 2021-01-21
DHFPROD-6329: Unconverted custom match action fails

fsnow:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/examples/smart-mastering-complete/src/main/ml-modules/root/custom-modules/custom/dob-match.xqy
+++ b/examples/smart-mastering-complete/src/main/ml-modules/root/custom-modules/custom/dob-match.xqy
@@ -14,14 +14,12 @@ Example of a simple DOB matching algorithm that expands the query by bumping eac
 :)
 declare function algorithms:dob-match(
   $expand-values as xs:string*,
-  $expand-xml as node(),
-  $options-xml as node()
+  $expand-xml as element(matcher:expand),
+  $options-xml as element(matcher:options)
 ) as cts:query*
 {
-  let $property-name := $expand-xml/entityPropertyPath
-
-  (: A custom match action is no longer being passed the weight, so it's hardcoded here. DHFPROD-6324 captures the issue. :)
-  let $weight := 10
+  let $property-name := $expand-xml/@property-name
+  let $weight := $expand-xml/@weight
 
   for $value in $expand-values
   where $value castable as xs:date
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/com.marklogic.smart-mastering/matcher-impl/options-impl.xqy
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/com.marklogic.smart-mastering/matcher-impl/options-impl.xqy
@@ -147,7 +147,11 @@ declare function opt-impl:save-options(
 (: Convert JSON match options to XML :)
 declare function opt-impl:options-from-json($options-json)
 {
-  let $options-root := $options-json/options
+  let $options-root :=
+    if (fn:exists($options-json/options)) then
+      $options-json/options
+    else
+      $options-json
   return
     element matcher:options {
       if (fn:exists($options-root/targetEntity)) then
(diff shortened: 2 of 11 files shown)


ryanjdew (approved):

marklogic-data-hub/src/main/resources/ml-modules/root/com.marklogic.smart-mastering/matcher-impl/options-impl.xqy
```diff
@@ -578,6 +584,40 @@ declare function opt-impl:compile-match-options(
     )
 };
 
+declare function opt-impl:convert-match-rule-for-custom-module($match-rule, $match-options, $custom-algorithm)
```
rjrudin, line 587: Is it possible to declare types for these inputs? I always like to see those in XQuery functions, unless the args can be anything (although even item() is useful to signify whether one or multiple are expected).

examples/smart-mastering-complete/src/main/ml-modules/root/custom-modules/custom/dob-match.xqy
```diff
@@ -1,33 +1,31 @@
...
+at "/com.marklogic.smart-mastering/matcher-impl/helper-impl.xqy";
+
+declare namespace matcher = "http://marklogic.com/smart-mastering/matcher";
+
+declare option xdmp:mapping "false";
+
+(:
+Example of a simple DOB matching algorithm that expands the query by bumping each date up by one year.
+:)
+declare function algorithms:dob-match(
```
rjrudin, line 15: Nice - this means that this function will still work without being modified?

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/smart-mastering/matching/custom-xqy-matching-algo.xqy
```diff
@@ -0,0 +1,22 @@
+xquery version "1.0-ml";
```
rjrudin, line 1: Can this go under smart-mastering/algorithms/custom instead? I think that gives you a nice clean slate to work with too, instead of modifying the plumbing for all the existing tests in the matching directory.

rjrudin (changes requested):

examples/smart-mastering-complete/src/main/ml-modules/root/custom-modules/custom/dob-match.xqy
fsnow: Yes, I reverted to the version in the 5.2 branch.

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/smart-mastering/matching/custom-xqy-matching-algo.xqy
fsnow: I think of /algorithms as having tests that are limited to algorithm implementations. These changes test how matching calls algorithms, so I don't think it belongs under algorithms. The plumbing changes match what other mastering suites are doing (loading artifacts and custom modules).

marklogic-data-hub/src/main/resources/ml-modules/root/com.marklogic.smart-mastering/matcher-impl/options-impl.xqy
fsnow: The XQuery functions in SM *should* have declared types, but where do we start? Do we have test coverage to handle all the possible inputs to SM (3 flavors of config, custom modules in XQ and JS) to catch type errors now instead of our customers seeing errors later?

rjrudin: Yeah, that's fair.

rjrudin (approved):

PR #5147, opened 2021-01-21
DHFPROD-6511: fixed input textbox lag in new step dialogs

jyangML:
### Description
Original problem: new/edit step dialogs would send a payload on every keystroke.  steps.tsx can't process this very quickly, so there's a lag between user keystroke and when it appears on screen.  This fix replaces the update criteria to when the user shifts focus instead (for ex. user presses Save, or user moves onto the next input).

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- N/A Added Tests
  

- ##### Reviewer:

- N/A Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/src/components/advanced-settings/advanced-settings.tsx
+++ b/marklogic-data-hub-central/ui/src/components/advanced-settings/advanced-settings.tsx
@@ -211,11 +211,17 @@ const AdvancedSettings: React.FC<Props> = (props) => {
     props.onCancel();
   };
 
+  /* sends payload to steps.tsx */
+  const sendPayload = () => {
+    props.setHasChanged(hasFormChanged());
+    props.setPayload(getPayload());
+  };
+
   // On change of any form field (or on init), update the changed flag for parent
   useEffect(() => {
     props.setHasChanged(hasFormChanged());
     props.setPayload(getPayload());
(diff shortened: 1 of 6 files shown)


wooldridge (approved):
Nice simple fix! I'm not seeing the delay anymore in all new/edit steps.

wooldridge (changes requested):
Actually, I'm seeing some settings-related tests fail now (in steps.test and Load.test). I suspect it's not enough to just change field content to test discard changes functionality, you'll need to unfocus the field after changing.

wooldridge:
@jyangML, looks like RTL doesn't always play nice with document.activeElement: https://github.com/testing-library/react-testing-library/issues/276

I was able to fix the failing tests by wrapping certain form interactions in focus/blur. This triggers the payload-saving useEffect. PR to your branch here: https://github.com/jyangML/marklogic-data-hub/pull/2

jyangML:
Squashed Mike's RTL fixes

brucean52 (dismissed):

wooldridge (approved):

jyangML:
Rebased from develop

wooldridge (dismissed):

briantang (dismissed):

brucean52 (dismissed):

briantang (changes requested):
Looks better, but for a new map step, seeing the 'Advanced' tab disabled (but still clickable) even though the form is valid. I think maybe a validation flag is getting lost with the latest change. Works fine for all other steps though.

briantang (approved):
Looks good now.

brucean52 (approved):

PR #5156, opened 2021-01-22
DHFPROD-6560: mlWatch now generates function metadata

rjrudin:
### Description

I tried to get a test in place for this, but I don't know how to test mlWatch in a test (which enters an infinite loop). Instead, I added some logging to at least indicate when there's a misconfiguration -  i.e. the command or context are null. 

The actual bug fix is just to populate generateFunctionMetadataCommand when the other beans are populated in DataHubPlugin. 

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


srinathgit (approved):
I guess we can only do manual testing for this fix.

rjrudin:
@srinathgit  Yup, that's what I did - just tried it out on the reference project.

ryanjdew (approved):

PR #5159, opened 2021-01-22
DHFPROD-6492: Fix regression in QuickStart mastering related test

ryanjdew:
### Description
Items Covered:

- The number step definitions increased by 1
- targetEntityType was no longer picked up when the mastering step was run causing the wrong collections to be applied
- The targetTypeEntity not being found hid an issue with blocking future merges. Once the mastering step was fixed, the step wasn't finding the merged record that was archived.
- Using the compiled merge options as a standard way to apply collections caused some merge options related tests to be updated to ensure they call valid functions. This didn't matter before because those tests covered translating between JSON and XML formats.

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/com.marklogic.smart-mastering/survivorship/merging/base.xqy
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/com.marklogic.smart-mastering/survivorship/merging/base.xqy
@@ -129,11 +129,8 @@ declare function merge-impl:save-merge-models-by-uri(
       ") were already write-locked", "debug")
   else
     let $start-elapsed := xdmp:elapsed-time()
-    let $merge-options :=
-      if ($merge-options instance of object-node()) then
-        merge-impl:options-from-json($merge-options)
-      else
-        $merge-options
+    let $compiled-options := merge-impl:compile-merge-options($merge-options)
+    let $merge-options := $compiled-options => map:get("mergeOptionsNode")
     let $merge-write-object :=  merge-impl:build-merge-models-by-uri($uris, $merge-options, $id)
     let $merged-document := $merge-write-object => map:get("value")
     let $merge-uri := $merge-write-object => map:get("uri")
@@ -146,7 +143,7 @@ declare function merge-impl:save-merge-models-by-uri(
         $audit-trace => map:get("context") => map:get("permissions"),
         $audit-trace => map:get("context") => map:get("collections")
       ),
-      let $on-merge-options := $merge-options/merging:algorithms/merging:collections/merging:on-merge
+      let $on-merge-options := $compiled-options => map:get("onMerge")
       let $distinct-uris := fn:distinct-values(($uris, $uris))[fn:doc-available(.)][fn:not(. = $merge-uri)]
       let $_archive := $distinct-uris ! merge-impl:archive-document(., $merge-options)
       return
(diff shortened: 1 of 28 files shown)


rjrudin (approved):

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mastering/default/matching.sjs
```diff
@@ -35,7 +35,7 @@ function filterContentAlreadyProcessed(content, summaryCollection, collectionInf
   const jobIdQuery = cts.fieldWordQuery('datahubCreatedByJob', jobID);
   let auditingNotificationsInSourceQuery = false;
   for (let item of content) {
-    const collections = item.context ? item.context.collections || [] : [];
+    const collections = item.context ? item.context.originalCollections || [] : [];
```
rjrudin, line 38: This seems like a significant change here, since "originalCollections" is what the document is currently in and "collections" is what the document is supposed to be written to. Has this line always been correct? Or did a change in 5.4 cause this to break?

marklogic-data-hub/src/main/resources/ml-modules/root/com.marklogic.smart-mastering/survivorship/merging/base.xqy
```diff
@@ -352,16 +350,46 @@ declare function merge-impl:rollback-merge(
...
+          return (
+            xdmp:trace($const:TRACE-MERGE-RESULTS, $doc-prefix || " previous URIs: " || xdmp:to-json-string($previous-uris)),
+            xdmp:trace($const:TRACE-MERGE-RESULTS, $doc-prefix || " last merge dateTime: " || xdmp:to-json-string($last-merge-dateTime)),
+            xdmp:trace($const:TRACE-MERGE-RESULTS, $doc-prefix || " latest auditing receipt: " || xdmp:to-json-string($latest-auditing-receipt-for-doc))
+          )
+        else
+          ()
   where fn:exists(($latest-auditing-receipt-for-doc,$last-merge-dateTime))
   return (
+    let $remerged-doc :=
```
rjrudin, line 364: This seems like a lot of significant changes based on failures in QS e2e tests; the fact that no tests were failing in the ./marklogic-data-hub is a real concern, because we don't want to be relying on QS tests that are poorly understood (I'm looking at ./web/e2e and can't figure out what's going on - the ./qa-data folder makes this look like a DHF 4 project). 

What are the actual problems surfaced by this JIRA ticket? The main test fix I see is that "targetEntity" is now correctly changed to "targetEntityType", but it looks like we have a lot more logic being implemented here, and I can't tell what problem is being solved by looking at the tests and at the JIRA ticket.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mastering/default/matching.sjs
ryanjdew: Technically, the matching and mastering steps have set the collections to originalCollections by this point. The change here is to mirror the mastering version that also does locking on document URI. originalCollections makes it more clear that we're interested in the state of collections of the document currently in the database.

fsnow (approved):

rjrudin (approved):

PR #5181, opened 2021-01-28
DHFPROD-6585:Deploy only modified artifact when using QS

srinathgit:
### Description
Deploy only modified artifact when using QS
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/LoadUserArtifactsCommand.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/LoadUserArtifactsCommand.java
@@ -51,8 +51,10 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.SimpleFileVisitor;
 import java.nio.file.attribute.BasicFileAttributes;
+import java.util.ArrayList;
 import java.util.Date;
 import java.util.Iterator;
+import java.util.List;
 import java.util.regex.Pattern;
 
 /**
@@ -67,6 +69,7 @@ public class LoadUserArtifactsCommand extends AbstractCommand {
     private DocumentPermissionsParser documentPermissionsParser = new DefaultDocumentPermissionsParser();
     private ObjectMapper objectMapper;
     private TokenReplacer tokenReplacer;
+    private PropertiesModuleManager propertiesModuleManager;
 
     public void setForceLoad(boolean forceLoad) {
         this.forceLoad = forceLoad;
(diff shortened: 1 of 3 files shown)


rjrudin (changes requested):
This change looks good, but I'd much rather that LUAC default to not using a ModulesManager so that all user artifacts are loaded in every scenario except for when using QS. QS is the oddball here (it shouldn't be using this command in the first place), and it's the one that needs the optimization of remembering when files were last modified. PropertiesModuleManager is an optimization, and optimizations can introduce bugs, and I don't want any chance of bugs being introduced for the normal case of using LUAC, where we simply want to load all user artifacts and there's no problem in doing so. 

So I think this command should default to forceLoad=true, and if forceLoad=true, don't do anything with a ModulesManager - don't even instantiate it. QS can then set it to false.

srinathgit:
@rjrudin ,

```
So I think this command should default to forceLoad=true, and if forceLoad=true, don't do anything with a ModulesManager - don't even instantiate it. QS can then set it to false.
```

1. So, with forceLoad set to 'true' should we still should delete the timestamp file but not create an instance of 'PropertiesModuleManager' ?
2. Also if we don't create an instance of 'PropertiesModuleManager' when forceLoad=true, we will not be able to record timestamps of artifacts loaded . So, if artifacts are loaded again now with forceLoad =false, we will be reloading artifacts once again. Is the ok ? (QS loads artifacts with forceLoad = true when users click on "Redeploy" button and forceLoad = false when updating flows, mappings)

rjrudin:
@srinathgit  For QS - it makes sense that redeploy does forceLoad=true, and that a properties file should be updated as well due to its behavior of loading all artifacts whenever the user changes only one. 

For Gradle - it should always do forceLoad=true so that all artifacts are always loaded, and it's fine if a properties file is updated as well - Gradle will never care about it though. 

So it sounds like two scenarios:

# If forceLoad=true, still create a PMM so timestamps can be written out, but don't check the timestamps
# If forceLoad=false, create a PMM so timestamps can be both written and checked

Bottom line though is we don't want any of this to break LUAC in the Gradle case, which is far more important than whatever QS is doing.

srinathgit:
@rjrudin ,
In the current PR

1.  If forceLoad=true, we first delete the properties file, load all artifacts, record timestamp. I am setting forceLoad=true for 'hubDeploy' gradle command (mlDeploy and hubDeployUserArtifacts already set forceLoad=true). I think that should cover all the gradle commands.
2. If forceLoad=false, we look into timestamp file and write only the modified artifact.

This seems to satisfy the scenarios you had mentioned. Or am I missing anything ?

rjrudin (approved):
@srinathgit  That sounds good, I had forgotten that Gradle sets forceLoad=true. Really, that should be the default behavior of LUAC, as only QS has a need for it to not be true.

ryanjdew (approved):

PR #5192, opened 2021-01-29
DHFPROD-6410: and DHFPROD-6534 Construct HC reduce rules appropriately

ryanjdew:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/examples/smart-mastering-complete/steps/matching/match-persons.step.json
+++ b/examples/smart-mastering-complete/steps/matching/match-persons.step.json
@@ -73,7 +73,7 @@
     "weight" : 5,
     "reduce" : true,
     "matchRules" : [ {
-      "entityPropertyPath" : [ "Address" ],
+      "entityPropertyPath" : "Address",
       "matchType" : "exact",
       "options" : { }
     } ]
--- a/marklogic-data-hub-central/ui/src/components/entities/matching/matching-step-detail/matching-step-detail.tsx
+++ b/marklogic-data-hub-central/ui/src/components/entities/matching/matching-step-detail/matching-step-detail.tsx
@@ -89,10 +89,14 @@ const MatchingStepDetail: React.FC = () => {
   };
 
   const matchRuleSetOptions = matchingStep.matchRulesets && matchingStep.matchRulesets.map((i) => {
+    const firstMatchRule = i.matchRules[0];
+    const firstMatchRuleType = firstMatchRule ? firstMatchRule.matchType : "";
+    // reduce is a special case that applies to an entire ruleset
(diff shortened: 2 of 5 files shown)


fsnow (dismissed):

examples/smart-mastering-complete/steps/matching/match-persons.step.json
```diff
@@ -73,7 +73,7 @@
     "weight" : 5,
     "reduce" : true,
     "matchRules" : [ {
-      "entityPropertyPath" : [ "Address" ],
+      "entityPropertyPath" : "Address",
```
rjrudin, line 76: Doesn't this suggest the conversion process needs to be updated? In the unconverted project, here's the JSON for this:

```
"reduce" : [ {
              "allMatch" : {
                "property" : [ "Address" ]
              },
              "algorithmRef" : "standard-reduction",
              "weight" : "5"
            } ]
```

The use of an array there is consistent with what's in the SM docs at https://marklogic-community.github.io/smart-mastering-core/docs/match-algorithms/#standard-reduction 

So my guess would be that if anyone is using Reduce in their current project, they will have an array here, and it would not work either. 

Also, the SM docs show two properties being referenced - what happens during conversion when the user has two properties? Does that lead to 2 Reduce rulesets?

rjrudin (changes requested):

examples/smart-mastering-complete/steps/matching/match-persons.step.json
ryanjdew: The conversion process iterates over each property under allMatch and creates a rule in the ruleset. So if 2 properties are referenced there will be a ruleset with reduce set to true and an exact rule for each property. 

You can see the test for that conversion scenario here: https://github.com/marklogic/marklogic-data-hub/blob/5405ad04a04a84ec8578ca473134e9af01f8f6e7/marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/data-services/mastering/updateMatchOptions.sjs#L216

rjrudin: Got it - but if there's a single property - like in the SMC project - then the conversion process produces (I just tested this in the SMC project):

    "entityPropertyPath" : [ "Address" ]

And this change is suggesting that this is not correct, right? In which case conversion needs to be fixed?

rjrudin (dismissed):

rjrudin (approved):

fsnow (approved):

PR #5198, opened 2021-01-29
DHFPROD-6590:Entity model property names should be an NCName

srinathgit:
### Description
Entity model property names should be an NCName
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/entity-lib.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/entity-lib.sjs
@@ -370,8 +370,11 @@ function validateModelDefinitions(definitions) {
     }
     if (definitions[entityName].properties) {
       Object.keys(definitions[entityName].properties).forEach(propertyName => {
-        if (!pattern.test(propertyName)) {
-          throw new Error(`Invalid property name: ${propertyName}; must start with a letter and can only contain letters, numbers, hyphens, and underscores.`);
+        try{
+          fn.QName('',propertyName)
+        }
+        catch(ex){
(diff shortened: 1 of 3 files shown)


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/entity-lib.sjs
```diff
@@ -370,8 +370,11 @@ function validateModelDefinitions(definitions) {
     }
     if (definitions[entityName].properties) {
       Object.keys(definitions[entityName].properties).forEach(propertyName => {
-        if (!pattern.test(propertyName)) {
-          throw new Error(`Invalid property name: ${propertyName}; must start with a letter and can only contain letters, numbers, hyphens, and underscores.`);
+        try{
+          fn.QName('',propertyName)
+        }
+        catch(ex){
+          throw new Error(`Invalid property name: ${propertyName} in entity model ${entityName}; It must be a valid NCName.`);
```
rjrudin, line 377: I think we need a little more info here, because I don't anticipate that Stewart the modeler will have ever heard of "NCName" before. 

How about: "; it must be a valid NCName as defined at http://www.datypic.com/sc/xsd/t-xsd_Name.html" (that page provides a nice description at the top; I'm open to including that description too - i.e. must start with a letter, underscore(_), or colon (:), and may contain only letters, digits, underscores (_), colons (:), hyphens (-), and periods (.) "

rjrudin (changes requested):
This is DHFPROD-6590, right? Not 6602?

srinathgit:
> This is DHFPROD-6590, right? Not 6602?

My bad, It is DHFPROD-6590.

rjrudin (approved):

ryanjdew (approved):

PR #5224, opened 2021-02-09
DHFPROD-6549: Explorer shows datahub artifacts when All Entities is selected

xnikhil08:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/src/pages/Browse.tsx
+++ b/marklogic-data-hub-central/ui/src/pages/Browse.tsx
@@ -54,7 +54,7 @@ const Browse: React.FC<Props> = ({location}) => {
   const [data, setData] = useState<any[]>([]);
   const [entities, setEntites] = useState<any[]>([]);
   const [entityDefArray, setEntityDefArray] = useState<any[]>([]);
-  const [facets, setFacets] = useState();
+  const [facets, setFacets] = useState<any>();
   const [isLoading, setIsLoading] = useState(true);
   const [totalDocuments, setTotalDocuments] = useState(0);
   const [tableView, toggleTableView] = useState(JSON.parse(getUserPreferences(user.name)).tableView);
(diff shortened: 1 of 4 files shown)


marklogic-data-hub-central/ui/src/pages/Browse.tsx
```diff
@@ -148,13 +148,19 @@ const Browse: React.FC<Props> = ({location}) => {
     if (searchOptions.nextEntityType && searchOptions.nextEntityType !== "All Data") {
       setCardView(false);
     }
-    if ((entities.length || (searchOptions.nextEntityType === "All Data" || searchOptions.nextEntityType === "All Entities" || searchOptions.nextEntityType === undefined)) &&
-      (!searchOptions.nextEntityType ||
-        (searchOptions.nextEntityType === "All Entities" && !searchOptions.entityTypeIds.length && !searchOptions.selectedTableProperties.length && !cardView) ||
+    if ((entities.length > 0 || (searchOptions.nextEntityType === "All Data" || searchOptions.nextEntityType === "All Entities" || searchOptions.nextEntityType === undefined)) &&
```
rjrudin, line 151: Something to consider here - there are a lot of conditionals here, making it hard for a human to read this and understand what the conditions are. 

The Clean Code book recommends defining local variables to capture the names of the conditions here, which then makes the overall conditional much easier to read. I think that would help a lot here. 

For example, the first condition above is saying - if at least one entity type exists (I am assuming that "entities" is really "entityTypes" or the user has selected All Data or All Entities or has not selected anything (and I'm assuming that "nextEntityType" is really "what the user selected" - "next" is confusing here). So how about:

    const entityTypesExistOrNoEntityTypeIsSelected = (all the above logic)

Reading that... I'm wondering when that would ever be false?? Maybe "nextEntityType" isn't really "what the user has selected in the search type dropdown"? I'm confused.

Regardless, It looks like all of the logic can then be simplified to something that reads like this:

    if (entityTypesExistOrNoEntityTypeIsSelected && (nameForFirstCondition || nameForSecondCondition || nameForThirdCondition || nameForFourthCondition))

Which would be far easier to understand than the current code.

marklogic-data-hub-central/ui/src/pages/Browse.tsx
```diff
@@ -148,13 +148,19 @@ const Browse: React.FC<Props> = ({location}) => {
...
+      ((!searchOptions.nextEntityType && entities.length > 0) ||
+        (searchOptions.nextEntityType === "All Entities" && !searchOptions.entityTypeIds.length && !searchOptions.selectedTableProperties.length && !cardView && entities.length > 0) ||
         (searchOptions.nextEntityType === "All Data" && !searchOptions.entityTypeIds.length && !searchOptions.selectedTableProperties.length && cardView) ||
-        (!["All Entities", "All Data"].includes(searchOptions.nextEntityType) && searchOptions.entityTypeIds[0] === searchOptions.nextEntityType)
+        (searchOptions.nextEntityType && !["All Entities", "All Data"].includes(searchOptions.nextEntityType) && searchOptions.entityTypeIds[0] === searchOptions.nextEntityType)
       )) {
       getSearchResults(entities);
+    } else {
+      if (!cardView && (!searchOptions.entityTypeIds.length || !searchOptions.nextEntityType)) {
```
timur-isangulov, line 159: This condition crashes the app due to setting facet object to be empty if apply facets and refresh the page or open detail page and return back to browse. 
Screen Shot 2021-02-09 at 8 29 08 AM

timur-isangulov (changes requested):
There is no search request call if select All Data option on Explorer zero state page. 
https://drive.google.com/file/d/1IS5TDhyxV9lz9ES-h0HH-K3R9EqWJT6D/view

marklogic-data-hub-central/ui/src/pages/Browse.tsx
xnikhil08: Refactored the code a bit to better understand the logic.

marklogic-data-hub-central/ui/src/pages/Browse.tsx
xnikhil08: This is resolved.

xnikhil08:
> There is no entity search request call if select All Data option on Explorer zero state page.
> https://drive.google.com/file/d/1IS5TDhyxV9lz9ES-h0HH-K3R9EqWJT6D/view

@timur-isangulov - This is already working fine for me. Can you please check again now with updated refactoring ?

timur-isangulov (approved):
Checked out and tested, looks good.

rjrudin (approved):

timur-isangulov:
@xnikhil08, one of the manage query test fails on pipeline, I can reproduce it manually. Can you please take a look.

rahulvudutala (approved):

xnikhil08:
> @xnikhil08, one of the manage query test fails on pipeline, I can reproduce it manually. Can you please take a look.

@timur-isangulov - Pushed a fix. Another review ?

timur-isangulov (approved):

Sanjeevani19 (approved):

PR #5233, opened 2021-02-11
DHFPROD-6668:Add "eval-search-string" privilege to Datahub

srinathgit:
### Description
Add "eval-search-string" privilege to Datahub
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/test/java/com/marklogic/hub/security/DataHubOperatorTest.java
+++ b/marklogic-data-hub/src/test/java/com/marklogic/hub/security/DataHubOperatorTest.java
@@ -1,7 +1,15 @@
 package com.marklogic.hub.security;
 
+import com.marklogic.client.ForbiddenUserException;
+import com.marklogic.client.io.Format;
+import com.marklogic.client.io.SearchHandle;
+import com.marklogic.client.io.StringHandle;
+import com.marklogic.client.query.QueryManager;
+import com.marklogic.client.query.RawQueryByExampleDefinition;
 import org.junit.jupiter.api.Test;
 
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 /**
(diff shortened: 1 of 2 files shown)


marklogic-data-hub/src/test/java/com/marklogic/hub/security/DataHubOperatorTest.java
```diff
@@ -32,4 +40,25 @@ void readSchemas() {
...
+        String rawJSONQuery = "{ \"$query\": {\n" +
+            "    \"$and\": [\n" +
+            "        {\"itemPrice\" : { \"$ge\": 6 } },\n" +
+            "        {\"itemPrice\" : { \"$le\": 12 } } ],\n" +
+            "    \"$filtered\": true\n" +
+            "}}";
+        StringHandle rawHandle = new StringHandle(rawJSONQuery).withFormat(Format.JSON);
+        RawQueryByExampleDefinition querydef = queryMgr.newRawQueryByExampleDefinition(rawHandle);
+        SearchHandle resultsHandle = queryMgr.search(querydef, new SearchHandle());
+        assertEquals(1, resultsHandle.getTotalResults());
```
rjrudin, line 59: Can you add an assertion message stating that this works because the user has the eval-search-string privilege?

marklogic-data-hub/src/test/java/com/marklogic/hub/security/DataHubOperatorTest.java
```diff
@@ -32,4 +40,25 @@ void readSchemas() {
...
+
+        QueryManager queryManager = runAsTestUserWithRoles("data-hub-common").getFinalClient().newQueryManager();
+        assertThrows(ForbiddenUserException.class, ()->queryManager.search(queryManager.newRawQueryByExampleDefinition(rawHandle), new SearchHandle()));
```
rjrudin, line 62: Same thing here - use an assertion message to explain why this doesn't work. A benefit of the message is that when someone searches on "eval-search-string" in the codebase, they'll find both the role that inherits it and the test that explains why it's inherited

rjrudin (changes requested):

rjrudin (approved):

ryanjdew (approved):

PR #5237, opened 2021-02-12
DHFPROD-3280: fixed timestamps for batch jobs report

jyangML:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/jobs.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/jobs.sjs
@@ -401,7 +401,7 @@ module.exports.updateBatch = module.amp(
     }
 
     if (batchStatus === "finished" || batchStatus === "finished_with_errors" || batchStatus === "failed") {
-      docObj.batch.timeEnded = fn.currentDateTime();
+      docObj.batch.timeEnded = fn.currentDateTime().add(xdmp.elapsedTime());
     }
     if(error){
       // Sometimes we don't get the stackFrames
--- a/marklogic-data-hub/src/test/java/com/marklogic/hub/flow/VerifyBatchTimestampTest.java
+++ b/marklogic-data-hub/src/test/java/com/marklogic/hub/flow/VerifyBatchTimestampTest.java
@@ -1,0 +1,51 @@
+package com.marklogic.hub.flow;
+
+import com.marklogic.client.io.Format;
+import com.marklogic.hub.AbstractHubCoreTest;
+import com.marklogic.hub.step.RunStepResponse;
+import com.marklogic.hub.test.Customer;
+import com.marklogic.hub.test.ReferenceModelProject;
+import org.junit.jupiter.api.Test;
+
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeFormatterBuilder;
+import java.time.temporal.ChronoField;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+
+
+public class VerifyBatchTimestampTest extends AbstractHubCoreTest {
+
+    @Test
+    void test() {
+        installProjectInFolder("test-projects/custom-step-with-sleep");
+        
+        ReferenceModelProject project = new ReferenceModelProject(getHubClient());
+        doWithWriteBatcher(getHubClient().getStagingClient(), writeBatcher -> {
+        	writeBatcher.add(project.buildCustomerInstanceToWrite(new Customer(0, "Jane"), Format.JSON, null));
+        });
+        
+        RunFlowResponse response = runFlow(new FlowInputs("customFlowWithSleep"));
+        RunStepResponse stepResponse = response.getStepResponses().get("1");
+
+        assertEquals("completed step 1", stepResponse.getStatus());
+        assertEquals(1, stepResponse.getSuccessfulBatches());
+        
+        DateTimeFormatter formatter = new DateTimeFormatterBuilder()
+                .appendPattern("yyyy-MM-dd'T'HH:mm:ss")
+                .appendFraction(ChronoField.MICRO_OF_SECOND, 0, 7, true)
+                .appendPattern("[XXX][X]")
+                .parseLenient()
+                .toFormatter();
+        
+        LocalDateTime startTime = LocalDateTime.parse(stepResponse.getStepStartTime(), formatter);
+        LocalDateTime endTime = LocalDateTime.parse(stepResponse.getStepEndTime(), formatter);
+        
+        assertNotNull(startTime);
+        assertNotNull(endTime);
+        assertTrue(endTime.isAfter(startTime));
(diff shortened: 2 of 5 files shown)


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/jobs.sjs
```diff
@@ -401,7 +401,7 @@ module.exports.updateBatch = module.amp(
     }
 
     if (batchStatus === "finished" || batchStatus === "finished_with_errors" || batchStatus === "failed") {
-      docObj.batch.timeEnded = fn.currentDateTime();
+      docObj.batch.timeEnded = fn.currentDateTime().add(xdmp.elapsedTime());
```
rjrudin, line 404: We definitely want to get a test in place for this. We don't care about exact times, just that timeEnded is after timeStarted. 

A good way to test this is via a custom step that just does an xdmp.sleep(10) call in it; that ensures there's no way that these timestamps should be the same. 

Check with @srinathgit  about how to write tests; also see the "Writing Data Hub Tests" internal Wiki page.

rjrudin (changes requested):

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/jobs.sjs
jyangML: Thank you.  Will take a look

marklogic-data-hub/src/test/java/com/marklogic/hub/flow/VerifyBatchTimestampTest.java
```diff
@@ -0,0 +1,28 @@
...
+    @Test
+    void test() {
+        installProjectInFolder("test-projects/custom-step-with-sleep");
+        
+        // writeManyCustomers(customerCount);
+        RunFlowResponse response = runFlow(new FlowInputs("customFlowWithSleep"));
+        RunStepResponse stepResponse = response.getStepResponses().get("1");
+
+        assertEquals("completed step 1", stepResponse.getStatus());
+        assertNotEquals(stepResponse.getStepStartTime(), stepResponse.getStepEndTime());
```
rjrudin, line 26: I think we want 3 assertions here:

1. The start time is not null
2. The end time is not null
3. The end time is after the start time 

To perform assertion 3, you'll need to get each time as a LocalDateTime. That would be a useful method to add to RunStepResponse. 

Unfortunately, we don't seem to have any good helper methods for parsing these times using the more reliable DateTimeFormatter in Java 8 (as opposed to SimpleDateFormat). I just whipped up this quick test to verify how to construct a DateTimeFormatter that works on ML timestamps, which can have variable lengths of microseconds. You can use this code in your test to verify that the end time is after the start time - no need to add to RunStepResponse for now, I'll take care of that in a future PR.

```
import java.time.LocalDateTime;
import java.time.format.DateTimeFormatter;
import java.time.format.DateTimeFormatterBuilder;
import java.time.temporal.ChronoField;

public class MyTest {

    public static void main(String[] args) {
        String before = "2021-02-18T14:28:26.123456Z";
        String after = "2021-02-18T14:28:27.526055Z";

        DateTimeFormatter formatter = new DateTimeFormatterBuilder()
            .appendPattern("yyyy-MM-dd'T'HH:mm:ss")
            .appendFraction(ChronoField.MICRO_OF_SECOND, 0, 6, true)
            .appendPattern("'Z'")
            .parseLenient()
            .toFormatter();
        LocalDateTime beforeTime = LocalDateTime.parse(before, formatter);
        LocalDateTime afterTime = LocalDateTime.parse(after, formatter);
        System.out.println(beforeTime);
        System.out.println(afterTime);
        System.out.println(afterTime.isAfter(beforeTime));
    }
}
```

rjrudin (changes requested):

marklogic-data-hub/src/test/resources/test-projects/custom-step-with-sleep/modules/root/custom-modules/custom/customStepWithSleep/customStepWithSleep.sjs
```diff
@@ -0,0 +1,8 @@
+function main(contentItem, options) {
+  xdmp.sleep(100000);
```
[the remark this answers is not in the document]
jyangML: Sorry Rob.  That was me saving my current work because I was still trying to work out why Eclipse doesn't like my project.  I just updated to my v1 of the test.  Reopening the PR

marklogic-data-hub/src/test/java/com/marklogic/hub/flow/VerifyBatchTimestampTest.java
```diff
@@ -0,0 +1,43 @@
...
+        RunStepResponse stepResponse = response.getStepResponses().get("1");
+
+        assertEquals("completed step 1", stepResponse.getStatus());
+        assertEquals(1, stepResponse.getSuccessfulBatches());
+        
+        String start_time = stepResponse.getStepStartTime();
+        String end_time = stepResponse.getStepEndTime();
+        
+        // extraction out seconds value from timestamp, then round to nearest second
+        int t1 = (int) (Float.parseFloat(start_time.split(":")[2].split("-")[0]) + 0.5);
```
rjrudin, line 36: Let's just do before/after with LocalDateTime classes; that's more reliable and more readable - i.e. looking at this, I have to do some thinking to figure out why either 5 or 55 is a correct value.

jyangML: Used your other suggestion

marklogic-data-hub/src/test/java/com/marklogic/hub/flow/VerifyBatchTimestampTest.java
```diff
@@ -0,0 +1,51 @@
...
+        	writeBatcher.add(project.buildCustomerInstanceToWrite(new Customer(0, "Jane"), Format.JSON, null));
+        });
+        
+        RunFlowResponse response = runFlow(new FlowInputs("customFlowWithSleep"));
+        RunStepResponse stepResponse = response.getStepResponses().get("1");
+
+        assertEquals("completed step 1", stepResponse.getStatus());
+        assertEquals(1, stepResponse.getSuccessfulBatches());
+        
+        DateTimeFormatter formatter = new DateTimeFormatterBuilder()
```
rjrudin, line 37: Looks good - and in a future PR, I'll refactor this a bit so we have a helper method to make this easier in future tests.

rjrudin (dismissed):

marklogic-data-hub/src/test/java/com/marklogic/hub/flow/VerifyBatchTimestampTest.java
jyangML: Sounds great.  I'm reopening the PR

jyangML:
Reopened PR.  Ticket has been changed to 5.4

rjrudin (approved):

srinathgit (approved):

PR #5244, opened 2021-02-16
DHFPROD-6636: Hide data hub artifacts by default

rahulvudutala:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/entities/search/EntitySearchManager.java
+++ b/marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/entities/search/EntitySearchManager.java
@@ -53,6 +53,7 @@ import java.io.Reader;
 import java.time.ZonedDateTime;
 import java.time.format.DateTimeFormatter;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -62,6 +63,13 @@ public class EntitySearchManager {
 
     private static final String MASTERING_AUDIT_COLLECTION_NAME = "mdm-auditing";
     private static final String[] IGNORED_SM_COLLECTION_SUFFIX = {"auditing", "notification"};
+    private static final List<String> ARTIFACT_COLLECTION_NAMES = Arrays.asList(
+            "http://marklogic.com/data-hub/flow",
+            "http://marklogic.com/data-hub/step-definition",
+            "http://marklogic.com/data-hub/steps",
+            "http://marklogic.com/entity-services/models",
+            "http://marklogic.com/data-hub/mappings"
+    );
 
     private static final String CSV_CONTENT_TYPE = "text/csv";
     private static final String CSV_FILE_EXTENSION = ".csv";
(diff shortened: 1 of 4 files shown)


marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/entities/search/EntitySearchManager.java
```diff
@@ -62,6 +63,13 @@
 
     private static final String MASTERING_AUDIT_COLLECTION_NAME = "mdm-auditing";
     private static final String[] IGNORED_SM_COLLECTION_SUFFIX = {"auditing", "notification"};
+    private static final List<String> ARTIFACT_COLLECTION_NAMES = Arrays.asList(
+            "hub-core-artifact",
```
rjrudin, line 67: I don't think this should be included because it raises the question of "What's in hub-core-artifact that's not in the 4 collections listed below?"

rjrudin (approved):

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/entities/search/EntitySearchManager.java
srinathgit: Should these collections be included as well ?
 http://marklogic.com/data-hub/steps/mapping|ingestion|mastering|matching|merging, 
http://marklogic.com/data-hub/mappings .

rjrudin: The steps collection covers all of those, and "mappings" won't exist in an HC app. Though - they could exist - so probably better to exclude those just to be safe.

rahulvudutala: @srinathgit artifacts in these collections  http://marklogic.com/data-hub/steps/mapping|ingestion|mastering|matching|merging and http://marklogic.com/data-hub/mappings also have http://marklogic.com/data-hub/steps collection added

rahulvudutala: @srinathgit @rjrudin Added the collections as suggested. Also removed hub-core-artifact collection

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/entities/search/EntitySearchManager.java
```diff
@@ -62,6 +63,18 @@
 
     private static final String MASTERING_AUDIT_COLLECTION_NAME = "mdm-auditing";
     private static final String[] IGNORED_SM_COLLECTION_SUFFIX = {"auditing", "notification"};
+    private static final List<String> ARTIFACT_COLLECTION_NAMES = Arrays.asList(
+            "http://marklogic.com/data-hub/flow",
+            "http://marklogic.com/data-hub/step-definition",
+            "http://marklogic.com/data-hub/steps",
+            "http://marklogic.com/entity-services/models",
+            "http://marklogic.com/data-hub/mappings",
+            "http://marklogic.com/data-hub/steps/mapping",
```
rjrudin, line 72: We don't need all of these other step collections - we only need "steps".

We do need "mappings" because legacy mappings - which are not in steps - may exist, and they shouldn't show up either.

rjrudin (changes requested):

srinathgit (dismissed):

rjrudin (dismissed):

srinathgit (approved):

xnikhil08 (dismissed):
Its ok for now to comment out the failing test, which will be fixed in my upcoming PR.

xnikhil08 (approved):
Its ok for now to comment out the failing test, which will be fixed in my upcoming PR.

srinathgit (approved):

xnikhil08 (approved):

PR #5259, opened 2021-02-18
DHFPROD-6397: Retrieve Pendo key from cloud service

akshaysonvane:
### Description
- Introduced two new properties:
```
hubPendoEnabled = true/false <If true HC will try to retireve the pendo api key>
hubRetrieveCertificate = true/false <if true HC will try to retrieve the certificates keystore and password>
```
- Earlier the certificates keystore and password retrieval was based on just the spring profile but now it depends on `hubRetrieveCertificate` property as we need to support the following service combinations:
1. Services with no certificate and no Pendo
2. Services with a certificate but no Pendo
3. Services with no certificate but with Pendo
4. Services with both certificate and Pendo

- DHS now always needs to apply the appropriate spring profile using `spring.profiles.active=aws/azure` and then can use the `hubPendoEnabled/hubRetrieveCertificate` properties to configure Pendo and certificate retrieval. This change will also help us to align with the on-prem HC release where cloud env will use aws/azure profiles and on-prem will use production profile.

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/build.gradle
+++ b/marklogic-data-hub-central/build.gradle
@@ -78,10 +78,8 @@ dependencies {
     implementation "com.squareup.okhttp3:okhttp:4.7.2"
     implementation "com.squareup.okhttp3:logging-interceptor:4.7.2"
 
-    // Excluding okhttp3 dependency as it bumps down the dependency from 4.4.0 to 3.x
-    implementation ('com.microsoft.azure:azure-keyvault-secrets-spring-boot-starter:2.2.4') {
-        exclude group: "com.squareup.okhttp3", module: "okhttp"
-    }
+    implementation 'com.azure:azure-identity:1.2.0'
+    implementation 'com.azure:azure-security-keyvault-secrets:4.2.3'
     implementation 'com.azure:azure-storage-blob:12.8.0'
     implementation 'com.amazonaws:aws-java-sdk-s3:1.11.847'
     implementation 'com.amazonaws:aws-java-sdk-ssm:1.11.847'
--- a/marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/cloud/aws/AWSConfiguration.java
+++ b/marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/cloud/aws/AWSConfiguration.java
@@ -17,20 +17,23 @@
 package com.marklogic.hub.central.cloud.aws;
 
 import com.amazonaws.services.s3.AmazonS3ClientBuilder;
-import com.amazonaws.services.simplesystemsmanagement.AWSSimpleSystemsManagementClientBuilder;
-import com.amazonaws.services.simplesystemsmanagement.model.GetParameterRequest;
 import com.marklogic.hub.central.web.SslUtil;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.boot.autoconfigure.web.ServerProperties;
 import org.springframework.boot.web.embedded.tomcat.TomcatServletWebServerFactory;
 import org.springframework.boot.web.server.WebServerFactoryCustomizer;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.Primary;
 import org.springframework.context.annotation.Profile;
 
 import java.io.InputStream;
+
+import static com.marklogic.hub.central.cloud.Constants.CERTIFICATE_KEY_STORE_PASSWORD;
 
 @Configuration
 @Profile("aws")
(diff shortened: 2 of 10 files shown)


marklogic-data-hub-central/build.gradle
```diff
@@ -78,10 +78,8 @@ dependencies {
     implementation "com.squareup.okhttp3:okhttp:4.7.2"
     implementation "com.squareup.okhttp3:logging-interceptor:4.7.2"
 
-    // Excluding okhttp3 dependency as it bumps down the dependency from 4.4.0 to 3.x
-    implementation ('com.microsoft.azure:azure-keyvault-secrets-spring-boot-starter:2.2.4') {
-        exclude group: "com.squareup.okhttp3", module: "okhttp"
-    }
```
akshaysonvane, line 84: Replaced this with azure's identity and keyvault secret java libraries as the earlier library was injecting the value from the key directly using `@Value`. 
This is not required in the case where both hubPendo and hubRetrieveCertificate is set to false.

rjrudin (changes requested):
Providing ideas on Slack

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/controllers/EnvironmentController.java
```diff
@@ -86,12 +92,14 @@ public SystemInfo getSystemInfo(HttpSession session) {
...
         info.marklogicVersion = versionInfo.getMarkLogicVersion();
         info.host = hubCentral.getHost();
-        info.pendoKey = "dummy-key-f7d836d7-2afa-40a6-44e4-f32f58cb35a";
         info.sessionTimeout = environment.getProperty("server.servlet.session.timeout");
         Object hubCentralSessionToken = session.getAttribute("hubCentralSessionToken");
         if (hubCentralSessionToken != null) {
             info.sessionToken = hubCentralSessionToken.toString();
         }
+        if (pendoKeyProvider != null) {
+            info.pendoKey = pendoKeyProvider.getPendoKey(PENDO_API_KEY);
```
rjrudin, line 101: I don't think an argument should be needed here - we want EC to be as "dumb" as possible; it shouldn't know what the param name is or even that one is required. So the PKP interface should just be "getPendoKey()".

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/cloud/azure/AzureKeyVaultSecret.java
```diff
@@ -0,0 +1,70 @@
...
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.context.annotation.Profile;
+import org.springframework.stereotype.Component;
+
+import java.util.HashMap;
+import java.util.Map;
+
+@Component
+@Profile("azure")
+public class AzureKeyVaultSecret {
```
rjrudin, line 33: I think "AzureKeyVault" is sufficient for a name; if "Secret" were used, it seems it should really be "SecretClient", but I don't think that's necessary to add here. "AzureKeyVault" implies what this class represents, which is an interface for getting parameter values from an Azure key vault.

rjrudin (dismissed):
I think this is good for 5.4.0, but for 5.5.0, I'm thinking we should have a blend of what you initially had and what I suggested. I think a PendoKeyProvider class makes sense, but it can be a concrete class. It would autowire in a ParameterSource object (I don't think that needs to be cloud-specific though), which AWSParameterStore and AzureKeyValue would both implement. PKP would then depend on Environment to do the hubPendoEnabled check, and if that's true, it would call ParameterSource.getParameter("the pendo param name"). PKP could also log what's going on. 

I think that design gets the responsibilities organized - EC doesn't do anything other than ask for a pendo key; PKP checks if pendo is enabled and also logs stuff; a ParameterSource abstraction that handles getting the PKP key. 

You can make those changes for 5.4, but I think it's fine to go forward with what we have here.

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/cloud/azure/AzureKeyVaultSecret.java
akshaysonvane: Azure KeyVault service can be used to store keys, secrets, and certificates. This class will only retrieve a "secret" hence the name.

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/controllers/EnvironmentController.java
akshaysonvane: Updated PR

akshaysonvane:
> I think this is good for 5.4.0, but for 5.5.0, I'm thinking we should have a blend of what you initially had and what I suggested. I think a PendoKeyProvider class makes sense, but it can be a concrete class. It would autowire in a ParameterSource object (I don't think that needs to be cloud-specific though), which AWSParameterStore and AzureKeyValue would both implement. PKP would then depend on Environment to do the hubPendoEnabled check, and if that's true, it would call ParameterSource.getParameter("the pendo param name"). PKP could also log what's going on.
> 
> I think that design gets the responsibilities organized - EC doesn't do anything other than ask for a pendo key; PKP checks if pendo is enabled and also logs stuff; a ParameterSource abstraction that handles getting the PKP key.
> 
> You can make those changes for 5.4, but I think it's fine to go forward with what we have here.

@rjrudin made the suggested changes.

rjrudin (dismissed):

ryanjdew (approved):

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/cloud/aws/AWSConfiguration.java
```diff
@@ -17,63 +17,54 @@
...
     private String bucketName;
 
-    @Value("${aws.s3.keyName}")
+    @Value("${aws.s3.keyName:}")
     private String keyName;
 
-    private String keyStorePassword;
-
     @Bean
+    @Primary
```
akshaysonvane, line 54: Needed to make this primary since Spring 2.3.8 was bringing in a new ServerProperties bean.

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/cloud/azure/AzureConfiguration.java
```diff
@@ -21,42 +21,54 @@
...
-    @Value("${azure.storage.containerName}")
+    @Value("${azure.storage.containerName:}")
     private String containerName;
 
-    @Value("${azure.storage.blobName}")
+    @Value("${azure.storage.blobName:}")
     private String blobName;
 
     @Bean
+    @Primary
```
akshaysonvane, line 58: Same as above.

rjrudin (approved):

ryanjdew (approved):

PR #5261, opened 2021-02-19
DHFPROD-5968: Fixing Jackson dependency issues

rjrudin:
### Description

This combines DHFPROD-4854 to remove the Spring Boot dependency from core and bump up SB in HC to 2.3.8, thus ensuring that HC uses Jackson 2.11, just like the rest of DHF (minus QS), which fixes the 5968 bug. 
Temporarily removing some projects

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [ ] Reviewed Tests


rjrudin:
Don't need reviews on this yet, haven't figured out why Jenkins is failing when running clientJar (doesn't happen locally anymore).

PR #5262, opened 2021-02-19
DHFPROD-5968: Fixing CSV ingest issue via Spring Boot upgrade

rjrudin:
### Description

Was curiously not able to reproduce this in a test. But in manual testing, with Jackson 2.10 used via Spring Boot 2.2.x, ingest of e.g. 1000 rows in a CSV file consistently did not ingest everything (931 rows would be returned). Once Jackson 2.11 is used, the problem goes away. 

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/build.gradle
+++ b/marklogic-data-hub-central/build.gradle
@@ -4,7 +4,7 @@ buildscript {
 		maven { url "https://plugins.gradle.org/m2/" }
 	}
 	dependencies {
-		classpath "org.springframework.boot:spring-boot-gradle-plugin:2.2.8.RELEASE"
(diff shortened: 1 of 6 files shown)


ryanjdew (dismissed):

srinathgit (commented):
Should an ingestion test be added with the csv in the ticket and confirm count is right ?

srinathgit (approved):

rjrudin:
@srinathgit  Good point - I had written a test, but it didn't fail, so I didn't include it. I've added it back in.

akshaysonvane (approved):

PR #5270, opened 2021-02-22
DHFPROD-6739: updated nodejs version for palamida report

jyangML:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [NA] Added Tests
  

- ##### Reviewer:

- [NA] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/build.gradle
+++ b/marklogic-data-hub-central/build.gradle
@@ -150,10 +150,10 @@ task generateSchemaPojos {
 
 node {
     // Version of node to use.
-    version = '12.18.3'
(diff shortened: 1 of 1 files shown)


rjrudin (commented):
@wooldridge I don't know the impact here, but I'm assuming that these are only used for building the HC UI files? Which means if we can still build and run HC without the build process blowing up, this seems safe to me.

ryanjdew (approved):

wooldridge (approved):
Builds fine, unit tests pass.

PR #5276, opened 2021-02-23
DHFPROD-6962: Only enable hubDhs for aws/azure spring profile

akshaysonvane:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -53,7 +53,7 @@ def dhfCypressE2ETests(String mlVersion, String type){
                     cd marklogic-data-hub-central/ui/e2e;
                     chmod +x setup.sh;
                     ./setup.sh dhs=false mlHost=localhost;
-                    nohup java -jar $WORKSPACE/$WAR_NAME --hubUseLocalDefaults=true >> nohup.out &
+                    nohup java -jar $WORKSPACE/$WAR_NAME >> nohup.out &
                     sleep 10s;
                     mkdir -p output;
                     docker build . -t cypresstest;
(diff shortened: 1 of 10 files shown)


srinathgit (commented):
1. Will the property 'hubUseLocalDefaults' be relevant after this change ? If not, can that and related code be removed ?
2. Should 'hubSsl' be removed from  in marklogic-data-hub-central/gradle.properties ?

akshaysonvane:
> 1. Will the property 'hubUseLocalDefaults' be relevant after this change ? If not, can that and related code be removed ?
> 2. Should 'hubSsl' be removed from  in marklogic-data-hub-central/gradle.properties ?

yes, and yes. Removed both.

rjrudin (approved):
Looks good! I can QA it tomorrow morning

rjrudin (changes requested):
Can you make one change - gradle.properties has a comment about "production" being the default profile, that needs to be updated to say that Spring Boot defaults to "default" for the profile.

SameeraPriyathamTadikonda:
@akshaysonvane what values should I set if I'm pointing HC on local to ML on DHS.
hubDHS=true?

akshaysonvane:
> Can you make one change - gradle.properties has a comment about "production" being the default profile, that needs to be updated to say that Spring Boot defaults to "default" for the profile.

Good catch, updated.

rjrudin:
@SameeraPriyathamTadikonda  All you'd need is -PspringProfiles=aws, which selects application-aws.properties, which configures hubDhs=true.

akshaysonvane:
> @akshaysonvane what values should I set if I'm pointing HC on local to ML on DHS.
> hubDHS=true?

yes, that's right.
Also, hubDhs enables hubSsl as well.
So based on the service type you may need to explicitly turn off hubSsl via `hubSsl=false`

rjrudin (approved):

srinathgit (approved):

SameeraPriyathamTadikonda (approved):

ryanjdew (approved):

PR #5281, opened 2021-02-24
DHFPROD-6693:Updates to step schema to support related entity mappings

srinathgit:
### Description
Initial draft of updates to step schema to support related entity mappings
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [n/a ] Code passes ESLint tests
- [n/a ] Added Tests
  

- ##### Reviewer:

- [n/a ] Reviewed Tests


specs/models/Step.schema.json
```diff
@@ -132,6 +132,56 @@
     },
     "name": {
       "type": "string"
+    },
+    "properties": {
```
rjrudin, line 136: I think it'd be better to make a MappingStep.schema.json that uses the example at the end of this comment - https://github.com/json-schema-org/json-schema-spec/issues/348#issuecomment-322940347 - to grab the Step properties via "allOf". We can do likewise for the other concrete types of steps when we define schemas for those.

specs/models/Step.schema.json
```diff
@@ -132,6 +132,56 @@
...
+      "type": "object",
+      "$ref": "./MappingProperties.schema.json"
+    },
+    "expressionContext": {
```
rjrudin, line 140: I don't think we need this - i.e. I don't think we have a requirement yet for a user to define a top-level context.

specs/models/Step.schema.json
```diff
@@ -132,6 +132,56 @@
...
+          "permissions": {
+            "type": "string",
+            "description": "Comma-delimited string of role,capability,role,capability,etc for the entity instance generated by this mapping"
+          },
+          "targetEntityType": {
+            "type": "string",
+            "description": "The identifier of the related Entity Type. (IRI, with title as fallback)"
+          }
+        },
+        "required": ["relatedEntityMappingId", "properties", "uriExpression"]
```
rjrudin, line 183: I think targetEntityType has to be required as well . I don't think uriExpression has to be required though - the mapper can always use a default approach if no URI expression is provided.

rjrudin (changes requested):

specs/models/Step.schema.json
srinathgit: If you consider an entity model 'User' with one property “name” of type "string". If we want to map it to the source file, 
```
{
	"names": ["A", "B", "C", "D"]
}
```
so as to generate 4 different entity instances corresponding to each of the names, we would want "expressionContext" set to "/names" so that we get a mapping like this and 4 entity instances can be created. Currently, we set the "expressionContext" to "/" always at top level

```
<m:mapping xmlns:map="http://marklogic.com/xdmp/map" xmlns:m="http://marklogic.com/entity-services/mapping">
  <m:entity name="User">
    <User xmlns="" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
          <m:optional><name xsi:type="xs:integer"><m:val>.</m:val></name></m:optional>
    </User>
  </m:entity>
  <m:output>
    <m:for-each>
      <m:select>/names</m:select>
        <m:call-template name="User"/>
    </m:for-each>
  </m:output>
</m:mapping>
```

rjrudin: I believe that requirement is tangential to the "Map related entities" feature; that requirement is really part of a "Map multiple entities of the target entity type" feature which I don't believe is part of 5.5 yet. It could be; but it's not part of "Map related entities", so I don't think we should try to support it yet.

specs/models/Step.schema.json
rjrudin: I think that's okay for now. We can see if a springfox upgrade fixes it later.

rjrudin (approved):

ryanjdew (dismissed):

rjrudin (approved):

ryanjdew (approved):

PR #5291, opened 2021-02-26
DHFPROD-6941: Add new Job Monitor roles

AkshayBajajML:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/resources/hub-internal-config/security/roles/data-hub-job-monitor.json
+++ b/marklogic-data-hub/src/main/resources/hub-internal-config/security/roles/data-hub-job-monitor.json
@@ -1,0 +1,14 @@
+{
+  "role-name": "data-hub-job-monitor",
+  "description": "Permits access to job monitoring features in DHF",
+  "role": [
+    "data-hub-job-reader"
+  ],
+  "privilege": [
+    {
+      "privilege-name": "data-hub-monitor-jobs",
+      "action": "http://marklogic.com/data-hub/privileges/monitor-jobs",
+      "kind": "execute"
+    }
+  ]
+}
--- a/marklogic-data-hub/src/main/resources/hub-internal-config/security/privileges/data-hub-monitor-jobs.json
+++ b/marklogic-data-hub/src/main/resources/hub-internal-config/security/privileges/data-hub-monitor-jobs.json
@@ -1,0 +1,6 @@
+{
+  "privilege-name": "data-hub-monitor-jobs",
+  "action": "http://marklogic.com/data-hub/privileges/monitor-jobs",
+  "kind": "execute"
+}
+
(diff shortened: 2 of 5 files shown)


ryanjdew (approved):

rjrudin (commented):
@AkshayBajajML  @ryanjdew  Should this still be done given the new direction for the Jobs team? I'd rather hold off on this if we're going to set the Jobs work aside for now.

ryanjdew:
@rjrudin I'd prefer this be merged in. It is low-impact and we still haven't determined that we won't do any Jobs work for 5.5 yet. We'll definitely hold off on future feature tickets for the moment.

marklogic-data-hub/src/main/resources/hub-internal-config/security/roles/data-hub-job-monitor.json
```diff
@@ -0,0 +1,14 @@
+{
+  "role-name": "data-hub-job-monitor",
+  "description": "This role grants access to read job monitoring related data",
```
rjrudin, line 3: Is the intent of this role to distinguish between users that can read job data for the purpose of e.g. viewing records and when they've been processed by steps, vs users that want to perform "monitoring" actions that are should not be accessible to all of the former group of users?

I ask because if so, I don't think this description really conveys that. "read job monitoring related data" is already provided by the data-hub-job-reader role.

I am guessing the description should be something like "Permits access to job monitoring features in DHF" (note that it should start with "Permits" for consistency with existing DHF roles). That conveys that - hey, just because you can read job data doesn't mean you're able to access job monitoring features; you're going to need this role if you want to do that.

rjrudin (changes requested):
I'll add a comment to the JIRA ticket about this - I am wondering if data-hub-monitor can be reused here (so add the privilege, and add it to that role), as that already has a description of "Permits monitoring Data Hub activity, such as viewing job data and system logs".

marklogic-data-hub/src/main/resources/hub-internal-config/security/privileges/data-hub-job-monitor.json
```diff
@@ -0,0 +1,6 @@
+{
+  "privilege-name": "data-hub-job-monitor",
```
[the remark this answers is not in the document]
AkshayBajajML: Updated privilege name to data-hub-monitor-jobs

marklogic-data-hub/src/main/resources/hub-internal-config/security/roles/data-hub-job-monitor.json
AkshayBajajML: Updated description

ryanjdew (approved):

rjrudin (approved):

PR #5300, opened 2021-03-01
DHFPROD-6947: Move the source navigation buttons in mapping UI to a s…

timur-isangulov:
…eparate component

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/src/components/entities/mapping/mapping-card.tsx
+++ b/marklogic-data-hub-central/ui/src/components/entities/mapping/mapping-card.tsx
@@ -71,11 +71,6 @@ const MappingCard: React.FC<Props> = (props) => {
   //For storing docURIs
   const [docUris, setDocUris] = useState<any[]>([]);
 
-  //For handling docUris navigation
-  const [disableURINavLeft, setDisableURINavLeft] = useState(true);
-  const [disableURINavRight, setDisableURINavRight] = useState(false);
-
-
   //For storing  mapping functions
   const [mapFunctions, setMapFunctions] = useState([]);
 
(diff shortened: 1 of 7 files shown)


wooldridge (changes requested):
This is great. If it isn't happening elsewhere, can we test buttons clicks, similar to what we do here? https://github.com/marklogic/marklogic-data-hub/blob/develop/marklogic-data-hub-central/ui/src/components/common/confirm-yes-no/confirm-yes-no.test.tsx#L25

Also check that the right index value is passed on a button click?

timur-isangulov:
> This is great. If it isn't happening elsewhere, can we test buttons clicks, similar to what we do here? https://github.com/marklogic/marklogic-data-hub/blob/develop/marklogic-data-hub-central/ui/src/components/common/confirm-yes-no/confirm-yes-no.test.tsx#L25
> 
> Also check that the right index value is passed on a button click?

Added tests, can you please review again. Thanks.

wooldridge (changes requested):
@timur-isangulov, thanks! Here's the test modified slightly to check the function arg as well:
```
  test("Verify handleSelection is called with correct index", () => {
    const handleSelectionMock = jest.fn();
    const {getByTestId} = render(<SourceNavigation currentIndex={1} startIndex={0} endIndex={2} handleSelection={handleSelectionMock} />);

    fireEvent.click(getByTestId("navigate-uris-left"));
    expect(handleSelectionMock).toHaveBeenCalledTimes(1);
    expect(handleSelectionMock).toHaveBeenLastCalledWith(0);

    fireEvent.click(getByTestId("navigate-uris-right"));
    expect(handleSelectionMock).toHaveBeenCalledTimes(2);
    expect(handleSelectionMock).toHaveBeenLastCalledWith(2);
  });
```

wooldridge (changes requested):
I think we can streamline the logic more if we can assume the following constraint:

startIndex <= currentIndex <= endIndex

I made a PR to your branch with some suggested changes: https://github.com/timur-isangulov/marklogic-data-hub/pull/1

wooldridge (approved):

xnikhil08 (approved):
Looks good.

PR #5312, opened 2021-03-03
DHFPROD-6606: ML is allowing data-hub-security-admin can create role with xdmp:shutdown privilege

devinnicholson:
### Description
This is a test created to verify the bug fix for DHFPROD-6606. 

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- N/A Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/test/java/com/marklogic/hub/security/DataHubSecurityAdminTest.java
+++ b/marklogic-data-hub/src/test/java/com/marklogic/hub/security/DataHubSecurityAdminTest.java
@@ -5,6 +5,7 @@ import com.marklogic.mgmt.api.security.Privilege;
 import com.marklogic.mgmt.api.security.Role;
 import com.marklogic.mgmt.api.security.RolePrivilege;
 import com.marklogic.mgmt.resource.security.PrivilegeManager;
+import com.marklogic.hub.MarkLogicVersion;
 import org.junit.jupiter.api.Assumptions;
 import org.junit.jupiter.api.Test;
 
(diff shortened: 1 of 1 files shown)


marklogic-data-hub/src/test/java/com/marklogic/hub/security/DataHubSecurityAdminTest.java
```diff
@@ -271,4 +272,28 @@ void task5CreateRoleWithCustomPrivilegeInDisallowedNamespace() {
...
+
+        try {
+            customRole.save();
+            fail("The role creation should have failed because a DHSA user shouldn’t be able to create a custom role with xdmp:shutdown privilege");
+        }
+        catch (Exception ex) {
+            logger.info("Caught expected exception: " + ex.getMessage());
+
+            // Just in case the role was created (role was being created even after exception thrown -> bug: 55765).
+            assertNull(customRole.getRole());
```
rjrudin, line 294: This is a good check to make, but what this is currently doing is verifying that the customRole doesn't inherit any roles. What we want to do instead is verify that the role was not created:

```
assertFalse(customRole.exists(), "The role should not have been created");
```

The exists() call will under the hood make a call to the Manage API to see if the resource exists or not.

marklogic-data-hub/src/test/java/com/marklogic/hub/security/DataHubSecurityAdminTest.java
```diff
@@ -271,4 +272,28 @@ void task5CreateRoleWithCustomPrivilegeInDisallowedNamespace() {
...
+        Assumptions.assumeTrue(version.getMajor() >= 10 &&
+            (version.isNightly() || version.getMinor() >= 601), "Bug<55765> is fixed in 10.0-6.1");
+
+        final String roleName = "custom-role-with-xdmp-shutdown-privilege";
+        Role customRole = new Role(userWithRoleBeingTestedApi, roleName);
+        customRole.addPrivilege(new RolePrivilege("xdmp:shutdown", "http://marklogic.com/xdmp/privileges/xdmp-shutdown", "execute"));
+
+        try {
+            customRole.save();
+            fail("The role creation should have failed because a DHSA user shouldn’t be able to create a custom role with xdmp:shutdown privilege");
```
rjrudin, line 288: This is good - assertion messages are critical because they explain why something is expected (an assertion message that just restates what the assertion itself is stating doesn't add any value).

rjrudin (changes requested):

marklogic-data-hub/src/test/java/com/marklogic/hub/security/DataHubSecurityAdminTest.java
devinnicholson: Thanks Rob! This is what I was intending to do.

srinathgit (approved):

rjrudin (approved):

PR #5326, opened 2021-03-05
DHFPROD-6702:Map multiple entities from one json source doc

srinathgit:
### Description
Map multiple entities from one json source doc
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


rjrudin:
Will try this out locally

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/builtins/steps/mapping/map-multiple-entities/createRelatedEntityMapping.sjs
```diff
@@ -0,0 +1,186 @@
...
+            <m:call-template name="mapping2-Order" />
+         </m:for-each>
+      </entityInstance2>
+   </m:output>
+</m:mapping>
+
+`);
+
+template = constructRelatedEntityMapping();
+assertions.push(
```
rjrudin, line 182: I recommend adding more assertions to explain why certain lines are what they are. Doing a full-XML comparison is useful to verify correctness, but it doesn't provide any explanation as to why it's correct. I think assertions on each entity name would be useful to explain why they are what they are.

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/builtins/steps/mapping/map-multiple-entities/runMapping.sjs
```diff
@@ -0,0 +1,107 @@
...
+  let instance5 = result[5].value.root.envelope.instance;
+
+  try {
+    assertions = assertions.concat([
+      test.assertEqual(202, fn.number(instance0.Customer.customerId)),
+      test.assertEqual("Cynthia", fn.string(instance0.Customer.firstName)),
+      test.assertEqual('Waters', fn.string(instance0.Customer.lastName))
+    ]);
+  } catch (e) {
+    assertions.push(
```
rjrudin, line 25: This seems unusual - is the problem it's trying to solve that you can't see the instance that failed? If so, I think the better way to do it is this (pseudocoding):

```
const message = "instance: " + xdmp.describe(instance0);
test.assertEqual(202, fn.number(...), "customerId is not 202 in " + message);
```

That way you know exactly which assertion failed, along with why, and what the instance actually is.

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/builtins/steps/mapping/map-multiple-entities/runMapping.sjs
```diff
@@ -0,0 +1,107 @@
...
+      test.assertFalse(
+        fn.true(),
+        `Error "${e.toString()}" encountered testing instance5 '${xdmp.describe(instance5, Sequence.from([]), Sequence.from([]))}'`
+      )
+    );
+  }
+}
+runMapping();
```
rjrudin, line 106: Since there's only the one test (which is a good practice), let's remove the function - just have this all in the "main" part of the module, no need for a function.

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/builtins/steps/mapping/map-multiple-entities/suite-setup.xqy
```diff
@@ -0,0 +1,19 @@
...
+hub-test:load-entities($test:__CALLER_FILE__)
+;
+
+xquery version "1.0-ml";
+
+import module namespace hub-test = "http://marklogic.com/data-hub/test" at "/test/data-hub-test-helper.xqy";
+import module namespace test = "http://marklogic.com/test" at "/test/test-helper.xqy";
+
+hub-test:load-non-entities($test:__CALLER_FILE__),
+hub-test:clear-jobs-database()
```
rjrudin, line 19: No need for this, reset-hub calls clear-jobs-database too.

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/builtins/steps/mapping/map-multiple-entities/test-data/entities/Order.entity.json
```diff
@@ -0,0 +1,48 @@
...
+        "lineItems": {
+          "datatype": "array",
+          "items": {
+            "$ref": "#/definitions/LineItem"
+          }
+        }
+      }
+    },
+    "LineItem": {
+      "required": [],
```
rjrudin, line 31: To keep these tests as simple as possible (and thus easier to understand), remove these empty arrays, they're just adding noise.

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/builtins/steps/mapping/map-multiple-entities/writeEntityInstances.sjs
```diff
@@ -0,0 +1,42 @@
+const DataHubSingleton = require("/data-hub/5/datahub-singleton.sjs");
```
rjrudin, line 1: Since this test is running a flow, let's name this test "runFlowWithRelatedEntityMapping.sjs" - "writeEntityInstances" doesn't imply that a flow is being run.

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/builtins/steps/mapping/map-multiple-entities/writeEntityInstances.sjs
```diff
@@ -0,0 +1,42 @@
...
+  won't get applied before runFlow() is called. */
+  test.assertEqual("data-hub-operator,read,data-hub-operator,update", getDocPermissions(productUris[0])),
+  test.assertEqual("data-hub-common,read,data-hub-common,update", getDocPermissions(orderUris[0]))
+];
+
+function getUris(collection){
+  return fn.head(xdmp.eval('cts.uris(null, null, cts.collectionQuery("'+ collection+ '")).toArray()'))
+}
+
+function getDocPermissions(uri){
```
rjrudin, line 35: Use getRecord in data-hub-test-helper.sjs, it gets you all this data in an easy-to-consume way.

rjrudin (changes requested):
This looks good, I'm going to test it out locally as well. Had mostly cleanup-ish requests for changes.

rjrudin:
I think some trace events would be helpful for our own debugging here too. We have "hub-mapping", but I'm thinking "hub-mapping-debug" would be the right event for capturing the following things:

# The output of xqueryLib.dataHubMapToCanonical (arrayOfInstancesArray)
# Each entityContent object that's outputted 

Use hubUtils.hubTrace for this.

rjrudin (dismissed):
Looks good, I can do QA for it once it's in develop too

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mapping/entity-services/lib.sjs
```diff
@@ -46,7 +46,66 @@ function buildMappingXML(mappingDoc) {
...
+  }
+
+  let xml =
+    `<m:mapping xmlns:m="http://marklogic.com/entity-services/mapping" xmlns:map="http://marklogic.com/xdmp/map" ${namespaces.join(' ')}>
+      ${retrieveFunctionImports()}
+      <m:param name="URI"/>
+      ${entityTemplates}
+      <m:output>
+      ${allEntityMap.map((entityMap, index) =>
+        `<entityInstance${index}>
```
ryanjdew, line 77: I would recommend adding a specific namespace for these elements

srinathgit:
Changes since last commit:
1. Added a namespace "http://marklogic.com/datahub/entityInstance" for entity instances and updated the xpath
2. Made minor changes to impl/flow.sjs so that custom hooks work fine.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/flow.sjs
```diff
@@ -349,9 +349,10 @@ class Flow {
     if (!hook || !hook.module) {
       hook = processor.customHook;
     }
+    let parameters;
```
rjrudin, line 352: This is a change in how custom hooks work, and I've wondered before why it doesn't work this way - i.e. I think this change is correct. However, I think it should be handled via a separate ticket so we can think through what tests are needed for this.

rjrudin: The ticket would have a scope of verifying that a "after" custom hook has access to the content array that's returned by the step module. Again, that's probably how it always should have worked. But we should treat the change as a bug as it is changing the behavior in a way that could be view as a backwards incompatibility, but in way that we think was intended all along.

rjrudin (changes requested):
Looks good to me, I just had the comment about making the custom hook changes in a separate ticket.

ryanjdew (dismissed):

srinathgit:
Changes since last commit:

1. Removed changes to custom hook
2. Commented out an cypress test that runs a mapping step with interceptor and custom hook. I think the test is passing today due to a bug. Somehow the 'content' that is passed to custom hook has mapped documents. If the interceptor is removed, custom hook only gets the 'content' that is available prior to running the mapping main.sjs. 
3. Used 'useContextCollectionsOnly' as so that context.collections is not concatenated with step collections.

rjrudin (approved):

ryanjdew (approved):

PR #5337, opened 2021-03-10
DHFPROD-6526: Job times are incorrect when reusing job ID

AkshayBajajML:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/jobs.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/jobs.sjs
@@ -316,6 +316,10 @@ module.exports.updateJob = module.amp(
         throw new Error("Cannot create job document. Incorrect options.");
       }
     }
+    else if(jobDoc.job.timeEnded) {
+      jobDoc.job.timeStarted = fn.currentDateTime();
+      jobDoc.job.timeEnded = null;
+    }
     //update job status at the end of flow run
     if(status === "finished"|| status === "finished_with_errors" || status === "failed"|| status === "canceled"|| status === "stop-on-error") {
       jobDoc.job.timeEnded = fn.currentDateTime();
@@ -326,6 +330,8 @@ module.exports.updateJob = module.amp(
       jobDoc.job.lastAttemptedStep = step;
       if(lastCompleted) {
        jobDoc.job.lastCompletedStep = lastCompleted;
+       jobDoc.job.stepResponses[step].stepStartTime = fn.currentDateTime();
+       jobDoc.job.stepResponses[step].stepEndTime = null;
       }
       if(!jobDoc.job.stepResponses[step]){
        jobDoc.job.stepResponses[step] = {};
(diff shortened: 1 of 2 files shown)


ryanjdew (changes requested):
We'll want to add a test for this. A JUnit test would likely be the best option. It could be added to /marklogic-data-hub/src/test/java/com/marklogic/hub/job/JobDocManagerTest.java

rjrudin:
I think a JUnit test is likely best here. The flow can be extremely simply; a single step that processes a single document will work fine (it may not even need to process any documents). Need to make sure that the start/end times on the overwritten Job doc are not the same as the first Job doc that was created. See https://wiki.marklogic.com/display/ENGINEERING/Writing+Data+Hub+Tests for guidance on writing a JUnit test.

ryanjdew (dismissed):

marklogic-data-hub/src/test/java/com/marklogic/hub/job/JobDocManagerTest.java
```diff
@@ -32,6 +32,24 @@ public void testGetLatestJobDocumentForFlows() {
         Assertions.assertEquals("newestJob", latestJobsForFlow.get("job").get("jobId").asText(), "The latest JobId for 'hub' should be 'newestJob'");
     }
 
+    @Test
+    public void testJobTimesForSameJobId() {
```
rjrudin, line 36: For this test, we really need to run a flow, not just make arbitrary calls to create a job. 

To do that, you'll need to install a project first; there are a bunch of them under src/test/resources/test-projects. 

A good simple one is "simple-customer-flow":

```
installProjectInFolder("test-projects/simple-customer-flow");
```

You can then run a flow like this, which runs step "1" (and you don't need it to process any data:

runFlow(new FlowInputs("simpleCustomerFlow", "1").withJobId("abc123"));

After running the flow, you can use jobDocManager to get the job doc and grab the times on it. Then you can run the flow again with the same jobId, and compare the times.

Also, the test is difficult to read right now because there's no separation between the lines of code. Think about the steps in your test here:

- Run a flow
- Get the job times
- Run the flow again
- Get the job times again
- Compare the job times

So you have 5 steps to your test - that suggests that 5 blocks of code (i.e. with blank lines in between them) is a good starting point for making the test easier to read.

rjrudin: Also, let's put the test into com.marklogic.hub.flow - make a new test class called "RunFlowWithSameJobIdTest", as that's the exact thing we're testing.

rjrudin (changes requested):

marklogic-data-hub/src/test/java/com/marklogic/hub/job/JobDocManagerTest.java
AkshayBajajML: Thank you Rob for the comments and detailed guidance for the test to be added.
I have added test RunFlowWithSameJobIdTest as mentioned.

rjrudin (approved):
Can you assign this one to me for QA? I want to do some refactoring with VerifyBatchTimestampTest, which has code in it for verifying that a timestamp is after another one. That needs to move elsewhere so that tests like this can reuse it. So I'll submit another PR after this one is merged into develop to do that refactoring. And I'll QA it while I'm at it.

AkshayBajajML:
> VerifyBatchTimestampTest

Yes sure!! I will ask it you for QA

srinathgit (approved):

PR #5338, opened 2021-03-10
DHFPROD-7006:Extract Gradle subprojects for DHS installer and client JAR

yafeldman:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [ ] Added Tests
  

- ##### Reviewer:

- [ ] Reviewed Tests


installer-for-dhs/build.gradle
```diff
@@ -2,6 +2,7 @@ plugins {
     id "java"
     id 'com.github.johnrengelman.shadow' version '5.2.0'
     id "io.snyk.gradle.plugin.snykplugin" version "0.4"
+    id 'maven-publish'
```
rjrudin, line 5: We don't use maven-publish for the HC project, so I don't think we need it for either of these projects. Both produce an executable jar. The client-jar only needs to be available on a github release page, just like the HC war file. The installer jar is only for internal use by the DHS team.

PR #5344, opened 2021-03-10
DHFPROD-6630: Upgrade to Cypress 6.0.1 to fix bad video artifacts

ngodugu-marklogic:
### Description
Upgraded Cypress version to latest version(6.6.0). The method **should("not.be.visible")** does not work for versions >= 6.0.0. Have replace it with **should("not.exist")/should("have.length", 0)/should('have.css', 'width', '0px')**
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


Sanjeevani19 (commented):
@ngodugu-marklogic I see that the cypress framework has been upgraded, Could you please add the list of deprecated methods and their replacement as a part of the description in the PR so that we are aware of new methods.

Sanjeevani19 (commented):
@ngodugu-marklogic Looks like there are few test test failures on the pipeline on different builds, Could you please confirm if those are related ?

Sanjeevani19 (approved):

timur-isangulov (approved):

PR #5363, opened 2021-03-15
DHFPROD-6957:Update example project with a flow that runs related entity mappings

srinathgit:
### Description
Update example project with a flow that runs related entity mappings
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/examples/reference-entity-model/entities/BabyRegistry.entity.json
+++ b/examples/reference-entity-model/entities/BabyRegistry.entity.json
@@ -1,0 +1,34 @@
+{
+  "info": {
+    "title": "BabyRegistry",
+    "version": "0.0.1",
+    "baseUri": "http://marklogic.com/example/"
+  },
+  "definitions": {
+    "BabyRegistry": {
+      "primaryKey": "babyRegistryId",
+      "properties": {
+        "babyRegistryId": {
+          "datatype": "integer"
+        },
+        "arrivalDate": {
+          "datatype": "date"
+        },
+        "ownedBy": {
+          "datatype": "integer",
+          "relatedEntityType": "http://example.org/Customer-0.0.1/Customer",
+          "joinPropertyName": "customerId"
+        },
+        "hasProduct": {
+          "datatype": "array",
+          "items": {
+            "datatype": "integer",
+            "relatedEntityType": "http://example.org/Product-0.0.1/Product",
+            "joinPropertyName": "customerId"
+          }
+        }
+      }
+    }
+  }
+}
+
--- a/examples/reference-entity-model/entities/Order.entity.json
+++ b/examples/reference-entity-model/entities/Order.entity.json
@@ -6,24 +6,39 @@
   },
   "definitions": {
     "Order": {
-      "required": [ ],
       "primaryKey": "orderId",
       "properties": {
         "orderId": {
-          "datatype": "string"
+          "datatype": "integer"
         },
-        "address": {
-          "$ref": "#/definitions/Address"
+        "orderDateTime": {
+          "datatype": "dateTime"
+        },
+        "orderedBy": {
+          "datatype": "integer",
+          "relatedEntityType": "http://example.org/Customer-0.0.1/Customer",
+          "joinPropertyName": "customerId"
+        },
+        "deliveredTo": {
+          "datatype": "integer"
+        },
+        "lineItems": {
+          "datatype": "array",
+          "items": {
+            "$ref": "#/definitions/LineItem"
+          }
         }
       }
     },
-    "Address": {
+    "LineItem": {
       "properties": {
-        "city": {
-          "datatype": "string"
+        "quantity": {
+          "datatype": "integer"
         },
-        "state": {
-          "datatype": "string"
+        "hasProduct": {
+          "datatype": "integer",
+          "relatedEntityType": "http://example.org/Product-0.0.1/Product",
+          "joinPropertyName": "productId"
         }
       }
     }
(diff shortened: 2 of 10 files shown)


ryanjdew (dismissed):

wooldridge (changes requested):
@srinathgit @rjrudin @sbayatpur 

When I create a related entity relationship in the UI that has multiple selected:

Screen Shot 2021-03-15 at 10 51 26 AM

This results in an entity property that looks like this:
```
"products": {
	"datatype": "array", 
	"facetable": false, 
	"sortable": false, 
	"items": {
		"datatype": "string", 
		"relatedEntityType": "http://example.org/Product-0.0.1/Product", 
		"joinPropertyName": "productId"
	}
}, 
```
Can we represent this type of relationship in the example? With the addition of the lineItems property, this type of relationship has gone away in the example, maybe we need another entity relationship in the example.

srinathgit:
> @srinathgit @rjrudin @sbayatpur
> 
> When I create a related entity relationship in the UI that has multiple selected:
> 
> Screen Shot 2021-03-15 at 10 51 26 AM
> 
> This results in an entity property that looks like this:
> 
> ```
> "products": {
> 	"datatype": "array", 
> 	"facetable": false, 
> 	"sortable": false, 
> 	"items": {
> 		"datatype": "string", 
> 		"relatedEntityType": "http://example.org/Product-0.0.1/Product", 
> 		"joinPropertyName": "productId"
> 	}
> }, 
> ```
> 
> Can we represent this type of relationship in the example? With the addition of the lineItems property, this type of relationship has gone away in the example, maybe we need another entity relationship in the example.

I have used the entity models from https://wiki.marklogic.com/display/PM/Mapping+Example+%28Data+Hub+5.5%29%3A+Map+multiple+entities+from+a+single+source . In the latest commit, I have added 'BabyRegistry' which has a property 'babyRegistryIncludes' that is an array and is related to Product.productId.

wooldridge (approved):
Looks good, entities are showing up in my updated modeling UI, e.g.:
Screen Shot 2021-03-16 at 12 53 28 AM

examples/reference-entity-model/entities/BabyRegistry.entity.json
```diff
@@ -0,0 +1,34 @@
...
+        },
+        "arrivalDate": {
+          "datatype": "date"
+        },
+        "ownedBy": {
+          "datatype": "integer",
+          "relatedEntityType": "http://example.org/Customer-0.0.1/Customer",
+          "joinPropertyName": "customerId"
+        },
+        "babyRegistryIncludes": {
```
rjrudin, line 22: Let's call this "hasProduct" so that the property identifies the type of property; "babyRegistryIncludes" is redundant because we already know this is being included by a baby registry by virtue of being a BabyRegistry property.

examples/reference-entity-model/entities/Order.entity.json
```diff
@@ -1,31 +1,46 @@
...
+    "baseUri": "http://marklogic.com/example/"
+  },
+  "definitions": {
+    "Order": {
+      "primaryKey": "orderId",
+      "properties": {
+        "orderId": {
+          "datatype": "integer"
+        },
+        "orderTime": {
```
rjrudin, line 14: Let's call this orderDateTime; across all of DHF, when possible, I'd like to use "Date" / "Time" / "DateTime" as a suffix when the type is date/time/dateTime.

examples/reference-entity-model/entities/Order.entity.json
```diff
@@ -1,31 +1,46 @@
...
+          }
+        }
+      }
+    },
+    "LineItem": {
+      "properties": {
+        "quantity": {
+          "datatype": "integer"
+        },
+        "orderIncludes": {
```
rjrudin, line 38: Same thing, let's do "includesProduct". Actually, let's do "hasProduct" - "has" is more typical here.

examples/reference-entity-model/input/jsonCustomersWithRelatedEntities/customer1WithRelatedEntities.json
```diff
@@ -0,0 +1,88 @@
...
+  ],
+  "Phone": "(740) 407-2081",
+  "Created": "04/03/2017",
+  "Status": "Old Customer",
+  "SSN": "333-33-3333",
+  "Orders": [
+    {
+      "OrderId": "2002",
+      "DateAndTime":"2020/08/14-07:45:22",
+      "FirstName": "Audrey",
```
rjrudin, line 38: Let's remove data that isn't going to be mapped here; it adds complexity without any value. I think it's good to remove some of the customer properties as well that we don't care about here - e.g. phone / created / status / ssn / email. Do keep nicknames though, as that will be interesting to include in a SQL query.

examples/reference-entity-model/steps/mapping/mapCustomersWithRelatedEntitiesJSON.step.json
```diff
@@ -0,0 +1,179 @@
...
+            "plusFour": {
+              "sourcedFrom": "substring(Postal, 7)"
+            }
+          },
+          "targetEntityType": "#/definitions/Zip"
+        }
+      },
+      "targetEntityType": "#/definitions/Address"
+    },
+    "customerSince": {
```
rjrudin, line 83: Remove these properties from the mapping too, just to simplify the mapping as much as possible so we can focus on related entities.

rjrudin (changes requested):

wooldridge:
@srinathgit, thanks for the update, and @rjrudin, great rec's. This example project is looking really good!

ryanjdew (approved):

wooldridge (approved):

PR #5364, opened 2021-03-15
DHFPROD-6413: Provenance records persisted for each step in a flow with unique id

rahulvudutala:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/prov.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/prov.sjs
@@ -44,6 +44,10 @@ class Provenance {
     return this.config.granularityLevel;
   }
 
+  _newProvId(jobId, flowId, stepType, docUri) {
+    return `${jobId + flowId + stepType + docUri}_${sem.uuidString()}`;
+  }
+
   /**
    * Get array of provTypes for a given step type for provenance record creation
    * @param {string} stepDefinitionType - step definition type ['ingestion','mapping','mastering','matching','merging','custom']
@@ -241,7 +245,7 @@ class Provenance {
    *      - location (doc URI)
    */
   _createIngestionStepRecord(jobId, flowId, stepName, stepDefinitionName, docURI, info) {
-    let provId = `${jobId + flowId + 'ingestion' + docURI}`;
+    let provId = this._newProvId(jobId, flowId, 'ingestion', docURI);
     let provTypes = ['ps:Flow','ps:Entity','dhf:Entity','dhf:IngestionStep','dhf:IngestionStepEntity'];
     if (info && info.status)
       provTypes.push('dhf:Doc' + hubUtils.capitalize(info.status));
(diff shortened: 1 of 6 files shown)


ryanjdew (approved):

srinathgit (dismissed):

Sanjeevani19 (commented):
@rahulvudutala Few core unit tests are failing on this PR https://jenkins.marklogic.com/blue/organizations/jenkins/Datahub_CI/detail/PR-5364/6/pipeline, Please look into them

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/prov.sjs
```diff
@@ -241,7 +241,7 @@ class Provenance {
    *      - location (doc URI)
    */
   _createIngestionStepRecord(jobId, flowId, stepName, stepDefinitionName, docURI, info) {
-    let provId = `${jobId + flowId + 'ingestion' + docURI}`;
+    let provId = `${jobId + flowId + 'ingestion' + docURI}_${sem.uuidString()}`;
```
rjrudin, line 244: Per the write-up in 6413, I think all we need is a UUID. The jobId itself is a UUID, so we now have two UUIDs in this. Also see the "How provenance works" Wiki page for details on how this is constructed. 

There's also a lot of duplicated code in this library; we don't want to continue that pattern, we want to clean that up. 

So add a "newProvId()" function that each of these functions uses, and that can just return a UUID.

marklogic-data-hub/src/test/java/com/marklogic/hub/flow/ProvenanceDocumentsForEachStepTest.java
```diff
@@ -0,0 +1,18 @@
...
+import static org.junit.jupiter.api.Assertions.*;
+
+public class ProvenanceDocumentsForEachStepTest extends AbstractHubCoreTest {
+    @Test
+    void verifyProvenanceDocumentsAreGeneratedForEachStep() {
+        installProjectInFolder("test-projects/addSources-step-definition-test");
+        FlowInputs inputs = new FlowInputs("addSourcesFlow", "1", "2");
+        addAbsoluteInputFilePath(inputs, "data");
+        runFlow(inputs);
+        assertEquals(2, getDocCount(HubConfig.DEFAULT_JOB_NAME, "http://marklogic.com/provenance-services/record"));
```
rjrudin, line 16: This really needs an assertion message to explain what's going on - i.e. why is 2 expected?

rjrudin (changes requested):

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/prov.sjs
rahulvudutala: The provenanceId being a UUID will be addressed in a separate ticket

rjrudin: Let's still get a "newProvId(jobId, flowId, stepType, docUri)" function in place - that avoids duplicating the logic of how the prov ID is constructed in many different places.

rjrudin (changes requested):

ryanjdew (approved):

rjrudin (approved):

rjrudin (approved):

Sanjeevani19 (approved):

PR #5369, opened 2021-03-15
DHFPROD-2445: adjust flow options to include overrides before writing batch report

crism:
### Description
Fixes DHFPROD-2445 by adding overridden options to flow object options before writing initial batch report. The initial batch report is included in any intermediate updates.

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/flow.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/flow.sjs
@@ -258,7 +258,12 @@ class Flow {
     this.globalContext.sourceDatabase = combinedOptions.sourceDatabase || this.globalContext.sourceDatabase;
 
     if (!(combinedOptions.noBatchWrite || combinedOptions.disableJobOutput)) {
-      let batchDoc = this.datahub.jobs.createBatch(jobDoc, flowStep, stepNumber);
+      const flowStepWithOptions = Object.assign(
+        {},
+        flowStep,
+        {"options": Object.assign({}, flowStep.options, combinedOptions)}
+      );
+      let batchDoc = this.datahub.jobs.createBatch(jobDoc, flowStepWithOptions, stepNumber);
       this.globalContext.batchId = batchDoc.batch.batchId;
     }
 
--- a/marklogic-data-hub/src/test/java/com/marklogic/hub/flow/impl/FlowRunnerTest.java
+++ b/marklogic-data-hub/src/test/java/com/marklogic/hub/flow/impl/FlowRunnerTest.java
@@ -316,10 +316,13 @@ public class FlowRunnerTest extends AbstractHubCoreTest {
 
         Map<String,Object> opts = new HashMap<>();
         List<String> coll = new ArrayList<>();
+        String mappingJobId = UUID.randomUUID().toString();
+
         coll.add("test-collection");
         opts.put("targetDatabase", HubConfig.DEFAULT_FINAL_NAME);
         opts.put("collections", coll);
         opts.put("sourceQuery", "cts.collectionQuery('test-collection')");
+
         RunFlowResponse resp = runFlow("testFlow", "1,2", UUID.randomUUID().toString(), opts, null);
         flowRunner.awaitCompletion();
         Assertions.assertEquals(2, getDocCount(HubConfig.DEFAULT_FINAL_NAME, "test-collection"));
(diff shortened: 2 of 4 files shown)


srinathgit (changes requested):
You may want to include a test in FlowRunnerTest.testRunFlowOptions() to check the batch doc after the flow is run and see if the options is updated with runtime options

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/flow.sjs
```diff
@@ -258,6 +258,7 @@ class Flow {
     this.globalContext.sourceDatabase = combinedOptions.sourceDatabase || this.globalContext.sourceDatabase;
 
     if (!(combinedOptions.noBatchWrite || combinedOptions.disableJobOutput)) {
+      Object.assign(flowStep.options, combinedOptions);
```
rjrudin, line 261: We need to get a test in place for this first; it's likely best for that to be a JUnit test so that we can test a user providing runtime options. 

@srinathgit  @ryanjdew  I'm also wondering about the fix here. The schema for "Batch.schema.json" describes "step" in the Batch document as having a copy of the step. That is an accurate description. 

I think what 2445 is really saying is "The Batch document doesn't account for the runtime options that I provide". 

I am wondering if the right solution here is to add "combinedOptions" to the Batch document. We could also add just "runtimeOptions" so that the user can see what they provided.

rjrudin (changes requested):

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/flow.sjs
crism: Done.

srinathgit: @rjrudin 
I think it would be best if @yasht2010 can weigh in and confirm if it's ok to add "combinedOptions" and "runtimeOptions" in addition to "options" from step or if "options" in batch doc is the "combinedOptions" that is used in the step.

rjrudin, line 261: I think that because flowStep is used later on, modifying it with combinedOptions could lead to unexpected results. 

How about this instead:

```
const flowStepWithOptions = Object.assign({}, flowStep.options, combinedOptions);
```

And then pass in flowStepWithOptions instead of flowStep.

I can't identify a bug this is preventing, but I think it's safer not to modify flowStep since it's used for a lot of other stuff.

rjrudin (changes requested):
Test looks good, just one tweak

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/flow.sjs
crism: I suppose a step could be used twice in the same flow, and altering it could mess things up. I’ll change this.

srinathgit (changes requested):
Can you please make one more change ? 
https://github.com/marklogic/marklogic-data-hub/blob/develop/specs/models/Batch.schema.json#L25 
The description has to be updated to "Added in 5.5.0; A copy of the step from its flow with  step runtime options"

crism:
> Can you please make one more change ?
> https://github.com/marklogic/marklogic-data-hub/blob/develop/specs/models/Batch.schema.json#L25
> The description has to be updated to "Added in 5.5.0; A copy of the step from its flow with step runtime options"

OK… the step schema is also out of date, and pretty badly.

crism:
> OK… the step schema is also out of date, and pretty badly.

In fact, the step schema is confusing. It lacks `options` at all, but a lot of the top-level properties are apparently also allowed inside `options`. I’m adding `options` and calling it a day… if there is a normative schema somewhere, that would be nice, but it is quickly leaving the scope of this issue.

rjrudin:
@crism The Flow schema captures inline steps, where "options" exists. The Step schema captures the new Step documents introduced in 5.4, which allow steps to live outside of flows. They do not have "options", as there was no real distinction between what goes into "options" and what doesn't. One might think that "options" indicate step properties that are allowed to be overridden at runtime, but there are plenty of examples where no user would ever do that - e.g. stepUpdate, acceptsBatch, mapping, etc. And on the flip side, there are step properties outside of "options" that are often overridden at runtime - e.g. batchSize, threadCount, and fileLocations/inputFilePath. So "options" in the inline step schema is effectively meaningless.

However, to minimize changes to how flows are run, Step documents are converted into inline steps when a flow is run. That means that the "options" property is created based on knowledge of what Step properties do not go in there. 

I agree that the entire model is very confusing. I think what would have made more sense is to not have any "options" property at all, and then runtime options could have overridden any part of the step (with possible support for step properties that can never be overridden, such as stepDefinitionName/stepDefinitionType).

rjrudin (approved):

srinathgit (approved):

PR #5393, opened 2021-03-22
DHFPROD-7043: Establish Session Storage for Retaining Current View in Hub Central

devinnicholson:
### Description
This PR adds sessionStorage for the Load tile. It establishes the data structure and functions that will be used for future session storage stories.

- Persistence on view mode (card/list)
- Persistence on page number
- Persistence on sort order

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/e2e/cypress/integration/curation/load/defaultIngestionListView.spec.tsx
+++ b/marklogic-data-hub-central/ui/e2e/cypress/integration/curation/load/defaultIngestionListView.spec.tsx
@@ -144,7 +144,7 @@ describe("Validate CRUD functionality from list view", () => {
     loadPage.confirmationOptions("Yes").click();
     cy.waitForAsyncRequest();
     cy.waitUntil(() => toolbar.getLoadToolbarIcon()).click();
-    cy.waitUntil(() => loadPage.addNewButton("card").should("be.visible"));
+    cy.waitUntil(() => loadPage.addNewButton("list").should("be.visible"));
   });
   it("Verify Run in an existing flow", () => {
     loadPage.loadView("table").click();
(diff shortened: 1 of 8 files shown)


briantang (changes requested):
Looks awesome! Seeing session storage working and Load tile states being preserved across the app. Just one thing...RTL tests look good but can we add an e2e test for the Load tile? 
Maybe something as simple as changing to list view, applying a sort, navigating to a different tile, and then returning to see everything preserved. There may already be a similar scenario in e2e that you can reference. Let me know @devinnicholson

wooldridge (changes requested):
Recommendations on nomenclature. Feel free to push back if you have opinions. 

- I'm a fan of conciseness. 
- No need to include extra info that can be gleaned from where the code is imported from and the filename. 
- In this case, no need to expose implementation details (e.g., whether UI info is being stored in the session cache or, say, the local cache is irrelevant to the components; the storage location could theoretically change and the UI code could still use the same object).
- Let's be consistent with existing browser storage mechanisms (e.g., use the "dataHub" key prefix as I see we do for most of the local storage keys)

Since we're storing lots of settings related to each view, I'd go with:

DataHubViewSessionStorageType -->  dataHubViewSettings

Also:

getUserHCSessionStorage -->  getViewSettings
setUserHCSessionStorage -->  setViewSettings

For the type definition...

- I would leave off "DataHub". No need to call it DataHub within the Data Hub UI codebase (as opposed to in a browser cache where that distinction might be useful).
- Again, where the info ends up is an implementation detail, so we can ditch "SessionStorage" on the type.
- We seem to be half-and-half right now wrt adding "Interface" and "Type" suffixes to our interface and type definitions. I guess adding a suffix can be helpful.

So...

DataHubViewSessionStorageType -->  ViewSettingsType

briantang (changes requested):
Looks like there is a related e2e failure. The defaultIngestionListView spec is not expecting list view state to be maintained when navigating back to the Load tile. 

defaultIngestionListView.spec.tsx:147 should be changed to "list" instead of "card"

@devinnicholson

briantang (approved):
Looks good, all tests are passing as well.

wooldridge (approved):

PR #5407, opened 2021-03-24
DHFPROD-7100:Stop using Object.values() as it doesn't work in 9.x server

srinathgit:
### Description
Stop using Object.values() as it doesn't work in 9.x server . I don't think it needs test, all existing tests should run fine
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [ n/a] Added Tests
  

- ##### Reviewer:

- [n/a ] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/hub-utils.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/hub-utils.sjs
@@ -146,6 +146,19 @@ function writeDocument(docUri, content, permissions, collections, database) {
   }));
 }
 
+/**
+ * ML 9 does not support Object.values(). This function serves as its replacement so that datahub can be supported in ML 9
+ * @param object
+ * @returns  an array of a given object's property values
+ */
+function getObjectValues(object){
+    let valuesArray = [];
+    for (const property in object) {
+      valuesArray.push(object[property]);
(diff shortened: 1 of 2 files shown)


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/hub-utils.sjs
```diff
@@ -146,6 +146,14 @@ function writeDocument(docUri, content, permissions, collections, database) {
   }));
 }
 
+function getObjectValues(object){
```
rjrudin, line 149: Can you include a comment on this stating that the reason for its existence is that we have to support ML 9, and ML 9 does not support Object.values?

rjrudin (changes requested):

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/hub-utils.sjs
rjrudin: Great, thanks - this is the kind of stuff that's easy to forget later, such that a future developer says "Well we don't need this, we can just use Object.values", not knowing that it won't work on ML 9.

rjrudin (approved):

ryanjdew (approved):

PR #5410, opened 2021-03-25
DHFPROD-6634: Jobs data is not generated when ingested using MLCP DHF transform

rahulvudutala:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [ ] Reviewed Tests


rjrudin (changes requested):
@rahulvudutala @ryanjdew I apologize, I didn't realize this ticket had moved to groomed. As noted in the JIRA ticket, I don't know what the right resolution is here. Disabling job output completely seems problematic because of the point I mentioned in the ticket - not creating Batch documents makes debugging more difficult, if not impossible. 

I think we need more analysis here before moving forward to determine what a user can do if an error occurs, but there's no Batch document to capture that error. 

It may be that we have to live with the Job document being in this very incomplete state. If we can't get rid of Batch docs because they are needed for error handling, I think we have to have a Job document since we don't want Batch docs without a Job doc.

PR #5422, opened 2021-03-26
DHFPROD-7110: Replaced globalContext with stepExecutionContext

rjrudin:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mastering/default/matching.sjs
```diff
@@ -51,7 +51,7 @@ function filterContentAlreadyProcessed(content, summaryCollection, collectionInf
   return Sequence.from(filteredContent);
 }
 
-function main(content, options) {
+function main(content, options, stepExecutionContext) {
```
rjrudin, line 54: This is the only OOTB step module that needs this, as it needs to know the jobId. It's currently getting it from the globalContext. By passing this in, every step module has access to all the data that's relevant to the step execution. And it's trivial for us to add more in the future. 

The one awkward part is that stepExecutionContext contains combinedOptions, and that's the second argument.

rjrudin:
Fixed failed tests here Object.assign doesn't work on an error object

fsnow (approved):

akshaysonvane (approved):

PR #5434, opened 2021-03-29
DHFPROD-7002: GH 5308 - targetEntity still being set on new match/mer…

fsnow:
…ge steps instead of targetEntityType

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [ ] Added Tests
  

- ##### Reviewer:

- [ ] Reviewed Tests


marklogic-data-hub/src/main/resources/scaffolding/defaultFlow.flow.json
```diff
@@ -65,7 +65,7 @@
       "options": {
         "sourceDatabase": "%%mlFinalDbName%%",
         "targetDatabase": "%%mlFinalDbName%%",
-        "targetEntity": "Change this to a valid entity type name; e.g. Customer",
+        "targetEntityType": "Change this to a valid entity type name; e.g. Customer",
```
ryanjdew, line 68: I believe this change will also requiring updating this test: com/marklogic/hub/scaffolding/CreateDefaultFlowTest.java:24

ryanjdew (changes requested):

marklogic-data-hub/src/main/resources/scaffolding/defaultFlow.flow.json
fsnow: What about QuickStart support? Does this change break it? See final assert message in CreateDefaultFlowTest.

ryanjdew: I meant to comment on the higher-up option for mapping, but I'm thinking since the defaultFlow is still doing the older inline steps at this point that we should leave that alone since that will be handled when converted. Updating the createStepTest makes sense since it is testing the separated step format that QS doesn't support anyway.

fsnow: ok, I backed out the changes to defaultFlow.

ryanjdew (approved):

srinathgit (approved):

PR #5436, opened 2021-03-29
DHFPROD-7062: Incorrect scrolling after creating a modeling property …

timur-isangulov:
…with multiple setting

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- N/A Added Tests
  

- ##### Reviewer:

- N/A Reviewed Tests


wooldridge (changes requested):
The case described in the bug ticket is fixed, cool. But I'm still seeing cases where sometimes the browser window scrolls down and sometimes it doesn't scroll:

1. Go to Modeling and edit Customer in the ref project.
2. Add a property: test1 string -> SCROLLS
3. Add a property: test2 string -> NO SCROLL
4. Add a property: test3 integer -> SCROLLS
5. Add a property: test4 string -> NO SCROLL

I would expect the browser window to always scroll in my example. Are there duplicated or conflicting newRowKey values in the table rows?

wooldridge (approved):
It's working for me now, thanks!

briantang (approved):

PR #5461, opened 2021-04-02
DHFPROD-7053: ML library using Optic For Querying Jobs Data

rahulvudutala:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/jobs/job-results-lib.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/jobs/job-results-lib.sjs
@@ -1,0 +1,54 @@
+/**
+ Copyright (c) 2021 MarkLogic Corporation
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+ http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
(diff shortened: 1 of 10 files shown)


ryanjdew (approved):

Sanjeevani19 (approved):

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/jobs/job-results-lib.sjs
```diff
@@ -0,0 +1,54 @@
...
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+'use strict';
+
+const op = require('/MarkLogic/optic');
+
+function getJobData(start, pageLength, sortColumn, sortDirection) {
```
rjrudin, line 20: "JobData" is too generic here. We know exactly what we're returning - step responses. And I prefer "find" when we're doing a query, so this should be "findStepResponses".

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/jobs/job-results-lib.sjs
```diff
@@ -0,0 +1,54 @@
+/**
```
rjrudin, line 1: This library is currently based on step responses, which are part of job documents. And the scope is querying for data. "job-results" is a little vague - are these query results, or results from executing a job?

We have "entity-search-lib" already. "search" unfortunately implies use of the Search API.

I am thinking "(noun)-query-lib" is thus the better direction, as "query" doesn't imply a particular API that we're using - which shouldn't matter for the caller either.

So I think this should be "job-query-lib".

rjrudin (changes requested):

PR #5464, opened 2021-04-05
DHFPROD-7151: Add previewMatchingActivity test suite using patientMat…

fsnow:
…ch data

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


ryanjdew (approved):

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/data-services/mastering/previewMatchingActivity2/previewMatchingActivity.sjs
```diff
@@ -0,0 +1,18 @@
...
+  uris.push(uri);
+}
+
+let options = cts.doc("/steps/matching/patientMatch.step.json").root;
+xdmp.log(options);
+let results = pma.previewMatchingActivity(options, uris, 0);
+xdmp.log(results.actionPreview.length);
+xdmp.log(results);
+[
+  test.assertEqual(19, results.actionPreview.length, "There should be 19 matching pairs")
```
rjrudin, line 17: I think this should be simplified to expect 2 pairs - like there should be 2 that match, and 1 that does not. And the assertion message should make it clear which 2 are expected and why they should match. 

Otherwise, if this breaks, it'll be difficult to figure out what the problem is and why it broke. This is similar to the challenge we've had with MasterTest - when we don't get 205 merges, it's nearly impossible to figure out why we're not because we don't know exactly which 205 are expected, nor why. 

That should also remove the need for the log statements, which I assume are there to facilitate debugging.

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/data-services/mastering/previewMatchingActivity2/setup.xqy
```diff
@@ -0,0 +1,33 @@
...
+import module namespace test = "http://marklogic.com/test" at "/test/test-helper.xqy";
+hub-test:load-artifacts($test:__CALLER_FILE__)
+
+;
+
+xquery version "1.0-ml";
+import module namespace test = "http://marklogic.com/test" at "/test/test-helper.xqy";
+import module namespace const = "http://marklogic.com/smart-mastering/constants"
+  at "/com.marklogic.smart-mastering/constants.xqy";
+for $uri in ("/dictionary/last-names.xml", "/thesaurus/nicknames.xml")
```
rjrudin, line 25: If CONTENT-COLL doesn't actually matter for functionality, then you can get both of these docs loaded by tossing them under test-data/content .

rjrudin (changes requested):

rjrudin (approved):

PR #5466, opened 2021-04-05
DHFPROD-7053: Code refactor for ML libraries using Optic For Querying Jobs Data

rahulvudutala:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


ryanjdew (approved):

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/jobs/job-query-lib.sjs
```diff
@@ -17,26 +17,26 @@
...
-      'Job.jobMonitor.flowName as flowName' +
-      ' from Job.jobMonitor';
+  const totalCountQuery = 'select count(*) as total from Job.StepResponse';
+  const jobsDataQuery = 'select Job.StepResponse.stepName as stepName,' +
+      'Job.StepResponse.stepType as stepDefinitionType,' +
+      'Job.StepResponse.stepStatus as status,' +
+      'Job.StepResponse.targetEntityType as entityName,' +
+      'Job.StepResponse.stepStartTime as startTime,' +
+      'Job.StepResponse.stepEndTime - Job.StepResponse.stepStartTime as duration,' +
+      'Job.StepResponse.successfulEvents as successfulItemCount,' +
```
rjrudin, line 34: I think the template needs to be updated with these names also - the template is projecting column names on top of the data, where we can rename things that have misleading names.

marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/LoadHubTemplatesCommand.java
```diff
@@ -36,13 +36,13 @@ public void execute(CommandContext context) {
         try {
             JSONDocumentManager documentManager = hubConfig.newStagingClient(hubConfig.getStagingSchemasDbName()).newJSONDocumentManager();
             ResourcePatternResolver resolver = new PathMatchingResourcePatternResolver(getClass().getClassLoader());
-            for (Resource r : resolver.getResources("classpath*:/tdes/**/*.json")) {
-                File file = new File("tdes/" + r.getFilename());
+            for (Resource r : resolver.getResources("classpath*:/hub-templates/**/*.json")) {
```
rjrudin, line 39: Taking a second look at this - this approach of finding files on the classpath makes sense for e.g. figuring out what databases/roles/etc to deploy. But once we've loaded the DHF modules into the modules database, it seems the simplest thing to do is have an e.g. "installJobTemplates" function in job-query-lib, which will write the template to both schema databases.

We then just need a single command to call a single DS endpoint to finish deployment of DHF after all the resources/modules have been deployed - e.g. "finishHubDeployment". The concept is that once we've got all the modules in place, we can invoke a DS endpoint that knows what else needs to be done. 

That also moves all of this logic from Java code into ML, which is typically what we want to do. 

Thus, I'm proposing the following:

- Change this command to "FinishHubDeploymentCommand"
- Make a new endpoint in the "data-services/system" directory called "finishHubDeployment"
- Have that endpoint call job-query-lib:installJobTemplates() (there's only one template, but there could clearly be a Job/Job template in the near future)

I know that's some rework now, but it'll make life easy for when we have future deployment needs that we can handle via DHF modules - we just invoke that code in the finishHubDeployment endpoint, and we don't have to touch any Java code.

rjrudin (changes requested):
@ryanjdew @srinathgit  See my comments about a new DS endpoint to finish hub deployment - I think that's consistent with our shift of logic from Java to ML, and it also makes life easier in the future. 

The main difference here is we really don't need to be loading anything dynamically from the classpath - we know exactly what templates we want to install, and so we may as well toss knowledge of that into DHF modules as opposed to Java code.

marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/FinishHubDeploymentCommand.java
```diff
@@ -0,0 +1,23 @@
...
+import com.marklogic.hub.dataservices.SystemService;
+import com.marklogic.hub.impl.HubConfigImpl;
+
+public class FinishHubDeploymentCommand extends AbstractCommand {
+
+    private HubConfigImpl hubConfig;
+
+    public FinishHubDeploymentCommand(HubConfig hubConfig) {
+        this.hubConfig = (HubConfigImpl) hubConfig;
+        setExecuteSortOrder(SortOrderConstants.DEPLOY_TRIGGERS + 1);
```
rjrudin, line 16: I think this should be max integer to ensure this runs after everything else is done, which is what the name implies.

rjrudin (changes requested):
Looks good, just one tweak for the sort order

rjrudin (approved):

ryanjdew (approved):

PR #5463, opened 2021-04-05
DHFPROD-7068:Fetch all related entities for UI

srinathgit:
### Description
BE: Fetch all related entities for UI (DHFPROD-7068)
UI: View related entity tables via filter (DHFPROD-6706)
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] JIRA_ID included in all the commit messages
- [x] PR title is in the format JIRA_ID:Title
- [x] Rebase the branch with upstream
- [x] Squashed all commits into a single commit
- [x] Code passes ESLint tests
- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

PR #5471, opened 2021-04-06
DHFPROD-6706: View related entity types in Hub Central

briantang:
### Description
- Able to view related entity tables by selecting them via filter if related entities to the primary entity exist
- Screencastify here: https://drive.google.com/file/d/1jFHXHL7_dBrZj5_Yua4RsR2rlg_yQXRp/view
- Includes @srinathgit 's backend changes from https://github.com/marklogic/marklogic-data-hub/pull/5463
- Reviewers please do a clean build in order to pick up the backend changes
- Added RTL tests for the related entity filter, more extensive tests to be done in this [ticket](https://project.marklogic.com/jira/browse/DHFPROD-7155) 

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [ ] Reviewed Tests


briantang:
This PR is ready for review, but leaving in Draft state because I will combine this PR with @srinathgit 's [PR](https://github.com/marklogic/marklogic-data-hub/pull/5463) once it is fully reviewed. Reviewers please proceed with reviews.

wooldridge (changes requested):
There seems to be some issues with deselection and filter state. Starting from initial opening of the mapCustomersJSON: 

1. In Customer filter, SELECT: Order (orderedBy Customer)
2. In Customer filter, SELECT: BabyRegistry (ownedBy Customer)
3. In Order (orderedBy Customer) filter, SELECT: Product (Order hasProduct)
4. In Customer filter, DESELECT: Order (orderedBy Customer)

Result: The correct related entities are hidden after deselection (great!). BUT, BabyRegistry (ownedBy Customer) now has a "Product (Order hasProduct)" selection in its filter, which was empty before. Should the filter remain empty?

Are we testing deselection of the filter in the unit tests? I don't see this on a cursory look of mapping-step-detail.test.

The alignment of the primary entity disclosure caret and its label are too far to the left (not wireframes where they are indented more).

Screen Shot 2021-04-06 at 8 38 44 AM

Why are some Related Entity labels "(PROPERTY ENTITY)" and others "(ENTITY PROPERTY)"? E.g., "(orderedBy Customer)" and "(Order hasProduct)". I think this just inconsistent dummy data, should it aways be "(PROPERTY ENTITY)"?

briantang:
Thanks for the review @wooldridge. To address your last question, there is a distinction between those labels which was specifically requested by PM. If a related entity like Order is using Customer, the label will be property first, e.g. (orderedBy Customer). If the entity directly uses another entity like Order containing Product under the 'includes' property, the label will be property second, e.g. (Order includes). Hope this made sense, but the distinction is intended.

wooldridge:
@briantang, thanks for the explanation. I now see in your description that tests will be fleshed out in the following, so disregard my comment about testing deselection in this PR: https://project.marklogic.com/jira/browse/DHFPROD-7155

brucean52 (approved):
I use MongoDB Atlas https://www.mongodb.com/

timur-isangulov:
The table column color doesn't look right when the sort is applied. 
Screen Shot 2021-04-06 at 11 20 12 AM

briantang:
Have made the corresponding fixes and pushed onto Srinath's PR #5463 with our joint BE + UI changes. @wooldridge Created a separate bug [here](https://project.marklogic.com/jira/browse/DHFPROD-7163) to address the deselection issue.

Will close this PR now as it is ready for review on Srinath's PR.

PR #5480, opened 2021-04-07
DHFPROD-5764: Data Service/REST API for previewing match step against…

fsnow:
… documents

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


rjrudin (approved):

AkshayBajajML (dismissed):

ryanjdew (dismissed):

rjrudin (dismissed):

ryanjdew (approved):

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/mastering/util.sjs
```diff
@@ -0,0 +1,10 @@
+'use strict';
+
+function evalSourceQuery(step) {
```
rjrudin, line 3: I think we should make this non-step specific. And it isn't mastering-specific. 

We have a "hub-utils" library, and I think it might as well go in there (I wish it could go into data-hub/5/util, but I think it would be too confusing if it's not in hub-utils). So can you convert this to "evalInDatabase(script, database)" and toss the amp on that? 

I think it's worth changing now because once we release an amp, it's hard to get rid of / change.

rjrudin (changes requested):

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/mastering/util.sjs
fsnow: done

rjrudin (approved):

AkshayBajajML (approved):

ryanjdew (approved):

srinathgit (approved):

PR #5487, opened 2021-04-08
DHFPROD-6943: Bulk Data Service For Querying Jobs Data

rahulvudutala:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/system/finishHubDeployment.sjs
```diff
@@ -17,5 +17,5 @@
 
 // No privilege required: For now it is only used during installation and installation requires admin privileges.
 
-const jobQueryLib = require("/data-hub/5/jobs/job-query-lib.sjs");
+const jobQueryLib = require("/data-hub/5/flow/jobQueryLib.sjs");
```
rjrudin, line 20: Let's stick with job-query-lib.sjs as the name - the naming isn't consistent in the DHF ML code, but things like entity-lib and hub-utils are widely used, so "worm case" is safe to use

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/job/findStepResponses.api
```diff
@@ -0,0 +1,20 @@
+{
+  "endpoint": "/data-hub/5/data-services/stepResponses/findStepResponses.sjs",
```
rjrudin, line 2: I'm surprised this works, but the endpoint should be "data-services/job" instead of "data-services/stepResponses", right?

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/system/finishHubDeployment.sjs
rjrudin: Though "jobQueryLib" is good as a prefix, which is consistent with entityLib as a prefix.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/flow/jobQueryLib.sjs
```diff
@@ -18,36 +18,46 @@
...
-function findStepResponses(start, pageLength, sortColumn, sortDirection) {
+function findStepResponses(query) {
+  const start = query.start;
+  const pageLength = query.pageLength;
+  const sortColumn = query.sortColumn;
+  const sortDirection = query.sortDirection;
+
   let orderByConstraint = op.desc('startTime');
-  if(sortColumn) {
+  if(query.sortColumn) {
```
rjrudin, line 28: You can just reference sortColumn here, as you have it declared above.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/flow/jobQueryLib.sjs
```diff
@@ -18,36 +18,46 @@
...
-  const totalCount = op.fromSQL(totalCountQuery).result().toObject();
-  const jobsDataResults = op.fromSQL(jobsDataQuery).orderBy(orderByConstraint).offset(pageLength * (start-1)).limit(pageLength).result();
-
-  response["total"] = totalCount[0]["total"];
-  response["start"] = start;
-  response["pageLength"] = pageLength;
-  response["results"] = jobsDataResults.toObject();
-
-  return response;
+  return fn.head(xdmp.invokeFunction(
```
rjrudin, line 32: I don't think there's a use case for invokeFunction here. This is an internal API, and we know we want to use the DatabaseClient for the jobs database when we do this. So I think the way you had it before was better.

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/data-services/lib/jobService.sjs
```diff
@@ -0,0 +1,12 @@
+function invokeFindStepResponsesModule(module, args) {
+  console.log(args);
```
rjrudin, line 2: Can remove this log statement

rjrudin (changes requested):

rjrudin (approved):

Sanjeevani19 (approved):

PR #5488, opened 2021-04-08
DHFPROD-7130: Display error messaging if user wants to create an enti…

devinnicholson:
…ection" entity

### Description
This PR adds error messaging to prevent a bug that throws an error if a user tries to create these entity names:
"Collection", "createdByJob", "createdByStep", and "entity-type"

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [ ] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/src/components/modeling/entity-type-modal/entity-type-modal.tsx
+++ b/marklogic-data-hub-central/ui/src/components/modeling/entity-type-modal/entity-type-modal.tsx
@@ -32,6 +32,8 @@ const EntityTypeModal: React.FC<Props> = (props) => {
   const [errorMessage, setErrorMessage] = useState("");
   const [loading, toggleLoading] = useState(false);
 
+  const reservedEntityNamesArray = ["Collection", "createdByJob", "createdByStep", "entity-type"];
+
   useEffect(() => {
     if (props.isVisible) {
       if (props.isEditModal) {
@@ -105,6 +107,8 @@ const EntityTypeModal: React.FC<Props> = (props) => {
     } else {
       if (!NAME_REGEX.test(name)) {
         setErrorMessage(ModelingTooltips.nameRegex);
+      } else if (reservedEntityNamesArray.indexOf(name) > -1) {
+        setErrorMessage(name + ModelingTooltips.reservedEntityNames);
       } else {
         toggleLoading(true);
         createEntityType(name, description);
--- a/marklogic-data-hub-central/ui/src/config/tooltips.config.tsx
+++ b/marklogic-data-hub-central/ui/src/config/tooltips.config.tsx
@@ -43,6 +43,7 @@ const ModelingTooltips = {
   nameEntityProperty: 'The name of this entity property. ' +
     'Names must start with a letter and can contain letters, numbers, hyphens, and underscores.',  /* intended dupe: all names */
   descriptionEntityProperty: 'A description of this entity property.',
+  reservedEntityNames: ' is not an allowed entity name. Choose another.',
 
   /* Form fields */
   joinProperty: 'Structured type properties and arrays cannot be used as join properties.',
--- a/marklogic-data-hub-central/ui/src/components/modeling/entity-type-modal/entity-type-modal.test.tsx
+++ b/marklogic-data-hub-central/ui/src/components/modeling/entity-type-modal/entity-type-modal.test.tsx
@@ -163,5 +163,29 @@ describe("EntityTypeModal Component", () => {
     expect(axiosMock.put).toHaveBeenCalledWith(url, payload);
     expect(axiosMock.put).toHaveBeenCalledTimes(1);
   });
+
+
+  test("Adding a restricted Entity name shows error message", async () => {
+    const {getByText, getByPlaceholderText} = render(
+      <EntityTypeModal
+        isVisible={true}
+        toggleModal={jest.fn()}
+        updateEntityTypesAndHideModal={jest.fn()}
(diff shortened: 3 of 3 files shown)


marklogic-data-hub-central/ui/src/components/modeling/entity-type-modal/entity-type-modal.tsx
```diff
@@ -32,6 +32,8 @@ const EntityTypeModal: React.FC<Props> = (props) => {
   const [errorMessage, setErrorMessage] = useState("");
   const [loading, toggleLoading] = useState(false);
 
+  const reservedEntityNamesArray = ["Collection", "createdByJob", "createdByStep", "entity-type"];
```
rjrudin, line 35: Pinging @wooldridge  on this - we really ought to have the backend provide this knowledge to the front end, as the backend understands why these shouldn't be used. I think ideally, this is part of the validation check on the backend when an entity is saved. Worst case, the backend provides this data to the UI for the UI to enforce, but I'd really rather see the backend enforce it, as there could be other interfaces in the future for creating entity types that need this same logic.

devinnicholson: I can do that instead. If there's a concern that the array were to be consistently updated in the future, that would definitely become tedious and error-prone.

wooldridge: @rjrudin, @devinnicholson, definitely important to consider. We already do a client-side validation for bad chars (spaces et al.) and that there's something in the field, and this client-side validation pattern is what UX has prescribed across much of the app. Doing some validation client-side and some server side would be extra work for the frontend and an inconsistent user experience. I see this strategy as manageable for the frontend so I prefer what Devin is doing here.

marklogic-data-hub-central/ui/src/config/tooltips.config.tsx
```diff
@@ -43,6 +43,7 @@ const ModelingTooltips = {
   nameEntityProperty: 'The name of this entity property. ' +
     'Names must start with a letter and can contain letters, numbers, hyphens, and underscores.',  /* intended dupe: all names */
   descriptionEntityProperty: 'A description of this entity property.',
+  reservedEntityNames: ' is not an allowed entity name. Choose another.',
```
wooldridge, line 46: @devinnicholson, if you haven't already, can you have docs review this (I'd tag Al but I don't see his GitHub). I suspect they'll want to massage the wording.

wooldridge (changes requested):

marklogic-data-hub-central/ui/src/components/modeling/entity-type-modal/entity-type-modal.tsx
rjrudin: Client-side validation is fine, I'm more interested in where the knowledge of this is stored. Is there an endpoint call that the UI is making here, before the entity is saved, where the backend could include the list of forbidden names? 

My intent here is that Pari can also create entity models via a text editor and then deploying them via Gradle. We need this same validation applied when she does that.

rjrudin: @devinnicholson If you can point me to an appropriate endpoint - I'm thinking the one that returns all the entity types - then I can submit a small sub-task PR to enhance that endpoint to include this list of "forbiddenEntityNames". I'll bake that into the backend and get a test in place for it as well. Then the UI can refer to that list and still perform client side validation.

wooldridge: @rjrudin let's put this sort of information somewhere in the code base that both the UI and gradle can access directly.

devinnicholson: There's getPrimaryEntityTypes() found in ModelsService, which I believe is what you're looking for. @rjrudin

marklogic-data-hub-central/ui/src/config/tooltips.config.tsx
devinnicholson: Al confirmed wording is OK.

marklogic-data-hub-central/ui/src/components/modeling/entity-type-modal/entity-type-modal.tsx
devinnicholson: Oh, to answer your first question, the call when creating an entity is to /api/models passing name and description (createModel) after a client side validation check has passed for naming convention (createEntityType in entity-type-modal.tsx)

wooldridge: @rjrudin @devinnicholson I'm fine with adding this info to the primaryEntityTypes endpoint payload.

We'll want to update the mock test data for that endpoint, which I know includes this:

https://github.com/marklogic/marklogic-data-hub/blob/develop/marklogic-data-hub-central/ui/src/assets/mock-data/curation/flows.data.ts#L183

rjrudin: I've created https://project.marklogic.com/jira/browse/DHFPROD-7174 and will submit a PR for that this morning.

PR #5490, opened 2021-04-09
DHFPROD-3097: added option to maintain collections in steps

jyangML:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [X] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/flow/flowRunner.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/flow/flowRunner.sjs
@@ -150,6 +150,10 @@ function processContentWithStep(stepExecutionContext, contentArray, databaseWrit
     return [];
   }
 
+  if (String(stepExecutionContext.combinedOptions.targetCollectionsAdditivity) == "true") {
+    applyTargetCollectionsAdditivity(contentArray);
+  }
+
   if (databaseWriteQueue != null) {
     // Since DHF 5.0, collections from options have been added after step processing as opposed to before step processing.
     // The reason for this is not known.
@@ -174,6 +178,15 @@ function invokeBeforeMain(stepExecutionContext, contentArray) {
         `in flow '${stepExecutionContext.flow.name}'; cause: ${error.message}`);
     }
   }
+}
+
+function applyTargetCollectionsAdditivity(contentArray) {
+  contentArray.forEach(content => {
+    if (content.context.originalCollections) {
+      let collections = content.context.collections || [];
+      content.context.collections = collections.concat(content.context.originalCollections);
+    }
+  });
 }
 
 /**
--- a/marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/flow/testTargetCollectionAdditivity/flowWithTargetCollectionsAdditivity.sjs
+++ b/marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/flow/testTargetCollectionAdditivity/flowWithTargetCollectionsAdditivity.sjs
@@ -1,0 +1,31 @@
+const test = require("/test/test-helper.xqy");
+const hubTest = require("/test/data-hub-test-helper.sjs");
+const flowRunner = require("/data-hub/5/flow/flowRunner.sjs");
+
+let assertions = [];
+let results, collections;
+
+const options = {
+  sourceDatabase: "data-hub-FINAL",
+  targetDatabase: "data-hub-FINAL",
+  targetCollectionsAdditivity: true,
+  outputFormat: 'json'
+};
+
+const content = [{uri:"/customer1.json",
+  value:{"hello": "world"},
+  context: {originalCollections: ["source"]}
+}];
+results = flowRunner.processContentWithFlow("simpleMappingFlow", content, "1", options);
+collections = hubTest.getRecord("/customer1.json").collections;
+
+assertions.push(
+  test.assertEqual("finished", results.jobStatus),
+  test.assertEqual(2, collections.length,
+    "doc should have collections from both mapping step and from originalCollections"),
+  test.assertTrue(collections.includes("Customer"),
+    "doc should have collection from mapping step"),
+  test.assertTrue(collections.includes("source"),
+    "doc should have collection from source doc when targetCollectionsAdditivity is true")
+);
+assertions;
--- a/marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/flow/testTargetCollectionAdditivity/flowWithoutTargetCollectionsAdditivity.sjs
+++ b/marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/flow/testTargetCollectionAdditivity/flowWithoutTargetCollectionsAdditivity.sjs
@@ -1,0 +1,33 @@
+const test = require("/test/test-helper.xqy");
+const hubTest = require("/test/data-hub-test-helper.sjs");
+const flowRunner = require("/data-hub/5/flow/flowRunner.sjs");
+
+let assertions = [];
+let results, collections;
+
+const options = {
+  sourceDatabase: "data-hub-FINAL",
+  targetDatabase: "data-hub-FINAL",
+  targetCollectionsAdditivity: false,
+  outputFormat: 'json'
+};
+
+const content = [{uri:"/customer1.json",
+  value:{"hello": "world"},
+  context: {originalCollections: ["source"]}
+}];
+
+results = flowRunner.processContentWithFlow("simpleMappingFlow", content, "1", options);
+collections = hubTest.getRecord("/customer1.json").collections;
+
+assertions.push(
+  test.assertEqual("finished", results.jobStatus),
+  test.assertEqual(1, collections.length,
+    "doc should only have collection from mapping step"),
+  test.assertTrue(collections.includes("Customer"),
+    "doc should have collection from mapping step"),
+  test.assertFalse(collections.includes("source"),
+    "doc should not have collection from source doc when targetCollectionsAdditivity is false")
+);
+
+assertions;
--- a/marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/flow/testTargetCollectionAdditivity/setup.sjs
+++ b/marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/flow/testTargetCollectionAdditivity/setup.sjs
@@ -1,0 +1,7 @@
+declareUpdate();
+
+const hubTest = require("/test/data-hub-test-helper.sjs");
+const hubTestX = require("/test/data-hub-test-helper.xqy");
+
+hubTestX.resetHub();
+hubTest.createSimpleMappingProject([{}]);


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/flow-utils.sjs
```diff
@@ -718,8 +718,8 @@ function addMetadataToContent(content, flowName, stepName, jobId) {
  * @param configCollections the collections to add to each content object based on step definition / flow / step / runtime config
  * @return An object consisting of two properties - "transaction" and "dateTime"
  */
-function writeContentArray(contentArray, databaseName, configCollections = []) {
-  const vars = {contentArray, configCollections};
+function writeContentArray(contentArray, databaseName, configCollections = [], targetCollectionsAdditivity = false) {
```
srinathgit, line 721: 'targetCollectionsAdditivity' shouldn't be a parameter to writeContentArray() method. It can be part of the context within a content object.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/flow.sjs
```diff
@@ -243,6 +243,8 @@ class Flow {
         ].reduce((previousValue, currentValue) => (previousValue || []).concat((currentValue || [])))
           // filter out any null/empty collections that may exist
           .filter((col) => !!col);
+        
+        const targetCollectionsAdditivity = options.targetCollectionsAdditivity && String(options.targetCollectionsAdditivity) == "true";
```
srinathgit, line 247: We will have to check for the property in 'combinedOptions' not just 'options'

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/hub-utils/invoke-queue-write.sjs
```diff
@@ -23,6 +23,9 @@ var contentArray;
 // The collections to add to each content object, based on step definition / flow / step / runtime config
 var configCollections;
 
+// boolean to indicate whether source document collections are preserved in target document
+var targetCollectionsAdditivity;
```
srinathgit, line 27: Again, this can be part of the context.

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/flow/testTargetCollectionAdditivity/test-data/entities/Customer.entity.json
```diff
@@ -0,0 +1,26 @@
+{
```
srinathgit, line 1: can you remove all the entities/flow/step and try to reuse use the new createSimpleMappingProject helper in data-hub-test-helper.sjs ?

srinathgit (changes requested):

srinathgit:
@rjrudin ,
Currently with mapping steps (or custom steps), a source doc can write more than one document in target db. So, if this property is set, is the expectation that each of the written document will retain the collection of the source document ?

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/hub-utils/invoke-queue-write.sjs
rjrudin: I don't think it should be part of the context, because then it has to be a part of the context object on every content object being persisted. 

I'm doing some research on this now, as I don't think this module should be in the business of figuring out collections - that should be figured out before we get to this module. It's fine for this module to fiddle with temporal collections, as that's specific to document insertion. 

Also, we want this change to be made in a place so that it's reused by flow.sjs and flowRunner.sjs (and most of flow.sjs is in the process of going into flowRunner.sjs).

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/flow/testTargetCollectionAdditivity/test-data/entities/Customer.entity.json
rjrudin: Definitely - should be fine to just use a mapping step as part of this. You can have a document under test-data/content, which means it'll have a collection of "raw-content" on it. And you can verify that it's either preserved or not preserved.

rjrudin (changes requested):
I'm looking into this, will get back to you soon.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/flow.sjs
jyangML: Fixed

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/flow/flowRunner.sjs
```diff
@@ -132,6 +132,7 @@ function processContentWithStep(stepExecutionContext, contentArray, databaseWrit
   }
 
   invokeBeforeMain(stepExecutionContext, contentArray);
+  applyTargetCollectionsAdditivity(stepExecutionContext, contentArray);
```
rjrudin, line 135: This needs to happen after the step main function is run, as we want to apply the logic to the output content array, not the input content array.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/flow/flowRunner.sjs
```diff
@@ -176,6 +177,34 @@ function invokeBeforeMain(stepExecutionContext, contentArray) {
...
 }
 
+/**
+ * Applies source collections to written documents if targetCollectionsAdditivity is set to true
+ * @param stepExecutionContext 
+ * @param contentArray 
+ */
+function applyTargetCollectionsAdditivity(stepExecutionContext, contentArray) {
+  
+  if (!stepExecutionContext.combinedOptions.targetCollectionsAdditivity ||
```
rjrudin, line 187: It's simpler and reads easier if this is just:

```
if (String(stepExecutionContext.combinedOptions.targetCollectionsAdditivity) === "true") {
  do stuff
}
```

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/flow/flowRunner.sjs
```diff
@@ -176,6 +177,34 @@ function invokeBeforeMain(stepExecutionContext, contentArray) {
...
+    String(stepExecutionContext.combinedOptions.targetCollectionsAdditivity) != "true")
+  {
+    // do nothing
+    return;
+  }
+
+  // append context.originalCollections to context.collections
```
rjrudin, line 194: No need for this comment - it's only correct some of the time (if collections doesn't exist, then it's a "set", not an "append") and it's easy enough to understand the code that no comment is needed.

rjrudin: You can also simplify the code by avoiding the if/else:

```
if (content.context.originalCollections) {
   const collections = content.context.collections || [];
   content.context.collections = collections.concat(content.context.originalCollection);
}
```

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/flow/testTargetCollectionAdditivity/flowWithTargetCollectionsAdditivity.sjs
```diff
@@ -0,0 +1,74 @@
+const DataHubSingleton = require("/data-hub/5/datahub-singleton.sjs");
```
rjrudin, line 1: This test will be a lot simpler by using flowRunner.sjs . With processContentWithFlow, you can pass in your own contentArray, which means you can set content.context.collections to whatever you want - no need to persist anything. So you can do something like this:

```
const content = [{"uri":"whatever", "value":{"hello":"world"}, context: {collections: ["one", "two"]}];
flowRunner.processContentWithFlow("simpleMappingFlow", content, jobId, options);
```

Then use hubTest.getRecord(uri) to get an object that has the document and collections in it. Much easier to test with that.

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/flow/testTargetCollectionAdditivity/flowWithTargetCollectionsAdditivity.sjs
```diff
@@ -0,0 +1,74 @@
...
+  
+  // document should have "Customer" collection as before
+  test.assertEqual(1, output_collection.length),
+
+  // document should have "customer-input" collection preserved by targetCollectionsAddivity
+  test.assertEqual(1, source_collection.length),
+  test.assertTrue(source_collection[0] == "/customer1.json")
+);
+
+function getUris(collection){
```
rjrudin, line 70: No need for this - use hubTest.getRecord and then verify that the collections on the record are correct - i.e. that the collections are either retained or not retained.

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/flow/testTargetCollectionAdditivity/suite-setup.xqy
```diff
@@ -0,0 +1,12 @@
+xquery version "1.0-ml";
+import module namespace hub-test = "http://marklogic.com/data-hub/test" at "/test/data-hub-test-helper.xqy";
```
rjrudin, line 2: You can do this in setup.sjs, so you don't need this file (since you don't need to insert content below either, since flowRunner lets you pass in in-memory content).

rjrudin (changes requested):

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/flow/flowRunner.sjs
```diff
@@ -150,6 +150,11 @@ function processContentWithStep(stepExecutionContext, contentArray, databaseWrit
     return [];
   }
 
+  // if targetCollectionsAdditivity is set, then target documents should inherit collections from source documents
```
rjrudin, line 153: No need for a comment here - it's only restating what the code is already saying.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/flow/flowRunner.sjs
```diff
@@ -176,6 +181,20 @@ function invokeBeforeMain(stepExecutionContext, contentArray) {
   }
 }
 
+/**
+ * Applies source collections to written documents
```
rjrudin, line 185: Because the code uses the term "originalCollections", we don't want to introduce "source collections" as a different term, as that raises the question of - What's the difference between source collections and original collections? 

The comment isn't quite accurate either - these aren't written documents yet, it's just an array of content objects. 

What the function is doing is - for each content object, it's concatenating original collections to the content object's collections if original collections exists. But if you look at the code, the code is basically saying that. So I don't think any comment is needed, the code reads well enough.

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/flow/testTargetCollectionAdditivity/flowWithTargetCollectionsAdditivity.sjs
```diff
@@ -0,0 +1,31 @@
...
+
+const content = [{uri:"/customer1.json",
+  value:{"hello": "world"},
+  context: {originalCollections: ["source"]}
+}];
+results = flowRunner.processContentWithFlow("simpleMappingFlow", content, "1", options);
+collections = hubTest.getRecord("/customer1.json").collections;
+
+assertions.push(
+  // doc should have been written
```
rjrudin, line 23: With tests, assertion messages are more valuable than comments, as they explain why something is expected and they'll show up in the failure. So instead of a comment, just do:

    test.assertEqual("finished", results.jobStatus)

And no message is really needed there, it's clear what is expected.

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/flow/testTargetCollectionAdditivity/flowWithTargetCollectionsAdditivity.sjs
```diff
@@ -0,0 +1,31 @@
...
+  test.assertEqual("1", results.lastCompletedStep),
+
+  // doc should have both collection from mapping step and collection from source doc
+  test.assertEqual(2, collections.length),
+  test.assertTrue(collections.includes("Customer")),
```
rjrudin, line 28: getRecord will alphabetize the collections, and we'd rather use an assertion message here, so:

    test.assertEqual("Customer", collections[0]),
    test.assertEqual("source", collections[1], "The existing collection should have been retained since targetCollectionsAdditivity=true")

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/flow/testTargetCollectionAdditivity/flowWithoutTargetCollectionsAdditivity.sjs
```diff
@@ -0,0 +1,33 @@
...
+collections = hubTest.getRecord("/customer1.json").collections;
+
+assertions.push(
+  // doc should have been written
+  test.assertEqual("1", results.lastCompletedStep),
+
+  // doc should only have collection from mapping step
+  test.assertEqual(1, collections.length),
+  test.assertTrue(collections.includes("Customer")),
+  test.assertFalse(collections.includes("source"))
```
rjrudin, line 30: Same thing - use an assertion message instead of a comment

rjrudin (changes requested):
Looks good, just some little tweaks to the test

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/flow/testTargetCollectionAdditivity/flowWithTargetCollectionsAdditivity.sjs
jyangML: I think it makes the code more readable to say `collections.includes("Customer")` than `collections[0] == "Customer"`.  Unless you are majorly against this, I'd rather keep the `.includes` and just add a message instead

rjrudin (approved):

srinathgit (approved):

PR #5496, opened 2021-04-09
DHFPROD-5850:Pass sequence to custom SJS mapping function

srinathgit:
### Description
Pass sequence to custom SJS mapping function
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mapping/entity-services/xquery-lib.xqy
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mapping/entity-services/xquery-lib.xqy
@@ -19,6 +19,7 @@ xquery version "1.0-ml";
 module namespace xquery-lib = "http://marklogic.com/mapping/es/xquery";
 
 import module namespace inst="http://marklogic.com/entity-services-instance" at "/MarkLogic/entity-services/entity-services-instance.xqy";
+import module namespace es="http://marklogic.com/entity-services" at "/MarkLogic/entity-services/entity-services.xqy";
 import module namespace json = "http://marklogic.com/xdmp/json" at "/MarkLogic/json/json.xqy";
 
 declare function document-with-nodes($nodes as node()*) {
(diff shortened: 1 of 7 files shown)


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mapping/entity-services/xquery-lib.xqy
```diff
@@ -105,3 +106,52 @@ declare function remove-empty-structured-properties($element as item()*) as docu
...
+  return
+    xdmp:eval($put,
+      map:map()=>map:with("uri",$uri)=>
+      map:with("node",$compiled)=>
+      map:with("collection",$es:FUNCTIONDEF_COLLECTION)
+      ,
+      map:map()=>map:with("database",xdmp:modules-database()))
+};
+
+declare function function-metadata-compile(
```
rjrudin, line 149: Is the only difference with these functions that they're using the DHF version of the stylesheet?

Whatever the difference is, please add comments here to explain why we've copied/tweaked these ES functions so a future developer understands the intent.

rjrudin (changes requested):
Looks good, just need a comment on the modified ES code

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mapping/entity-services/xquery-lib.xqy
srinathgit: Essentially we are using xqueryLib.functionMetadataPut instead of OOTB es.functionMetadataPut. I will add a comment both in this file and also in mlGenerateFunctionMetadata.sjs where it is used.

srinathgit: Added a comment both in this file and in mlGenerateFunctionMetadata.sjs

rjrudin (approved):

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/mapping/entity-services/function-metadata.xsl
```diff
@@ -97,7 +97,7 @@
               <xsl:value-of select="@name"/>
             </xsl:attribute>
             <xsl:attribute name="select">
-              <xsl:value-of select="concat('($',@name,',null-node{})[1]')"/>
+                <xsl:value-of select="concat('if (empty($', @name ,')) then null-node{} else $', @name)"/>
```
srinathgit, line 100: @ryanjdew
This seems to work fine. I didn't have to add xsl:choose element. Let me know if you see any issues with this.

ryanjdew (approved):

PR #5502, opened 2021-04-12
DHFPROD-6981:Enable running 'hubDeploy' on prem with non-default resource names/values

srinathgit:
### Description
Enable running 'hubDeploy' on prem with non-default resource names/values

@rjrudin 
I am not sure what test can be written for this. I deployed datahub with non default names for dbs/ servers and port numbers. Then ran hubDeploy, it worked fine.

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/DeployHubDatabaseCommand.java
```diff
@@ -40,6 +40,7 @@
     private File databaseFile;
     private boolean mergeEntityConfigFiles = true;
     private boolean mergeExistingArrayProperties = false;
+    private boolean removeSchemaAndTriggersDatabaseSettings = false;
```
rjrudin, line 43: My analysis of 6981 was that the root cause here is that DhsDeployer should not call setKnownValuesForDhsDeployment unless mlIsProvisionedEnvironment=true. I'm not clear on what problem is being fixed here. This logic is saying "If this is DHS, then we know that including schema-database / triggers-database will result in a privilege error, so we don't include them". That seems unrelated to the problem in 6981.

rjrudin (changes requested):

marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/DeployHubDatabaseCommand.java
srinathgit: It's not just dhs, the command(hubDeployAsDeveloper) would fail in on prem environments as well with a user with "data-hub-developer". Currently, we are working around it by setting hubConfig.setIsProvisionedEnvironment(true) so we are always removing staging and triggers db settings from payload. Now if we want the command to work for both for on prem and dhs, we cannot always set hubConfig.setIsProvisionedEnvironment(true). We can set removeSchemaAndTriggersDatabaseSettings=true when running this command since we know "data-hub-developer"  would always want the staging and triggers db settings removed from payload.

rjrudin:
I'm trying this out locally now.

marklogic-data-hub/src/main/java/com/marklogic/hub/dhs/DhsDeployer.java
```diff
@@ -99,20 +90,24 @@ protected void prepareAppConfigForDeployingToDhs(HubConfig hubConfig) {
         removeHubInternalConfigFromConfigDirs(appConfig);
         addEntityConfigToConfigDirs(hubConfig.getHubProject(), appConfig);
 
-        // 8000 is not available in DHS
-        int port = hubConfig.getPort(DatabaseKind.STAGING);
-        logger.info("Setting App-Services port to: " + port);
-        appConfig.setAppServicesPort(port);
+        if(hubConfig.getIsProvisionedEnvironment()){
```
rjrudin, line 93: For testing this, you can add a scenario to DeployAsDeveloperTest (which is currently failing) where this is false, and you call prepareAppConfig and then verify that the below modifications were not made.

rjrudin (changes requested):

marklogic-data-hub/src/main/java/com/marklogic/hub/dhs/DhsDeployer.java
srinathgit: I've added a test for on prem (vanilla datahub)

rjrudin (approved):

ryanjdew (approved):

PR #5504, opened 2021-04-12
DHFPROD-7015:Entity models should not be loaded from subfolders

srinathgit:
### Description
Entity models should not be loaded from subfolders. No test needed, manual tests would suffice.
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/LoadUserArtifactsCommand.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/LoadUserArtifactsCommand.java
@@ -129,30 +129,22 @@ public class LoadUserArtifactsCommand extends AbstractCommand {
      * @throws IOException
      */
     private void loadModels(HubClient hubClient) throws IOException {
-        final Path modelsPath = hubConfig.getHubEntitiesDir();
-        if (modelsPath.toFile().exists()) {
-            ArrayNode modelsArray = objectMapper.createArrayNode();
-            EntityDefModulesFinder modulesFinder = new EntityDefModulesFinder();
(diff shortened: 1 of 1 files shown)


ryanjdew (dismissed):

marklogic-data-hub/src/test/java/com/marklogic/hub/deploy/commands/LoadUserArtifactsCommandTest.java
```diff
@@ -205,4 +207,26 @@ public void validateArtifactNames() throws Exception {
             () -> loadUserArtifactsCommand.execute(commandContext));
         assertEquals(errorMessage, ex.getServerMessage());
     }
+
+    @Test
+    void entityModelShouldNotBeLoadedFromSubFolder(){
```
rjrudin, line 212: I actually don't think we should include a test for this - the behavior of loading from a subfolder is not expected and is inconsistent with all other artifact types. I think deleting the code is sufficient here, we don't need to verify that DHF isn't doing something we don't expect it to do (we'd have an infinite number of tests to write otherwise).

rjrudin (changes requested):

marklogic-data-hub/src/test/java/com/marklogic/hub/deploy/commands/LoadUserArtifactsCommandTest.java
srinathgit: I agree, I have removed the test.

rjrudin (approved):

ryanjdew (approved):

PR #5508, opened 2021-04-13
DHFPROD-6843: add quality to documents created by step

jyangML:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [X] Added Tests
  

- ##### Reviewer:

- [ ] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/e2e/cypress/integration/persistence/persistence.spec.tsx
+++ b/marklogic-data-hub-central/ui/e2e/cypress/integration/persistence/persistence.spec.tsx
@@ -9,7 +9,6 @@ describe("Validate persistence across Hub Central", () => {
     cy.contains(Application.title);
     cy.loginAsTestUserWithRoles("hub-central-flow-writer", "hub-central-match-merge-writer", "hub-central-mapping-writer", "hub-central-load-writer", "hub-central-entity-model-reader", "hub-central-entity-model-writer", "hub-central-saved-query-user").withRequest();
     LoginPage.postLogin();
-    cy.waitForAsyncRequest();
   });
   beforeEach(() => {
     cy.loginAsTestUserWithRoles("hub-central-flow-writer", "hub-central-match-merge-writer", "hub-central-mapping-writer", "hub-central-load-writer", "hub-central-entity-model-reader", "hub-central-entity-model-writer", "hub-central-saved-query-user").withRequest();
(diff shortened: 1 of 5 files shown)


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/hub-utils/invoke-queue-write.sjs
```diff
@@ -88,6 +88,15 @@ for (let content of contentArray) {
     deleteContent(content, temporalCollection);
   } else {
     const metadata = context.metadata;
+    const quality = context.quality || 0;
+
+    if (!Number.isInteger(quality)) {
```
srinathgit, line 93: I think if(!Number(quality)) is sufficient , other checks can be removed. Also the error message can provide more information like doc uri. Something like `Unable to insert document: ${content.uri}; cause: invalid value for 'quality' : String(quality)`

srinathgit (changes requested):
Schema for context object has to be updated to include quality

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/flow/stepWithQuality/runStepWithQuality.sjs
```diff
@@ -0,0 +1,29 @@
+const test = require("/test/test-helper.xqy");
+const flowUtils = require("/data-hub/5/impl/flow-utils.sjs");
+
+let assertions = [];
+
+const content = [
```
srinathgit, line 6: Can you add a test with invalid value for quality as well ?

srinathgit (changes requested):

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/hub-utils/invoke-queue-write.sjs
rjrudin: I actually don't think we need any code here; ML will throw an error itself. We should see what that error looks like before writing any code. Otherwise, we're getting into the business of duplicating the logic that ML already enforces on valid values for quality.

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/flow/stepWithQuality/runStepWithQuality.sjs
rjrudin: I think that's good, and that test can determine whether or not the ML error is good enough.

Note that an erroneous value would be really unusual here. "quality" is not used much IMO, and when Pari does use it, we can typically count on her not shooting herself in the foot with an invalid value.

rjrudin (changes requested):

rjrudin (dismissed):

srinathgit (changes requested):
Can you update context's schema to include 'quality'  ? And also run generateSchemaPojos task

srinathgit (changes requested):
The file that you should checkin is Context.java. There is no change at all in marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/schemas/ContentObjectSchema.java except for spaces.

jyangML:
Not sure what's going on.  I can't reopen the PR anymore

PR #5532, opened 2021-04-16
DHFPROD-7051: Sorting when Querying Jobs Data

rahulvudutala:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/flow/job-query-lib.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/flow/job-query-lib.sjs
@@ -24,16 +24,20 @@ function findStepResponses(query) {
   const sortColumn = query.sortColumn;
   const sortDirection = query.sortDirection;
   const response = {};
+  const orderByConstraint = [];
 
-  let orderByConstraint = op.desc('startTime');
   if(sortColumn) {
-    orderByConstraint = sortDirection === 'ascending' ? op.asc(sortColumn) : op.desc(sortColumn);
+    orderByConstraint.push(sortDirection === 'ascending' ? op.asc(sortColumn) : op.desc(sortColumn));
+  }
+
+  if(sortColumn !== 'startTime') {
+    orderByConstraint.push(op.desc('startTime'));
   }
 
   const totalCountQuery = 'select count(*) as total from Job.StepResponse';
   const jobsDataQuery = 'select Job.StepResponse.stepName as stepName,' +
       'Job.StepResponse.stepDefinitionType as stepDefinitionType,' +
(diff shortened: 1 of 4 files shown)


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/flow/job-query-lib.sjs
```diff
@@ -24,16 +24,20 @@ function findStepResponses(query) {
...
+
+  if(sortColumn !== 'startTime') {
+    orderByConstraint.push(op.desc('startTime'));
   }
 
   const totalCountQuery = 'select count(*) as total from Job.StepResponse';
   const jobsDataQuery = 'select Job.StepResponse.stepName as stepName,' +
       'Job.StepResponse.stepDefinitionType as stepDefinitionType,' +
-      'Job.StepResponse.stepStatus as status,' +
+      'Job.StepResponse.jobStatus as status,' +
```
rjrudin, line 40: I don't think "status" is good to use here, because that begs the question - Is this the job status, or is it the step status? 

Given that this is a StepResponse, I think it's expected that "status" would refer to the step status. But it's not - it's the job status. So this should really be "as jobStatus".

rjrudin: Also, the fact that this code changed, but there's no accompanying change in a test, means that we don't have any test coverage on this column. That should be easy to add to your test below - you'd just verify that the expected job status is returned.

rjrudin (changes requested):

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/data-services/job/testFindStepResponses.sjs
```diff
@@ -137,7 +191,46 @@ function verifyPaginationInStepResponse() {
     return assertions;
 }
 
+function verifyJobStatusInStepResponse() {
```
rjrudin, line 194: You don't need a whole separate test just to verify one additional column. 

What I'll typically do here is have a "basic happy path" test where I validate every detail, and I use a basic set of inputs. Then I'll have other tests with varying sets of inputs that produce different outcomes that I need to test for. 

It looks like verifyStepResponse should be that "basic happy path" test should be that test, but it's not really validating the columns - it's just verifying that the value is null. We need at least one test that verifies every column - we can't assume that any column is properly configured.

rjrudin (changes requested):

ryanjdew (approved):

rjrudin (approved):

PR #5538, opened 2021-04-19
DHFPROD-7115: Generate facets for structured properties

akshaysonvane:
### Description
- I am storing the range index path for facetable structured properties in the generated primary entity type in the uber-model as a result we can now generate facets for structured properties with the same name in two different entity types as well.

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/hub-entities.xqy
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/hub-entities.xqy
@@ -186,7 +186,7 @@ declare %private function hent:fix-options-for-explorer(
         where fn:not($container-for-entity-property-generated-by-es)
         return
           element {fn:node-name($n)} {
-            $n/@*,
+            $n/@* ! attribute {fn:node-name()} { fn:replace(., "/", ".")},
             let $path-expression := fix-path-expression(fn:string($n/search:range/search:path-index))
             let $search-range-node := $n/search:range
             let $is-sortable-only :=
@@ -775,6 +775,33 @@ declare function hent:json-schema-generate($entity-title as xs:string, $uber-mod
   return xdmp:to-json($uber-model)
 };
 
+declare %private function hent:build-indexes-for-structured-entity-properties($entity-path as xs:string, $definiton-name as xs:string, $definitions as map:map, $primary-entity-definition as map:map) {
+  let $entity-type := map:get($definitions, $definiton-name)
+  let $entity-type-properties := map:get($entity-type, "properties")
+
+  let $_ :=
+    for $entity-type-property in map:keys($entity-type-properties)
+    let $ref := map:get($entity-type-properties, $entity-type-property)=>map:get("$ref")
+    let $items := map:get($entity-type-properties, $entity-type-property)=>map:get("items")
+    return
+      if (fn:empty($ref) and (fn:empty($items) or fn:not(fn:starts-with(map:get($items, "$ref"), "#")))) then
+        if (map:get($entity-type-properties, $entity-type-property)=>map:get("facetable")) then
+          json:array-push(map:get($primary-entity-definition, "rangeIndex"), $entity-path || "/" || $entity-type-property)
+        else ()
+      else
+        let $definiton-name :=
+          if (fn:empty($items)) then
+            fn:substring-after($ref, "#/definitions/")
+          else
+            fn:substring-after(map:get($items, "$ref"), "#/definitions/")
+        where $definiton-name
+        return
+          let $path := $entity-path || "/" || $entity-type-property || "/" || $definiton-name
+          let $_ := hent:build-indexes-for-structured-entity-properties($path, $definiton-name, $definitions, $primary-entity-definition)
+          return ()
+  return $primary-entity-definition
+};
+
 (:
   this function finds the first level facetable entityType properties and constrcuts and adds the rangeIndex array to
   the entityModel. All the structured properties are ignored for now even if a property is modeled as facetable as per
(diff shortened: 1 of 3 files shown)


akshaysonvane:
Converted to draft status to fix the failing test.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/hub-entities.xqy
```diff
@@ -775,6 +775,35 @@ declare function hent:json-schema-generate($entity-title as xs:string, $uber-mod
...
+          json:array-push(map:get($primary-entity, "rangeIndex"), $entity-path || "/" || $entity-type-property)
+        else ()
+      else
+        let $definiton-name :=
+          if (fn:empty($items)) then
+            fn:substring-after($ref, "#/definitions/")
+          else
+            fn:substring-after(map:get($items, "$ref"), "#/definitions/")
+        let $_ :=
+          if (fn:not(fn:not($definiton-name))) then
```
rjrudin, line 798: I think this can be cleaner if you just do:

```
let $definition-name := ...
where $definition-name
return 
  let $path := ...
  ...
  return ()
```

That replaces an empty "else" with a where clause, and I'm 99% certain you don't need a double fn:not.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/hub-entities.xqy
```diff
@@ -775,6 +775,35 @@ declare function hent:json-schema-generate($entity-title as xs:string, $uber-mod
...
+            let $path := $entity-path || "/" || $entity-type-property || "/" || $definiton-name
+            let $_ := hent:add-indexes-for-structured-entity-properties($path, $definiton-name, $definitions, $primary-entity)
+            return ()
+          else ()
+        return ()
+  return $primary-entity
```
rjrudin, line 804: The function name starts with "add", which typically implies that there's no return value - i.e. you're adding something to something else that already exists.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/hub-entities.xqy
```diff
@@ -808,25 +837,38 @@ declare %private function hent:add-indexes-for-entity-properties($entities as js
...
+                let $definiton-name :=
+                  if (fn:empty($items)) then
+                    fn:substring-after($ref, "#/definitions/")
+                  else
+                    fn:substring-after(map:get($items, "$ref"), "#/definitions/")
+                let $_ :=
+                  if (fn:not(fn:not($definiton-name))) then
+                    let $definitions := map:get($model, "definitions")
+                    let $entity-path := $entity-type-property || "/" || $definiton-name
+                    let $model := hent:add-indexes-for-structured-entity-properties($entity-path, $definiton-name, $definitions, $entity-type)
```
rjrudin, line 855: This is confusing - it implies that something useful is being returned that's assigned to "$model", but then "$model" is ignored. But based on the function def, "$entity-type" is what indexes are being added to.

I think the main confusion is over the name "entity-type". That implies that it's a string, but it's not - it's a map, it's the actual entity definition. So that should be renamed to "$entity-definition" (or just "$entity-def", as "def" is an intuitive abbreviation for "$definition").

"$model" should only refer to what a model is - i.e. the entire model descriptor, which contains "definitions" and "info".

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/impl/hub-entities/generateIndexes.sjs
```diff
@@ -232,22 +232,89 @@ function generateIndexConfigWithStructuredProperties() {
...
+              "items": {
+                "datatype": "string"
+              }
+            },
+            "id": {
+              "datatype": "string",
+              "facetable": false,
+              "collation": "http://marklogic.com/collation/"
+            },
+            "customer": {
```
rjrudin, line 259: I think this should be called "addresses", not "customer", since it's an array of Addresses.

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/impl/hub-entities/generateIndexes.sjs
```diff
@@ -232,22 +232,89 @@ function generateIndexConfigWithStructuredProperties() {
...
+              }
+            },
+            "address": {
+              "$ref": "#/definitions/Address"
+            }
+          }
+        },
+        "Address": {
+          "properties": {
+            "street": {
```
rjrudin, line 272: I would simplify this model by removing everything you're not testing against, as it's just noise - so remove "street". But "state" is useful in that we want to verify that there's not a range index generated for it.

rjrudin (changes requested):

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/hub-entities.xqy
```diff
@@ -186,7 +186,7 @@ declare %private function hent:fix-options-for-explorer(
         where fn:not($container-for-entity-property-generated-by-es)
         return
           element {fn:node-name($n)} {
-            $n/@*,
+            $n/@* ! attribute {fn:node-name()} { fn:replace(., "/", ".")},
```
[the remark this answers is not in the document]
akshaysonvane: The forward slashes are used by the `es:database-properties-generate()` method to generate the correct range index path.
As a side-effect, `es.searchOptionsGenerate()` method treats the range index path as a single property and thus we need to replace all the forward slashes with periods.

rjrudin (approved):

srinathgit (approved):

PR #5539, opened 2021-04-20
DHFPROD-6959:Enable use of hubURI() function in mapping expression

srinathgit:
Enable use of hubURI() function in mapping expression. In addition, this commit includes 
1. Backend work for DHFPROD-6711 as it's required for the functionality 
2. DHFPROD-7056 which doesn't require  any implementation, included a test for verification.

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/mapping-functions/core.sjs
```diff
@@ -201,23 +201,34 @@ function isNullOrEmpty(value) {
...
+ * @param object
+ * @returns
  */
 function lookup(key, object) {
   return object ? object[key] : null;
 }
 
+function hubURI(entityType) {
+  if(typeof entityType != "string"){
+    httpUtils.throwBadRequest(`Unable to generate uri: 'entityType' should be of type 'string'`);
```
rjrudin, line 219: Can you verify that the Test capability in HC supports 400 errors, and also the generic Error that is thrown in other places in this library? That doesn't matter for this PR, just something to test.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/mapping-functions/core.sjs
```diff
@@ -201,23 +201,34 @@ function isNullOrEmpty(value) {
...
+}
+
 module.exports = {
   documentLookup,
   lookup,
   memoryLookup,
   parseDate,
-  parseDateTime
+  parseDateTime,
+  hubURI
```
rjrudin, line 233: Let's keep these in alphabetical order, makes life easier.

marklogic-data-hub/src/test/ml-modules/root/test/data-hub-test-helper.sjs
```diff
@@ -250,5 +254,6 @@ module.exports = {
   getStagingRecord,
   stagingDocumentExists,
   verifyJson,
+  getUrisInCollection,
```
rjrudin, line 257: Same thing, use alphabetical order (though I don't mind the amped functions being at the end since they're special)

rjrudin (approved):
I'm approving - no need to change the functions to be in alphabetical order, just something to keep in mind in the future.

ryanjdew (approved):

PR #5541, opened 2021-04-20
DHFPROD-7233:Middle tier work for creating/updating entity model with namespace/prefix

srinathgit:
### Description
Middle tier work for creating/updating entity model with namespace/prefix
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/controllers/ModelController.java
+++ b/marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/controllers/ModelController.java
@@ -91,9 +91,13 @@ public class ModelController extends BaseController {
     }
 
     @RequestMapping(value = "/{modelName}/info", method = RequestMethod.PUT)
+    @ApiImplicitParam(required = true, paramType = "body", dataType = "UpdateModelInfoInput")
     @Secured("ROLE_writeEntityModel")
-    public ResponseEntity<Void> updateModelInfo(@PathVariable String modelName, @RequestBody UpdateModelInfoInput input) {
-        newService().updateModelInfo(modelName, input.description);
+    public ResponseEntity<Void> updateModelInfo(@PathVariable String modelName, @RequestBody @ApiParam(hidden = true) JsonNode input) {
+        if(input.get("name") != null && !(input.get("name").asText().equals(modelName))){
+            throw new RuntimeException("Unable to update entity model; incorrect model name: " + input.get("name").asText());
+        }
+        newService().updateModelInfo(modelName, input);
         return new ResponseEntity<>(HttpStatus.OK);
     }
 
@@ -258,10 +262,15 @@ public class ModelController extends BaseController {
     public static class CreateModelInput {
         public String name;
         public String description;
+        public String namespace;
+        public String namespacePrefix;
     }
 
     public static class UpdateModelInfoInput {
+        public String name;
         public String description;
+        public String namespace;
+        public String namespacePrefix;
     }
 
     public static class ModelReferencesInfo {
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/models/createModel.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/models/createModel.sjs
@@ -21,11 +21,14 @@ xdmp.securityAssert("http://marklogic.com/data-hub/privileges/write-entity-model
 
 const httpUtils = require("/data-hub/5/impl/http-utils.sjs");
 const entityLib = require("/data-hub/5/impl/entity-lib.sjs");
+const hubUtils = require("/data-hub/5/impl/hub-utils.sjs");
 
 var input = fn.head(xdmp.fromJSON(input));
 
 const name = input.name;
 const description = input.description;
+const namespace = input.namespace ? input.namespace : null;
+const namespacePrefix = input.namespacePrefix ? input.namespacePrefix : null;
 
 if (name == null) {
   httpUtils.throwBadRequest("The model must have an info object with a title property");
@@ -46,10 +49,26 @@ model.definitions[name] = {
   properties: {}
 };
 
+if(namespace || namespacePrefix){
+  if(!namespace){
+    httpUtils.throwBadRequest(`You cannot enter a prefix without specifying a namespace URI `);
+  }
+  if(!namespacePrefix){
+    httpUtils.throwBadRequest(`Since you entered a namespace, you must specify a prefix.`);
+  }
+  model.definitions[name].namespace = namespace;
+  model.definitions[name].namespacePrefix = namespacePrefix;
+}
+
 if (input.description) {
   model.definitions[name].description = description;
 }
 
-entityLib.writeModel(name, model);
+try{
+  entityLib.writeModel(name, model);
+}
+catch (e){
+  httpUtils.throwBadRequest(hubUtils.getErrorMessage(e));
+}
 
 model;
--- a/specs/models/ModelDefinitions.json
+++ b/specs/models/ModelDefinitions.json
@@ -36,6 +36,12 @@
             "type": "string"
           }
         },
+        "namespace": {
+          "type": "string"
+        },
+        "namespacePrefix": {
+          "type": "string"
+        },
         "properties": {
           "type": "object",
           "additionalProperties": false,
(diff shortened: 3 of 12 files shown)


marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/controllers/ModelController.java
```diff
@@ -92,8 +92,8 @@
 
     @RequestMapping(value = "/{modelName}/info", method = RequestMethod.PUT)
```
srinathgit, line 93: @rjrudin 
Is it ok to leave this endpoint as  /{modelName}/info or should it be renamed ( /{modelName}/infoAndNamepace)?

ryanjdew (dismissed):

specs/models/ModelDefinitions.json
```diff
@@ -75,6 +75,12 @@
                 "description": {
                   "type": "string"
                 },
+                "namespace": {
```
rjrudin, line 78: This doesn't seem like the right location - these are definition-level properties right, not properties-level properties?

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/controllers/ModelController.java
rjrudin: I think something is off here - "info" refers to model-level data, such as name/description (and possibly in the future, version and baseUri). 

namespace/namespacePrefix are definition-level data. So it doesn't make sense to save those together with model-level data. 

ES allows for each definition to have its own namespace/namespacePrefix. That certainly doesn't seem common, but it's allowable. So I think we need to allow that here as well - which means I think the story needs to be adjusted. 

If the story stays as-is, then the assumption would be that whatever the user specifies here should only be added to the "primary" entity definition. But - a user would likely still need a way to set this for the structured-type entity definitions as well.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/models/createModel.sjs
```diff
@@ -21,11 +21,14 @@ xdmp.securityAssert("http://marklogic.com/data-hub/privileges/write-entity-model
 
 const httpUtils = require("/data-hub/5/impl/http-utils.sjs");
 const entityLib = require("/data-hub/5/impl/entity-lib.sjs");
+const hubUtils = require("/data-hub/5/impl/hub-utils.sjs");
 
 var input = fn.head(xdmp.fromJSON(input));
 
 const name = input.name;
 const description = input.description;
+const namespace = input.namespace ? input.namespace : null;
```
rjrudin, line 30: If we do end up adding these 2 inputs, I think we should turn this into a JsonNode that gets passed to the DS endpoint.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/models/createModel.sjs
```diff
@@ -46,10 +49,26 @@ model.definitions[name] = {
...
 };
 
+if(namespace || namespacePrefix){
+  if(!namespace){
+    httpUtils.throwBadRequest(`You cannot enter a prefix without specifying a namespace URI `);
+  }
+  if(!namespacePrefix){
+    httpUtils.throwBadRequest(`Since you entered a namespace, you must specify a prefix.`);
+  }
+  model.definitions[name].namespace = namespace;
```
rjrudin, line 59: This assumption - that the namespace/namespacePrefix are to be applied to the primary entity type - seems reasonable to do. But it raises the question of what to do about structured type properties.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/models/updateModelInfoAndNamespace.api
```diff
@@ -0,0 +1,24 @@
+{
+  "functionName" : "updateModelInfoAndNamespace",
```
rjrudin, line 2: updateModelInfo is fine and really preferable, since there technically is no such thing as a "model namespace", there's only a "definition namespace".

rjrudin (changes requested):

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/controllers/ModelController.java
wooldridge: @rjrudin @srinathgit so on the UI side, if we want to enable the user to add a namespace and prefix to a structured property, that would go in this form:

Screen Shot 2021-04-21 at 9 12 35 AM

For the UI work, if we are to do it, that should be a separate story. FYI: @sbayatpur

wooldridge:
I've asked @sbayatpur to decide whether to move forward with just the non-structured entity support for this feature.

wooldridge (dismissed):
I'm seeing the POST and PUT versions working for entities with namespace and namespacePrefix info.

ryanjdew (dismissed):

rjrudin (approved):

Sanjeevani19 (commented):
@srinathgit Looks like this test is continuously failing, https://jenkins.marklogic.com/blue/organizations/jenkins/Datahub_CI/detail/PR-5541/14/tests, could you please look into that

wooldridge:
> @srinathgit Looks like this test is continuously failing, https://jenkins.marklogic.com/blue/organizations/jenkins/Datahub_CI/detail/PR-5541/14/tests, could you please look into that

I can take a look.

srinathgit:
> In writerScenario1 in Modeling, when the entity description is edited and the updated entity is saved, the server is throwing a 500 error. Looks like it is a problem with the PUT:
> 
> ```
> 2021-04-22 09:56:27.706 [http-nio-8080-exec-4] DEBUG o.s.w.f.CommonsRequestLoggingFilter - Before request [PUT /api/models/Buyer/info, client=127.0.0.1, session=9EED871EEF24B00DA7E9CCFE1D790D83, user=hc-test-user]
> 2021-04-22 09:56:27.715 [http-nio-8080-exec-4] ERROR c.m.h.c.e.CustomExceptionHandler - null
> java.lang.NullPointerException: null
>         at com.marklogic.hub.central.controllers.ModelController.updateModelInfo(ModelController.java:97)
>         at com.marklogic.hub.central.controllers.ModelController$$FastClassBySpringCGLIB$$27e9bf12.invoke(<generated>)
>         at org.springframework.cglib.proxy.MethodProxy.invoke(MethodProxy.java:218)
>         at org.springframework.aop.framework.CglibAopProxy$CglibMethodInvocation.invokeJoinpoint(CglibAopProxy.java:771)
>         at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:163)
>         at org.springframework.aop.framework.CglibAopProxy$CglibMethodInvocation.proceed(CglibAopProxy.java:749)
>         at org.springframework.security.access.intercept.aopalliance.MethodSecurityInterceptor.invoke(MethodSecurityInterceptor.java:69)
>         at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:186)
> ```

I think the latest commit should take care of the NPE

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/controllers/ModelController.java
```diff
@@ -91,9 +91,13 @@
     }
 
     @RequestMapping(value = "/{modelName}/info", method = RequestMethod.PUT)
+    @ApiImplicitParam(required = true, paramType = "body", dataType = "UpdateModelInfoInput")
     @Secured("ROLE_writeEntityModel")
-    public ResponseEntity<Void> updateModelInfo(@PathVariable String modelName, @RequestBody UpdateModelInfoInput input) {
-        newService().updateModelInfo(modelName, input.description);
+    public ResponseEntity<Void> updateModelInfo(@PathVariable String modelName, @RequestBody @ApiParam(hidden = true) JsonNode input) {
+        if(input.get("name") != null && !(input.get("name").asText().equals(modelName))){
+            throw new RuntimeException("Invalid payload: Incorrect model name: "+ input.get("name").asText());
```
rjrudin, line 98: We don't want to use the term "payload" here, which is not meaningful to a user. The fact that the model name is incorrect is meaningful, though it appears to be either due to programmer error or malicious use. So a better message would be "Unable to update entity model; incorrect model name: " + name.

rjrudin (changes requested):
Looks good, just one little tweak for an error message.

rjrudin (approved):

wooldridge:
On this PR branch, I'm still seeing the 500 error when editing entity description, both in the e2e test and manually on my local: https://drive.google.com/file/d/1aMPqgGxWDKDqY1yVrKJu7FLEfWXDZm6q/view

wooldridge (approved):

PR #5549, opened 2021-04-21
DHFPROD-7114: Cleaning up custom step template and starting public API

rjrudin:
### Description

Trimmed down main-custom.sjs to focus on what the inputs and outputs are, along with a few simple things that can be done in conjunction with makeEnvelope. 

Taking a stab at a true public API in DHF/ML code as well via flowApi.sjs. 

Removing scaffold templates for mapping/mastering - those are OBE since QS is gone.

Not touching custom-ingestion.sjs yet - I think the value of that template is extremely low since ingestion steps don't have much value in the first place. 

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


srinathgit (approved):

marklogic-data-hub/src/main/resources/scaffolding/custom-module/sjs/main-custom.sjs
```diff
@@ -1,125 +1,64 @@
...
-  return newContent;
+  const inputDocument = content.value;
+
+  // DHF recommends wrapping documents in an envelope, particularly for curated documents based on an entity model.
+  // If your input document is already an envelope, you can typically just call toObject() it and modify what you need.
+  // Otherwise, the below code is a starting point for constructing the 3 parts of an envelope.
+  // If your input document is XML and you need to modify it, it is recommended to generate an XQuery custom step instead.´
+  const instance = inputDocument.toObject();
+  const headers = {};
+  const triples = [];
```
rjrudin, line 35: Let's include an example here of how to use .xpath() to access data in an XML document, as that is a common use case and typically not well-known by a new ML developer.

akshaysonvane (approved):

PR #5551, opened 2021-04-22
DHFPROD-7088:Paginate Jobs Data UI

Sanjeevani19:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


ryanjdew (approved):

timur-isangulov (changes requested):
@Sanjeevani19, do monitorOptions need to be part of the Explorer search-context ? Since it's separate page probably it's better to create its own context file.

I am reusing some of the existing code components from explorer, so it's ideal to add it to the searchContext just to avoid repeating code

Sanjeevani19:
> @Sanjeevani19, do monitorOptions need to be part of the Explorer search-context ? Since it's separate page probably it's better to create its own context file.
> 
> @timur-isangulov I am reusing some of the existing code components from explorer, so it's ideal to add it to the searchContext just to avoid repeating code

rahulvudutala (approved):

timur-isangulov (approved):

PR #5554, opened 2021-04-23
DHFPROD-7123: Add job documents REST endpoint

ryanjdew:
### Description

Still needs tests but wanted earlier feedback, so creating this draft PR.

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [ ] Added Tests
  

- ##### Reviewer:

- [ ] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/flow/job-query-lib.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/flow/job-query-lib.sjs
@@ -59,6 +59,43 @@ function findStepResponses(query) {
   return response;
 }
 
+function valuesExist(values) {
+  return !(values || values.length);
+}
+
+function findJobDocuments({start, pageLength, jobId, status, startTimeBegin, startTimeEnd, endTimeBegin, endTimeEnd, user}) {
+  const queries = [cts.collectionQuery('Job')];
+  if (valuesExist(jobId)) {
+    queries.push(cts.jsonPropertyValueQuery('jobId', jobId));
+  }
+  if (valuesExist(status)) {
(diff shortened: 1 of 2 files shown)


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/flow/job-query-lib.sjs
```diff
@@ -59,6 +59,43 @@ function findStepResponses(query) {
   return response;
 }
 
+function valuesExist(values) {
+  return !(values || values.length);
+}
+
+function findJobDocuments({start, pageLength, jobId, status, startTimeBegin, startTimeEnd, endTimeBegin, endTimeEnd, user}) {
```
rjrudin, line 66: Seems fine so far, but just keep in mind whether /v1/search would be simpler. Search options have the benefit of being self-documenting too - i.e. our docs can say "Just look at host:8013/v1/config/query/hubJobDocuments" instead of having to spell out all the parameters and also maintain that list.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/flow/job-query-lib.sjs
```diff
@@ -59,6 +59,43 @@ function findStepResponses(query) {
...
+  if (valuesExist(user)) {
+    queries.push(cts.jsonPropertyValueQuery('user', user));
+  }
+  const finalQuery = cts.andQuery(queries);
+  const total = cts.estimate(finalQuery);
+  return {
+    total,
+    start,
+    pageLength,
+    results: fn.subsequence(cts.search(finalQuery), start, pageLength)
```
rjrudin, line 95: Thoughts on using jsearch instead? In theory that should be a simpler interface to code against.

PR #5569, opened 2021-04-26
DHFPROD-7164:Test Mapping expressions for related entities

srinathgit:
### Description
Test Mapping expressions for related entities
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mapping/entity-services/lib.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mapping/entity-services/lib.sjs
@@ -373,7 +373,7 @@ function escapeXML(input = '') {
  * @param uri
  * @returns
  */
-function validateAndRunMapping(mapping, uri) {
+function validateAndTestMapping(mapping, uri) {
   if (!fn.docAvailable(uri)) {
     throw Error(`Unable to validate and run mapping; could not find source document with URI '${uri}'`);
   }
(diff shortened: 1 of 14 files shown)


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mapping/entity-services/lib.sjs
```diff
@@ -396,10 +396,32 @@ function validateAndRunMapping(mapping, uri) {
   }
 
   const parameterMap = Object.assign({}, {"URI":uri}, userParameterMap);
+  const mappingCount = 1 + (mapping.relatedEntityMappings ? mapping.relatedEntityMappings.length : 0);
```
rjrudin, line 399: This looks like it can go into a separate method, which will help keep validateAndRunMapping fairly lean and mean. 

Specifically, it looks like you can extract a "validateMappings" function, which makes it very clear that there's a need to validate more than one mapping. And then you can extract a "testMappings" function. Those two functions will reduce the size of validateAndRunMapping - which should likely also be renamed to "validateAndTestMapping" - and make it clear that we first need to validate all the mappings, and then we need to test all the mappings.

rjrudin (changes requested):

rjrudin (approved):

briantang (approved):
Nice, response looks good for the UI to build on.

PR #5570, opened 2021-04-26
DHFPROD-3754: Add namespace fields to Entity form

wooldridge:
### Description

Allow users to specify a namespace and namespace prefix for non-structured properties in Modeling.

The backend handles the following errors:

- Namespace and prefix must be defined if either one is defined.
- Prefix cannot be a reserved word (e.g., "xml").
- Namespace needs to be an absolute URI (e.g., "http://example.org/test").
- Entity name cannot be a duplicate of existing entity name (this validation already exists).

Server-side errors are displayed in a common place at the bottom of the form. @jbelonoj has approved the UX.

Tests for this feature are handled with RTL unit tests. 

Entity payload is now handled slightly differently now so relevant tests were updated. Also cleaned up entity-type-modal tests to avoid repetition and hardcoding. 

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/e2e/cypress/integration/modeling/writerScenario1.spec.tsx
+++ b/marklogic-data-hub-central/ui/e2e/cypress/integration/modeling/writerScenario1.spec.tsx
@@ -117,8 +117,6 @@ describe("Entity Modeling Senario 1: Writer Role", () => {
     entityTypeModal.newEntityDescription("Description has changed");
     entityTypeModal.getAddButton().click();
     entityTypeModal.getAddButton().should("have.css", "width", "0px");
-    cy.findByLabelText(`Buyer-add-property`).should("not.exist");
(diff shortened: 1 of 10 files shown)


jyangML (approved):

wooldridge:
@timur-isangulov, thanks for catching that. I've added the tooltip and pushed the commit.

timur-isangulov (approved):
Looks good.

jyangML (dismissed):

wooldridge:
E2E failures. I will investigate.

jyangML (approved):

timur-isangulov (approved):

timur-isangulov (approved):

jyangML (approved):

PR #5575, opened 2021-04-26
DHFPROD-7196: Change tooltip for foreign key properties in modeler an…

timur-isangulov:
…d mapper

### Description
Includes: DHFPROD-7195: Add entity name to the foreign key property type
Opened a separate jira task for e2e tests DHFPROD-7275.

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/src/assets/mock-data/curation/entity-definitions-mock.js
+++ b/marklogic-data-hub-central/ui/src/assets/mock-data/curation/entity-definitions-mock.js
@@ -418,7 +418,7 @@ export const personRelatedEntityDef = [{
           },
           "orderedBy": {
             "datatype": "integer",
-            "relatedEntityType": "http://example.org/",
+            "relatedEntityType": "http://example.org/Person-0.0.1/Person",
             "joinPropertyName": "items"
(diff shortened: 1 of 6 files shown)


wooldridge (changes requested):
Since we're inserting the same dynamic content in two places (across two components), the content should probably live in a single place, and that place should be with the other tooltip content. I created a PR to your branch that defines a tooltip function for this in the tooltip config file: https://github.com/timur-isangulov/marklogic-data-hub/pull/3

timur-isangulov:
> Since we're inserting the same dynamic content in two places (across two components), the content should probably live in a single place, and that place should be with the other tooltip content. I created a PR to your branch that defines a tooltip function for this in the tooltip config file: [timur-isangulov#3](https://github.com/timur-isangulov/marklogic-data-hub/pull/3)

Thanks for the comment, I added the change.

wooldridge (approved):

Sanjeevani19 (approved):

PR #5594, opened 2021-05-03
DHFPROD-4489: Changing mapping step to acceptsBatch=true

rjrudin:
### Description

This is for user mapping params, and it avoids having to add the new "beforeMain" function. The mapping step main function then iterates over each item. If one throws an error, then it call addStepError and keeps going, unless stopOnError is true. This preserves the existing functionality for a mapping step. 

This is essentially a refactoring, as it doesn't change any tests except a new one, which is now simpler in terms of the error it's checking for. 

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


srinathgit (approved):
Looks good. Adding "acceptsBatch":true in step definition ensures that we don't have to do anything for upgrade scenarios right ?

ryanjdew (approved):

PR #5598, opened 2021-05-04
DHFPROD-6994: Backend work for separate merge warnings

ryanjdew:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/mastering/validateMergingStep.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/mastering/validateMergingStep.sjs
@@ -17,7 +17,7 @@
 
 const validateMergeLib = require("/data-hub/5/data-services/mastering/validateMergingStepLib.sjs");
 
-var stepName;
+var stepName, view;
 
 xdmp.securityAssert("http://marklogic.com/data-hub/privileges/read-match-merge", "execute");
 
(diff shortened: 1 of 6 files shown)


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/mastering/validateMergingStep.sjs
```diff
@@ -17,43 +17,44 @@
...
-    .concat(targetCollections.onArchive)
-    .concat(targetCollections.onNoMatch)
-    .concat(targetCollections.onMerge)
-    .concat(targetCollections.onNotification)
-    .concat(targetCollections.onAuditing)
-if (allCollections.length) {
-    let targetTypeWarning = common.targetEntityCollectionWarning(step.targetEntityType, allCollections);
-    if (targetTypeWarning) {
-        warnings.push(targetTypeWarning);
+if (view === "settings") {
```
rjrudin, line 28: If this code migrates out into a library, I think it'd be worth breaking this down into smaller functions for readability - i.e. validateSettings and validateRules. 

Also, I've been wondering if we should refactor at some point to having classes for steps - e.g. we'd have a MergingStep class, and we'd be adding new "validateSettings" and "validateRules" functions to it. The DS endpoint would then load the step from the database, pass the JSON into a new instance of MergingStep, and call either of those functions. Just something to think about - possibly a direction to head in during 5.6.

rjrudin (approved):

srinathgit (dismissed):

AkshayBajajML (approved):
Can now see the expected warning messages in UI
Screen Shot 2021-05-05 at 7 40 41 PM

Sanjeevani19 (approved):

PR #5604, opened 2021-05-05
DEVO-240: Investigate connection issues in nightly regressions

yafeldman:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [ ] Added Tests
  

- ##### Reviewer:

- [ ] Reviewed Tests


Jenkinsfile
```diff
@@ -936,22 +943,45 @@ pipeline{
...
+        environment{
+                    JAVA_HOME="$JAVA_HOME_DIR"
+                    PATH="$JAVA_HOME:$PATH"
+                }
+        steps {timeout(time: 3, unit: 'HOURS') {
+                  catchError(buildResult: 'SUCCESS', catchInterruptions: true, stageResult: 'FAILURE') {
+                            cleanWs()
+                            copyRPM 'Release', '9.0-11'
+                            setUpML '$WORKSPACE/xdmp/src/Mark*.rpm'
+                            sh 'cd $WORKSPACE/data-hub;./gradlew -g ./cache-build clean ml-data-hub:testFullCycle -i --stacktrace'
```
SameeraPriyathamTadikonda, line 969: Why do we need another stage for running full cycle tests? Can't we. just run after running all the tests in the same stage?

SameeraPriyathamTadikonda (changes requested):
Combine two stages

Jenkinsfile
yafeldman: 1) new stage would make executable time faster and better monitoring.
2)FullCycle is separate test and technically requires  different environment

SameeraPriyathamTadikonda (approved):

ngodugu-marklogic (approved):

PR #5607, opened 2021-05-05
DHFPROD-7038: Added examples for using DHF test support

rjrudin:
### Description

Also moved a bit of furniture around with the flow code:

- stepExecutionContext now handles finalizing collections and permissions, as opposed to invoke-queue-write.sjs handling this. This allows a test to run a step and see the "true" content objects that will be written to ML.
- Added flowRunner.runStepAgainstSourceDatabase to cut down on duplication between flowRunner and flow.sjs, which allowed for deleting invoke-step.sjs
- Added runFlowStepOnContent to flow-api, which seems like a very useful function for tests (in fact, we could probably stand to use it ourselves more often)

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/hub-utils/invoke-queue-write.sjs
```diff
@@ -49,40 +46,18 @@ function deleteContent(content, temporalCollection) {
...
-  if (context.useContextCollectionsOnly) {
-    targetCollections = fn.distinctValues(Sequence.from(contextCollections)).toArray();
-  } else {
-    targetCollections = fn.distinctValues(Sequence.from(configCollections.concat(contextCollections))).toArray();
-  }
-
-  return targetCollections.length > 0 ? targetCollections : xdmp.defaultCollections().toArray();
-}
-
 // Create a map of all temporal collections for quick checks on whether or not a collection is a temporal one
```
rjrudin, line 49: Add a comment here about how we need to be careful about moving this, because this needs to run against the target database so it sees the correct temporal collections

akshaysonvane (approved):

ryanjdew (approved):

PR #5608, opened 2021-05-06
DHFPROD-7197: Avoid throwing an error if job doc not found

akshaysonvane:
### Description
- Avoids a race condition if the job doc is retrieved before it's written.
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [N/A] Added Tests
  

- ##### Reviewer:

- [N/A] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/src/pages/Run.tsx
+++ b/marklogic-data-hub-central/ui/src/pages/Run.tsx
@@ -373,6 +373,10 @@ const Run = (props) => {
     let checkStatus = (resolve, reject) => {
       let promise = fn();
       promise.then(function(response) {
+        if (!response.data) {
+          throw new Error("Empty response body received");
+        }
+
         let status = response.data.jobStatus;
         if (status === Statuses.FINISHED || status === Statuses.CANCELED ||
                     status === Statuses.FAILED || status === Statuses.FINISHED_WITH_ERRORS) {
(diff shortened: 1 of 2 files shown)


marklogic-data-hub-central/ui/src/pages/Run.tsx
```diff
@@ -373,6 +373,10 @@ const Run = (props) => {
     let checkStatus = (resolve, reject) => {
       let promise = fn();
       promise.then(function(response) {
+        if (!response.data) {
+          throw new Error("Empty response body received");
```
akshaysonvane, line 377: We throw an error here if the data is empty so that it can be caught by the FE's retry mechanism.

ryanjdew (approved):

marklogic-data-hub-central/ui/src/pages/Run.tsx
rjrudin: Does it show up in the browser's console log though? That would seem odd, since this isn't an error condition; the job document just hasn't been written yet.

akshaysonvane: No it does not, its caught by the retry mechanism which only throws its console error after repeated polling attempts.

rjrudin (approved):

PR #5612, opened 2021-05-07
DHFPROD-7025: Test Mapping expressions for related entities

briantang:
### Description

- Test button functionality for related entity tables, this includes the URI expressions in the target and related entity tables

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [ ] Reviewed Tests


srinathgit (changes requested):
1.When I hit "Test", the URI error message is not displayed for primary mapping (the mapping for the 'targetEntity')
2. When the URI field for a related entity mapping returns an error, the text is not shown in red like the errors in xpath for entity property. You can probably consult with Shannon or UX to determine where/how to show errors in URI field

briantang:
> 1.When I hit "Test", the URI error message is not displayed for primary mapping (the mapping for the 'targetEntity')
> 2. When the URI field for a related entity mapping returns an error, the text is not shown in red like the errors in xpath for entity property. You can probably consult with Shannon or UX to determine where/how to show errors in URI field

Thanks @srinathgit. I recall Shannon mentioning that the error message for URI' should be shown in red as with the entity properties. Good catch, just pushed the fix.

wooldridge (changes requested):
A couple things...

1. If I put in an XPath expression for the URI field that doesn't evaluate to a good URI (e.g. "URL") and I test it, nothing it shown in the URI test field. If I put the same value in a property field, I get null for a test value. 

https://drive.google.com/file/d/1rUy3AsA-nzvB48TvRaFT5rQNELmtiskn/view

When I try to run the mapping, it errors upon execution with an error that includes the text: "The URI xpath expression for the mapping evaluates to null document"

Should testing the URI field for a bad expression like "URL" at least show null in the test field to be consistent with what we show in a regular property field? It would also make the test value be consistent with the error message that eventually appears when the mapping is run. @srinathgit 

2. When I put a value like CustomerID in the URI field and test it, I get 301.JSON as a test value. But the URI that is eventually generated when the mapping is run is 301.json. The test value should have a lowercase extension consistent with the URI generated.

Screen Shot 2021-05-10 at 1 09 23 AM

wooldridge (dismissed):

srinathgit (dismissed):

srinathgit (approved):

wooldridge (approved):

PR #5636, opened 2021-05-10
DHFPROD-7306:Provide error message when mapping uri expression evaluates to null

srinathgit:
### Description
Provide error message when mapping uri expression evaluates to null
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mapping/entity-services/lib.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mapping/entity-services/lib.sjs
@@ -477,30 +477,34 @@ function validateAndTestUriExpressions(mapping, validatedMappingsArray, sourceIn
     try {
       if(!response){
         const uriString = String(fn.head(testXmlMapping(xmlMapping, sourceInstance, parameterMap)).xpath('*:uris/*:uri[1]/text()'));
-        response = flowUtils.properExtensionURI(uriString, mapping.targetFormat);
+        response = flowUtils.properExtensionURI(uriString, fn.lowerCase(mapping.targetFormat));
       }
     }
     catch(e){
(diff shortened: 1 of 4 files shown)


srinathgit:
> Screen Shot 2021-05-10 at 11 09 55 PM
> 
> Error message for null URI expressions in the UI looks good! One thing @srinathgit, I'm seeing that when nothing is sent as the payload for the uriExpression (e.g. ""), the output test value being returned is the source doc URI. Is this expected? If not, I believe an error message for this scenario may be needed as well.
> 
> Screen Shot 2021-05-10 at 11 12 44 PM

The default URI for the targetEntity mapping when uri expression is not specified is the doc URI. So, it is correct here. What is returned when uriExpression is  "" for a related entity mapping ?

briantang (dismissed):

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mapping/entity-services/lib.sjs
```diff
@@ -477,30 +477,34 @@ function validateAndTestUriExpressions(mapping, validatedMappingsArray, sourceIn
...
+        response = flowUtils.properExtensionURI(uriString, fn.lowerCase(mapping.targetFormat));
       }
     }
     catch(e){
       response = hubUtils.getErrorMessage(e);
       errorEvaluatingExpression = true;
     }
+    if(!errorEvaluatingExpression && !response){
+      errorEvaluatingExpression = true;
+      response = "The URI xpath expression for the mapping evaluates to null";
```
rjrudin, line 489: Should be "XPath" instead of "xpath"

rjrudin (approved):

wooldridge (approved):

PR #5637, opened 2021-05-11
DHFPROD-7169: Add primary keys to matching preview response

ryanjdew:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/mastering/preview-matching-activity-lib.xqy
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/mastering/preview-matching-activity-lib.xqy
@@ -2,8 +2,10 @@ xquery version "1.0-ml";
 
 module namespace pma = "http://marklogic.com/smart-mastering/preview-matching-activity-lib";
 
+import module namespace hent = "http://marklogic.com/data-hub/hub-entities"
+  at "/data-hub/5/impl/hub-entities.xqy";
 import module namespace matcher = "http://marklogic.com/smart-mastering/matcher"
-  at "/com.marklogic.smart-mastering/matcher.xqy";
+at "/com.marklogic.smart-mastering/matcher.xqy";
 
 declare variable $PMA-MAX-RESULTS := 100;
 declare variable $DEFAULT-URI-SAMPLE-SIZE := 20;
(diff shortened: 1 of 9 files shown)


ryanjdew:
@rjrudin They are for a more friendly display of identifiers than URIs. If a field is configured as a primary key and it isn't intended to be unique that seems to indicate a misconfiguration on the part of the user.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/mastering/preview-matching-activity-lib.xqy
```diff
@@ -158,11 +156,47 @@ declare function pma:preview-matching-activity(
...
   (
     map:put($obj, "sampleSize", $sample-size),
+    map:put($obj, "primaryKeys", $primary-keys),
     map:put($obj, "uris", json:to-array($uris)),
     map:put($obj, "actionPreview", json:to-array($all-results))
   )
   return xdmp:to-json($obj)/node()
 };
+
+declare function pma:primary-keys-for-uris(
```
rjrudin, line 172: This seems generic to entity instances - i.e. "I have N entity instance URIs, and I know the entity type; I want identifiers for all of them. 

That seems like a candidate for adding to entity-lib - e.g. "getEntityIdentifiers(uris, entityType)" - but that would have the downside here of XQuery having to call JS. 

Thoughts on that sort of function living in hub-entities.xqy, which at least is in XQuery?

ryanjdew: That makes sense. I'll move it to hub-entities.xqy and provide also add a JS function in entity-lib that calls the XQuery.

Should I make the signature `find-entity-identifiers($uris, $entity-type)/findEntityIdentifiers(uris, entityType) ` since there is a SPARQL query?

rjrudin: Yes, "find" is a good choice.

ryanjdew: @rjrudin Okay. I've updated the PR with the function moved and added additional tests specifically for the common JS function.

rjrudin (approved):

AkshayBajajML (approved):

PR #5646, opened 2021-05-11
DHFPROD-7263:Prevent deletion of entities that have join properties associated with them

srinathgit:
…ssociated with them

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/entity-lib.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/entity-lib.sjs
@@ -263,6 +263,15 @@ function findModelReferencesInOtherModels(entityModelUri, entityTypeId) {
   return [...affectedModels];
 }
 
+function findForeignKeyReferencesInOtherModels(entityModel){
+  const entityTypeId = getEntityTypeId(entityModel, entityModel.info.title);
+  const entityModelsWithForeignKeyReferences = cts.search(cts.andQuery([cts.collectionQuery(consts.ENTITY_MODEL_COLLECTION),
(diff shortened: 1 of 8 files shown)


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/entity-lib.sjs
```diff
@@ -263,6 +263,15 @@ function findModelReferencesInOtherModels(entityModelUri, entityTypeId) {
   return [...affectedModels];
 }
 
+function findForeignKeyReferencesInOtherModels(entityModel){
+  const entityTypeId = getEntityTypeId(entityModel, entityModel.info.title);
+  const entityModelsWithForeignKeyReferences = cts.search(cts.andQuery([cts.collectionQuery(consts.ENTITY_MODEL_COLLECTION),
```
rjrudin, line 268: @ryanjdew Should a sparql query here be possible, and presumably more efficient too since we wouldn't be loading docs? 

@srinathgit  Check what's in /v1/graphs/things when you run your test - I think ML may project triples from the entity model automatically that allow for using sparql here.

srinathgit: @rjrudin 
I looked into the triples in "v1/graphs?graph=http://marklogic.com/entity-services/models". There aren't any triples for properties that we define like "relatedEntityType", "facetable", "sortable".

rjrudin (approved):

briantang (approved):
Functionality and tests on the UI side are looking good. 
https://drive.google.com/file/d/1fW6c-dfbYLF_rU9-xBaPXum4VuQZ53se/view

PR #5652, opened 2021-05-12
DHFPROD-7333: No longer scaffolding default users

rjrudin:
### Description

For test purposes, now creating test-flow-developer and test-flow-operator. 

Note that this is only affects scaffolding; existing projects will still have these files in their filesystems, and it's up to the user whether to keep them or not. 

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


srinathgit (commented):
mlUser in gradle.properties file in both marklogic-data-hub and ml-data-hub-plugin project is still "flow-developer". Since we are not creating those users, wouldn't that result in some test failures ?

rjrudin:
@srinathgit  I'm looking into that - I think many of those properties aren't used and can be removed; I think we only need mlUsername/mlPassword in marklogic-data-hub/build.gradle set to admin to support deploying the test app and running mlWatch.

rjrudin:
Fixed one test failure - GetProjectFilesAsZipTest was still expecting the user files to exist.

ryanjdew (approved):

srinathgit (approved):

PR #5656, opened 2021-05-13
DHFPROD-7256: Fixing custom rewriter for jobs app server

rjrudin:
### Description

The rewriter is now created based on the OOTB rewriter plus the custom routes needing for the trace-ui app. Was able to remove code that chose whether or not to use a version 9 or version 10 copy of the rewriter. 

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


rjrudin:
Recreated this as a bunch of tests were failing due to Unauthorized errors, which could be because the parallel test support was failing somehow. Not sure what's up yet.

ryanjdew (approved):

akshaysonvane (approved):

PR #5657, opened 2021-05-13
DHFPROD-7179:Ensure mapping step respects 'targetCollectionsAdditivity' option

srinathgit:
### Description
Ensure mapping step respects 'targetCollectionsAdditivity' option
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mapping/entity-services/main.sjs
```diff
@@ -102,6 +102,11 @@ function main(contentSequence, options, stepExecutionContext) {
           entityName = lib.getEntityName(fn.string(currentRelatedMapping.targetEntityType));
           entityContext = createContextForRelatedEntityInstance(currentRelatedMapping);
         }
+
+        if(content.context && content.context.originalCollections){
```
rjrudin, line 106: I think this should be done in the createContextForRelatedEntityInstance function, as doesn't it only apply for a related entity mapping?

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mapping/entity-services/main.sjs
```diff
@@ -111,7 +116,8 @@ function main(contentSequence, options, stepExecutionContext) {
           }
           entityContent["value"] = entityInstance.value;
           entityContent["uri"] = buildUri(entityInstance, entityName, outputFormat);
-          entityContent = validateEntityInstanceAndBuildEnvelope(doc, entityContent, entityContext, entityModel, outputFormat, options);
+          //Create a new context object for every instance generated by a mapping
```
rjrudin, line 119: This comment is a little confusing because it only applies to a little bit of the code in the next line. I think it'd be better to do something like this:

```
const entityInstanceContext = Object.assign({}, entityContext);
entityContent = validateEntity...
```

Then I don't think you need a comment because it's clear that each entity instance needs its own context. 

That makes me wonder then if "entityContext" needs a more self-documenting name - perhaps entityTypeContext, or entityContextTemplate, or initialEntityContext?

rjrudin (changes requested):

rjrudin (approved):

ryanjdew (approved):

PR #5659, opened 2021-05-13
DHFPROD-6731: and DHFPROD-6732: Bulk Data Service and Gradle Task to delete jobs data

Elgaeb:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/jobs.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/jobs.sjs
@@ -22,9 +22,9 @@ const StepDefinition = require("/data-hub/5/impl/stepDefinition.sjs");
 const XS_NAMESPACE = "http://www.w3.org/2001/XMLSchema";
 
 /**
- * 
- * @param flowName 
- * @param jobId 
+ *
+ * @param flowName
+ * @param jobId
  * @returns {object} the Job object (object, not a document node)
  */
 function createJob(flowName, jobId = null ) {
@@ -39,8 +39,8 @@ function createJob(flowName, jobId = null ) {
 
 /**
  * Saves a newly constructed Job; use createJob to both construct and save a Job.
- * 
- * @param job 
+ *
+ * @param job
  */
 function saveNewJob(job) {
   const jobUri = "/jobs/" + job.job.jobId + ".json";
@@ -324,9 +324,9 @@ module.exports = {
 module.exports.updateJob = module.amp(
   /**
    * Only updates the document, does not make any modifications to it, so nothing is returned.
-   * 
+   *
    * @param jobDoc
-   * @returns 
+   * @returns
    */
   function updateJob(jobDoc) {
     const jobId = jobDoc.job.jobId;
@@ -339,3 +339,16 @@ module.exports.updateJob = module.amp(
   }
 );
 
+module.exports.deleteJobs = module.amp(
+  /**
+   * Delete all job and batch document URIs passed in.
+   *
+   * @param jobDocumentURIs URIs of job documents to delete.
+   *
+   */
+  function deleteJobs(jobDocumentURIs) {
+    xdmp.securityAssert("http://marklogic.com/data-hub/privileges/delete-jobs", "execute");
+    for (let uri of jobDocumentURIs) {
+      xdmp.documentDelete(uri);
+    }
+  });
--- a/ml-data-hub-plugin/src/main/groovy/com/marklogic/gradle/DataHubPlugin.groovy
+++ b/ml-data-hub-plugin/src/main/groovy/com/marklogic/gradle/DataHubPlugin.groovy
@@ -144,6 +144,10 @@ class DataHubPlugin implements Plugin<Project> {
                 "This consists of downloading user configuration files (entity models, entity model-based files, flows, and steps) as a zip, " +
                 "deleting all such user configuration files in the project directory, "+
                 "and finally extracting the contents of the downloaded zip into the project directory.")
+        project.task("hubDeleteJobs", type: DeleteJobsTask, group: developGroup,
+            description: "Clears job and batch in the job database older than a given duration " +
+                "specify a duration to retain via -PretainDuration=(duration) " +
+                "(e.g. -PretainDuration=P30D)")
         project.task("hubClearUserData", type: ClearUserDataTask, group: developGroup,
             description: "Clears user data in the staging, final, and job databases, only leaving behind hub and user " +
                 "artifacts. Requires sufficient privilege to be able to clear each of the databases. " +
@@ -232,11 +236,11 @@ class DataHubPlugin implements Plugin<Project> {
             description: "Create a DHF 4 input flow in the project directory")
         project.task("hubRunLegacyFlow", group: legacyFlowGroup, type: RunLegacyFlowTask,
             description: "Run a DHF 4 flow")
-        project.task("hubDeleteJobs", group: legacyFlowGroup, type: DeleteJobsTask,
+        project.task("hubDeleteLegacyJobs", group: legacyFlowGroup, type: DeleteLegacyJobsTask,
             description: "Delete DHF 4 jobs data")
         project.task("hubExportLegacyJobs", group: legacyFlowGroup, type: ExportLegacyJobsTask,
             description: "Export DHF jobs data")
-        project.task("hubImportJobs", group: legacyFlowGroup, type: ImportJobsTask,
+        project.task("hubImportLegacyJobs", group: legacyFlowGroup, type: ImportJobsTask,
             description: "Import DHF 4 jobs data")
 
         ((UpdateIndexesTask)project.tasks.getByName("mlUpdateIndexes")).command = new HubUpdateIndexesCommand(dataHub)
--- a/ml-data-hub-plugin/src/main/groovy/com/marklogic/gradle/task/DeleteJobsTask.groovy
+++ b/ml-data-hub-plugin/src/main/groovy/com/marklogic/gradle/task/DeleteJobsTask.groovy
@@ -17,40 +17,16 @@
 
 package com.marklogic.gradle.task
 
-import com.marklogic.gradle.exception.JobIdsRequiredException
-import com.marklogic.hub.legacy.job.JobDeleteResponse
-import org.gradle.api.tasks.Input
-import org.gradle.api.tasks.Optional
+import com.marklogic.hub.flow.JobManager
 import org.gradle.api.tasks.TaskAction
 
 class DeleteJobsTask extends HubTask {
 
-    @Input
-    @Optional
-    public String jobIds
-
-    String getJobIds() {
-        return jobIds
-    }
-
     @TaskAction
-    void deleteJobs() {
-        if (jobIds == null) {
-            jobIds = project.hasProperty("jobIds") ? project.property("jobIds") : null
-        }
-        if (jobIds == null) {
-            throw new JobIdsRequiredException()
-        }
-
-        println("Deleting jobs: " + jobIds)
-        def jobManager = getJobManager()
-        def dh = getDataHub()
-        if (!isHubInstalled()) {
-            println("Data Hub is not installed.")
-            return
-        }
-        def jobDeleteResponse = jobManager.deleteJobs(jobIds)
-        print jobDeleteResponse
+    void cleanJobsData() {
+        def propName = "retainDuration"
+        def retainDuration = project.hasProperty(propName) ? project.property(propName) : null
+        JobManager jobManager = new JobManager(getHubConfig().newHubClient());
+        jobManager.deleteJobs(retainDuration);
     }
-
 }
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/flow/JobManager.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/flow/JobManager.java
@@ -1,0 +1,61 @@
+package com.marklogic.hub.flow;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import com.marklogic.client.dataservices.ExecCaller;
+import com.marklogic.client.dataservices.IOEndpoint;
+import com.marklogic.client.io.JacksonHandle;
+import com.marklogic.hub.HubClient;
+
+import java.io.IOException;
+import java.io.InputStreamReader;
+
+public class JobManager {
+
+    private ExecCaller deleteJobsCaller;
+    private ObjectMapper objectMapper = new ObjectMapper();
+
+    public JobManager(HubClient hubClient) {
+        ObjectNode apiNode;
+        final String apiPath = "/ml-modules/root/data-hub/5/data-services/bulk/deleteJobs.api";
+        try (InputStreamReader apiReader = new InputStreamReader(this.getClass().getResourceAsStream(apiPath))) {
+            apiNode = objectMapper.readValue(apiReader, ObjectNode.class);
+        } catch (IOException e) {
+            throw new RuntimeException(
+                    "Unable to find API module " + apiPath + " on classpath; cause: " + e.getMessage(), e);
+        }
+        deleteJobsCaller = ExecCaller.on(hubClient.getJobsClient(), new JacksonHandle(apiNode));
+    }
+
+    public void deleteJobs(String retainDuration) {
+        // See https://stackoverflow.com/a/52645128/535924 for more information about
+        // the duration regex
+        if (retainDuration == null || !retainDuration.matches(
+                "^(-?)P(?=.)((\\d*)Y)?((\\d+)M)?((\\d*)D)?(T(?=\\d)(\\d+H)?(([\\d]+)M)?(([\\d]+(?:\\.\\d+)?)S)?)?$")) {
+            throw new IllegalArgumentException(
+                    "retainDuration must be a duration in the format of PnYnM or PnDTnHnMnS");
+        }
+        ObjectNode endpointConstants = objectMapper.createObjectNode().put("batchSize", 250).put("retainDuration",
+                retainDuration);
+        CapturingErrorListener errorListener = new CapturingErrorListener();
+
+        ExecCaller.BulkExecCaller bulkCaller = this.deleteJobsCaller
+                .bulkCaller(this.deleteJobsCaller.newCallContext().withEndpointConstantsAs(endpointConstants));
+        bulkCaller.setErrorListener(errorListener);
+        bulkCaller.awaitCompletion();
+
+        Throwable throwable = errorListener.throwable;
+        if (throwable != null) {
+            throw new RuntimeException("Unable to delete jobs, cause: " + throwable.getMessage(), throwable);
+        }
+    }
+
+    static class CapturingErrorListener implements ExecCaller.BulkExecCaller.ErrorListener {
+        Throwable throwable;
+        public IOEndpoint.BulkIOEndpointCaller.ErrorDisposition processError(int retryCount, Throwable throwable,
+                IOEndpoint.CallContext callContext) {
+            this.throwable = throwable;
+            return IOEndpoint.BulkIOEndpointCaller.ErrorDisposition.STOP_ALL_CALLS;
+        }
+    }
+}
--- a/marklogic-data-hub/src/main/resources/hub-internal-config/security/amps/deleteJobs.json
+++ b/marklogic-data-hub/src/main/resources/hub-internal-config/security/amps/deleteJobs.json
@@ -1,0 +1,6 @@
+{
+  "local-name" : "deleteJobs",
+  "document-uri" : "/data-hub/5/impl/jobs.sjs",
+  "modules-database" : "%%mlModulesDbName%%",
+  "role" : [ "data-hub-job-internal" ]
+}
--- a/marklogic-data-hub/src/main/resources/hub-internal-config/security/privileges/data-hub-delete-jobs.json
+++ b/marklogic-data-hub/src/main/resources/hub-internal-config/security/privileges/data-hub-delete-jobs.json
@@ -1,0 +1,6 @@
+{
+  "privilege-name": "data-hub-delete-jobs",
+  "action": "http://marklogic.com/data-hub/privileges/delete-jobs",
+  "kind": "execute"
+}
+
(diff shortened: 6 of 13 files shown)


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/bulk-data-services/job/deleteJobs.api
```diff
@@ -0,0 +1,24 @@
+{
+    "endpoint": "/data-hub/5/bulk-data-services/job/deleteJobs.sjs",
```
rjrudin, line 2: We have a data-hub/5/data-services/bulk directory already; let's toss this there as well. We may end up reorganizing those later, but since that directory exists, it's consistent for now to keep using it.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/jobs.sjs
```diff
@@ -296,3 +296,62 @@ module.exports.updateJob = module.amp(
...
+          cts.collectionQuery([ 'Job' ]),
+          cts.rangeQuery(cts.jsonPropertyReference('jobId'), '=', jobIds)
+        ]);
+        const jobUrisToDelete = cts.uris(null, null, jobUriQuery).toArray();
+        return {
+          urisToDelete: jobUrisToDelete,
+          remaining: remainingJobCount > jobUrisToDelete.length
+        }
+      }
+    }, config.JOBDATABASE));
```
rjrudin, line 348: We don't need to support this running against some other app server - we can assume the client (in this case a Gradle task that we control) is hitting the jobs app server.

ml-data-hub-plugin/src/main/groovy/com/marklogic/gradle/task/CleanJobsDataTask.groovy
```diff
@@ -0,0 +1,84 @@
...
+import com.fasterxml.jackson.databind.node.ObjectNode
+import com.marklogic.client.dataservices.ExecCaller
+import com.marklogic.client.dataservices.IOEndpoint
+import com.marklogic.client.io.JacksonHandle
+import com.marklogic.gradle.exception.RetainDurationRequiredException
+import com.marklogic.hub.MappingManager
+import com.marklogic.hub.scaffold.Scaffolding
+import org.gradle.api.tasks.TaskAction
+
+class CapturingErrorListener implements ExecCaller.BulkExecCaller.ErrorListener {
```
rjrudin, line 30: All of this needs to shift into Java code; we want as little as possible in the Gradle task. All that the Gradle task should do is collect the retainDuration value and call a Java method. 

For the location in src/main/java - while we have a com.marklogic.hub.job package, jobs are really a part of the flow subsystem; i.e. jobs are not separate from flows (the only way to end up with a job is to run a flow; so really, the JobStatus class should move to hub.flow, but that's a separate matter). So let's do com.marklogic.hub.flow.JobManager and add a deleteJobs(retainDuration) method to it. No interface needed either, this isn't yet a public API.

ml-data-hub-plugin/src/test/groovy/com/marklogic/gradle/task/CleanJobsDataTaskTest.groovy
```diff
@@ -0,0 +1,53 @@
+package com.marklogic.gradle.task
```
rjrudin, line 1: This should all be done in JUnit in the ./marklogic-data-hub project as well, where there's better support for writing and running tests too. There's an internal Wiki page with a guide on writing DHF tests.

ml-data-hub-plugin/src/main/groovy/com/marklogic/gradle/DataHubPlugin.groovy
```diff
@@ -145,6 +145,10 @@ class DataHubPlugin implements Plugin<Project> {
                 "This consists of downloading user configuration files (entity models, entity model-based files, flows, and steps) as a zip, " +
                 "deleting all such user configuration files in the project directory, "+
                 "and finally extracting the contents of the downloaded zip into the project directory.")
+        project.task("hubCleanJobsData", type: CleanJobsDataTask, group: developGroup,
```
rjrudin, line 148: The privilege is named "delete-jobs", and the role is "job-deleter", so the expected name for this task would naturally be "hubDeleteJobs". But - we already have such a task, which is specific to DHF 4. 

Now, we have "hubRunLegacyFlow" and also "hubExportLegacyJobs". I'd argue that a DHF 4 user making use of these tasks is already confused that 2 of the 4 "legacy" DHF 4 tasks have the word "Legacy" in them and the other 2 do not. And that's confusing for poor DHF engineers like me, that see "hubDeleteJobs" but have to realize that's actually DHF 4-specific. 

I think we can kill 2 birds with 1 stone here - let's rename hubDeleteJobs/hubImportJobs to hubDeleteLegacyJobs and hubImportLegacyJobs. We'll capture that in our release notes and update the docs - and the docs are already incorrect - https://docs.marklogic.com/datahub/5.4/tools/gradle/gradle-tasks.html - they show "hubExportJobs" instead of "hubExportLegacyJobs", and hubImportJobs isn't even listed. 

We can then call this new task what it should be and what users expect it to be - hubDeleteJobs. And the task is DeleteJobsTask.

ml-data-hub-plugin/src/main/groovy/com/marklogic/gradle/exception/RetainDurationRequiredException.groovy
```diff
@@ -0,0 +1,26 @@
...
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ */
+
+package com.marklogic.gradle.exception
+
+import org.gradle.api.GradleException
+
+class RetainDurationRequiredException extends GradleException {
```
rjrudin, line 22: No need for this class; the Java code can just throw an IllegalArgumentException

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/jobs.sjs
```diff
@@ -296,3 +296,62 @@ module.exports.updateJob = module.amp(
...
+
+    hubUtils.deleteDocuments(urisToDelete, config.JOBDATABASE);
```
rjrudin, line 350: Since we can assume that the user is hitting the jobs app server, no need for this either. 

Also, I am thinking that jobs.sjs should have a "deleteJobUris" function that only does this code, and the amp sits on that. The rest of this code can go into the bulk service module. That keeps this module (which I'm not a fan of) from growing more, and it also restricts the amp to the smallest scope possible.

rjrudin (changes requested):
This is a good start, just have comments about naming and moving some code around

ryanjdew:
@rjrudin I believe I've made the changes that you requested. If could review it when you have a moment, that would be great. Thanks!

rjrudin:
Will try this out locally tomorrow morning

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/hub-utils.sjs
```diff
@@ -28,6 +28,14 @@ function deleteDocument(docUri, database) {
   });
 }
 
+function deleteDocuments(docUris, database) {
```
rjrudin, line 31: This can be removed, it's unused now

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/hub-utils/invoke-multiple-delete.sjs
```diff
@@ -0,0 +1,26 @@
+/**
```
rjrudin, line 1: This can be removed, it's unused now

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/jobs.sjs
```diff
@@ -296,3 +296,16 @@ module.exports.updateJob = module.amp(
   }
 );
 
+module.exports.deleteJobDocs = module.amp(
```
rjrudin, line 299: Let's just call this "deleteJobs" - technically, it's deleting Job and Batch docs, so "JobDocs" is a little misleading because it makes the reader think it only deletes Job docs. Also, "deleteJobs" is reasonable because both Job and Batch docs are in a collection named "Jobs".

marklogic-data-hub/src/main/resources/hub-internal-config/security/amps/amps-dhf-delete-job-docs.json
```diff
@@ -0,0 +1,6 @@
+{
```
rjrudin, line 1: See the comment below about calling this deleteJobs - then just call this file deleteJobs.json

rjrudin (changes requested):

marklogic-data-hub/src/main/java/com/marklogic/hub/job/JobDataManager.java
```diff
@@ -0,0 +1,66 @@
...
+import java.io.IOException;
+import java.io.InputStreamReader;
+
+public class JobDataManager {
+    ObjectNode apiObj;
+    String apiPath = "/ml-modules/root/data-hub/5/data-services/bulk/deleteJobs.api";
+    ObjectMapper mapper = new ObjectMapper();
+    HubConfig hubConfig;
+
+    JobDataManager(HubConfig hubConfig) {
```
rjrudin, line 20: Use HubClient instead of HubConfig here; HubConfig is appropriate when we need access to a HubProject. HubClient is much lighter-weight and is good for when you only need to connect to ML.

marklogic-data-hub/src/main/java/com/marklogic/hub/job/JobDataManager.java
```diff
@@ -0,0 +1,66 @@
...
+        this.hubConfig = hubConfig;
+        try (InputStreamReader apiReader = new InputStreamReader(this.getClass().getResourceAsStream(apiPath))) {
+            apiObj = mapper.readValue(apiReader, ObjectNode.class);
+        } catch (IOException e) {
+            throw new RuntimeException(e);
```
rjrudin, line 25: Include a message in the first argument here to explain the context - e.g.

```
throw new RuntimeException("Unable to find API module " + apiPath + " on classpath; cause: " + e.getMessage(), e);
```

marklogic-data-hub/src/main/java/com/marklogic/hub/job/JobDataManager.java
```diff
@@ -0,0 +1,66 @@
...
+        ExecCaller endpoint = ExecCaller.on(jobsClient, new JacksonHandle(apiObj));
+        CapturingErrorListener errorListener = new CapturingErrorListener();
+
+        ExecCaller.BulkExecCaller bulkCaller = endpoint.bulkCaller(endpoint.newCallContext()
+                .withEndpointConstantsAs(endpointConstants));
+        bulkCaller.setErrorListener(errorListener);
+        bulkCaller.awaitCompletion();
+
+        if(errorListener.didThrow()) {
+            throw new RuntimeException(errorListener.getThrowable());
```
rjrudin, line 45: Same thing, need a message here to provide context - e.g. "Unable to delete jobs, cause: "

marklogic-data-hub/src/test/java/com/marklogic/hub/dataservices/job/DeleteJobsTest.java
```diff
@@ -0,0 +1,129 @@
...
+import java.io.InputStreamReader;
+import java.io.Reader;
+import java.text.MessageFormat;
+import java.util.Arrays;
+import java.util.stream.Collectors;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.fail;
+
+public class DeleteJobsTest extends AbstractHubCoreTest {
```
rjrudin, line 31: I don't see any need for this test - we don't expect anyone to use the DS directly, we expect clients to all use JobDataManager. So we only need to test that class, with that class handling all of the DS plumbing for us. Also avoids duplicating things like the error listener below.

marklogic-data-hub/src/test/java/com/marklogic/hub/dataservices/job/DeleteJobsTest.java
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    @BeforeEach
+     void beforeEach() throws IOException {
+        Reader apiReader = new InputStreamReader(this.getClass().getResourceAsStream(apiPath));
+        apiObj = mapper.readValue(apiReader, ObjectNode.class);
+        setupAndRunFlow();
+    }
+
+    protected int getCollectionCount(DatabaseClient client, String... collections) {
```
rjrudin, line 63: Should be able to use getDocCount or getJobsDocCount or getJobDocCount or getBatchDocCount

marklogic-data-hub/src/test/java/com/marklogic/hub/dataservices/job/DeleteJobsTest.java
```diff
@@ -0,0 +1,129 @@
...
+            Arrays.stream(collections).map(c -> "'" + c + "'").collect(Collectors.joining(","))
+        );
+        return Integer.parseInt(client.newServerEval().javascript(query).evalAs(String.class));
+    }
+
+    protected void setupAndRunFlow() {
+        installProjectInFolder("test-projects/simple-custom-step");
+        new ReferenceModelProject(getHubClient()).createCustomerInstance(new Customer(1, "Jane"), "staging");
+
+        RunFlowResponse response = runFlow(new FlowInputs("simpleCustomStepFlow"));
```
rjrudin, line 74: Can use runSuccessfulFlow, which will verify that the flow succeeded for you

marklogic-data-hub/src/test/java/com/marklogic/hub/dataservices/job/DeleteJobsTest.java
```diff
@@ -0,0 +1,129 @@
...
+        deleteJobs("P1D");
+        int count = getCollectionCount(getHubClient().getJobsClient(), "Jobs");
+        assertEquals(2, count, "Jobs database should have 2 job and batch documents");
+    }
+
+    @Test
+    void testAsUserWhoCannotDeleteJobs() throws Throwable {
+        runAsDataHubOperator();
+        try {
+            deleteJobs("PT0S");
```
rjrudin, line 117: I think you can use verifyUserIsForbiddenTo, which is in HubTestBase

marklogic-data-hub/src/test/java/com/marklogic/hub/job/JobDataManagerTest.java
```diff
@@ -0,0 +1,56 @@
...
+import static com.marklogic.client.io.DocumentMetadataHandle.Capability.EXECUTE;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+public class JobDataManagerTest extends AbstractHubCoreTest {
+
+    @BeforeEach
+    public void setup(){
+        runAsAdmin();
+        clearDatabase(getHubClient().getStagingClient());
```
rjrudin, line 19: No need to do these 3 lines of code, the parent class does this for you automatically

ml-data-hub-plugin/src/main/groovy/com/marklogic/gradle/task/DeleteJobsTask.groovy
```diff
@@ -17,40 +17,20 @@
...
-        if (jobIds == null) {
-            jobIds = project.hasProperty("jobIds") ? project.property("jobIds") : null
-        }
-        if (jobIds == null) {
-            throw new JobIdsRequiredException()
+    void cleanJobsData() {
+        def propName = "retainDuration"
+        def retainDuration = project.hasProperty(propName) ? project.property(propName) : null
+        if (retainDuration == null) {
+            throw new RetainDurationRequiredException()
```
rjrudin, line 31: Just throw a GradleException here with the desired message - there's no value in the indirection from a separate class that won't ever be used except by this task class.

rjrudin (changes requested):

marklogic-data-hub/src/main/resources/hub-internal-config/security/amps/amps-dhf-delete-jobs.json
```diff
@@ -0,0 +1,6 @@
+{
```
rjrudin, line 1: Can you also name the file "deleteJobs.json" for consistency with the other amps?

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/jobs.sjs
```diff
@@ -296,3 +296,16 @@ module.exports.updateJob = module.amp(
   }
 );
 
+module.exports.deleteJobs = module.amp(
```
rjrudin, line 299: Something to consider during beta testing is whether this should become a public API; it might be nice to expose this for Pari to make use of however she wants. I think we'd put it under data-hub/public/flow/jobApi.sjs, something like that.

marklogic-data-hub/src/test/java/com/marklogic/hub/dataservices/job/DeleteJobsTest.java
rjrudin: I think this test should still be merged with JobDataManagerTest - I would call the merged test "DeleteJobsTest". I don't think there's any value in having all this duplicated Bulk IO plumbing here, as JobDataManager takes care of all that.

rjrudin (changes requested):

rjrudin:
I'm working on changes based on the comments I made on this

rjrudin:
@Elgaeb  @ryanjdew  I submitted a PR at https://github.com/Elgaeb/marklogic-data-hub/pull/1 with the changes based on my comments. DeleteJobsTest is passing for me locally.

If those work for you, squash merge those with the other 2 commits in this PR, then rebase off develop, and I'll approve.

rjrudin:
I'd still like the tests to be merged as shown in https://github.com/Elgaeb/marklogic-data-hub/pull/1 , which removes the duplication of Bulk IO code as well.

ryanjdew (approved):

ryanjdew:
@rjrudin Your changes have been incorporated.

rjrudin (approved):

Sanjeevani19 (commented):
@ryanjdew @elgaeb the data-hub build is continuously failing on this PR, https://jenkins.marklogic.com/blue/organizations/jenkins/Datahub_CI/detail/PR-5659/45/pipeline, the logs show Some UI dependencies are missing, Could you run npm run lint . -- --fix and see if that fixes the build, If not the other approach would be to close this PR and open a new one

ngodugu-marklogic:
> @ryanjdew @Elgaeb the data-hub build is continuously failing on this PR, https://jenkins.marklogic.com/blue/organizations/jenkins/Datahub_CI/detail/PR-5659/45/pipeline, the logs show Some UI dependencies are missing, Could you run npm run lint . -- --fix and see if that fixes the build, If not the other approach would be to close this PR and open a new one

@ryanjdew @Elgaeb Also this PR seems to have 2 commits, can you please squash them

ml-data-hub-plugin/src/main/groovy/com/marklogic/gradle/task/DeleteJobsTask.groovy
```diff
@@ -17,40 +17,16 @@
...
-        if (!isHubInstalled()) {
-            println("Data Hub is not installed.")
-            return
-        }
-        def jobDeleteResponse = jobManager.deleteJobs(jobIds)
-        print jobDeleteResponse
+    void cleanJobsData() {
+        def propName = "retainDuration"
+        def retainDuration = project.hasProperty(propName) ? project.property(propName) : null
+        JobDataManager jobDataManager = new JobDataManager(getHubConfig().newHubClient());
```
rahulvudutala, line 29: This should be 
`JobManager jobManager = new JobManager(getHubConfig().newHubClient());
        jobManager.deleteJobs(retainDuration);`

rahulvudutala (changes requested):

ml-data-hub-plugin/src/main/groovy/com/marklogic/gradle/task/DeleteJobsTask.groovy
ryanjdew: @rahulvudutala Thanks for catching that.

ryanjdew (dismissed):

rahulvudutala (approved):

ml-data-hub-plugin/src/main/groovy/com/marklogic/gradle/task/DeleteJobsTask.groovy
```diff
@@ -17,40 +17,16 @@
...
-        println("Deleting jobs: " + jobIds)
-        def jobManager = getJobManager()
-        def dh = getDataHub()
-        if (!isHubInstalled()) {
-            println("Data Hub is not installed.")
-            return
-        }
-        def jobDeleteResponse = jobManager.deleteJobs(jobIds)
-        print jobDeleteResponse
+    void cleanJobsData() {
```
rjrudin, line 26: I'm approving, and I'll submit a separate PR with a test in place for this - I think we just need a simple smoke test to make sure it compiles, basically.

rjrudin (dismissed):

rjrudin (dismissed):

rahulvudutala (approved):

rahulvudutala (changes requested):
The existing JobDeleteTaskTest should be refactored as LegacyJobDeleteTaskTest and the test should be running hubDeleteLegacyJobs task in the test instead of hubDeleteJobs.

ngodugu-marklogic:
@ryanjdew @Elgaeb Any update on this PR. Can we please address the changes requested and merge the PR.

ryanjdew (approved):

ryanjdew:
@rahulvudutala The hubDeleteLegacyJobs test has been updated.

rjrudin (approved):

rahulvudutala (approved):

PR #5669, opened 2021-05-15
DHFPROD-7131: Handle entities with updated props when building join menu

wooldridge:
### Description

Updated properties in entities that have not been saved yet will now show up in the Join Property menu in Modeling.

To test:

1. Create a new property for an entity (e.g. Customer). Do not save the entity.
2. Create a new entity relationship that points from one entity (Order) to the entity with the new property (Customer).
3. Open the Join Property menu to confirm that the new property is included in the list of properties. (Previously, it was not.)

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/e2e/cypress/integration/modeling/writerScenario1.spec.tsx
+++ b/marklogic-data-hub-central/ui/e2e/cypress/integration/modeling/writerScenario1.spec.tsx
@@ -55,7 +55,7 @@ describe("Entity Modeling Senario 1: Writer Role", () => {
     propertyModal.openPropertyDropdown();
     propertyModal.getTypeFromDropdown("Related Entity").click();
     propertyModal.getCascadedTypeFromDropdown("Person").click();
-    propertyModal.openJoinPropertyDropdown();
+    propertyModal.toggleJoinPropertyDropdown();
     propertyModal.getJoinProperty("id").click();
     propertyModal.getYesRadio("multiple").click();
     propertyModal.getSubmitButton().click();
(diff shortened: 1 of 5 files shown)


srinathgit (approved):
Verified that the property shows up in 'Order'

briantang (approved):

xnikhil08 (approved):

srinathgit (changes requested):
PR has to be updated based on the discussion we just had

wooldridge:
In discussion with the team, we decided that when a new property hasn't been saved to the backend yet in Modeling, that property shouldn't be allowed as a Join Property value. The property will be displayed in the menu but be disabled (just like multiple and structured properties are disabled). 

The tooltip associated with the Join Property field currently reads:

"Structured type properties and arrays cannot be used as join properties."

I suggest updating the tooltip to the following:

"Structured type properties, arrays, and unsaved properties cannot be used as join properties."

@sbayatpur does that work for you?

wooldridge:
> PR has to be updated based on the discussion we just had

@srinathgit I updated the branch to show but also disable unsaved properties in the Join Property menu. Also updated tooltip to explain unsaved properties cannot be used as join properties.

srinathgit (dismissed):
Looks good !

briantang (dismissed):
Nice!

Sanjeevani19 (approved):

devinnicholson (approved):

PR #5672, opened 2021-05-16
DHFPROD-7264: Delete mapping expressions for a related entity

briantang:
### Description
- Existing related entities that have mappings in the mapping step can now be deleted from the artifact via either the X button on the related entity's table or X button on the related entity's label from the referred entity's filter.
- RTL tests added, will also add more coverage in e2e together once this is started https://project.marklogic.com/jira/browse/DHFPROD-7267

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/src/components/entities/mapping/entity-map-table/entity-map-table.module.scss
+++ b/marklogic-data-hub-central/ui/src/components/entities/mapping/entity-map-table/entity-map-table.module.scss
@@ -24,13 +24,28 @@
 
 .entitySettingsLink {
   display: inline-block;
-  padding-right: 20px;
+  padding-right: 17px;
   font-size: 20px !important;
   color: #44499C;
   cursor: pointer;
   &:hover{
       color: var(--hoverColor);
   }
+}
+
+.deleteEntityLink {
+  display: inline-block;
+  margin-left: -10px;
+  font-size: 20px !important;
+  color: #44499C;
+  cursor: pointer;
+  &:hover{
+      color: var(--hoverColor);
+  }
+}
+
+.deleteTableIcon {
+  padding-right: 10px;
 }
 
 .mapRelatedEntitiesText {
--- a/marklogic-data-hub-central/ui/src/components/entities/mapping/entity-map-table/entity-map-table.tsx
+++ b/marklogic-data-hub-central/ui/src/components/entities/mapping/entity-map-table/entity-map-table.tsx
@@ -1,7 +1,7 @@
 import React, {useState, useEffect, CSSProperties} from "react";
 import styles from "./entity-map-table.module.scss";
 import "./entity-map-table.scss";
-import {Icon, Table, Popover, Input, Select, Dropdown} from "antd";
+import {Icon, Table, Popover, Input, Select, Dropdown, Modal} from "antd";
 import {MLButton, MLTooltip} from "@marklogic/design-system";
 import {FontAwesomeIcon} from "@fortawesome/react-fontawesome";
 import DropDownWithSearch from "../../../common/dropdown-with-search/dropdownWithSearch";
@@ -42,7 +42,6 @@ interface Props {
   relatedEntityTypeProperties: any;
   relatedEntitiesSelected: any;
   setRelatedEntitiesSelected: any;
-  setRemovedEntities: any;
   isRelatedEntity: boolean;
   tableColor: any;
   firstRowTableKeyIndex: any;
@@ -52,6 +51,8 @@ interface Props {
   setAllRelatedEntitiesKeys: any;
   mapFunctions : any;
   savedMappingArt : any;
+  deleteRelatedEntity: any;
+  labelRemoved: any;
 }
 
 const EntityMapTable: React.FC<Props> = (props) => {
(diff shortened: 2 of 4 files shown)


wooldridge:
Performed a bunch of manual tests using reference project, see below:

TL;DR: Testing checks out, things look fine, but the unrelated persistence issue (which is known: https://project.marklogic.com/jira/browse/DHFPROD-7304) is arising when returning to the mapping UI.

1. Confirmed mapCustomersWithRelatedEntitiesJSON showed up as expected in UI, all entity tables open.
2. Confirmed mapCustomersWithRelatedEntitiesJSON is as expected in QConsole.
3. After clearing user data, ran load and mapping step for mapCustomersWithRelatedEntitiesJSON, verified entity instances:
Customer (1)
Order (2)
BabyRegistry (1)
Product (6)
4. Deleted Product under BabyRegistry in mapCustomersWithRelatedEntitiesJSON via table X. 
5. Confirmed Product removed from BabyRegistry filter box.
6. Confirmed mapCustomersWithRelatedEntitiesJSON is as expected in QConsole (related Product gone under BabyRegistry).
7. After clearing user data, ran load and mapping step for mapCustomersWithRelatedEntitiesJSON, verified entity instances:
Customer (1)
Order (2)
BabyRegistry (1)
Product (4) (IDs 50, 60 are GONE)
8. Deleted Order under Customer in mapCustomersWithRelatedEntitiesJSON via filter pill in Customer.
9. Confirmed I see warning about needing to delete Product under Order first.
10. Deleted Product under Order in mapCustomersWithRelatedEntitiesJSON via filter pill in Order.
11. Confirmed Product table is gone under Order in UI.
12. Deleted Order under Customer in mapCustomersWithRelatedEntitiesJSON via filter pill in Customer.
13. Confirmed Order table is gone under Customer in UI.
14. Confirmed mapCustomersWithRelatedEntitiesJSON is as expected in QConsole (Customer and BabyRegistry only).
15. After clearing user data, ran load and mapping step for mapCustomersWithRelatedEntitiesJSON, verified entity instances:
Customer (1)
BabyRegistry (1)
16. When I return to the mapping step view for mapCustomersWithRelatedEntitiesJSON, the Order and Product tables ARE BACK, while the Product under BabyRegistry is still GONE.
17. When I hard refresh the browser window, mapCustomersWithRelatedEntitiesJSON goes back to the expected configuration, with  Order and Product tables GONE.
Note: This is a persistence bug being handled here: https://project.marklogic.com/jira/browse/DHFPROD-7304
18. Deleted BabyRegistry under Customer, now no more related entities remain (and relatedEntityMappings property in artifact is now []), and ran the same steps, results are good:
Customer (1)
19. Ran similar tests with mapCustomersWithRelatedEntitiesXML. Results are fine (but still seeing unrelated persistence bug).

wooldridge (changes requested):
@srinathgit @briantang 

General modeling/mapping bug?

1. Start with a fresh reference project (I redeployed artifacts and wiped user data).
2. In Modeling, create a new relationship between Order and BabyRegistry. Create a new property in BabyRegistry "orderRel" that points to Order, with Join Property "orderId". (Don't worry about whether this makes sense wrt to the source data, we're not testing any mapping expressions here, just relationships.) Save the all changes in Modeling.
3. Go to mapCustomersWithRelatedEntitiesJSON. 
4. I don't see a relationship. Either the related Order entity in the BabyRegistry filter or the related BabyRegistry entity in the Order Filter.

DHFPROD-7264-Order
DHFPROD-7264-BabyRegistry

wooldridge:
@briantang I see that we do a mapping save (the alert appears) when a related entity is DELETED via the filter or the close icon. But we don't do a save when a related entity is ADDED via the filter:

https://drive.google.com/file/d/1LiTL9CPDpvlW7KI41POnQOFnaaFAxnvp/view

I'm assuming that's because this relationship only has an effect when a property expression in that related entity has been defined. And I see that's when things are saved. Was that the thinking?

briantang:
> @briantang I see that we do a mapping save (the alert appears) when a related entity is DELETED via the filter or the close icon. But we don't do a save when a related entity is ADDED via the filter:
> 
> https://drive.google.com/file/d/1LiTL9CPDpvlW7KI41POnQOFnaaFAxnvp/view
> 
> I'm assuming that's because this relationship only has an effect when a property expression in that related entity has been defined. And I see that's when things are saved. Was that the thinking?

@wooldridge This is correct. Saving is only necessary to update the artifact when an expression has been defined or something is deleted, so we don't see a save happening when a related entity is added.

wooldridge (changes requested):
@briantang for the mapCustomersWithRelatedEntitiesJSON mapping in the reference project:

1. I've added a second Product relationship so that BabyRegistry now has two entities related to it.
2. I try to delete BabyRegistry at the Customer level and get a popup saying BabyRegistry has related entities, but it only lists the first Product. 
3. I delete the first Product.
4. Now I try to delete BabyRegistry at the Customer level and I can, successfully. 
5. This leaves a dangling Product that used to be related to BabyRegistry but it now unrelated to anything in view. 

Seems like a bug?

briantang:
Thanks @wooldridge, pushed a fix for the above issue.

wooldridge:
Looks good:

https://drive.google.com/file/d/1aS0o3AjH2J84_qORpwTXkkmQuzKexOxG/view

wooldridge (approved):

wooldridge (approved):

Sanjeevani19 (approved):

PR #5693, opened 2021-05-21
DHFPROD-7370: Fix for Export legacy jobs on ML 9

rahulvudutala:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

### DIFF
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/legacy/job/impl/LegacyJobManagerImpl.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/legacy/job/impl/LegacyJobManagerImpl.java
@@ -32,7 +32,6 @@ import com.marklogic.client.io.Format;
 import com.marklogic.client.io.JacksonDatabindHandle;
 import com.marklogic.client.io.StringHandle;
 import com.marklogic.client.query.QueryManager;
-import com.marklogic.client.query.StringQueryDefinition;
 import com.marklogic.client.query.StructuredQueryBuilder;
 import com.marklogic.client.query.StructuredQueryDefinition;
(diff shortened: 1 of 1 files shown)


marklogic-data-hub/src/main/java/com/marklogic/hub/legacy/job/impl/LegacyJobManagerImpl.java
```diff
@@ -120,7 +120,7 @@ public LegacyJobManagerImpl(DatabaseClient jobClient) {
         QueryBatcher batcher = jobIds == null ?
             newEmptyQueryBatcher(dmm, jobClient) :
             dmm.newQueryBatcher(sqb.value(sqb.jsonProperty("jobId"), jobIds));
-        batcher.onUrisReady(new ExportListener().onDocumentReady(zipConsumer));
+        batcher.withBatchSize(100).withThreadCount(4).onUrisReady(new ExportListener().onDocumentReady(zipConsumer));
```
rjrudin, line 123: Can you add a comment to this to explain why these are being set? Better yet - this is duplicated code, let's make a private method that does this in one place, and add a comment to that method stating e.g. "If batchSize and threadCount are not set, Java Client 5.4 will try to call an ML server function that only exists in ML 10.0-5 and higher; because this feature must work on ML 9.x, batchSize and threadCount are given sensible default values to avoid the Java Client trying to call a server function that may not exist."

Also, let's do a thread count of 16 - 4 is way too low for almost all use cases.

rjrudin (changes requested):

rjrudin (approved):

srinathgit (approved):

ryanjdew (approved):

PR #5698, opened 2021-05-21
DHFPROD-7283:Support for selecting references in mapping UI in HC

srinathgit:
### Description
Support for selecting references in mapping UI in HC
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/controllers/steps/MappingStepController.java
+++ b/marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/controllers/steps/MappingStepController.java
@@ -67,6 +67,13 @@ public class MappingStepController extends BaseController {
         return emptyOk();
     }
 
+    @RequestMapping(value = "/{stepName}/references",method = RequestMethod.GET)
+    @ApiOperation(value = "Get mapping references associated with the step")
+    @Secured("ROLE_readMapping")
+    public ResponseEntity<JsonNode> getReferences(@PathVariable String stepName) {
+        return ResponseEntity.ok(MappingService.on(getHubClient().getStagingClient()).getReferences(stepName));
+    }
+
     @RequestMapping(value = "/{stepName}/uris",method = RequestMethod.GET)
     @ApiOperation(value = "Get uris associated with source query of the step; uris count is determined by 'limit'", response = ArrayList.class)
     @Secured("ROLE_readMapping")
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/mapping/getReferences.api
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/mapping/getReferences.api
@@ -1,0 +1,15 @@
+{
+    "functionName": "getReferences",
+    "desc": "Gets the references  associated with given mapping step name.",
+    "params": [
+        {
+           "name": "stepName",
+           "datatype": "string"
+        }
+    ],
+    "return": {
+        "datatype": "jsonDocument",
+        "$javaClass": "com.fasterxml.jackson.databind.JsonNode",
+        "desc": "An array of objects, where each object has a reference name and an optional description"
+    }
+}
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/mapping/getReferences.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/mapping/getReferences.sjs
@@ -1,0 +1,33 @@
+/**
+ Copyright (c) 2021 MarkLogic Corporation
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+ http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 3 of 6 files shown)


wooldridge (approved):
Tested API functionality in Postman. Seeing the following response for mapCustomersJSON: 
```
[
    {
        "name": "$URI",
        "description": "The uri of the source document"
    },
    {
        "name": "$ZIP_POINTS",
        "description": "Maps zip codes to points"
    }
]
```
Also seeing appropriate responses for mapCustomersXML and a non-existent mapping step (404 error with message).

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/controllers/steps/MappingStepController.java
```diff
@@ -67,6 +67,14 @@
         return emptyOk();
     }
 
+    @RequestMapping(value = "/{stepName}/references",method = RequestMethod.GET)
+    @ApiOperation(value = "Get mapping references associated with the step")
+    @Secured("ROLE_readMapping")
+    public ResponseEntity<JsonNode> getReferences(@PathVariable String stepName) {
+        JsonNode response = MappingService.on(getHubClient().getStagingClient()).getReferences(stepName);
```
rjrudin, line 74: No need to change, but I would just one-line this - i.e. there's no value in declaring the JsonNode response variable.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/mapping/getReferences.api
```diff
@@ -0,0 +1,14 @@
...
+    "functionName": "getReferences",
+    "desc": "Gets the references  associated with given mapping step name.",
+    "params": [
+        {
+           "name": "stepName",
+           "datatype": "string"
+        }
+    ],
+    "return": {
+        "datatype": "jsonDocument",
```
rjrudin, line 11: Let's add a description here so it's clear what's returned - i.e. an array of objects, where each object has a name and an optional description

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/mapping/getReferences.sjs
```diff
@@ -0,0 +1,33 @@
...
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+'use strict';
+xdmp.securityAssert("http://marklogic.com/data-hub/privileges/read-mapping", "execute");
+
+const core = require('/data-hub/5/artifacts/core.sjs')
```
rjrudin, line 16: Both "const Artifacts" and "const core" are used for this import; we should only use one.  I don't like either of those - "Artifacts" implies a class, and "core" doesn't mean anything. 

Let's use "artifacts" as the prefix since that implies a module library and is descriptive. A separate PR can clean up the inconsistency of different prefixes for the same import.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/mapping/getReferences.sjs
```diff
@@ -0,0 +1,33 @@
...
+
+var stepName;
+const mappingStep = core.getArtifact("mapping", stepName);
+let resp = [{"name": "$URI", "description": "The uri of the source document"}];
```
rjrudin, line 20: Should be "URI", not "uri". And let's not abbreviate to "resp" - there's nothing gained there, just do "response", which is what's also done in the Java controller.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/mapping/getReferences.sjs
```diff
@@ -0,0 +1,33 @@
...
+const modulePath = mappingStep.mappingParametersModulePath;
+if (modulePath) {
+  try {
+    const paramsFunction = require(modulePath)["getParameterDefinitions"];
+    const userParams = paramsFunction(mappingStep);
+    userParams.map(userParam => userParam.name = "$" + userParam.name);
```
rjrudin, line 26: The use of "map" is a little confusing here because "map" is typically used to produce an output array. I think "forEach" is better here, which makes it clear we're modifying the array and not making a new one.

rjrudin (changes requested):
We need a Java-level test for the controller as well to verify that everything's working in the middle tier. I think it's fine if that test doesn't have a mapping with user-defined references; it can just verify that "URI" exists.

rjrudin (approved):

wooldridge (approved):
Seeing the following when GETting the references for mapCustomerJSON:
```
[
    {
        "name": "$URI",
        "description": "The URI of the source document"
    },
    {
        "name": "$ZIP_POINTS",
        "description": "Maps zip codes to points"
    }
]
```
Other tests look good as well.

PR #5705, opened 2021-05-24
DHFPROD-7120: View job details

ryanjdew:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/jobs.sjs
```diff
@@ -98,7 +100,31 @@ function saveNewJob(job) {
...
+              return xs.dateTime(artifact.lastUpdated).gt(timeStarted);
+            }
+          });
+        } catch (e) {
+          // the flow no longer exists
+          jobDoc.job.flowOrStepsUpdatedSinceRun = true;
+        }
+        if (xdmp.castableAs(XS_NAMESPACE, "dateTime", jobDoc.job.timeEnded)) {
+          // Calculate the duration
+          jobDoc.job.duration = xs.dateTime(jobDoc.job.timeEnded).subtract(timeStarted);
```
rjrudin, line 124: This function is currently "get job", which just grabs the job doc as-is from the jobs database. But it's no longer that now - now it's "get job and some other data about the job that HC cares about". That means every client of getJobDocWithId is paying for the cost of calculating "updated since". 

Thus, I think that only the client that cares about this additional data should have this logic in it. Right now, JobController is using a getJob DS endpoint, but so is FlowRunnerImpl, which doesn't care about this data. 

New DS endpoints are cheap - since the story has a title of "See the details", then I think an expected DS endpoint here would be "getJobWithDetails", where the description of the endpoint can provide some clue as to what "details" means. 

That also means that my above comment about extracting a function would apply to the new DS module - i.e. jobs.sjs shouldn't be modified at all. And the unit tests below are still fine, but they should invoke the DS module instead.

rjrudin (changes requested):

rjrudin (dismissed):

rahulvudutala (approved):

Sanjeevani19 (commented):
@ryanjdew https://jenkins.marklogic.com/blue/organizations/jenkins/Datahub_CI/detail/PR-5705/8/tests, Around 99 core tests are failing on the pipeline, Could you please look into them

rjrudin (dismissed):

Sanjeevani19 (dismissed):

Sanjeevani19 (commented):
@ryanjdew One of the test failure is related to monitor tile https://jenkins.marklogic.com/blue/organizations/jenkins/Datahub_CI/detail/PR-5705/10/tests, could you please look into it.

Sanjeevani19 (approved):

rjrudin (approved):

PR #5718, opened 2021-05-26
DHFPROD-7023: Retain state when navigating away and back to the Explo…

devinnicholson:
…re Tile

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [ ] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/e2e/cypress/integration/persistence/persistence.spec.tsx
+++ b/marklogic-data-hub-central/ui/e2e/cypress/integration/persistence/persistence.spec.tsx
@@ -1,5 +1,6 @@
 import {Application} from "../../support/application.config";
 import {toolbar} from "../../support/components/common";
+import browsePage from "../../support/pages/browse";
 import curatePage from "../../support/pages/curate";
 import loadPage from "../../support/pages/load";
 import LoginPage from "../../support/pages/login";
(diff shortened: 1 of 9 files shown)


wooldridge (changes requested):
@devinnicholson, not seeing search box text persisted in Explore, can we add that?

xnikhil08 (changes requested):
Not sure if saving the currently applied facets is part of the story, but seems like something we should do. It's not happening currently.

Sanjeevani19 (changes requested):
There are couple of issues I noticed on this PR
Query persistence 
1. Once the query is saved, the query should be highlighted in the dropdown with bold text
2. Edit the query and save it with new query name , it is not updated with latest name in the dropdown
3. Now shift between tiles, saved query is not persisted with facets
https://drive.google.com/file/d/1lwgzCki3xR7jZBFSNV3-ubFJ8wzSA9A_/view 


Clearing session and local storage
- If a user manually clears session and local storage and logs back in the explore screen keeps spinning without returning any results, this is highly unlikely but something that needs to be addressed
https://drive.google.com/file/d/1HBgaLfw7AXNImyraEpEcYQCEeUSx1T35/view

devinnicholson:
> Not sure if saving the currently applied facets is part of the story, but seems like something we should do. It's not happening currently.

The reason the facet AC is crossed off in the PR is because there's a separate story (https://project.marklogic.com/jira/browse/DHFPROD-7363). This story is the foundation so I can find foundational bugs before building off it.

devinnicholson:
> There are couple of issues I noticed on this PR
> Query persistence
> 
> 1. Once the query is saved, the query should be highlighted in the dropdown with bold text
> 2. Edit the query and save it with new query name , it is not updated with latest name in the dropdown
> 3. Now shift between tiles, saved query is not persisted with facets
>    https://drive.google.com/file/d/1lwgzCki3xR7jZBFSNV3-ubFJ8wzSA9A_/view
> 
> Clearing session and local storage
> 
> * If a user manually clears session and local storage and logs back in the explore screen keeps spinning without returning any results, this is highly unlikely but something that needs to be addressed
>   https://drive.google.com/file/d/1HBgaLfw7AXNImyraEpEcYQCEeUSx1T35/view

See above response in regards to facet persistence. Will fix the rest, thank you Sanjeevani.

devinnicholson:
> @devinnicholson, not seeing search box text persisted in Explore, can we add that?

This addition was force pushed earlier, just so reviewers know to check that as well.

PR #5755, opened 2021-06-04
DHFPROD-7420:Ensure 'additionalCollections' setting works for related entity mapping

srinathgit:
### Description
Ensure 'additionalCollections' setting works for related entity mapping
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mapping/entity-services/main.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mapping/entity-services/main.sjs
@@ -166,6 +166,9 @@ function createContextForRelatedEntityInstance(relatedEntityMapping, content){
   let entityContext = {};
   let relatedEntityPermissions = fn.string(relatedEntityMapping.permissions);
   let relatedEntityCollections = relatedEntityMapping.collections;
+  if(relatedEntityMapping.additionalCollections){
+    relatedEntityCollections = relatedEntityCollections.concat(relatedEntityMapping.additionalCollections);
+  }
   entityContext["permissions"] = hubUtils.parsePermissions(relatedEntityPermissions);
   entityContext["collections"] = relatedEntityCollections;
   if(content.context && content.context.originalCollections){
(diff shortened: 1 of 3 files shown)


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mapping/entity-services/main.sjs
```diff
@@ -166,6 +166,9 @@ function createContextForRelatedEntityInstance(relatedEntityMapping, content){
   let entityContext = {};
   let relatedEntityPermissions = fn.string(relatedEntityMapping.permissions);
   let relatedEntityCollections = relatedEntityMapping.collections;
+  if(relatedEntityMapping.additionalCollections){
```
rjrudin, line 169: This is a second "leak" now, where the mapping step is having to account for something the flow framework should handle - it was targetCollectionsAdditivity first, now additionalCollections. 

Is there here though that "collections" on a mapping step is supposed to contain the target entity name, and thus we can't just copy "collections" over to each related mapping content object? 

If so, that seems like another reason for the flow framework to automatically add target entity name as a collection based on the presence of targetEntityType. I believe that means we could get rid of additionalCollections. Not something we can achieve for 5.5 though.

srinathgit: Yeah, "collections" in related entity mapping should have target entity name so we can't copy it. 
Yeah, we will have to get rid of "additionalCollections", serves no purpose.

ryanjdew (approved):

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mapping/entity-services/main.sjs
rjrudin: Yeah, I think this is fine then, it's just an unfortunate consequence of the existence of "additionalCollections".

rjrudin (approved):

PR #5759, opened 2021-06-04
DHFPROD-7148:Handle XDMP_UNDFUN error in mapping step

srinathgit:
### Description
Throw better when encountering XDMP_UNDFUN in mappings. This is not be merged until error messages are finalized and respective stories groomed. Creating PR just for review.

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mapping/entity-services/lib.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mapping/entity-services/lib.sjs
@@ -5,8 +5,9 @@ const datahub = DataHubSingleton.instance();
 const entityLib = require("/data-hub/5/impl/entity-lib.sjs");
 const httpUtils = require("/data-hub/5/impl/http-utils.sjs");
 const hubUtils = require("/data-hub/5/impl/hub-utils.sjs");
+const mappingLib = require("/data-hub/5/mapping/mapping-lib.sjs");
 const inst = require('/MarkLogic/entity-services/entity-services-instance');
-const mappingLib = require('/data-hub/5/builtins/steps/mapping/default/lib.sjs');
+const mappingStepLib = require('/data-hub/5/builtins/steps/mapping/default/lib.sjs');
 const flowUtils = require("/data-hub/5/impl/flow-utils.sjs");
 const infoEvent = datahub.consts.TRACE_MAPPING;
 const infoEnabled = xdmp.traceEnabled(infoEvent);
@@ -350,7 +351,7 @@ function fallbackLegacyEntityLookup(targetEntityType) {
   let entityName = targetArr[targetArr.length - 1];
   let tVersion = targetArr[targetArr.length - 2] ? targetArr[targetArr.length - 2].split('-') : '';
   let modelVersion = tVersion[tVersion.length - 1];
-  return fn.head(mappingLib.getModel(entityName, modelVersion));
+  return fn.head(mappingStepLib.getModel(entityName, modelVersion));
 }
 
 function escapeXML(input = '') {
@@ -586,7 +587,7 @@ function validatePropertyMapping(fullMapping, userParameterNames, propertyName, 
     // Thus, this is not using es.mappingCompile, which does validation, and just invokes the transform instead.
     validateXmlMapping(xmlMapping);
   } catch (e) {
-    return hubUtils.getErrorMessage(e);
+    return mappingLib.extractErrorMessageForMappingUI(e);
   }
 }
 
@@ -687,7 +688,7 @@ function testMappingExpression(mapping, propertyName, sourceInstance, userParame
     }
   }
   catch(e){
-    resp.errorMessage = hubUtils.getErrorMessage(e);
+    resp.errorMessage = mappingLib.extractErrorMessageForMappingUI(e);
   }
   return resp;
 }
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mapping/entity-services/main.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mapping/entity-services/main.sjs
@@ -2,6 +2,7 @@ const consts = require('/data-hub/5/impl/consts.sjs');
 const defaultLib = require('/data-hub/5/builtins/steps/mapping/default/lib.sjs');
 const flowUtils = require("/data-hub/5/impl/flow-utils.sjs");
 const lib = require('/data-hub/5/builtins/steps/mapping/entity-services/lib.sjs');
+const mappingLibrary = require("/data-hub/5/mapping/mapping-lib.sjs");
 const entityValidationLib = require('entity-validation-lib.sjs');
 const xqueryLib = require('xquery-lib.xqy')
 const hubUtils = require("/data-hub/5/impl/hub-utils.sjs");
@@ -79,9 +80,14 @@ function main(contentSequence, options, stepExecutionContext) {
       try {
         arrayOfInstanceArrays = xqueryLib.dataHubMapToCanonical(instance, mappingURIforXML, mappingParams, {"format":outputFormat});
       } catch (e) {
-        throw Error(e);
+        const errorMessage = mappingLibrary.extractFriendlyErrorMessage(e);
+        if(errorMessage){
+          throw Error(errorMessage);
+        }
+        else{
+          throw Error(e);
+        }
       }
-
       hubUtils.hubTrace(traceEvent, `Entity instances with mapping ${mappingStep.name} and source document ${currentContentUri}: ${arrayOfInstanceArrays}`);
 
       let counter = 0;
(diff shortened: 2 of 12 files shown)


rjrudin (changes requested):

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mapping/entity-services/lib.sjs
```diff
@@ -687,7 +689,8 @@ function testMappingExpression(mapping, propertyName, sourceInstance, userParame
     }
   }
   catch(e){
-    resp.errorMessage = hubUtils.getErrorMessage(e);
+    const error = mappingLibrary.parseMappingError(e);
+    resp.errorMessage = typeof error  === 'string' ? error : hubUtils.getErrorMessage(e);
```
rjrudin, line 693: This is repeated above, so I think it's worth making a single function for it. Also, "parseMappingError" implies that some set of things will be returned based on the parsing - that's not the case here. The use case is - I've got an error, and I want to get a user-friendly error message for it. So how about "extractUserFriendlyErrorMessage"? And that function will have a comment noting that if that cannot be done, it'll fall back to hubUtils.getErrorMessage.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mapping/entity-services/main.sjs
```diff
@@ -79,7 +80,7 @@ function main(contentSequence, options, stepExecutionContext) {
       try {
         arrayOfInstanceArrays = xqueryLib.dataHubMapToCanonical(instance, mappingURIforXML, mappingParams, {"format":outputFormat});
       } catch (e) {
-        throw Error(e);
+        throw Error(mappingLibrary.parseMappingError(e));
```
rjrudin, line 83: This is confusing - parseMappingError will either return a string or an error object. Shouldn't this always return a string so it can be passed to the Error constructor?

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/mapping/mapping-lib.sjs
```diff
@@ -0,0 +1,32 @@
...
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+'use strict';
+
+function parseMappingError(e){
+  let error;
```
rjrudin, line 19: Assuming that this function should only return a "string message" as opposed to either a string or an error object, this should be renamed to "message".

rjrudin (changes requested):

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mapping/entity-services/main.sjs
srinathgit: @rjrudin 
We were passing the error object as argument to 'Error' constructor (till 5.5). I was just following the same pattern except in cases where the error is one that's specified in the jira stories in which case I construct a more meaningful message and pass that as argument to 'Error' constructor. Should I do something differently ? If the error occurring is not listed in one of the stories, what string should be passed to 'Error' constructor ?

rjrudin: I added a comment above that "throw Error(e)" part to explain it:

```
// For not-yet-known reasons, catching this error and then simply rethrowing it causes the MappingTest JUnit class
      // to pass due to the "message" part of the JSON in the stepOutput containing the expected output when this is done.
```

This seems pretty brittle - and perhaps the MappingTest shouldn't really be caring about fairly technical stuff within the error contents. It certainly seems simpler if we just do "throw Error(message)". 

Can you take a look at the MappingTest test that makes an assertion on the contents of "stepOutput" and see if that can be reasonably adjusted based on the output of "throw Error(message)"?

srinathgit: So should I just 

1. 'throw Error(message)' in case where 'e' is one of the errors listed in jira stories  and
2. 'throw e' in case of other errors (the ones not listed in jira stories)

(and fix MappingTest) ?

rjrudin: I like that - "throw Error(message)", else "throw e".

srinathgit: I am seeing that
1. When error object 'e' is passed as argument to 'Error' constructor (current behavior), stepOutput.message is 

"RESTAPI-SRVEXERR: /*:stylesheet/*:template[2] <xsl:element name=\"instance:mapping0Instances\"> <xsl:for-each select=\"/\"> <xsl:element name=\"instance:entityInstance\"> <xsl:element name=\"uri\"> <xsl:value-of select=\"$URI\"/></xsl:element><xsl:element name=\"value\"> <xsl:call-template name=\"mapping0-Order\"/></xsl:element></xsl:element></xsl:for-each></xsl:element> -- Extension Error:  code: 400 message: The pattern 'YYYY.MM.DD' cannot be applied to the value '01/08/1996' document: %3"
Screen Shot 2021-06-14 at 3 16 24 PM


2. When we just 'throw e', stepOutput.message is just "Extension Error"
Screen Shot 2021-06-14 at 3 17 07 PM


Doing the latter (throw e) also affects the message displayed in HC when step fails as can be seen in the screenshots in each cases

rjrudin: It seems this weirdness of needing "throw Error(e)" is specific to this block of code here. Thus, I think the mapping-lib function should ignore this and have a scope of "I'll try to extract a message from this, and if I cannot, I'll return null so the client can decide what to do". That way, the mapping-lib function is only returning a string or null, and not a string or an error object. 

The client code would then be (pseudocode):

```
const message = mappingLib.getHumanFriendlyErrorMessage(error);
if (message) throw Error(message) else throw Error(error);
```

rjrudin: And actually, i think "extract" is a better verb than "get", as "extract" implies that we're going to dig around in the error and see if we can come up with a nice error message. And "human" is redundant here - "extractFriendlyErrorMessage" should do the trick.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mapping/entity-services/lib.sjs
```diff
@@ -586,7 +587,7 @@ function validatePropertyMapping(fullMapping, userParameterNames, propertyName,
     // Thus, this is not using es.mappingCompile, which does validation, and just invokes the transform instead.
     validateXmlMapping(xmlMapping);
   } catch (e) {
-    return hubUtils.getErrorMessage(e);
+    return mappingLibrary.extractErrorMessageForMappingUI(e);
```
rjrudin, line 590: We really need to do "mappingLib" here to be consistent with e.g. entity-lib.sjs. So for mapping/default/lib.sjs, let's call that "mappingStepLib".

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/mapping/mapping-lib.sjs
```diff
@@ -0,0 +1,38 @@
...
+  }
+  return errorMessage;
+}
+function extractErrorMessageForMappingUI(e){
+  let errorMessage = extractFriendlyErrorMessage(e);
+  return errorMessage ? errorMessage : hubUtils.getErrorMessage(e);
+}
+
+module.exports = {
+  extractErrorMessageForMappingUI,
```
rjrudin, line 36: This shouldn't have "UI" in it, because this functionality will be used regardless of how the step is run - e.g. if you run the step via Gradle and there's an error, it will show up in stepOutput. 

Perhaps just have one function here with a "boolean fallbackToGenericErrorMessage" that controls whether or not hubUtils.getErrorMessage is used.

rjrudin (changes requested):

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/mapping/mapping-lib.sjs
srinathgit: 1. extractErrorMessageForMappingUI() will be used exclusively in HC mapping UI when user clicks on "Test" button. The fallback when the error is not the one mentioned in the jira stories here is hubUtils.getErrorMessage(e) (as it was earlier)
2. extractFriendlyErrorMessage() will be used when you run the step from gradle or from HC run tile. The fallback when the error is not the one mentioned in the jira stories here is throw Error(e) (as it was earlier)

rjrudin (approved):

rjrudin (approved):

ryanjdew (approved):

PR #5760, opened 2021-06-07
DHFPROD-7430: Add time for trigger to previewMatchActivity test

ryanjdew:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- N/A Code passes ESLint tests
- N/A Added Tests
  

- ##### Reviewer:

- N/A Reviewed Tests


marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/data-services/mastering/previewMatchingActivity/setup.xqy
```diff
@@ -10,6 +10,8 @@ import module namespace test = "http://marklogic.com/test" at "/test/test-helper
 hub-test:load-entities($test:__CALLER_FILE__)
 
 ;
+(: Sleep added to give time for trigger that builds the entity-based TDEs. :)
```
rjrudin, line 13: These tests right now our fast enough that I think it's reasonable to wait a bit longer if necessary, like 100ms or 250ms. It should be an exception to the rule that we need to do this, so waiting a little bit longer is fine.

ngodugu-marklogic (approved):

PR #5769, opened 2021-06-09
DHFPROD-7463: Turning off positions for metadata field range indexes on staging and final databases

rahulvudutala:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

### DIFF
--- a/marklogic-data-hub/src/main/resources/hub-internal-config/database-fields/staging-database.xml
+++ b/marklogic-data-hub/src/main/resources/hub-internal-config/database-fields/staging-database.xml
@@ -13,7 +13,7 @@
       <stemmed-searches>off</stemmed-searches>
       <word-searches>false</word-searches>
       <field-value-searches>true</field-value-searches>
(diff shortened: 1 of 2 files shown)


srinathgit (approved):

akshaysonvane (approved):

rjrudin (changes requested):
I think this needs to be adjusted on datahubCreatedOn as well:

```
<field-value-positions>true</field-value-positions>
```

srinathgit (approved):

rjrudin (approved):
I think it's worth manually testing out these changes too - e.g. via Hub Central, which I believe makes use of these indexes.

akshaysonvane (approved):

Sanjeevani19 (commented):
As rob mentioned, @rahulvudutala Could you please test this manually to confirm everything works , if it affects something we shouldn't allow it to be merged ?

ryanjdew (approved):

PR #5772, opened 2021-06-09
DHFPROD-7467: Clear schemas databases before running the test

akshaysonvane:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [N/A] Added Tests
  

- ##### Reviewer:

- [N/A] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/flow/flowRunner/ingestMapMaster/setup.sjs
+++ b/marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/flow/flowRunner/ingestMapMaster/setup.sjs
@@ -5,6 +5,8 @@ const hubTestX = require("/test/data-hub-test-helper.xqy");
 const test = require("/test/test-helper.xqy");
 
 hubTestX.resetHub();
+// Clearing the schema databases so that the merge step does not pickup any TDE triples that might be left behind from any previous tests
+hubTestX.clearSchemasDatabases();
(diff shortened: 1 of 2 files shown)


marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/flow/flowRunner/ingestMapMaster/setup.sjs
```diff
@@ -5,6 +5,8 @@ const hubTestX = require("/test/data-hub-test-helper.xqy");
 const test = require("/test/test-helper.xqy");
 
 hubTestX.resetHub();
+// Clearing the schema databases so that the merge step does not pickup any TDE triples that might be left behind from any previous tests
```
rjrudin, line 8: Any reason not to include this in resetHub? This seems like a tricky problem to debug which could cause future tests to fail as well. The cost of doing this should be very small - i.e. something in the tens of milliseconds, since there's rarely much data in these databases.

akshaysonvane: I think we can, I just did not want to mess up other tests since we want to start the release tomorrow.

rjrudin (dismissed):

srinathgit (approved):

ryanjdew (approved):

PR #5775, opened 2021-06-10
DHFPROD-7471: Show page size menu in mapping UI pagination (source an…

briantang:
…d entity)

### Description

- Added page size menu for source and entity tables
- https://drive.google.com/file/d/1ESfMQGWdA4ESc9ZFjRa4W6HT25oZbrgj/view

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/src/components/entities/mapping/mapping-step-detail/mapping-step-detail.test.tsx
+++ b/marklogic-data-hub-central/ui/src/components/entities/mapping/mapping-step-detail/mapping-step-detail.test.tsx
@@ -1641,7 +1641,7 @@ describe("RTL Source-to-entity map tests", () => {
     mockGetSourceDoc.mockResolvedValue({status: 200, data: data.jsonSourceDataLargeDataset});
     mockGetNestedEntities.mockResolvedValue({status: 200, data: personRelatedEntityDefLargePropSet});
 
-    let getByText, queryByText, getByTestId, getByTitle, getAllByTitle, queryByTitle, getByLabelText, queryByTestId;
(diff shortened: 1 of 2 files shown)


wooldridge (changes requested):
@briantang, this is great, thanks for getting it in co quickly.

I think we want to set hideOnSinglePage to false since, if we don't, there's no way for user to set the page size to 1 (to limit the rows to possibly improve performance with lots of deep properties) if, say, there are only 10 top level rows and the default is 20. @sbayatpur @jbelonoj

briantang:
> @briantang, this is great, thanks for getting it in co quickly.
> 
> I think we want to set hideOnSinglePage to false since, if we don't, there's no way for user to set the page size to 1 (to limit the rows to possibly improve performance with lots of deep properties) if, say, there are only 10 top level rows and the default is 20. @sbayatpur @jbelonoj

Good point @wooldridge, pushed a fix.

Sanjeevani19 (approved):
Looks good, tests pass on pipeline. https://jenkins.marklogic.com/blue/organizations/jenkins/Datahub_CI/detail/PR-5775/2/pipeline

briantang:
Updated pagination options size to "small", as requested by @wooldridge and @jbelonoj

wooldridge (approved):
Looks great, and I'm seeing the updated test pass locally.

Sanjeevani19 (approved):

AkshayBajajML (approved):

PR #5795, opened 2021-06-16
DHFPROD-7343: Deleted index gets restored on final db

AkshayBajajML:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [ ] Reviewed Tests


marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/controllers/model/CreateAndUpdateModelTest.java
```diff
@@ -71,6 +72,15 @@ void testCreateNamespacedModel() {
         createNamespacedModel();
     }
 
+    @Test
+    @WithMockUser(roles = {"writeEntityModel"})
+    void testUpdatedIndexes() {
```
rjrudin, line 77: It's not clear what this test is testing. We need to identify the scenario somewhere - i.e. that we are verifying that if a user manually removes an index that was created for a property that no longer needs an index, the index isn't accidentally restored when the entity model is saved. 

Thus, I'd expect the test method to read like this (pseudocode):

- create a model with a facetable dateTime property
- change property type to date
- verify both date and dateTime indexes exist
- manually remove dateTime index from final database
- add another property
- verify that the dateTime index wasn't added back to the staging database

rjrudin (changes requested):

marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/controllers/model/CreateAndUpdateModelTest.java
AkshayBajajML: Done!!

marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/controllers/model/CreateAndUpdateModelTest.java
```diff
@@ -207,6 +221,101 @@ private void updateModelEntityTypes() {
...
+    private void addFacetableDateTimeProperty() {
+        Assumptions.assumeTrue(isVersionCompatibleWith520Roles());
+
+        String entityTypes = "[\n" +
+            "  {\n" +
+            "    \"entityName\": \"" + MODEL_NAME + "\",\n" +
+            "    \"modelDefinition\": {\n" +
+            "      \"Customer\": {\n" +
+            "        \"required\": [],\n" +
+            "        \"pii\": [\n" +
```
rjrudin, line 233: I'm going to be picky since this is one of your first JUnit tests. :)

We want every test to be minimized to only the details that matter. In this case, we don't care about pii or required, so both should be removed (neither need to be there). And since elementRangeIndex is being set, facetable doesn't need to be set either. Also, collation is meaningless unless it's a string property, so that should be removed as well.

marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/controllers/model/CreateAndUpdateModelTest.java
```diff
@@ -207,6 +221,101 @@ private void updateModelEntityTypes() {
...
+
+        controller.updateModelEntityTypes(readJsonArray(entityTypes));
+    }
+
+    private void updateDataType() {
+        Assumptions.assumeTrue(isVersionCompatibleWith520Roles());
+
+        String entityTypes = "[\n" +
+            "  {\n" +
+            "    \"entityName\": \"" + MODEL_NAME + "\",\n" +
```
rjrudin, line 259: Same thing here - the only thing we want is the the property and elementRangeIndex. And to reduce the duplication across these two requests, you should make a private method that constructs this payload based on a "String type" property, substituting that in for the datatype.

marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/controllers/model/CreateAndUpdateModelTest.java
```diff
@@ -207,6 +221,101 @@ private void updateModelEntityTypes() {
...
+            "        \"elementRangeIndex\": [\n" +
+            "          \"" + ENTITY_PROPERTY_1 + "\"\n" +
+            "        ],\n" +
+            "        \"properties\": {\n" +
+            "          \"" + ENTITY_PROPERTY_1 + "\": {\n" +
+            "            \"datatype\": \"date\",\n" +
+            "            \"facetable\": true,\n" +
+            "            \"collation\": \"http://marklogic.com/collation/codepoint\"\n" +
+            "          },\n" +
+            "          \"" + ENTITY_PROPERTY_2 + "\": {\n" +
```
rjrudin, line 305: Same thing here - minimize the payload, and remove facetable/collation for the new property, as whether or not the new property is facetable or not is irrelevant. We want to make it easy as possible to see that the only things that are changing are 1) the datatype in the first update, and then 2) the addition of a new property in the second update.

marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/controllers/model/CreateAndUpdateModelTest.java
```diff
@@ -230,6 +339,23 @@ private void assertIndexDeployment() {
         });
     }
 
+    private void getIndexes(Boolean isIndexDeleted) {
+        String json = new DatabaseManager(getHubClient().getManageClient()).getPropertiesAsJson(getHubConfig().getDbName(DatabaseKind.FINAL));
+        Database db = new DefaultResourceMapper(new API(getHubClient().getManageClient())).readResource(json, Database.class);
+        List<ElementIndex> rangeIndexes = db.getRangeElementIndex();
+        if(!isIndexDeleted) {
+            assertEquals("someProperty", rangeIndexes.get(rangeIndexes.size() - 2).getLocalname());
```
rjrudin, line 347: Assertions that assume that the index is somewhere in a list are brittle - they could break. We don't care what position in the list the index is in, we just care whether it exists or not. 

To achieve that, get the Database object, then add a private method like this:

```
private boolean rangeIndexExists(Database db, String name, String type) 
```

And then just iterate through each index object in the Database, and if one with the given name/type are found, return true. Else return false.

marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/controllers/model/CreateAndUpdateModelTest.java
```diff
@@ -254,6 +380,17 @@ private void verifyIndexes(String dbName) {
         assertTrue(pathExpressions.containsAll(Arrays.asList("//*:instance/testPathIndexForDHFPROD4704", "/(es:envelope|envelope)/(es:instance|instance)/Customer/someOtherProperty")));
     }
 
+    private void updateIndexes() {
```
rjrudin, line 383: This should have a more self-documenting name - it's not just updating the indexes, it's removing the path and element indexes. But - I think we only care about element indexes here. So this should only clear about the element indexes, and it should have a name that makes it clear that's what it does - i.e. "removeRangeElementIndexesFromFinalDatabase()".

marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/controllers/model/CreateAndUpdateModelTest.java
```diff
@@ -230,6 +339,23 @@ private void assertIndexDeployment() {
         });
     }
 
+    private void getIndexes(Boolean isIndexDeleted) {
```
rjrudin, line 342: This method name doesn't indicate what it's doing - i.e. a "get" method is expected to return something. Instead, I think you want two methods that will each use the method I mention below:

```
private void assertDateTimeIndexExists()
private void assertDateTimeIndexDoesntExist()
```

Those are the exact things you want to prove, so having methods with those names makes good sense.

rjrudin (changes requested):

AkshayBajajML:
Closing this and would be submitting new PR due to conflicts.

PR #5813, opened 2021-06-21
DHFPROD-7427: Add Help info to entity table title in mapping

AkshayBajajML:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [ ] Reviewed Tests


AkshayBajajML:
@wooldridge covered previous comments

devinnicholson (commented):
It looks good. Two quick questions:
1) Why are we linking to https://docs.marklogic.com/datahub/5.5/index.html instead of https://docs.marklogic.com/datahub/5.5/flows/about-mapping.html ?

2) Do we want to have an extra break in the tooltip between the sentences?

AkshayBajajML:
> It looks good. Two quick questions:
> 
> 1. Why are we linking to https://docs.marklogic.com/datahub/5.5/index.html instead of https://docs.marklogic.com/datahub/5.5/flows/about-mapping.html ?
> 2. Do we want to have an extra break in the tooltip between the sentences?

1) Updated the link.
2) I showed popover to @jbelonoj with extra break but we agreed to exclude it.

briantang (approved):

PR #5815, opened 2021-06-22
DHFPROD-7050: Gradle Task For Provenance Data Management

rahulvudutala:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/provenance/pruning.sjs
```diff
@@ -26,7 +26,7 @@ function validatePruneRequest({ retainDuration, batchSize = 100 }) {
 function pruneProvenance(pruneRequest, endpointState) {
     xdmp.securityAssert("http://marklogic.com/data-hub/privileges/prune-provenance", "execute");
```
ryanjdew, line 27: We should also change this privilege to be delete-provenance, along with this test file name and function names.

ryanjdew (changes requested):

marklogic-data-hub/src/main/java/com/marklogic/hub/flow/ProvenanceManager.java
```diff
@@ -0,0 +1,63 @@
+package com.marklogic.hub.flow;
```
rjrudin, line 1: While jobs are specific to flows, provenance is not - i.e. mastering can generate property-level provenance that has nothing to do with steps and flows. So provenance should be considered a top-level feature in DHF, and thus worthy of its own package under com.marklogic.hub - com.marklogic.hub.provenance.

marklogic-data-hub/src/main/java/com/marklogic/hub/flow/ProvenanceManager.java
```diff
@@ -0,0 +1,63 @@
...
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import com.marklogic.client.dataservices.ExecCaller;
+import com.marklogic.client.dataservices.IOEndpoint;
+import com.marklogic.client.io.JacksonHandle;
+import com.marklogic.hub.HubClient;
+
+import java.io.IOException;
+import java.io.InputStreamReader;
+
+public class ProvenanceManager {
```
rjrudin, line 13: Almost all of this code is the same as in JobManager - the only difference appears to be the API path. So we can make a static utility method like this:

```
public static void deleteData(String apiPath, String retainDuration)
```

Because that class will be specific to classes in the com.marklogic.client.dataservices package, and specifically the "bulk" classes, let's put that method into a new com.marklogic.hub.dataservices.BulkUtil class. Both JobManager and ProvenanceManager can use it and avoid all of this duplicated code.

marklogic-data-hub/src/main/java/com/marklogic/hub/flow/ProvenanceManager.java
```diff
@@ -0,0 +1,63 @@
...
+        ExecCaller.BulkExecCaller bulkCaller = this.deleteProvenanceCaller
+                .bulkCaller(this.deleteProvenanceCaller.newCallContext()
+                        .withEndpointStateAs(endpointState)
+                        .withEndpointConstantsAs(endpointConstants));
+        bulkCaller.setErrorListener(errorListener);
+        bulkCaller.awaitCompletion();
+
+        Throwable throwable = errorListener.throwable;
+        if (throwable != null) {
+            throw new RuntimeException("Unable to delete provenance, cause: " + throwable.getMessage(), throwable);
```
rjrudin, line 51: This can just be "Unable to delete data, cause: " in the new BulkUtil method.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/provenance/pruning.sjs
```diff
@@ -26,7 +26,7 @@ function validatePruneRequest({ retainDuration, batchSize = 100 }) {
...
     // update with validated request properties
-    Object.assign(pruneRequest, validatePruneRequest(pruneRequest));
+    pruneRequest = validatePruneRequest(pruneRequest);
```
rjrudin, line 29: I think this module should just be "prov-lib.sjs", and likely the future home of all the logic in "prov.sjs" today. That makes it consistent with e.g. "entity-lib.sjs".

marklogic-data-hub/src/test/java/com/marklogic/hub/flow/DeleteProvenanceTest.java
```diff
@@ -0,0 +1,52 @@
...
+
+import static org.junit.jupiter.api.Assertions.*;
+
+public class DeleteProvenanceTest extends AbstractHubCoreTest {
+
+    String provenanceCollection = "http://marklogic.com/provenance-services/record";
+
+    @BeforeEach
+    void installProjectAndRunFlow() {
+        installProjectInFolder("test-projects/provenance-test" );
```
rjrudin, line 18: We should only do this in the deleteAllProvenanceRecords test - installing the project and running the flow isn't necessary for the other two tests.

ml-data-hub-plugin/src/main/groovy/com/marklogic/gradle/DataHubPlugin.groovy
```diff
@@ -148,6 +148,9 @@ class DataHubPlugin implements Plugin<Project> {
         project.task("hubDeleteJobs", type: DeleteJobsTask, group: developGroup,
             description: "Deletes job and batch documents in the jobs database that are older than a duration specified  " +
                 "via -PretainDuration=(duration) (e.g. -PretainDuration=P30D to delete every document that is older than 30 days)")
+        project.task("hubDeleteProvenance", type: DeleteProvenanceTask, group: developGroup,
+            description: "Clears provenance records in the jobs database that are older than a duration specified " +
```
rjrudin, line 152: Use "Deletes", not "Clears"

rjrudin (changes requested):

marklogic-data-hub/src/main/java/com/marklogic/hub/dataservices/ProvenanceService.java
```diff
@@ -56,7 +56,7 @@ private ProvenanceServiceImpl(DatabaseClient dbClient, JSONWriteHandle servDecl)
                 this.baseProxy = new BaseProxy("/data-hub/5/data-services/provenance/", servDecl);
 
                 this.req_pruneProvenance = this.baseProxy.request(
-                    "pruneProvenance.sjs", BaseProxy.ParameterValuesKind.MULTIPLE_NODES);
+                    "deleteProvenance.sjs", BaseProxy.ParameterValuesKind.MULTIPLE_NODES);
```
rjrudin, line 59: Run "gradle generateProvenanceInterface", that will update this file, which should include updating the pruneProvenance references too.

rjrudin (changes requested):
Looks good, just run that generateProvenanceInterface task.

rjrudin (approved):

ryanjdew (approved):

PR #5864, opened 2021-07-06
DHFPROD-5878: Stop table scrolling in Explorer

devinnicholson:
### Description
This PR removes the multiple scrollbars in Explore. There should be 1 vertical scrollbar on the Explore tile (the scroll is not on the table anymore). When reducing size of the window, a horizontal scrollbar should appear. The search bar stays fixed in line with table view and snippet view, which now has dynamic width. Sticky header was added in this PR. This PR also fixes the placeholder that appears when scrolling in Modeling.

Reviewers: Please check scrolling in other tiles to make sure no other tile was affected.
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

### DIFF
--- a/marklogic-data-hub-central/ui/e2e/cypress/integration/explore/crudQueriesFromManageQueries.spec.tsx
+++ b/marklogic-data-hub-central/ui/e2e/cypress/integration/explore/crudQueriesFromManageQueries.spec.tsx
@@ -89,7 +89,7 @@ describe("manage queries modal scenarios, developer role", () => {
     browsePage.getSaveQueriesDropdown().click();
     browsePage.waitForSpinnerToDisappear();
     browsePage.selectEntity("Person");
-    cy.waitUntil(() => browsePage.getDetailInstanceViewIcon("/json/persons/last-name-dob-custom1.json"), {timeout: 10000}).click();
+    cy.waitUntil(() => browsePage.getDetailInstanceViewIcon("/json/persons/last-name-dob-custom1.json"), {timeout: 10000}).click({force: true});
     browsePage.waitForSpinnerToDisappear();
   });
   it("Navigate to detail page and verify if manage query modal opens up.", () => {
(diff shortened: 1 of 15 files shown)


Sanjeevani19 (dismissed):
@devinnicholson Here are couple of things I noticed
- There should be a horizontal scroll on the table to make it responsive on different screen widths, otherwise we are not able to get to detail view column when testing on 1024 screen widths and lower
- The top search-bar doesn't align with the table width when we see on 1024 screen widths and lower
https://drive.google.com/file/d/14vf0y3fm4lryQtnsJ24fKy3mImy2OINv/view 
- Since this is a 8 pointer story, it is worth addressing the fixed header also in this story so we can test everything together

Sanjeevani19 (changes requested):
@devinnicholson Couple of things I noticed on latest changes

- There is an extra white space between the table and tile on different screen widths, the space is more on smaller desktop widths of 1024 and the search bar overflows on bigger desktop widths of 2560
 https://drive.google.com/file/d/1VNr7R-PgzgR1ASUi_s4Loa0ICyAKArBF/view
-  Export and column selector icons should be visible for users to click on which is available when you select a particular entity
Screen Shot 2021-07-14 at 4 51 38 PM
- Collapsible side bar icon should be at top right of sidebar, currently it is shifted below
Screen Shot 2021-07-14 at 4 50 52 PM

briantang (dismissed):
Screen Shot 2021-07-15 at 2 28 48 PM

@devinnicholson Nice work! This is such a great improvement. One thing I noticed is that the width of the header does not adjust when the sidebar is collapsed. One solution you can try is giving the header a different className if the sidepanel is collapsed, and the styling of that class should set the width to 100%.

ex: `<div className={collapse? styles. stickyHeaderCollapsed : styles.stickyHeader}>`

wooldridge:
Looking good! I'm seeing the width adjust correctly on different browser widths and the collapse button is sticky. I just see the issue @briantang mentioned.

briantang (approved):
Looks great!

Sanjeevani19 (dismissed):

Sanjeevani19 (approved):

briantang (approved):

briantang (approved):

Sanjeevani19 (approved):

PR #5880, opened 2021-07-12
DHFPROD-7570: stopOnError now works when the collector fails

rjrudin:
### Description

I had to update a couple existing tests, which I think in hindsight were erroneously expecting the wrong job status. If stopOnError=true, and the collector fails, then a user expects the flow to be stopped and no other steps to execute, which means the job status should be stop-on-error. 

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


ryanjdew (approved):

rjrudin:
Removed some System.out.println's

rjrudin:
Removed some System.out.println's

ryanjdew (approved):

AkshayBajajML (approved):

PR #5886, opened 2021-07-13
DHFPROD-7448: Add legend to Properties tab in Graph View

briantang:
### Description
- https://drive.google.com/file/d/1eZL6L0IBE8NbZpahMbrU2tDZy_sZoKiy/view
- When legend is shown, the legend + "Hide Legend <<" text is set to wrap to a new line depending on the width of the side panel, as discussed with Jamie. 
- Design approved by Jamie

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/src/components/modeling/graph-view/properties-tab/properties-tab.tsx
+++ b/marklogic-data-hub-central/ui/src/components/modeling/graph-view/properties-tab/properties-tab.tsx
@@ -1,5 +1,9 @@
-import React from "react";
+import React, {useState} from "react";
 import PropertyTable from "../../property-table/property-table";
+import styles from "./properties-tab.module.scss";
+import {FontAwesomeIcon} from "@fortawesome/react-fontawesome";
+import {faLayerGroup, faKey} from "@fortawesome/free-solid-svg-icons";
+import arrayIcon from "../../../../assets/icon_array.png";
 
 interface Props {
   entityTypeData: any;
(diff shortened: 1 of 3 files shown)


wooldridge (changes requested):
Nice work, a couple comments:

Can we call it showLegend, that way we don't have the double negative when it is visible? (I think of "hide" as a negative and I find double negatives harder my brain to parse.)   

Can we have the CSS add horizontal space after the legend elements instead of before. That way when we have a linebreak, we avoid the left-side space?
Screen Shot 2021-07-15 at 9 32 54 AM

briantang:
Thanks @wooldridge, good points. Should be fixed now.

AkshayBajajML (dismissed):
Looks Good!

marklogic-data-hub-central/ui/src/components/modeling/graph-view/properties-tab/properties-tab.tsx
```diff
@@ -1,22 +1,61 @@
...
+    }
+  };
+
+  const LegendShowHide =
+    <div>
+      {showLegend ?
+        <div className={styles.legend}>
+          <div data-testid="foreignKeyIconLegend" className={styles.foreignKeyLegendText}>
+            <FontAwesomeIcon className={styles.foreignKeyIcon} icon={faKey}/> Foreign
+                                  Key Relationship
```
wooldridge, line 32: Can we clean up the label content so that it is always together (same line either with the icon or not)?

marklogic-data-hub-central/ui/src/components/modeling/graph-view/properties-tab/properties-tab.tsx
```diff
@@ -1,22 +1,61 @@
...
+          </div>
+          <div data-testid="multipleIconLegend" className={styles.legendText}><img
+            className={styles.arrayImage} src={arrayIcon} alt={""}/> Multiple
+          </div>
+          <div data-testid="structuredIconLegend" className={styles.legendText}>
+            <FontAwesomeIcon className={styles.structuredIcon}
```
wooldridge, line 38: Can we clean up the icon content so that it is always together (same line either with the label or not)?

marklogic-data-hub-central/ui/src/components/modeling/graph-view/properties-tab/properties-tab.tsx
```diff
@@ -1,22 +1,61 @@
...
+              icon={faLayerGroup}/> Structured Type
+          </div>
+          <a data-testid="hideLegendLink" onClick={() => toggleLegend()}>&nbsp;&nbsp;&nbsp;Hide Legend &lt;&lt;</a>
```
wooldridge, line 41: Using &nbsp; for layout (e.g. padding) is generally an anti-pattern and in this case results in clickable (linked) whitespace, which seems odd. Can we remove those HTML entities and format using CSS where needed?

wooldridge (changes requested):
I had a few more comments.

marklogic-data-hub-central/ui/src/components/modeling/graph-view/properties-tab/properties-tab.tsx
briantang: Right thanks, removed the &nbsp; as it's actually redundant. I believe the &lt; and &gt; are necessary though to display "<<" and ">>" without react picking it up as a tag.

wooldridge (approved):
Looks great!

marklogic-data-hub-central/ui/src/components/modeling/graph-view/properties-tab/properties-tab.tsx
wooldridge: Yeah, I just meant the space entities, thanks!

AkshayBajajML (approved):

PR #5895, opened 2021-07-16
DEVO-350: add publishing task to installer-for-dhs module

yafeldman:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/installer-for-dhs/build.gradle
+++ b/installer-for-dhs/build.gradle
@@ -2,6 +2,7 @@ plugins {
     id "java"
     id 'com.github.johnrengelman.shadow' version '5.2.0'
     id "io.snyk.gradle.plugin.snykplugin" version "0.4"
+    id 'maven-publish'
 }
 
 group = 'com.marklogic'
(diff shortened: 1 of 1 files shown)


installer-for-dhs/build.gradle
```diff
@@ -53,3 +54,27 @@ task installDhfIntoDhs(type: Exec) {
 }
 installDhfIntoDhs.dependsOn jar
 
+
+publishing {
+    publications {
+        maven(MavenPublication) {
+            artifact("build/libs/marklogic-data-hub-$version-installer"+".jar")
+            artifactId = 'marklogic-data-hub-installer'
```
rjrudin, line 62: This isn't a good choice, because it sounds like it's a generic DHF installer. But it's specifically for DHS. Granted, the line above should ideally have "installer-for-dhs" in it as well. But if this is only being published internally, "installer-for-dhs" should be used as the artifactId.

rjrudin (changes requested):

installer-for-dhs/build.gradle
yafeldman: fixed to artifact name installer-for-dhs

PR #5903, opened 2021-07-20
DHFPROD-7459: Create PROV records for Job/Step(s) at Job start

rahulvudutala:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/job/finishJob.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/job/finishJob.sjs
@@ -18,8 +18,44 @@
 xdmp.securityAssert("http://marklogic.com/data-hub/privileges/run-step", "execute");
 
 const Job = require("/data-hub/5/flow/job.sjs");
+const dhProv = require('/data-hub/5/provenance/dh-provenance.xqy');
+const config = require("/com.marklogic.hub/config.sjs");
+const hubUtils = require("/data-hub/5/impl/hub-utils.sjs");
 
 var jobId;
 var jobStatus;
 
-Job.getRequiredJob(jobId).finishJob(jobStatus, fn.currentDateTime()).update();
+function updateJobProvenance(jobId, jobEndTime) {
+  const options = {
+    "endTime": jobEndTime
+  }
+  const provCollectionQuery = cts.collectionQuery("http://marklogic.com/provenance-services/record");
+  const provIdAttributeQuery = cts.elementAttributeValueQuery(fn.QName("http://www.w3.org/ns/prov#", "activity"),
+    fn.QName("http://www.w3.org/ns/prov#", "id"), fn.concat("job:", jobId));
+  const jobProvQuery = cts.andQuery([provCollectionQuery, provIdAttributeQuery]);
+
+  const stagingJobProvRecordUri = fn.head(
+    hubUtils.invokeFunction(function () {
+      return cts.uris("", null, jobProvQuery);
+    }, config.STAGINGDATABASE)
+  );
+
+  let finalJobProvRecordUri = fn.head(
+    hubUtils.invokeFunction(function () {
+      return cts.uris("", null, jobProvQuery);
+    }, config.FINALDATABASE)
+  );
+
+  if(stagingJobProvRecordUri) {
+    dhProv.updateEndTimeInProvenanceRecord(stagingJobProvRecordUri, options, config.STAGINGDATABASE);
+  }
+
+  if(finalJobProvRecordUri) {
+    dhProv.updateEndTimeInProvenanceRecord(finalJobProvRecordUri, options, config.FINALDATABASE);
+  }
+}
+
+const jobEndTime = fn.currentDateTime();
+const finishedJob = Job.getRequiredJob(jobId).finishJob(jobStatus, jobEndTime).update();
+updateJobProvenance(jobId, jobEndTime);
+finishedJob;
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/job/startJob.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/job/startJob.sjs
@@ -14,24 +14,48 @@
  limitations under the License.
  */
 'use strict';
-
 xdmp.securityAssert("http://marklogic.com/data-hub/privileges/run-step", "execute");
 
 const consts = require('/data-hub/5/impl/consts.sjs');
+const config = require("/com.marklogic.hub/config.sjs");
 const hubUtils = require("/data-hub/5/impl/hub-utils.sjs");
 const Job = require("/data-hub/5/flow/job.sjs");
 const jobs = require("/data-hub/5/impl/jobs.sjs");
+const dhProv = require('/data-hub/5/provenance/dh-provenance.xqy');
+const coreLib = require('/data-hub/5/artifacts/core.sjs')
 
 var jobId;
 var flowName;
+var stepNumber;
+
+function insertJobProvenance(jobId, startDateTime) {
+    const targetDatabase = hubUtils.invokeFunction(function () {
+    const fullFlow = coreLib.getFullFlow(flowName);
+    return fullFlow["steps"][stepNumber]["options"]["targetDatabase"];
+  }, config.STAGINGDATABASE);
+
+  const options = {
+    "startDateTime": startDateTime,
+    "user": xdmp.getCurrentUser()
+  }
+  const record = dhProv.newProvenanceRecord(jobId, options);
+  if(targetDatabase.toString() === config.FINALDATABASE) {
+    dhProv.insertProvenanceRecord(record, config.FINALDATABASE);
+  } else {
+    dhProv.insertProvenanceRecord(record, config.STAGINGDATABASE);
+  }
+}
 
 // A user is allowed to reuse a jobId, in which case the existing job document will be overwritten.
 // The updateJob function must be used in that scenario so that the amp associated with that function can be
 // used to let the user overwrite the document.
+let newJob;
 if (jobs.getJob(jobId)) {
   hubUtils.hubTrace(consts.TRACE_FLOW, `Overwriting job '${jobId}' for flow '${flowName}'`);
-  Job.newJob(flowName, jobId).update();
+  newJob = Job.newJob(flowName, jobId).update();
 }
 else {
-  Job.newJob(flowName, jobId).create();
+  newJob = Job.newJob(flowName, jobId).create();
 }
+insertJobProvenance(jobId, newJob.job.timeStarted);
+newJob;
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/provenance/dh-provenance.xqy
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/provenance/dh-provenance.xqy
@@ -1,0 +1,145 @@
+(:
+  Copyright (c) 2021 MarkLogic Corporation
+
+  Licensed under the Apache License, Version 2.0 (the "License");
+  you may not use this file except in compliance with the License.
+  You may obtain a copy of the License at
+
+     http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+:)
+
+(: This is a general extension to MarkLogic PROV library to handle PROV record creation and insertion :)
+
+xquery version "1.0-ml";
+
+module namespace dhps = "http://marklogic.com/data-hub/dh-provenance-services";
+
+declare namespace prov = "http://www.w3.org/ns/prov#";
+declare namespace xs="http://www.w3.org/2001/XMLSchema";
+
+declare private variable $dhps-collection as xs:string := "http://marklogic.com/provenance-services/record";
+declare private variable $user-role as xs:string := "ps-user";
+declare private variable $internal-role as xs:string := "ps-internal";
+declare private variable $ps-dir as xs:string := "/provenance/";
+
+
+declare function new-provenance-record($id as xs:anyURI, $options as map:map) as node() {
+  let $record := map:map()
+  let $record := $record=>job-record-activity($id, $options)=>job-record-agent($options)=>job-record-associatedWith($id, $options)=>build()
+  return $record
+};
+
+declare function insert-provenance-record($doc as node(), $target-database as xs:string) as empty-sequence() {
+  let $uri := record-uri(fn:head($doc/prov:activity/@prov:id))
+  let $permissions := (xdmp:permission($internal-role,"update"), xdmp:permission($user-role,"read"))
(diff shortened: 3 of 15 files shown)


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/job/startJob.sjs
```diff
@@ -14,24 +14,39 @@
...
 var flowName;
 
+function recordJobProvenance(jobId, timeStarted) {
+  const options = {
+    "startTime": timeStarted,
+    "user": xdmp.getCurrentUser()
+  }
+  const record = dhProv.provenanceRecord(jobId, options);
+  dhProv.insertProvenanceRecord(record, config.FINALDATABASE);
+  dhProv.insertProvenanceRecord(record, config.STAGINGDATABASE);
```
rjrudin, line 36: This doesn't always seem appropriate - i.e. should this happen for a job that only writes to one database? The story doesn't explain what to do either.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/job/startJob.sjs
```diff
@@ -14,24 +14,39 @@
...
 const Job = require("/data-hub/5/flow/job.sjs");
 const jobs = require("/data-hub/5/impl/jobs.sjs");
+const dhProv = require('/data-hub/5/provenance/dh-provenance.xqy');
 
 var jobId;
 var flowName;
 
+function recordJobProvenance(jobId, timeStarted) {
+  const options = {
+    "startTime": timeStarted,
```
rjrudin, line 31: I highly recommend using "startDateTime" to make it clear that this is a dateTime value. If it's a date, do "startDate". If it's a time - e.g. "HH:mm:ss" - call it "startTime". Using explicit naming like that removes any uncertainty over whether the expected value is a date, time, or dateTime.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/provenance/dh-provenance.xqy
```diff
@@ -0,0 +1,128 @@
...
+};
+
+declare private function build($map as map:map) as node(){
+  wrap-in-prov-document(build-components($map))
+};
+
+declare private function job-record-activity($map as map:map, $id as xs:string, $options as map:map*) as map:map {
+  let $prov-id := fn:concat("job:", $id)
+  let $rec :=
+    element { xs:QName("prov:activity") } {
```
rjrudin, line 55: Something to consider, but not necessarily change - I prefer to only use this form of XML construction when dynamically building elements and attributes - e.g. "element {fn:node-name($some-var)}". Otherwise, I think this is easier to read, with the downside being the need to duplicate QNames in start and close tags:

```
let $rec := 
  <prov:activity prov:id="{$prov-id}">
    <prov:label>{$id}<prov:label>
    etc...
```

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/provenance/dh-provenance.xqy
```diff
@@ -0,0 +1,128 @@
...
+  let $rec :=
+    element { xs:QName("prov:wasAssociatedWith") } {
+      element prov:activity { attribute prov:ref {$prov-id} },
+      element prov:agent { attribute prov:ref {$user} }
+    }
+  let $_ := map:put($map, "wasAssociatedWith", $rec)
+  return $map
+};
+
+declare function update-endTime-in-provenance-record($uri as xs:string, $options as map:map*, $target-database as xs:string) as empty-sequence() {
```
rjrudin, line 90: Let's put public functions at top and private at the bottom, which is consistent with languages like Java. Makes it easier to understand the public API of this module. And generally, that improves readability - i.e. a user can read in a top-down fashion, starting with a public function and then scrolling down to find the private functions called by it.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/provenance/dh-provenance.xqy
```diff
@@ -0,0 +1,128 @@
...
+  let $path := "prov:document/prov:activity"
+  let $child-node := <prov:endTime>{map-get($options, "endTime")}</prov:endTime>
+  return dhps-document-update($uri, $child-node, $path, $target-database)
+};
+
+declare function provenance-record($id as xs:anyURI, $options as map:map) as node() {
```
rjrudin, line 96: A function does something, and thus its name should have a verb in it - this is building a new object, so let's do "new-provenance-record", which is consistent with "newJob" in job.sjs.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/provenance/dh-provenance.xqy
```diff
@@ -0,0 +1,128 @@
+xquery version "1.0-ml";
+
+module namespace dhps = "http://marklogic.com/data-hub/dh-provenance-services";
```
rjrudin, line 3: I think this library should have a comment to explain the purpose of it. It's sitting in a directory with another library named "prov-lib.sjs". And that begs the question - what's the purpose of each library? i.e. What purposes should I use dh-provenance.xqy for, and what purposes should I use prov-lib.shs for?

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/data-services/provenance/dhProvLibTest.sjs
```diff
@@ -0,0 +1,107 @@
...
+    test.assertNotEqual(null, cts.doc(uris.stagingRecordUri)),
+    test.assertNotEqual(null, cts.doc(uris.finalRecordUri)),
+    test.assertEqual("job:verifyProvenanceRecordCreatedAfterStartJob-job-id-2", xs.string(fn.head(record.xpath("/*:document/*:activity/@*:id")))),
+    test.assertEqual("verifyProvenanceRecordCreatedAfterStartJob-job-id-2", xs.string(fn.head(record.xpath("/*:document/*:activity/*:label")))),
+    test.assertNotEqual(null, xs.string(fn.head(record.xpath("/*:document/*:activity/*:startTime")))),
+    test.assertNotEqual(null, fn.head(record.xpath("/*:document/*:activity/*:endTime")))
+  ];
+}
+
+function queryProvenanceRecord(jobId) {
```
rjrudin, line 78: I think it'd be better to put each of these test scenarios into a separate module, and then use setup.xqy to clear out the databases before each test. Then, instead of duplicating logic that's in the application code, you can safely grab the first document in the collection below in each database. 

See https://wiki.marklogic.com/display/ENGINEERING/Writing+Data+Hub+Tests#WritingDataHubTests-HowtoorganizeMLunittests for more information too on the downsides of multiple scenarios per test module.

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/data-services/provenance/dhProvLibTest.sjs
```diff
@@ -0,0 +1,107 @@
...
+    declareUpdate();
+    jobService.startJob(jobId, flowName);
+  });
+  const uris = queryProvenanceRecord(jobId);
+  let record = fn.head(fn.doc(uris.finalRecordUri));
+
+  return [
+    test.assertNotEqual(null, cts.doc(uris.stagingRecordUri)),
+    test.assertNotEqual(null, cts.doc(uris.finalRecordUri)),
+    test.assertEqual("job:verifyProvenanceRecordCreatedAfterStartJob-job-id-1", xs.string(fn.head(record.xpath("/*:document/*:activity/@*:id")))),
```
rjrudin, line 43: For this test and any test that involves XML data, I'd strongly consider using XQuery for the tests. These assertions are pretty difficult to read because of the bloat from "xs.string / fn.head / record.xpath". An equivalent assertion in XQuery would be much nicer to read:

```
let $activity := $record/*:document/*:activity
return (
  test:assert-equal("dh:Job", $activity/*:label/text())
```

And I also wouldn't use a wildcard in the expressions above. A wildcard should only be used when we don't know what the namespace might be. We know exactly what's expected here, so we should use the specific namespace prefix for each element. 

Finally - you could write the test in SJS, and then just do all the assertions in an XQuery block of code. I think that'd still be a lot less code and a lot easier to read.

rjrudin (changes requested):

ryanjdew (approved):

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/job/startJob.sjs
rahulvudutala: Added the information to the story. The job provenance record is inserted to both the staging and final content databases during the job start as we are not aware of the target database. The unused provenance record will be deleted when the job finishes. The cleanup will be addressed as part of another ticket.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/job/finishJob.sjs
```diff
@@ -18,8 +18,46 @@
...
+
+  let finalJobProvRecordUri = fn.documentUri(fn.head(
+    xdmp.invokeFunction(function () {
+      return cts.search(jobProvQuery);
+    }, {
+      database: xdmp.database(config.FINALDATABASE)
+    })
+  ));
+
+  if(stagingJobProvRecordUri && fn.docAvailable(stagingJobProvRecordUri)) {
```
rjrudin, line 52: The fn.docAvailable call isn't necessary, and actually won't work for one of the two. 

Also, since you only want the URI back, use cts.uris instead of cts.search - it's faster to get the URI from the URI lexicon as opposed to loading the document from disk.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/job/startJob.sjs
rjrudin: Hmm - only a step has a target database, not a flow. Since a job can involve running multiple steps, a job may involve writing data to multiple databases. It is conceivable that step 1 writes to staging, step 2 writes to final, and step 3 writes to staging. What would the target database be in that scenario?

rjrudin (changes requested):

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/job/startJob.sjs
rahulvudutala: The job provenance record is initially written to both staging and final databases
1) step1 writes to staging, step2 writes to final -- then the records in both staging and final exists. If a step target database is staging, provenance is written into the record created in the staging database. If it is final database, then the provenance is written to the record created in the final database
2) step1 writes to final, step2 writes to final -- then the record in final exists and staging will be deleted and vice-versa
This is yet to be implemented. Will be implemented in the future when writing provenance for each step.

rjrudin: Okay - though would it make life easier if the call to startJob included the first step to be run, such that the endpoint could figure out the target database of the first step and thus avoid writing a document that needs to be deleted later? Note that there's no guarantee that that document will get deleted, as the user could stop the flow at any time.

rahulvudutala: 1) It will be easy if startJob included all the steps that the user wants to run
2) When user stops a flow, will the finishJob be called?  I am assuming that the job document will be updated when the user stops running the flow by calling finishJob which will also clean up the provenance records

rjrudin: If a user stops a flow run by Gradle or the client jar via ctrl-c, then the JVM will terminate immediately and finishJob won't be called. For that reason, I think it's better if we only create a job provenance document when we know we need one. 

As for passing along the first step, or any of them - the initializeFlow method in FlowRunnerImpl calls startJob. That code can get the steps from stepsMap.get(jobId). So you can then modify the startJob DS endpoint to accept any other data that you need, whether it's the first step to be run or all of the step numbers. The startJob DS endpoint can then look up the full flow definition to determine that the target databases are.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/job/finishJob.sjs
rahulvudutala: It works for both when a flow with steps having staging and final databases as target databases.

rahulvudutala: Also, changed cts.search to cts.uris

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/job/startJob.sjs
rahulvudutala: For example, if I all the step numbers are used to create the provenance records, then the issue with stopping the flow using Gradle or the client jar via ctrl-c still exists.

So its better to create the prov record based on first step. When recording job prov information for a step, we can create the job prov record if necessary or update an existing record. Am I right?

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/job/finishJob.sjs
rjrudin: But finishJob is called against the jobs database. The fn.docAvailable call will then be made against the jobs database. But these two URIs were just fetched from the staging and final databases. So I'm confused - I would think fn.docAvailable will always return false, unless the URIs are also in the jobs database??

ryanjdew (approved):

ryanjdew (approved):

ryanjdew (dismissed):

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/data-services/provenance/data-hub-provenance/test-data/steps/ingestion/loadCustomersJSON.step.json
```diff
@@ -0,0 +1,26 @@
+{
```
rjrudin, line 1: If this works, that's kind of okay... but it doesn't make sense to test an ingestion step from ML code since in order to run an ingestion step, you have to use the DHF Java code (WriteStepRunner and FlowRunner). So this test raises questions of - wait, what happens if you run an ingestion step from ML code? And is the result of that even valid for testing? 

I recommend instead using the createSimpleProject or createSimpleMappingProject function in data-hub-test-helper.sjs instead. Then you don't need to create a flow file or a step file. Since you don't care about what the mapping actually does, you don't need to pass anything into createSimpleMappingProject.

rjrudin (changes requested):

rjrudin (approved):

ryanjdew (approved):

PR #5920, opened 2021-07-26
DHFPROD-7428:Include number of entity instances created by mapping when "Test" button is clicked

srinathgit:
### Description
Include number of entity instances created by mapping when "Test" button is clicked
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
- [x] Run UI tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mapping/entity-services/lib.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mapping/entity-services/lib.sjs
@@ -115,6 +115,20 @@ function fetchNamespacesFromMappingStep(mappingStep){
   return namespaces;
 }
 
+function getMappingNamespacesObject(mappingStep){
+  let namespaces = {};
+  if (mappingStep.namespaces) {
+    for (const prefix of Object.keys(mappingStep.namespaces).sort()) {
+      if (mappingStep.namespaces.hasOwnProperty(prefix)) {
+        if (!reservedNamespaces.includes(prefix)) {
+          namespaces[prefix]=mappingStep.namespaces[prefix];
+        }
+      }
+    }
+  }
+  return namespaces;
+}
+
 /**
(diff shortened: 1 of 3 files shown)


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mapping/entity-services/lib.sjs
```diff
@@ -115,6 +115,20 @@ function fetchNamespacesFromMappingStep(mappingStep){
   return namespaces;
 }
 
+function fetchMappingNamespacesObject(mappingStep){
```
rjrudin, line 118: Let's use "get", which implies we're getting something from an object and not doing a query. "fetch" isn't used much elsewhere and thus raises the question of "What does fetch mean here?", which could imply a query.

rjrudin (changes requested):

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mapping/entity-services/lib.sjs
```diff
@@ -511,6 +526,21 @@ function validateAndTestUriExpressions(mapping, validatedMappingsArray, sourceIn
   return mapping;
 }
 
+function calculateInstanceCount(mapping, validatedMappingsArray, sourceInstance){
+  const namespaces = getMappingNamespacesObject(mapping);
+  validatedMappingsArray.forEach((entityMapping, mappingIndex) => {
+    if(mappingIndex != 0){
+      const sourceDocument = fn.head(xdmp.unquote(xdmp.quote(sourceInstance)));
```
rjrudin, line 533: Sorry just noticed this - it looks like this should be outside the for loop so it's only calculated once, as it doesn't depend on what we're iterating over.

rjrudin (changes requested):

ryanjdew (approved):

rjrudin (approved):

PR #5936, opened 2021-08-03
DHFPROD-7606: Update PROV records for Job/Step(s) at start of step run

rahulvudutala:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/job/startStep.sjs
```diff
@@ -17,9 +17,55 @@
...
 
-Job.getRequiredJob(jobId).startStep(stepNumber).update()
\ No newline at end of file
+function updateProvenance(flowName, stepNumber, jobId) {
+  const stepDetails = fn.head(hubUtils.invokeFunction(function () {
+    const fullFlow = coreLib.getFullFlow(flowName);
+    return fullFlow["steps"][stepNumber];
+  }, config.STAGINGDATABASE));
+
+  const latestProvenance = stepDetails["options"]["latestProvenance"];
```
rjrudin, line 37: This is ignoring runtime options, but I'm not sure if that matters for this release. The use of getFullFlow is also a little inefficient, because that's doing a query for all steps, while we only need one. getFullFlow (which should really be in flow.sjs, not core.sjs, but that's a separate matter) could accept an optional stepNumber argument to only query for that step. 

But the main issue is runtime options - but only if we're exposing this property for 5.6. If we're not, I think it would be worth adding a 5.7 ticket so we don't forget to support runtime options here.

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/jobs.sjs
```diff
@@ -299,17 +299,31 @@ function saveNewJob(job) {
...
             hubUtils.hubTrace(consts.TRACE_FLOW, `Inserting job report with URI: ${reportUri}`);
             hubUtils.writeDocument(reportUri, jobReport, permissions, collections, config.JOBDATABASE);
-          }  
+          }
         }
       }
     }
   }
 
+  function findProvenanceRecordUriFromJobId(jobId, targetDatabase) {
```
rjrudin, line 308: @ryanjdew  @rahulvudutala  A refactor to consider (not as part of this ticket) - this module should become "job-lib.sjs" in the data-hub/5/flow directory. We'd then have "job.sjs" as a class module for wrapping a job document, and then multiple library modules for various purposes (which job.sjs would delegate to as appropriate).

rjrudin (commented):
Looks good, I'm just wondering about runtime options support for latestProvenance

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/job/startStep.sjs
ryanjdew: @rahulvudutala How difficult would it be for you to adjust this to use the runtime options as part of this PR? Beyond needing to support it eventually, it would be more convenient for demos and testing if it was incorporated in this PR.

rahulvudutala: @rjrudin @ryanjdew Added changes taking care of runtime options and getFullFlow only for a step

marklogic-data-hub/src/main/java/com/marklogic/hub/step/impl/QueryStepRunner.java
```diff
@@ -188,7 +188,7 @@ public RunStepResponse run() {
         combinedOptions.put("jobId", jobId);
 
         if (jobOutputIsEnabled()) {
-            JobService.on(hubClient.getJobsClient()).startStep(jobId, step);
+            JobService.on(hubClient.getJobsClient()).startStep(jobId, step, flow.getName(), flow.getOptions());
```
rjrudin, line 191: You want this.combinedOptions here and in the other 3 places

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/artifacts/core.sjs
```diff
@@ -224,9 +224,12 @@ function generateArtifactKey(artifactType, artifactName, artifactVersion = 'late
     return `${artifactType}:${artifactName}:${artifactVersion}`;
 }
 
-function getFullFlow(flowName, artifactVersion = 'latest') {
+function getFullFlow(flowName, stepNumber, artifactVersion = 'latest') {
```
rjrudin, line 227: You can drop artifactVersion from here, it's never used

rjrudin (changes requested):

ryanjdew (approved):

rjrudin (approved):

PR #5937, opened 2021-08-03
DHFPROD-7620:CRUD for entity node coordinates to support graph display

srinathgit:
### Description
CRUD for entity node coordinates to support graph display
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/controllers/ModelController.java
+++ b/marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/controllers/ModelController.java
@@ -264,6 +264,18 @@ public class ModelController extends BaseController {
         public String description;
         public String namespace;
         public String namespacePrefix;
+        public HubCentralConfig hubCentral;
+    }
+
+    public static class HubCentralConfig {
+        public static Modeling modeling;
+    }
+
+    public static class Modeling{
+        public double graphX;
+        public double graphY;
+        public String icon;
+        public String color;
     }
 
     public static class UpdateModelInfoInput {
@@ -271,6 +283,7 @@ public class ModelController extends BaseController {
         public String description;
         public String namespace;
         public String namespacePrefix;
+        public HubCentralConfig hubCentral;
     }
 
     public static class ModelReferencesInfo {
--- a/marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/controllers/model/CreateAndUpdateModelTest.java
+++ b/marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/controllers/model/CreateAndUpdateModelTest.java
@@ -69,6 +69,7 @@ public class CreateAndUpdateModelTest extends AbstractModelTest {
     void testCreateNamespacedModel() {
         runAsTestUserWithRoles("hub-central-entity-model-writer");
         createNamespacedModel();
+        updateModelWithGraphConfig();
     }
 
     @Test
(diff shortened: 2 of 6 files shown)


marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/controllers/ModelController.java
```diff
@@ -264,13 +264,26 @@ private ModelsService newService() {
         public String description;
         public String namespace;
         public String namespacePrefix;
+        public HubCentral hubCentral;
```
rjrudin, line 267: This class name should be fine for now, but you may later want to make it "HubCentralConfig" just so it's not the same as the existing-and-very-important HubCentral class.

marklogic-data-hub-central/src/test/java/com/marklogic/hub/central/controllers/model/CreateAndUpdateModelTest.java
```diff
@@ -69,6 +69,7 @@ void testModelsServicesEndpoints() {
     void testCreateNamespacedModel() {
         runAsTestUserWithRoles("hub-central-entity-model-writer");
         createNamespacedModel();
+        updateHubCentral();
```
rjrudin, line 72: Hmm - in this context, I think we really need a qualifier on "HubCentral" because this sounds like "I'm going to update Hub Central the application". I think "hubCentral" is totally fine in the model definition itself, where it's obvious that it's configuration. 

Since we're only focused on adding graph config, how about calling the data that you're creating/updating/verifying "GraphConfig" - so "updateModelWithGraphConfig"?

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/models/createModel.sjs
```diff
@@ -29,6 +29,7 @@ const name = input.name;
 const description = input.description;
 const namespace = input.namespace ? input.namespace : null;
 const namespacePrefix = input.namespacePrefix ? input.namespacePrefix : null;
+const hubCentral = input.hubCentral ? input.hubCentral : null;
```
rjrudin, line 32: I think here, we should definitely say "hubCentralConfig".

specs/models/ModelDescriptor.json
```diff
@@ -43,6 +43,28 @@
     },
     "definitions": {
       "$ref": "./ModelDefinitions.json"
+    },
+    "hubCentral": {
+      "type": "object",
```
rjrudin, line 48: Let's add a description to this, both for hubCentral and for modeling - for the former, something like "Captures model configuration specific to the HubCentral application", and for the latter, "Captures the location of an entity model on the modeling graph in HubCentral"

rjrudin (changes requested):

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/controllers/ModelController.java
```diff
@@ -264,13 +264,26 @@ private ModelsService newService() {
         public String description;
         public String namespace;
         public String namespacePrefix;
+        public HubCentralConfig hubCentralConfig;
```
rjrudin, line 267: I think this should be "HubCentralConfig hubCentral", right?

rjrudin (approved):

ryanjdew (approved):

PR #5940, opened 2021-08-04
DHFPROD-7663: Can now run DHF on Gradle 7

rjrudin:
### Description

Most changes are to ml-data-hub tests and the example projects; Gradle 7 uses Groovy 3, so had to bump up spock test to 2.0-3.0. Then had to modify many tests to not use the junit4 TemporaryFolder class, but rather a plain File.

For example projects - bumped them to gradlew 7.1.1, and replaced jcenter with mavenCentral.

Fixed several DHF tasks; task properties apparently should not have "public" on them, because otherwise you need to include a public getter. So it's simpler / better just to have e.g. "String flowName" with @Input and @Optional on it. Also fixed every other task property that didn't already have @Input on it, or added @Internal if appropriate. 

DHF will now also scaffold out a gradlew properties file for Gradle 7.1.1. 

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  
- ##### Reviewer:

- [x] Reviewed Tests


ryanjdew (approved):

srinathgit (dismissed):

SameeraPriyathamTadikonda:
Build failed while downloading node from internal node-mirror which is setup with http.
Execution failed for task ':marklogic-data-hub-central:nodeSetup'.

[2021-08-04T16:45:54.109Z] > Could not resolve all dependencies for configuration ':marklogic-data-hub-central:detachedConfiguration1'.

[2021-08-04T16:45:54.109Z]    > Using insecure protocols with repositories, without explicit opt-in, is unsupported. Switch Ivy repository 'ivy(http://node-mirror.eng.marklogic.com:8080/)' to redirect to a secure protocol (like HTTPS) or allow insecure protocols. See https://docs.gradle.org/7.1.1/dsl/org.gradle.api.artifacts.repositories.UrlArtifactRepository.html#org.gradle.api.artifacts.repositories.UrlArtifactRepository:allowInsecureProtocol for more details. 

Please allow insecure protocols.

ryanjdew (approved):

srinathgit (approved):

PR #5946, opened 2021-08-05
DHFPROD-7645:Include timezone for parseDate() response

srinathgit:
### Description
Include timezone for parseDate() response
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x]  PR title is in the format JIRA_ID:Title

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/mapping/core-functions-impl.xqy
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/mapping/core-functions-impl.xqy
@@ -162,7 +162,7 @@ declare private function parse-standard-date($value as xs:string?, $pattern as x
     else $pattern
 
   let $picture := fn:tokenize(fn:replace($pattern, "YYYY", "yyyy"), "T")[1]
-  return fn:substring(fn:string(xdmp:parse-yymmdd($picture, $value)), 1, 10)
+  return xs:string(xs:date(xdmp:parse-yymmdd($picture, $value)))
 };
 
 declare private function parse-non-standard-date($value as xs:string?, $pattern as xs:string?) as xs:string?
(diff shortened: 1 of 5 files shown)


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/mapping/core-functions-impl.xqy
```diff
@@ -174,10 +174,12 @@ declare private function parse-non-standard-date($value as xs:string?, $pattern
       let $month := map:get($MONTH-MAP, fn:upper-case(fn:substring($value, 1, 3)))
       let $day := fn:substring($value, 5, 2)
       let $year := fn:substring($value, 8, 4)
-      return fn:string-join(($year, $month, $day), "-")
+      let $log := xdmp:log(fn:string-join(($year, $month, $day), "-"))
```
rjrudin, line 177: Should remove this and the log stmt below too

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/builtins/steps/mapping/entity-services/runMapping.sjs
```diff
@@ -13,7 +13,7 @@ function runMapping () {
   let instance = result.envelope.instance;
   try {
     assertions.push(
-      test.assertEqual('2019-12-07', fn.string(instance.OrderType.purchaseDate)),
+      test.assertEqual(xs.string(xs.date(`2019-12-07${serverTimezone}`)), fn.string(instance.OrderType.purchaseDate)),
```
rjrudin, line 16: No need to change, but it's good to make a function for something like this which is duplicated a couple times - e.g. 

    test.assertEqual(makeDate("2019-12-07"), ...)

Where "makeDate" would handle the timezone bit and calling xs.string/xs.date.

rjrudin (changes requested):
Looks good, just need to remove log statements

marklogic-data-hub/src/test/ml-modules/root/test/suites/data-hub/5/builtins/steps/mapping/entity-services/runMapping.sjs
srinathgit: It's the same value so I used a variable.

rjrudin (approved):

ryanjdew (approved):

rjrudin (approved):

PR #5966, opened 2021-08-10
DHFPROD-7530: Showing and editing relationships between entity types

briantang:
### Description
- Edit relationship via clicking on edge in graph view; can edit relationship name, join property, and cardinality
- Out of scope: Swapping, Delete relationship, target entity dropdown
- RTL and Cypress tests added
- https://drive.google.com/file/d/1krFjAQkg0j0-ivmyT8cYGRGSYo3rQvo1/view

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
- [x] Run UI tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/src/components/modeling/graph-view/graph-vis/graph-vis.tsx
+++ b/marklogic-data-hub-central/ui/src/components/modeling/graph-view/graph-vis/graph-vis.tsx
@@ -7,6 +7,7 @@ import {faFileExport} from "@fortawesome/free-solid-svg-icons";
 import {FontAwesomeIcon} from "@fortawesome/react-fontawesome";
 import NodeSvg from "./node-svg";
 import graphConfig from "../../../../config/graph-vis.config";
+import AddEditRelationship from "../relationship-modal/add-edit-relationship";
 
 type Props = {
   entityTypes: any;
@@ -14,6 +15,9 @@ type Props = {
   filteredEntityTypes: any;
   entitySelected: any;
   isEntitySelected: boolean;
+  updateSavedEntity: any;
+  toggleRelationshipModal: any;
+  relationshipModalVisible: any;
 };
 
 // TODO temp hardcoded node data, remove when retrieved from db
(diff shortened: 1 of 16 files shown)


wooldridge (changes requested):
This looks GREAT. Feature works, clean code. I noticed that when I create a new relationship in the tabular view I can't edit that relationship in Graph:

https://drive.google.com/file/d/1Z_P16NGpRQ-etVb50v8aRkYNjQ-uHFLv/view

I'm assuming this has to do with the current hardcoding and us not yet supporting color but is there a way to defensively code so that we ignore color if it is not on the relationship for now?

briantang:
@wooldridge Good catch. I've defensively fixed it so that nodes without supporting colors for now will default to the default color that we set.

marklogic-data-hub-central/ui/src/components/modeling/graph-view/graph-vis/graph-vis.tsx
```diff
@@ -205,6 +247,26 @@ const GraphVis: React.FC<Props> = (props) => {
         props.handleEntitySelection(nodes[0]);
       }
     },
+    click: (event) => {
+      //if click is on an edge
+      if (event.edges.length > 0 && event.nodes.length < 1) {
+        let connectedNodes = window.graphVisApi.getConnectedNodes(event.edges[0]);
```
xnikhil08, line 253: We should directly refer to the network function here like **network.getConnectedNodes(event.edges[0])** instead of adding it in window object first and then calling it from there. This will also improve the performance of the graph edge modal display.

xnikhil08 (changes requested):

briantang:
Thanks for pointing that out @xnikhil08. Fixed now.

xnikhil08 (dismissed):

AkshayBajajML (approved):
- Functionality works as expected!

wooldridge (approved):

PR #5994, opened 2021-08-17
DHFPROD-6147: Test performance of disableJobOutput

devinnicholson:
### Description
Note: the story itself will not be completed until I have written a performance report regarding the results.

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests

### DIFF
--- a/performance-tests/gradle.properties
+++ b/performance-tests/gradle.properties
@@ -10,8 +10,13 @@ mlHost=localhost
 mlUsername=test-data-hub-developer
 mlPassword=password
 
+# Parameters that apply to multiple performance tests 
+sourceDocumentCount=10000
+
 # Parameters for mapping functions test
-sourceDocumentCount=1000
 entityPropertyCount=100
 iterations=3
 mappingExpression=sourceProperty1
+
+# Parameters for disableJobOutput test
+disableJobOutput=false
--- a/performance-tests/src/main/jmeter/disable-job-output.jmx
+++ b/performance-tests/src/main/jmeter/disable-job-output.jmx
@@ -1,0 +1,283 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<jmeterTestPlan version="1.2" properties="5.0" jmeter="5.4.1">
+  <hashTree>
+    <TestPlan guiclass="TestPlanGui" testclass="TestPlan" testname="disableJobOutput performance test" enabled="true">
+      <stringProp name="TestPlan.comments">This test shows the effect of disableJobOutput with a value of &quot;true&quot; and &quot;false&quot; when a custom step that does nothing is run on N documents</stringProp>
+      <boolProp name="TestPlan.functional_mode">false</boolProp>
+      <boolProp name="TestPlan.tearDown_on_shutdown">true</boolProp>
+      <boolProp name="TestPlan.serialize_threadgroups">false</boolProp>
+      <elementProp name="TestPlan.user_defined_variables" elementType="Arguments" guiclass="ArgumentsPanel" testclass="Arguments" testname="User Defined Variables" enabled="true">
+        <collectionProp name="Arguments.arguments"/>
+      </elementProp>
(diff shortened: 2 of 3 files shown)


devinnicholson:
Adding auth username/password variables

performance-tests/gradle.properties
```diff
@@ -11,7 +11,11 @@ mlUsername=test-data-hub-developer
 mlPassword=password
 
 # Parameters for mapping functions test
-sourceDocumentCount=1000
+## sourceDocumentCount=1000
```
rjrudin, line 14: This should be in a section of parameters that apply to multiple tests, as it still applies to the mapping functions test

rjrudin (changes requested):
I'll try this out in the morning

performance-tests/src/main/jmeter/disable-job-output.jmx
```diff
@@ -0,0 +1,340 @@
...
+                <boolProp name="HTTPArgument.use_equals">true</boolProp>
+                <stringProp name="Argument.name">rs:options</stringProp>
+              </elementProp>
+            </collectionProp>
+          </elementProp>
+          <stringProp name="HTTPSampler.domain">${host}</stringProp>
+          <stringProp name="HTTPSampler.port">${port}</stringProp>
+          <stringProp name="HTTPSampler.protocol"></stringProp>
+          <stringProp name="HTTPSampler.contentEncoding"></stringProp>
+          <stringProp name="HTTPSampler.path">/v1/resources/mlRunFlow</stringProp>
```
rjrudin, line 252: We don't want to use this, this endpoint is deprecated (see the comments in it). The code that you have in the "Run flow" component will run the flow, and you can customize that to pass in disableJobOutput in the options Map.

performance-tests/src/main/jmeter/disable-job-output.jmx
```diff
@@ -0,0 +1,340 @@
...
+            <stringProp name="Argument.metadata">=</stringProp>
+          </elementProp>
+          <elementProp name="disableJobOutput" elementType="Argument">
+            <stringProp name="Argument.name">disableJobOutput</stringProp>
+            <stringProp name="Argument.value">${__P(disableJobOutput, false)}</stringProp>
+            <stringProp name="Argument.metadata">=</stringProp>
+          </elementProp>
+          <elementProp name="authUsername" elementType="Argument">
+            <stringProp name="Argument.name">authUsername</stringProp>
+            <stringProp name="Argument.value">${__P(authUsername, admin)}</stringProp>
```
rjrudin, line 64: Use test-data-hub-developer and password here; we know that user exists because it's created via "./gradlew bootstrap".

rjrudin (dismissed):

performance-tests/src/main/jmeter/disable-job-output.jmx
```diff
@@ -0,0 +1,340 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<jmeterTestPlan version="1.2" properties="5.0" jmeter="5.1.1 r1855137">
```
rjrudin, line 2: Use JMeter 5.4.1, that's the latest release.

rjrudin (changes requested):
This looks good but needs one change - each job fails because the module path associated with the step definition doesn't exist. That prevents any batch documents from being created, and batch documents are what's going to slow things down the most (since only 1 job doc is created, but N batch docs can be created). 

In the "Create flow" processor, you'll need to insert the following text into the modules database with a URI matching that in the step definition (this is the simplest possible step module):

```
function main(contentItem, options) {
  return contentItem;
}

module.exports = {
  main
};
```

You can use hubClient.getModulesClient().newDocumentManager() to get a DocumentManager that will let you write this.

Then try running the flow with job output enabled, and verify that you're getting both a job doc (with a status of "success") and N batch docs.

rjrudin (approved):

srinathgit (approved):

PR #6029, opened 2021-08-24
DHFPROD-7719: Can override how DHF gets entity property values

rjrudin:
### Description

Reworked entity-search-lib so that there's now a "getEntityDetails" function that returns the entityName and properties for a doc, if they exist. That function is where DHF supports a custom module for constructing entity details.

Added a "user-config" concept since the approach of using the "info" block in entity models doesn't work here, and it was awkward for the non-entity-type specific stuff - i.e. for post-processing database indexes and search options. 

Also moved ExploreCustomEntityInstancesTest and its files over to the HC project so that it could use EntitySearchManager instead of QueryManager. 

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests  

- ##### Reviewer:

- [ ] Reviewed Tests


ryanjdew (approved):

rjrudin:
Going to look into a different implementation.

PR #6050, opened 2021-08-30
DHFPROD-7805: Fixing instance count in modeling tile 

rjrudin:
### Description

This applies when a project uses a different technique for identifying entity instance documents; getPrimaryEntityTypes.sjs now supports that technique via the build-entity-query.xqy extension point. 

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/models/getPrimaryEntityTypes.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/models/getPrimaryEntityTypes.sjs
@@ -18,6 +18,7 @@
 xdmp.securityAssert("http://marklogic.com/data-hub/privileges/read-entity-model", "execute");
 
 const entityLib = require("/data-hub/5/impl/entity-lib.sjs");
+const ext = require("/data-hub/public/extensions/entity/build-entity-query.xqy");
 
 let modelResponseArr = fn.collection(entityLib.getModelCollection()).toArray().map(model => {
   model = model.toObject();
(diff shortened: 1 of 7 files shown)


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/models/getPrimaryEntityTypes.sjs
```diff
@@ -26,7 +27,7 @@ let modelResponseArr = fn.collection(entityLib.getModelCollection()).toArray().m
   const response = {
     entityName: entityName,
     entityTypeId: entityLib.getEntityTypeId(model, entityName),
-    entityInstanceCount: cts.estimate(cts.collectionQuery(entityName)),
+    entityInstanceCount: cts.estimate(ext.buildEntityQuery(entityName)),
```
rjrudin, line 30: @rahulvudutala  Any idea if there are other places in the codebase where DHF is assuming that it can do a collection query on the entityName?

ryanjdew: @rjrudin There is mastering but that will, unfortunately, be more difficult to patch due to it being in XQuery.

ryanjdew (approved):

srinathgit (approved):

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/models/getPrimaryEntityTypes.sjs
rjrudin: Yeah that's fine for now - I should have qualified my question by asking @rahulvudutala  if there was another Explore-related place.

PR #6052, opened 2021-08-31
DHFPROD-7574: Create relationship via Add relationship button

briantang:
### Description

- Can now create relationships via dragging edges between nodes or clicking on a node with edit mode
- Demo here: https://drive.google.com/file/d/1-q9HPD2c7YoJ1I8GYNbMm20IEn-PtbGR/view
- UX approved by Jamie

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
- [x] Run UI tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/e2e/cypress/integration/modeling/reader.spec.tsx
+++ b/marklogic-data-hub-central/ui/e2e/cypress/integration/modeling/reader.spec.tsx
@@ -7,11 +7,13 @@ import {
   graphView,
   graphViewSidePanel,
   propertyModal,
-  propertyTable
+  propertyTable,
+  relationshipModal
 } from "../../support/components/model/index";
 import {confirmationModal, toolbar} from "../../support/components/common/index";
 import {Application} from "../../support/application.config";
 import LoginPage from "../../support/pages/login";
+import graphVis from "../../support/components/model/graph-vis";
 import "cypress-wait-until";
 
 describe("Entity Modeling: Reader Role", () => {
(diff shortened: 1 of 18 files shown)


xnikhil08 (dismissed):
Looks good.

briantang:
Briefly added some missing styling/functionality across graph view for when no edit permissions are present. Ready for re-review

xnikhil08 (approved):

briantang:
@wooldridge The asterick was requested by Jamie and Matthew to show that the field is required. It is to be consistent with the relationship field placeholder which also has an asterick towards the end because it is a required field.

wooldridge:
> @wooldridge The asterick was requested by Jamie and Matthew to show that the field is required. It is to be consistent with the relationship field placeholder which also has an asterick towards the end because it is a required field.

If we have asterisked content, I would expect that asterisk to reference a legend (or whatever you want to call it) describing why the asterisk is there. I know an asterisk is common for required form fields, perhaps to the point where it is generally understood that it means, but in this case we're not dealing with a standard form and I was confused looking for what the asterisk meant.

wooldridge (changes requested):
Looks like if I don't select a Join Property I get an error in the browser? Is that being handled separately?

https://drive.google.com/file/d/1FpL2g94H6TcVmooT41582DFkn9_6eYLc/view

briantang:
@wooldridge Correct, there is no BE support for that yet, so it will be handled with DHFPROD-7575.

briantang:
Added footnote for asterick clarification, approved by UX.

wooldridge (approved):

Sanjeevani19 (approved):

PR #6073, opened 2021-09-07
DHFPROD-7528: Persist entity node coordinates for graph

wooldridge:
### Description

Nodes have physics applied initially to determine initial coordinates. 

(Position persistence during and immediately after stabilization may be buggy and can use some review by the team. For example: Select a node during stabilization or immediately after stabilization.)

Clicking and dragging a node updates the coordinates for the node in the local state and in the database (user can leave graph view and return and the position is retained).

Position changes due to panning the entire graph set and zooming are not supported yet and have been described in bug tickets: DHFPROD-7838, DHFPROD-7839

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
- [x] Run UI tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/e2e/cypress/integration/modeling/graphValidations.spec.tsx
+++ b/marklogic-data-hub-central/ui/e2e/cypress/integration/modeling/graphValidations.spec.tsx
@@ -161,4 +161,50 @@ describe("Graph Validations", () => {
       graphViewSidePanel.getDeleteIcon("Person").should("be.visible");
     });
   });
+
+  it("can select all available nodes in graph view, locations persist", () => {
+    let ids = ["BabyRegistry", "Client", "Customer", "Order", "Person"];
+    let savedCoords: any = {};
+    modelPage.selectView("project-diagram");
+
+    // Select each entity node using retrieved coords
+    ids.forEach(id => {
+      graphVis.getPositionsOfNodes(id).then((nodePositions: any) => {
+        let coords: any = nodePositions[id];
+        savedCoords[id] = {x: coords.x, y: coords.y};
+        graphVis.getGraphVisCanvas().click(coords.x, coords.y);
+      });
+      // Verify entity is shown in side panel and close
+      graphViewSidePanel.getSelectedEntityHeading(id).should("be.visible");
+      graphViewSidePanel.closeSidePanel();
+    });
+
+    // Exit graph view and return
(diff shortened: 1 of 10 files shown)


marklogic-data-hub-central/ui/e2e/cypress/integration/modeling/graphValidations.spec.tsx
```diff
@@ -161,4 +161,49 @@ describe("Graph Validations", () => {
...
+    });
+
+    // Exit graph view and return
+    cy.waitUntil(() => toolbar.getCurateToolbarIcon()).click();
+    cy.waitUntil(() => toolbar.getModelToolbarIcon()).click();
+    modelPage.selectView("project-diagram");
+
+    // Select entity nodes using previously saved coords
+    ids.forEach(id => {
+      graphVis.getPositionsOfNodes(id).then(() => {
```
xnikhil08, line 189: Look like we are using the previously stored positions only instead of positions from output of getPositionsOfNodes(). In this case we can avoid firing `graphVis.getPositionsOfNodes(id)` and instead, let's place the click action (`graphVis.getGraphVisCanvas().click(savedCoords[id].x, savedCoords[id].y)`) directly under the forEach loop.

xnikhil08 (changes requested):
Looks good. Just added a comment in the cypress test case. 
Other than that, we can handle the multiple server requests for saving coordinates, in a separate ticket.

marklogic-data-hub-central/ui/e2e/cypress/integration/modeling/graphValidations.spec.tsx
wooldridge: @xnikhil08 and I looked into this fix and it's not straightforward since removing that async wrapper in the second forEach loops breaks the async dependency from the first loop (savedCoords is not saved by the time the second loop is executed). We agreed to keep it as-is for now, I will add a TODO note.

xnikhil08 (approved):

briantang (dismissed):

wooldridge:
E2E test failures appear to be due to graph nodes not having time to stabilize and therefore overlapping. Added initial test that pauses to allow for stabilization in graph.spec.

xnikhil08 (approved):

briantang (approved):

PR #6094, opened 2021-09-13
DHFPROD-7117: Add endpoint for provenance graph

ryanjdew:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
- [x] Run UI tests
  

- ##### Reviewer:

- [x] Reviewed Tests


AkshayBajajML (approved):

marklogic-data-hub-central/src/main/java/com/marklogic/hub/central/controllers/ProvenanceController.java
```diff
@@ -0,0 +1,70 @@
...
+public class ProvenanceController extends BaseController {
+    /**
+     * Returns the provenance graph for a document.
+     *
+     * @return
+     */
+    @RequestMapping(value = "/getProvenanceGraph", method = RequestMethod.GET)
+    @ResponseBody
+    @ApiOperation(value = "Get the provenance data for a document in graph form", response = ProvenanceGraph.class)
+    public ResponseEntity<JsonNode> getProvenanceGraph(@RequestParam String documentURI) {
```
aniamy, line 28: Any reason why we are not pssing it as PathVariable? In most endpoints I saw we usually pass the "id"/"name" as part of the path.
I don't know the characters a documentURI can have. But to include dot you can do /{documentURI:.+}
Reference: https://www.baeldung.com/spring-mvc-pathvariable-dot

ryanjdew: The documentURI can be anything that complies with the URI standard which includes protocols, forward slashes, etc. I think it makes more sense, in this case, to have it be a request parameter to avoid potential issues with encoding and parsing the value.

aniamy (approved):

PR #6142, opened 2021-09-27
DHFPROD-7907: Viewing certain entities in Explore breaks UI

federicoRufrancosMarklogic:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [X] Run UI tests
  

- ##### Reviewer:

- [ ] Reviewed Tests


wooldridge (approved):

wooldridge:
@briantang FYI, this is working for me. No compile issues.

briantang:
@federicoRufrancosMarklogic @wooldridge Ah, looks like the tests still don't like ML Design System even if it's just this one instance.
Screen Shot 2021-09-28 at 6 33 27 PM

federicoRufrancosMarklogic:
> @federicoRufrancosMarklogic @wooldridge Ah, looks like the tests still don't like ML Design System even if it's just this one instance. Screen Shot 2021-09-28 at 6 33 27 PM

I see the error on my local environment, but I don't follow the source of the error. I switched to develop if I could reproduce and it's working fine there.
I'm guessing that maybe the error comes from the design-system library, right?

yubiselmarklogic (approved):

xnikhil08 (approved):
Works for me.

wooldridge:
@federicoRufrancosMarklogic, I discussed this with @briantang and we decided we should not worry about fixing this right now unless it is blocking the replacement work. (Eventually we will need to fix this as part of the Table component replacement, but that will involve using a new component altogether.) So we can close this PR and move the Jira ticket back to the Backlog.

PR #6144, opened 2021-09-27
DHFPROD-7947: Making Detail view its own tile

awanczowski:
### Description

The detail view shared the same options as the explore tile view. This caused issues with CSS that was targeting the Explore view to hide contents on the detail page.  This change set addresses this by making a detail view its own TileView.

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
- [x] Run UI tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/src/config/tiles.config.ts
+++ b/marklogic-data-hub-central/ui/src/config/tiles.config.ts
@@ -1,6 +1,6 @@
 import { faLongArrowAltRight, faCube, faCubes, faObjectUngroup, faProjectDiagram } from "@fortawesome/free-solid-svg-icons";
 
-export type TileId =  'load' | 'model' | 'curate' | 'run' | 'explore' | 'monitor';;
+export type TileId =  'load' | 'model' | 'curate' | 'run' | 'explore' | 'monitor' | 'detail';
 export type IconType = 'fa' | 'custom';
 export type ControlType = 'menu' | 'newTab' | 'maximize' | 'minimize' | 'close';
 
@@ -13,7 +13,8 @@ interface TileItem {
     border: string;
     controlColor: string;
     controls: ControlType[];
-    intro: string;
+    intro?: string;
+    toolbar: boolean;
 }
 
 const tiles: Record<TileId, TileItem>  = {
@@ -27,6 +28,7 @@ const tiles: Record<TileId, TileItem>  = {
         controlColor: '#777',
         controls: ['close'],
         intro: 'Create and configure steps that ingest raw data from multiple sources.',
+        toolbar: true
     },
     model: {
         title: 'Model',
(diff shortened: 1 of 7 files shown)


AkshayBajajML:
This PR is ready for review

xnikhil08 (changes requested):
@awanczowski - I can see a new Detail tile option in the tile view toolbar. Let's get rid of it, since the Detail page is explorer related only. Please see the screencast below.
https://drive.google.com/file/d/1hnzMgN0imm6YfFQaDp2kUN_l5S0BrIrs/view

awanczowski:
@xnikhil08 I added an extra config to hide tiles icons the toolbar and added a descriptor to the pane title "Explore - Details"

marklogic-data-hub-central/ui/src/config/tiles.config.ts
```diff
@@ -27,6 +28,7 @@ const tiles: Record<TileId, TileItem>  = {
         controlColor: '#777',
         controls: ['close'],
         intro: 'Create and configure steps that ingest raw data from multiple sources.',
+        toolbarHidden: false
```
wooldridge, line 31: Let's avoid a double negative and do:
```
toolbar: true (or false for the Details case)
```

marklogic-data-hub-central/ui/src/config/tiles.config.ts
```diff
@@ -82,6 +88,19 @@ const tiles: Record<TileId, TileItem>  = {
         controlColor: '#777',
         controls: ['menu', 'close'],
         intro: 'Monitor Flows and Steps',
+        toolbarHidden: false
+    },
+    detail: {
+        title: 'Explore - Details',
```
wooldridge, line 94: I think this PR should cover an implementation detail (we're handling the Details view as a separate tile definition) and not how content is presented in the UI. There's no problem with the current titles/organization and changing tile titles (if we wanted to do that) is a bigger question that should involve PM and UX. So let's keep this title as "Explore".

marklogic-data-hub-central/ui/src/config/tiles.config.ts
```diff
@@ -82,6 +88,19 @@ const tiles: Record<TileId, TileItem>  = {
...
+        iconType: 'custom',
+        icon: 'exploreIcon',
+        color: '#376F63',
+        bgColor: '#F4F6F8',
+        border: '#BFBFBF',
+        controlColor: '#777',
+        controls: ['menu', 'close'],
+        intro: 'Detail record view',
```
wooldridge, line 102: I don't see this actually displayed for Details so let's keep the intro value empty, "", to avoid confusion?

wooldridge (changes requested):

wooldridge (approved):
Looks good to me!

xnikhil08 (dismissed):

SameeraPriyathamTadikonda:
Missing ":" in the PR-title.

AkshayBajajML:
@awanczowski I see some of the e2e tests failing consistently.

wooldridge (approved):

briantang (approved):

PR #6164, opened 2021-10-04
DHFPROD-7979: (5.5) Allow ingest to write to DB other than STAGING or FINAL

ryanjdew:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-api/src/main/java/com/marklogic/hub/impl/HubClientImpl.java
+++ b/marklogic-data-hub-api/src/main/java/com/marklogic/hub/impl/HubClientImpl.java
@@ -25,6 +25,7 @@ import java.util.Map;
 
 public class HubClientImpl implements HubClient {
 
+    private final HubClientConfig hubClientConfig;
     private String username;
     private DatabaseClient stagingClient;
     private DatabaseClient finalClient;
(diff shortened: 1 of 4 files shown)


marklogic-data-hub-api/src/main/java/com/marklogic/hub/impl/HubClientImpl.java
```diff
@@ -58,6 +60,11 @@ public DatabaseClient getStagingClient() {
         return stagingClient;
     }
 
+    @Override
+    public DatabaseClient getStagingClient(String databaseName) {
```
aniamy, line 64: Shouldn't we do something similar to allow another database to be the final?
I recall in standup mention of a staging-test and final-test.

briantang (approved):
@ryanjdew This looks great! Fixes the customer issue perfectly.

marklogic-data-hub-api/src/main/java/com/marklogic/hub/impl/HubClientImpl.java
ryanjdew: @aniamy That is a good question. For performance reasons, we want to use an app server that is associated with a content database to avoid invoking across databases. So if the target database is FINAL, we want to use the FINAL app server, etc. We are only adding the ability to specify a database to the STAGING app server-client since functionally speaking there isn't a difference in performance or outcome between using the FINAL or STAGING app server when the target database isn't associated with either.

aniamy (approved):

PR #6262, opened 2021-11-04
DHFPROD-7948: Unable to deploy step definitions

AkshayBajajML:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- N/A Added Tests
- N/A Run UI tests
  

- ##### Reviewer:

- N/A Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/LoadUserArtifactsCommand.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/LoadUserArtifactsCommand.java
@@ -310,7 +310,7 @@ public class LoadUserArtifactsCommand extends AbstractCommand {
                 }
                 final String flowName = flow.get("name").asText();
                 logger.info(format("Loading flow with name '%s'", flowName));
-                service.setArtifact("flow", flowName, flow);
+                service.setArtifact("flow", flowName, flow, "");
             }
         }
     }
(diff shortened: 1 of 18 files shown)


ryanjdew (approved):

marklogic-data-hub/src/main/java/com/marklogic/hub/deploy/commands/LoadUserArtifactsCommand.java
```diff
@@ -322,19 +322,23 @@ private void loadStepDefinitions(HubClient hubClient) {
...
+                        File stepDefFile = new File(defDir, stepDefFileName);
+                        if (stepDefFile.exists()) {
+                            JsonNode stepDef = readArtifact(stepDefFile);
+                            if (!stepDef.has("name")) {
+                                throw new RuntimeException("Unable to load step definition from file: " + stepDefFile +
+                                    "; no 'name' property was found");
+                            }
+                            final String stepDefName = stepDef.get("name").asText();
+                            logger.info(format("Loading step definition with type '%s' and name '%s'", stepDefType, stepDefName));
+                            service.setArtifact("stepDefinition", stepDefName, stepDef, stepDefFileName.replace(".step.json",""));
```
aniamy, line 337: If I understood this right, this new parameter needs to be the parent folder. I have the following example:
..\step-definitions\custom\archiveDoc\afile.step.json
and it creates the following json in the database: /step-definitions/custom/**afile**/afile.step.json
Notice the double use of the step name. And it is used where the parent folder name should be.
Shouldn't we create a json named /step-definitions/custom/archiveDoc/afile.step.json ? Or we need to have the same name twice as a rule?

AkshayBajajML: As per my understanding afile directory can have multiple filenames may or may not be of same name as directory.
@ryanjdew what will be your views on that

ryanjdew: @aniamy This change allows for the situation where we have multiple files under the same directory. We don't generate step definitions that way, but in the scenario where a developer may create variations of a step definition configuration file we want to allow that.

ryanjdew (dismissed):

aniamy (dismissed):

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/artifacts/stepDefinition.sjs
```diff
@@ -48,11 +48,11 @@ function getArtifactNode(artifactName, artifactVersion) {
     return stepDef;
 }
 
-function getDirectory(artifactName, artifact) {
+function getDirectory(artifactName, artifact, artifactDirName) {
     let doc = getArtifactNode(artifactName, null);
     let dir = "/step-definitions/";
     if(!doc && artifact && artifactName) {
-        dir = dir + artifact.type.toLowerCase() + "/" + artifact.name +"/"
+        dir = dir + artifact.type.toLowerCase() + "/" + artifactDirName +"/"
```
ryanjdew, line 55: Try using: `dir = dir + artifact.type.toLowerCase() + "/" + (artifactDirName || artifact.name) +"/"`

AkshayBajajML:
@ryanjdew  @aniamy All tests are passing now!

ryanjdew (approved):

briantang (approved):

PR #6282, opened 2021-11-16
DHFPROD-8004: Query managing replacement

yubiselmarklogic:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
- [x] Run UI tests
  

- ##### Reviewer:

- [ ] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/src/components/queries/managing/edit-query-dialog/edit-query-dialog.tsx
+++ b/marklogic-data-hub-central/ui/src/components/queries/managing/edit-query-dialog/edit-query-dialog.tsx
@@ -1,6 +1,5 @@
 import React, {useState, useEffect, useContext} from "react";
-import {Modal} from "antd";
-import {Row, Col, Form, FormLabel} from "react-bootstrap";
+import {Row, Col, Modal, Form, FormLabel} from "react-bootstrap";
 import styles from "./edit-query-dialog.module.scss";
 import {UserContext} from "../../../../util/user-context";
 import {SearchContext} from "../../../../util/search-context";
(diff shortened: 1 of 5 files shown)


briantang (changes requested):
@yubiselmarklogic Looks good, but I think it is a little strange that there are two cancellation buttons ("No" & "Cancel"). This is on the nightly as well, but can we just choose one to keep with this PR? FYI @jbelonoj . And if we choose to keep "No" instead of "Cancel", do we follow the same theme of positioning both buttons in the center of the modal?

Screen Shot 2021-11-16 at 11 31 24 AM
s choose one to keep?
Screen Shot 2021-11-16 at 11 31 06 AM

jbelonoj:
> @yubiselmarklogic Looks good, but I think it is a little strange that there are two cancellation buttons ("No" & "Cancel"). This is on the nightly as well, but can we just choose one to keep with this PR? FYI @jbelonoj . And if we choose to keep "No" instead of "Cancel", do we follow the same theme of positioning both buttons in the center of the modal?

Thanks, @briantang. Yes, we should definitely have two buttons, No and Yes. They should be centered, just like on all the other Yes/No confirmations.

yubiselmarklogic:
@jbelonoj, @briantang  The format for yes/no dialog don't use title, in this case I should keep it?

jbelonoj:
@yubiselmarklogic I think we can keep the "Existing Query" title. We don't need the "Confirmation" title, let's remove it. Thanks!

yubiselmarklogic:
Close the pr to create new one when complete all modals related to query process in tasks 8004, 8005 and 8047.

PR #6294, opened 2021-11-19
DHFPROD-8091: Update TDE to create triple joins between entity instan…

prolonMarkLogic:
…ces and external instance references

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
- [x] Run UI tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/hub-entities.xqy
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/hub-entities.xqy
@@ -23,12 +23,20 @@ import module namespace es = "http://marklogic.com/entity-services"
 import module namespace ext = "http://marklogic.com/data-hub/extensions/entity"
   at "/data-hub/extensions/entity/post-process-search-options.xqy";
 
+
+import module namespace tg = "http://marklogic.com/data-hub/hub-entities" at "/data-hub/5/impl/template-generated.xqy";
+
+import module namespace sem = "http://marklogic.com/semantics" at "/MarkLogic/semantics.xqy";
+import module namespace functx = "http://www.functx.com" at "/MarkLogic/functx/functx-1.0.1-nodoc.xqy";
 declare namespace search = "http://marklogic.com/appservices/search";
 declare namespace tde = "http://marklogic.com/xdmp/tde";
 
+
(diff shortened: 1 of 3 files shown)


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/hub-entities.xqy
```diff
@@ -592,10 +599,931 @@ declare function hent:dump-tde($entities as json:array)
...
+                  <tde:triple>
+                    <tde:subject><tde:val>$subject-iri</tde:val></tde:subject>
+                    <tde:predicate><tde:val>sem:iri("http://www.w3.org/2000/01/rdf-schema#isDefinedBy")</tde:val></tde:predicate>
+                    <tde:object><tde:val>fn:base-uri(.)</tde:val></tde:object>
+                  </tde:triple>
+
+                {
+                 for $property-name in map:keys($properties)
+                    let $is-related-entity-type := map:contains( map:get($properties, $property-name), "relatedEntityType" )
+                    let $related-entity-type := map:get( map:get($properties, $property-name), "relatedEntityType" )
```
ryanjdew, line 849: @prolonMarkLogic Can you adjust to also look under the "items" property for "relatedEntityType" so we catch the scenario where the relationship is many to one? 

Something like:
`let $property-info := if (map:contains(map:get($properties, $property-name), "items")) then 
                                        map:get(map:get($properties, $property-name), "items")
                                    else
                                        map:get($properties, $property-name)
  let $is-related-entity-type := map:contains($property-info, "relatedEntityType" )
  let $related-entity-type := map:get($property-info, "relatedEntityType" )
`

ryanjdew (changes requested):

marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/impl/hub-entities.xqy
prolonMarkLogic: @ryanjdew @aniamy Hi I created another class to put all the tde template logic together, after that I put the changes that Ryan wrote me and I tested this with the babyRegistry entity of the reference-entity-model project and it worked correctly. Let me know what do you think about these changes. thanks

ryanjdew (approved):

aniamy (approved):

PR #6326, opened 2021-11-30
DHFPROD-8162: Clean up Ant components

lucianaabc-ml:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
- [x] Run UI tests
  

- ##### Reviewer:

- [x] Reviewed Tests


Sanjeevani19 (changes requested):
@lucianaabc-ml there are around 20 more files which are using antd components, we need to replace all of them

PR #6359, opened 2021-12-16
DHFPROD-7887: Remove Explorer landing page

Sanjeevani19:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
- [x] Run UI tests
  

- ##### Reviewer:

- [x] Reviewed Tests


marklogic-data-hub-central/ui/src/pages/Browse.tsx
```diff
@@ -321,27 +318,11 @@ const Browse: React.FC<Props> = ({location}) => {
...
-            start: parsedPreferences.start || 1,
-            pageNumber: parsedPreferences.pageNumber || 1,
-            pageLength: parsedPreferences.pageLength,
-            propertiesToDisplay: searchOptions.selectedTableProperties || [],
-            zeroState: parsedPreferences.zeroState,
-            sortOrder: parsedPreferences.sortOrder || [],
-            database: parsedPreferences.database
-          };
-          await setPageQueryOptions(options);
+        if (!parsedPreferences.zeroState) {
```
xnikhil08, line 325: I still see few zeroState specific conditions like this here. This condition seems irrelevant, now that there is no zeroState page. It would be better if we remove all such references of zeroState based variables and conditions.

xnikhil08 (changes requested):
Everything looks good, except few references to zeroState still available at few places.
Also, the search queries dropdown shouldn't be visible in the graph view, as its already available in the side panel now and we should use the same for all the views eventually. See the screenshot below.

image

Sanjeevani19:
> Everything looks good, except few references to zeroState still available at few places. Also, the search queries dropdown shouldn't be visible in the graph view, as its already available in the side panel now and we should use the same for all the views eventually. See the screenshot below.
> 
> image

@xnikhil08 removed the references to the zero state in the entire code base. Visibility of search queries dropdown is not working in develop either, Please create a separate bug to handle that.

xnikhil08 (approved):

briantang (approved):
Looks good!

xnikhil08 (approved):

PR #6378, opened 2021-12-22
DHFPROD-7756: Job-results-table-view

lucianaabc-ml:
DHFPROD-7756: Job-results-table

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
- [x] Run UI tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/e2e/cypress/support/pages/monitor.tsx
+++ b/marklogic-data-hub-central/ui/e2e/cypress/support/pages/monitor.tsx
@@ -1,7 +1,7 @@
 import "cypress-wait-until";
 class MonitorPage {
   getTableRows() {
-    return cy.get(".ant-table-row");
+    return cy.get(".hc-table_row");
   }
   waitForMonitorTableToLoad() {
     cy.waitUntil(() => this.getTableRows().should("have.length.gt", 0));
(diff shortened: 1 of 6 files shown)


MSBrignoni (dismissed):
Goof work @lucianaabc-ml, just some comments to work what can be done in this instance with this new table:

- The color of the header is different, and the hover i think too.
- Is possible with this table write the titles of the header in one line and not two?
- The pages could have an issue when 10 is chosen, is not showing 10 in the "combobox".

monitor table

mrodrigu-marklogic (commented):
One thing that I'm not sure why is the container padding there, the paginator should stick to the rightmost column in the worst case scenario that the scroll cannot scroll the table content. 

image

yubiselmarklogic (dismissed):

MSBrignoni (commented):
It's ok @lucianaabc-ml , just a little comment, depends of the new component, perhaps it's possible to add the hover only in columns with ordering like nightly?

image

yubiselmarklogic (approved):

MSBrignoni (approved):

mrodrigu-marklogic (approved):

PR #6413, opened 2022-01-10
DHFPROD-8193: TABLE-Structured property should expand

MSBrignoni:
### Description
Bug: TABLE-Structured property should expand when property is added and delete

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
- [x] Run UI tests
  

- ##### Reviewer:

- [x] Reviewed Tests

### DIFF
--- a/marklogic-data-hub-central/ui/e2e/cypress/integration/modeling/writerScenario2.spec.tsx
+++ b/marklogic-data-hub-central/ui/e2e/cypress/integration/modeling/writerScenario2.spec.tsx
@@ -75,7 +75,6 @@ describe("Entity Modeling: Writer Role", () => {
     propertyModal.getYesRadio("pii").click();
     //propertyModal.clickCheckbox('wildcard');
     propertyModal.getSubmitButton().click();
-    cy.waitUntil(() => propertyTable.getExpandIcon("address").click());
     propertyTable.getMultipleIcon("street").should("not.exist");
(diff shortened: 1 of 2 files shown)


MSBrignoni:
This line was causing the problems related to the nested table in specific cases, altering the key of the row and affecting that later the library could not handle the default expand when adding and deleting.

image

briantang (changes requested):
I think I see two tricky cases where the behavior is unexpected @MSBrignoni.

https://watch.screencastify.com/v/RctKJ4vM6FfWbKDS1Jjy

MSBrignoni:
> I think I see two tricky cases where the behavior is unexpected @MSBrignoni.
> 
> https://watch.screencastify.com/v/RctKJ4vM6FfWbKDS1Jjy

Good catch @briantang , i was looking the code and testing it, and it is coming from the [old ant table too](http://rh7v-10-dhe-2:8080/tiles/model), although the code has improved as ant, there are different inconsistencies in the "property-table.tsx". It seems to work good with shipping but not with others cases, example, if I create a new property in "billing", it expands "shipping" (and others situations), i think it is necessary restructure parts of this page, perhaps in another jira issue.

https://user-images.githubusercontent.com/89084298/149152484-293374d2-21ca-42ff-8bc6-1f6257153833.mp4

briantang (approved):

aniamy (approved):

PR #6430, opened 2022-01-14
DHFPROD-7258: Set namespaces for structured properties

ryanjdew:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
- [x] Run UI tests
  

- ##### Reviewer:

- [ ] Reviewed Tests


briantang (changes requested):
@ryanjdew Looks pretty good, there might just be some edge cases to handle on top of some RTL failures:

https://watch.screencastify.com/v/YEe3ambYcNHqe07mnw2q
1) I believe users should still be able to create structured properties without a namespace but a form error seems to be preventing that?
2) Cancelling the add structured type modal still inserts the type into the property modal and allows the user to create the property in a broken state. Comparing to the nightly, cancelling should insert a placeholder value and clicking submit should throw a form error.


Let me know if you need any assistance with this or the RTL tests.

aniamy:
@ryanjdew Adding to Brian: We should do a trim to the values. I was able to create a Structured type without prefix (I'm not sure if I added a space by accident). And re-resting it I could add spaces and it would get validated.

In a general comment: Do you know if we'll be able to edit it later? In case of mistyping something. In the story I noticed we have a screenshot where "Namespace URI" and "Prefix" also appear on the first modal (Add Property one) Is that screen related to the A/C of this story?

AkshayBajajML (changes requested):
Add New Structured Property Type Modal remains open when we enter values in all field including prefix and clicking on save. It closes when we not add any value to prefix

Sanjeevani19 (commented):
@ryanjdew looks like couple of RTL tests failing are related to this PR, Please look into it https://jenkins.marklogic.com/blue/organizations/jenkins/Datahub_CI/detail/PR-6430/3/tests

briantang:
@aniamy According to the [JIRA ticket](https://project.marklogic.com/jira/browse/DHFPROD-7258?focusedCommentId=199668&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-199668), looks like the screenshot you're referencing was a mistake and the PM clarified that the options should be added in the structured modal, which is what Ryan did so it is correct. 

Regarding the editing functionality, there was no plan for that and it would need new mockups and a separate user story. I left a comment referring to that in the ticket.

briantang:
@ryanjdew I think I'm still seeing the same issues all-across: https://watch.screencastify.com/v/u734ll3YRbO9LEi3bYOG

briantang (changes requested):
Working well now. Can we also add this line to the "Add" Button props so we get a spinner on the button until the request completes?

`      loading={!successfulSubmit}
` 
NOTE: I got this to work but had to initialize the successfulSubmit state to "true" and also only set it to false inside the onSubmit function on line 105.

aniamy:
@ryanjdew I think that Brian's request affected the "Add" button to appear spinning from the beginning. Now I'm not able to click it since as soon as I open the modal it is disabled and with the spinner :(
Maybe change the flag to loading while submitting is true?

Sanjeevani19 (commented):
@ryanjdew Looks like the UI tests are failing on the latest jenkins build https://jenkins.marklogic.com/blue/organizations/jenkins/Datahub_CI/detail/PR-6430/11/tests, Could you please look into it

marklogic-data-hub-central/ui/src/components/modeling/structured-type-modal/structured-type-modal.tsx
```diff
@@ -88,13 +146,15 @@ A structured type cannot use the same name as an existing structured type.</span
       type="primary"
       htmlType="submit"
       size="default"
+      loading={!isLoading}
```
aniamy, line 149: This should be loading={isLoading}

aniamy:
@ryanjdew See comment about the flag.
Another minor comment is that we are still hitting the backend when namespace or prefix is missing. (Story says we require none or both, so we can add a simple pre-check for both filled before calling the backend.)

aniamy (dismissed):

briantang (approved):

AkshayBajajML (changes requested):
There are validation error when user not mention uri and prefix. 
Even though it gets saved, modal does close as validation warning comes
Screen Shot 2022-01-18 at 2 53 40 PM
.

briantang:
> There are validation error when user not mention uri and prefix. Even though it gets saved, modal does close as validation warning comes Screen Shot 2022-01-18 at 2 53 40 PM .

I'll take a look. Thanks @AkshayBajajML

AkshayBajajML (approved):
Looks Good!

briantang (dismissed):

briantang (approved):
Tests pass

aniamy (approved):

PR #6527, opened 2022-02-17
DHFPROD-7876: Default views for first-time user experience

AkshayBajajML:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests  

- ##### Reviewer:

- [ ] Reviewed Tests
- [x] Added to Release Wiki

### DIFF
--- a/marklogic-data-hub-central/ui/e2e/cypress/integration/curation/curate/mappingWithCustomHeader.spec.tsx
+++ b/marklogic-data-hub-central/ui/e2e/cypress/integration/curation/curate/mappingWithCustomHeader.spec.tsx
@@ -225,6 +225,8 @@ describe("Create and verify load steps, map step and flows with a custom header"
     cy.waitUntil(() => runPage.getFlowName(flowName).should("be.visible"));
     cy.verifyStepRunResult("success", "Mapping", mapStep);
     runPage.explorerLink().click();
+    browsePage.getTableView().click();
+    browsePage.waitForSpinnerToDisappear();
     browsePage.getFirstTableViewInstanceIcon().should("be.visible").click({force: true});
     detailPage.getDocumentSource().should("contain", "backup-ABC123");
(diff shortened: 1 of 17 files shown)


AkshayBajajML:
Made updates to 5.7 release wiki

briantang (dismissed):
1) Default view with instances in final - sidebar + view defaults me to final instances
2) Deleted entity instances in final - sidebar + view defaults me to staging instances
3) Deleted entity instances in staging - sidebar + view defaults me to All Data documents

xnikhil08 (approved):

aniamy:
@xnikhil08 For this A/C
(Given) No entity instances exist (just raw data),
(Then) by default, the Card view is selected, the All Data toggle is selected, and the final toggle is selected.

In the scenario were I only used the load step the A/C is happening. I see "All data" and "Final". But since my data is in staging shouldn't it also check this case and automatically should Staging?
For a second I did not understand why my load step data wasn't showing.

aniamy:
(And given) given they exist in the staging database
Pre-setting the data:
1. I loaded data.
2. I ran the Map step with target database Staging.
3. I logged out and cleaned my local storage
First time I logging as a first time experience:
1 logged in as a first time experience and went to explore.
2. Entities is selected. Final is selected but switches to Staging.
3. No data is shown.
Expected: Data from pre-setting should be shown.
Graph search call seems ok but backend is returning no data (To check if the issue is payload or backend)
There is no search call pointing to staging so when I switch to table/snippet view I see no data. (What was gotten from the final call probably)

If I stay in table view, move to another tile and come back, it looks like it tries to show data, but doesn't look good.

briantang (dismissed):
Just pushed a conflict resolution for @AkshayBajajML. Tests passed on latest run, PR should be good to go now.

xnikhil08 (approved):

aniamy (dismissed):

PR #6549, opened 2022-02-25
DHFPROD-8506: In explore, right side panel must be closed on changes

xnikhil08:
### Description
Also contains the fix for the bug DHFPROD-8459, 8563

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [N/A] Added Tests  

- ##### Reviewer:

- [N/A] Reviewed Tests
- [N/A] Added to Release Wiki

### DIFF
--- a/marklogic-data-hub-central/ui/src/pages/Browse.tsx
+++ b/marklogic-data-hub-central/ui/src/pages/Browse.tsx
@@ -53,7 +53,9 @@ const Browse: React.FC<Props> = ({location}) => {
     setDatabase,
     setEntityDefinitionsArray,
     clearAllGreyFacets,
-    setEntityTypeIds
+    setEntityTypeIds,
+    savedNode,
+    setSavedNode
   } = useContext(SearchContext);
   const searchBarRef = useRef<HTMLDivElement>(null);
   const authorityService = useContext(AuthoritiesContext);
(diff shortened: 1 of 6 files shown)


MSBrignoni (dismissed):

yubiselmarklogic (approved):

xnikhil08:
Fixed the failing tests.

briantang (approved):

xnikhil08:
> @xnikhil08 Looking at the code I don't see when we call this new function when you apply a facet or a saved query or changing the entities from the dropdown. Are those actions also closing the entity specific panel?

@aniamy - The bug DHFPROD-8459 is about closing the entity specific side panel when user tries to navigate away from the graph view, which is happening now. However, when user is on the graph view only, then in order to change the entities from the dropdown, or saved queries, user will have to close the entity specific side panel anyway since only one of the side panels (either main or entity specific) can remain open at a time.

marklogic-data-hub-central/ui/src/pages/Browse.tsx
```diff
@@ -352,6 +353,7 @@ const Browse: React.FC<Props> = ({location}) => {
     }
     return () => {
       setGraphSearchData({});
+      setSavedNode(undefined);
```
aniamy, line 356: Can we move this line to line 173? It makes more sense to only apply this when we are in graph view and we have new graphSearchData.
In this line it is being called eveytime there is a change in the dependencies regardless if we are in graph or not.

xnikhil08: Placing it after the graphSearch endpoint might result in the side panel disappearing after few seconds if the dataset is huge. Added it in line 354 instead, to ensure it will be called only when needed.

aniamy (approved):

briantang (approved):

aniamy:
@xnikhil08 Looks like the existing test failure could be related to this changes. Can you check it?

xnikhil08:
> @xnikhil08 Looks like the existing test failure could be related to this changes. Can you check it?

@aniamy - Fixed it. Another review ?

briantang (approved):

MSBrignoni (dismissed):

xnikhil08:
Added the fix for the bug DHFPROD-8563 as well.

xnikhil08:
Updated the limit to 100 here.

briantang (approved):
Looks good, also seeing results count updating properly with node expansion.

briantang (approved):

Sanjeevani19 (approved):

PR #6568, opened 2022-03-03
DHFPROD-8437: FixCreatedByStepTest Core Failure

prolonMarkLogic:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [n/a] Code passes ESLint tests
- [x] Added Tests  

- ##### Reviewer:

- [x] Reviewed Tests
- [n/a] Added to Release Wiki

### DIFF
--- a/marklogic-data-hub/src/test/java/com/marklogic/hub/flow/FixCreatedByStepTest.java
+++ b/marklogic-data-hub/src/test/java/com/marklogic/hub/flow/FixCreatedByStepTest.java
@@ -68,11 +68,11 @@ public class FixCreatedByStepTest extends AbstractHubCoreTest {
     void queryStep() {
         ReferenceModelProject project = installReferenceModelProject();
 
-        // Creating 100 records for a very mild performance test - loading and process takes under a second still, but
+        // Creating 50 records for a very mild performance test - loading and process takes under a second still, but
         // it at least causes the QueryBatcher in the implementation to do some multi-threading work.
         // Can crank this up for manual performance tests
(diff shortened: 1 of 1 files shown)


marklogic-data-hub/src/test/java/com/marklogic/hub/flow/FixCreatedByStepTest.java
```diff
@@ -179,8 +179,9 @@ private void verifyCreatedByStepIsFixed(String expectedValue) {
     private void addAnotherJobIdToSomeDocuments() {
         for (int i = 1; i <= 10; i++) {
             String script = "declareUpdate(); const uri = '/echo/customer" + i + ".json'; " +
+                "if (fn.docAvailable(uri)) {" +
```
aniamy, line 182: Do we know why some might not be available?

prolonMarkLogic: All of them should be available because this test added them a some lines before,  but the thing is that one timeout is happening sometimes, for this reason I also reduced the number of records. To check if with these changes solve the problem.

ryanjdew: The check for docAvailable should also give us better messaging if the test fails so we can see if the counts are off and by how much they are off in the assertions.

ryanjdew (approved):

aniamy (approved):

PR #6583, opened 2022-03-09
DHFPROD-8502: Image Gallery 

iveyMarklogic:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [ ] Reviewed Tests
- [ ] Added to Release Wiki

### DIFF
--- a/marklogic-data-hub-central/ui-custom/src/components/ImageGallery/ImageGallery.scss
+++ b/marklogic-data-hub-central/ui-custom/src/components/ImageGallery/ImageGallery.scss
@@ -1,0 +1,20 @@
+.ImageGallery {
+    display: flex;
+    justify-content: center;
+    .carouselContainer {
+        width: 150px;
+        height: 150px;
+        img {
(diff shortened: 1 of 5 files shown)


marklogic-data-hub-central/ui-custom/src/components/ImageGallery/ImageGallery.scss
```diff
@@ -0,0 +1,20 @@
+.ImageGallery {
+    display: flex;
+    justify-content: center;
+    .carouselContainer {
+        width: 150px;
```
wooldridge, line 5: @iveyMarklogic, how does one change the size of the image carousel? For example, what if I wanted to make it twice as big? It would be good to allow this via config properties, how can we do this?

When I adjust this CSS, I don't see the size change.

wooldridge (changes requested):

PR #6607, opened 2022-03-17
DHFPROD-8645: BE - Defining related concepts in entity model

ryanjdew:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests
- [x] Added to Release Wiki

### DIFF
--- a/examples/reference-entity-model/entities/Product.entity.json
+++ b/examples/reference-entity-model/entities/Product.entity.json
@@ -14,8 +14,19 @@
         "productName": {
           "datatype": "string",
           "collation": "http://marklogic.com/collation/codepoint"
+        },
+        "category": {
+          "datatype": "string",
+          "collation": "http://marklogic.com/collation/codepoint"
         }
-      }
+      },
+      "relatedConcepts": [
+        {
+          "context": "category",
+          "predicate": "isCategory",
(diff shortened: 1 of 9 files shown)


examples/reference-entity-model/entities/Product.entity.json
```diff
@@ -14,8 +14,18 @@
         "productName": {
           "datatype": "string",
           "collation": "http://marklogic.com/collation/codepoint"
+        },
+        "category": {
```
aniamy, line 18: How would be tell it apart as a Concept from the "basic" properties? ( those that are integer, boolean, string, etc)
For properties related to strucuted properties we do not put datatype:string
`"address": {
          "$ref": "#/definitions/Address"
        },`
I think we should follow the same pattern and do not say the type is string when the type would be "concept".
Maybe we can add "$ref": "#/concepts/Category" ? Or something along thoes lines? A new $concept attribute with the base http://www.example.com/Category?

Overall, I like how you thought about adding the Concept as part of a property to have a smooth flow.

ryanjdew (commented):
@prolonMarkLogic This PR looks good to go from what I'm seeing. Thanks!

aniamy (approved):

prolonMarkLogic (approved):

PR #6630, opened 2022-03-22
DHFPROD-8589: Implement Matchable baselineQuery

ryanjdew:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [ ] Reviewed Tests
- [ ] Added to Release Wiki

Updated wiki here: https://wiki.marklogic.com/display/ENGINEERING/Data+Hub+5.8.0

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/mastering/matching/matchable.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/mastering/matching/matchable.sjs
@@ -1,4 +1,5 @@
 'use strict';
+const httpUtils = require("/data-hub/5/impl/http-utils.sjs");
 
 /*
  * A class that encapsulates the configurable portions of teh matching process.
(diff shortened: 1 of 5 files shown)


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/mastering/matching/matchable.sjs
```diff
@@ -15,7 +16,41 @@ class Matchable {
...
+      if (this.matchStep.targetEntityType) {
+        const targetEntityType = this.matchStep.targetEntityType;
+        const tripleQuery = cts.tripleRangeQuery(null, sem.iri("http://www.w3.org/1999/02/22-rdf-syntax-ns#type"), sem.iri(targetEntityType))
+        // check for TDE for entity to be enabled
+        if (cts.exists(tripleQuery)) {
+          firstBaseline = tripleQuery;
+        } else {
+          firstBaseline = cts.collectionQuery(targetEntityType.substring(targetEntityType.lastIndexOf("/") + 1));
+        }
+      } else if (this.matchStep.collections && this.matchStep.collections.content) {
```
aniamy, line 30: Why are we doing this filter in the else?
My understanding is that if a collection or a query is included in the step we need add that filter to the target entity type filter.
Currently we have that filter (by collection or by query) and the interceptors as two separate items.
image

aniamy (approved):

prolonMarkLogic (approved):

PR #6658, opened 2022-03-28
DHFPROD-8633: expand and collapse

iveyMarklogic:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [ ] Reviewed Tests
- [ ] Added to Release Wiki

### DIFF
--- a/marklogic-data-hub-central/ui-custom/src/views/Detail.scss
+++ b/marklogic-data-hub-central/ui-custom/src/views/Detail.scss
@@ -37,9 +37,15 @@
                     padding: 2px 10px;
                 }
                 .expand {
+                    button {
+                        width: 130px;
+                        text-align: right;
+                    }
                     margin-right: 12px;
                     .label {
                         display: inline-block;
+                        width: 92px;
+                        text-align: center;
                         padding-right: 6px;
                     }
                     .icon {


marklogic-data-hub-central/ui-custom/src/views/Detail.tsx
```diff
@@ -98,8 +117,8 @@ const Detail: React.FC<Props> = (props) => {
             <button onClick={handleExpandClick}>
               <span className="label">Expand All</span>
```
wooldridge, line 118: Let's have the button label read "Collapse All" when clicking it will collapse all sections. It should read "Expand All" when clicking it will expand all sections. 

The button width SHOULD NOT change when the label toggles between the two label types.

marklogic-data-hub-central/ui-custom/src/views/Detail.tsx
```diff
@@ -35,6 +35,12 @@ const COMPONENTS = {
   ImageGalleryMulti: ImageGalleryMulti,
   Membership: Membership
 };
+const EXPANDIDS = {
```
wooldridge, line 38: When I leave the Detail page and then return to the Detail page, any collapsed sections become expanded again. Can we persist the state of the sections across visits to the Detail page? 

Also, for naming consistency, let's name the EXPANDID properties the same names as the detail properties in config.json. Let's also order them the same way:
```
  membership: true,
  info: true,
  relationships: true,
  imageGallery: true
```

wooldridge (changes requested):

iveyMarklogic:
@wooldridge I fixed all of the comments, can you review it again, please?

wooldridge (changes requested):
Can we add the following updates to avoid the button resizing on toggle:

https://github.com/iveyMarklogic/marklogic-data-hub/pull/2

wooldridge (approved):

PR #6687, opened 2022-03-31
DHFPROD-8624: BE support for dateTime constraint

rahulvudutala:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

### DIFF
--- a/marklogic-data-hub-central/src/main/resources/explore-data/ml-search-lib/search.sjs
+++ b/marklogic-data-hub-central/src/main/resources/explore-data/ml-search-lib/search.sjs
@@ -34,7 +34,43 @@ class Search {
     let facets = searchParams.selectedFacets;
     let keys = Object.keys(facets);
     let constraintArr = [];
-    keys.forEach(key => constraintArr.push(facets[key].map(value => key+':' + '"' + value + '"').join(" OR ")))
+
+    keys.forEach(key => {
+      if(Array.isArray(facets[key])) {
+        if(xdmp.castableAs("http://www.w3.org/2001/XMLSchema", "date", facets[key][0])) {
+          facets[key].map(date => xdmp.parseDateTime("[Y0001]-[M01]-[D01]", date))
+        }
+        constraintArr.push(facets[key].map(value => key+':' + '"' + value + '"').join(" OR "));
+      } else {
+        let minVal = facets[key]["min"];
+        let maxVal = facets[key]["max"];
+        let minValConstraint = "";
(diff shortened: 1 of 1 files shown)


AkshayBajajML (changes requested):
I can see this error now.
Screen Shot 2022-03-31 at 4 55 33 PM

AkshayBajajML:
Payload is : 
Screen Shot 2022-03-31 at 4 56 16 PM

AkshayBajajML (approved):
Looks Good!! It is fetching results for particular date range now.

AkshayBajajML:
@wooldridge It can be merged.

marklogic-data-hub-central/src/main/resources/explore-data/ml-search-lib/search.sjs
```diff
@@ -34,7 +34,18 @@ class Search {
     let facets = searchParams.selectedFacets;
     let keys = Object.keys(facets);
     let constraintArr = [];
-    keys.forEach(key => constraintArr.push(facets[key].map(value => key+':' + '"' + value + '"').join(" OR ")))
+    keys.forEach(key => {
+
+      if(xdmp.castableAs("http://www.w3.org/2001/XMLSchema", "date", facets[key][0])) {
```
wooldridge, line 39: @AkshayBajajML and @rahulvudutala, we need to consider this API update in the context of the other facet selection types we will need to support, e.g.:
Screen Shot 2022-04-01 at 5 25 00 PM
Will we also be able to pass in:
- A SINGLE date or datatime to support a single (non-range) date facet?
- A range of two NUMERIC values to support a numeric range facet?

@rahulvudutala, would it be straightforward to support these two other cases with this PR?

wooldridge (changes requested):

marklogic-data-hub-central/src/main/resources/explore-data/ml-search-lib/search.sjs
rahulvudutala: @wooldridge It works with date and dateTime formats. I don't think it will work with numeric formats. We are not passing the data types in the search request. Validation is done to check if the string is a date or dateTime.
If the year is only expected to be passed, then it needs to convert into either date or dateTime formats before the request is sent. 

For example, 
1) if 1974 - 1995 is selected, then it needs to be converted to 1974-01-01 and 1995-12-31
2) If only 1995 is selected, then it needs to be converted to 1995-01-01 and 1995-12-31

PR #6702, opened 2022-04-04
DHFPROD-8706: Reset start state on new searches

wooldridge:
### Description

While we reset the pageNumber state on new searches, we aren't resetting the start state. This causes a search error. See the ticket for a description of a test that gives an error.

This fix resets the start state on new searches just as we reset the pageNumber state.

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [ ] Reviewed Tests
- [ ] Added to Release Wiki


AkshayBajajML (approved):
Records can be seen now and API call does not return error.

PR #6717, opened 2022-04-08
DHFPROD-8625: Widget— Recent Searches, Management (Local Storage)

AkshayBajajML:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [ ] Reviewed Tests
- [ ] Added to Release Wiki


marklogic-data-hub-central/ui-custom/src/store/SearchContext.tsx
```diff
@@ -227,16 +234,55 @@ const SearchProvider: React.FC = ({children}) => {
...
+      let obj;
       if (newObj && newObj[userContext.userid]) {
+        obj = newObj[userContext.userid];
         newObj[userContext.userid][key] = dt;
       } else {
         newObj[userContext.userid] = {};
         newObj[userContext.userid][key] = dt;
       }
       localStorage.setItem("search", JSON.stringify(newObj));
+      //To handle deletion of records in local storage when over certain count and time limit
```
wooldridge, line 246: Should be "To handle deletion of searches

wooldridge (changes requested):
I have a few requests. Also, for clearing Recent Searches, we need to use "maxEntries" and "maxTime" settings under recentSearches.config.

We can't use the same settings as recentRecords since a customer may not have both of those widgets turned on.

marklogic-data-hub-central/ui-custom/src/components/RecentClear/RecentClear.scss
```diff
@@ -8,4 +8,14 @@
     border: ridge;
     padding: 0px 20px 0px 20px;
   }
+}
+
+.recent-clear-searches {
```
wooldridge, line 13: I don't know why, but the search Clear button ends up being 22px in height while the records Clear button ends up being 26px in height. Maybe give both buttons an explicit 26px height?

wooldridge (changes requested):

wooldridge (approved):

PR #6718, opened 2022-04-08
DHFPROD-7889: Customize entity icons and change color

yubiselmarklogic:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests
- [x] Added to Release Wiki


marklogic-data-hub-central/ui/src/components/explore/entity-type-display-settings-modal/entity-type-display-settings-modal.tsx
```diff
@@ -54,8 +65,35 @@ const EntityTypeDisplaySettingsModal: React.FC<Props> = ({isVisible, hubCentralC
     return options;
   };
 
-  const onEntityLabelChange = (row, e) => {
-    setEntityLabels({...entityLabels, [row.entityType]: e});
+  const onColumnValueChange = (row, e, column: TableColumns) => {
+    setExploreSettingsData(exploreSettingsData => {
+      let settingsData = exploreSettingsData.map(entityData => Object.assign({}, entityData));
```
xnikhil08, line 70: I would avoid iterating over the entire array every time we are changing Icon, color label or propertiesOnHover. 
Perhaps keeping them separately in either one object with structure similar to "hubCentralConfig.modeling.entities" or in separate objects for each column (like we are doing in few other areas in HC might help. That would really help when the database scales up to include more that 100-1500 entities.

xnikhil08: **For Example**: If you create a separate state object variable (may be "settingsData") to store all metadata, then your logic here can be reduced to something like below.

`let tempSettingData = settingsData;
    let exploreSettingRow = tempSettingData[row.entityType];
    switch (column) {
          case TableColumns.Color:
          exploreSettingRow.color = e.color.hex;
          break;
        case TableColumns.Icon:
          exploreSettingRow.icon = e;
          break;
        case TableColumns.EntityLabel:
          exploreSettingRow.label = e.value;
          break;
        }
        setSettingsData(tempSettingsData);`

And then we can use this "settingsData" object to be refresh table columns and also avoid iterating over the Array in handleSave by doing something like this 
"updatedPayload.modeling.entities = exploreSettingsData"
This is just a high level layout of the approach and of course there would be some other adjustments in the code.

marklogic-data-hub-central/ui/src/components/explore/entity-type-display-settings-modal/entity-type-display-settings-modal.tsx
```diff
@@ -73,6 +111,35 @@ const EntityTypeDisplaySettingsModal: React.FC<Props> = ({isVisible, hubCentralC
     return headerLabel;
   };
 
+  const handleSave = () => {
+    try {
+      let updatedPayload = {...hubCentralConfig};
+      for (const key in exploreSettingsData) {
```
xnikhil08, line 117: See if we can avoid this iteration as well, like i mentioned in another comment above.

xnikhil08 (changes requested):
Really good work here. Other than the suggestions mentioned within review comments, i noticed few bugs when i am trying to update the color or icons when the entity-specific side panel is opened for any entity and also when i move from Explore to Modeling screen.
See the following screencast for more details.
https://watch.screencastify.com/v/HN8SWvLDukezMWhBCg1v

marklogic-data-hub-central/ui/src/components/explore/entity-type-display-settings-modal/entity-type-color-picker/entity-type-color-picker.tsx
```diff
@@ -0,0 +1,54 @@
+import React, {useState, useEffect, useRef, useCallback} from "react";
```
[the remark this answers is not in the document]
yubiselmarklogic: Ok. I created a subtask DHFPROD-8758

AkshayBajajML (approved):
Looks Good!! Above mentioned issue by @xnikhil08 is also addressed.

briantang (approved):

xnikhil08 (approved):

PR #6728, opened 2022-04-12
DHFPROD-8648: Widget: Data Range Facet

AkshayBajajML:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [ ] Reviewed Tests
- [ ] Added to Release Wiki


wooldridge (changes requested):

marklogic-data-hub-central/ui-custom/src/components/Facets/Facets.tsx
```diff
@@ -4,8 +4,11 @@ import Form from "react-bootstrap/Form";
 import Table from "react-bootstrap/Table";
 import OverlayTrigger from "react-bootstrap/OverlayTrigger";
 import Tooltip from "react-bootstrap/Tooltip";
-import {InfoCircleFill, ChevronDoubleRight, ChevronDoubleLeft} from "react-bootstrap-icons";
+import {InfoCircleFill, ChevronDoubleRight, ChevronDoubleLeft, Calendar4, XLg} from "react-bootstrap-icons";
 import "./Facets.scss";
+import moment from "moment";
+import DateRangePicker from "react-bootstrap-daterangepicker";
```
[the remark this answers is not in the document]
AkshayBajajML: One is used for DateRangePicker and one for this css file(bootstrap-daterangepicker/daterangepicker.css)

marklogic-data-hub-central/ui-custom/src/components/Facets/Facets.tsx
```diff
@@ -167,13 +177,117 @@ const Facets: React.FC<Props> = (props) => {
...
+        let placeHolderDate = input.map(dateValue => moment(dateValue).format(dateFormat)).join(" ~ ");
+        let flag;
+        searchContext.facetStrings.map((facet => {
+            if(facet.split(":")[0] === facetData?.name) flag=true;
+        }))
+        if(!flag) setDatePickerValue([null,null]);
+        return placeHolderDate;
+    };
+
+    const createdOnDateWidget =  (facet) => <div><div className="title" data-testid={facet?.name}>
```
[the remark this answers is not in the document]
AkshayBajajML: Done!

AkshayBajajML:
> When I leave the Search view and then return, a date range that is saved in the app state is not displayed in the data range facet widget. I would expect those dates to be loaded just as any checked facets are loaded. Can we do that? Here's a screencast:
> 
> https://watch.screencastify.com/v/nvMbuhzrphBDoDGlW41P

Date is displayed now when user comes back from other screen.

wooldridge (changes requested):
I updated the example app to better test this feature:

https://github.com/AkshayBajajML/marklogic-data-hub/pull/6

This adds a second range index for lastUpdate so we can have two date range widgets at once. (You can merge this PR with your branch so we have better testing/example code.)

I see a few problems:

https://watch.screencastify.com/v/K3GLT4OJ3rt9vStRaJJv

- Start end state for multiple widgets isn't handled
- When I search for 2022-02-01 - 2022-02-04, I get 2022-01-31 showing in results
- The console is complaining about a React key missing

wooldridge:
FYI, though I haven't reviewed the latest push, @AkshayBajajML, I do notice that the person record with the createdOn datetime of:

2022-01-01T05:30:42Z

has its datetime value shown as 2021-12-31 in the search results UI. So there appears to be some timezone-related conversion going on. At 2022-01-01T05:30:42Z in our Pacific timezone, the date would still be 2021-12-31.

marklogic-data-hub-central/ui-custom/package.json
```diff
@@ -24,10 +25,12 @@
     "jsdoc": "^3.6.7",
     "lodash": "^4.17.21",
     "luxon": "^2.3.0",
+    "moment": "^2.29.2",
```
xnikhil08, line 28: @AkshayBajajML - If we have to import a new date time library then let's avoid moment js, since it is a legacy library with few security vulnerabilities. Instead, let's try using dayjs here, since that has replaced moment js in HC as well.
Its API is more or less similar to moment js, so it can be as simple as replacing "moment" with "dayjs" keyword in most cases, except where you try to mutate an existing date variable, since dayjs is immutable as apposed to moment. Hope it helps.
@wooldridge - Just FYI. Hope thats alright ?

xnikhil08 (changes requested):

marklogic-data-hub-central/ui-custom/package.json
wooldridge: @xnikhil08, yes, that sounds good. We should get rid of moment.js.

marklogic-data-hub-central/ui-custom/src/components/DateRangeFacet/DateRangeFacet.tsx
```diff
@@ -0,0 +1,144 @@
...
+import { SearchContext } from "../../store/SearchContext";
+import moment from "moment";
+import DateRangePicker from "react-bootstrap-daterangepicker";
+import "bootstrap-daterangepicker/daterangepicker.css";
+
+type Props = {
+  data?: any;
+};
+
+const DateRangeFacet: React.FC<Props> = (props) => {
```
wooldridge, line 15: When I apply the DateRangeFacet to a range, and then remove that range, the range still persist in the calendar when I open the calendar up.

I expect for the calendar range to be cleared when I clear the facet. Can we do that? See this screencast:

https://watch.screencastify.com/v/aAEB0918ZLD6WYxCXV8i

marklogic-data-hub-central/ui-custom/src/components/DateRangeFacet/DateRangeFacet.scss
```diff
@@ -0,0 +1,79 @@
+.title {
```
wooldridge, line 1: Since this title CSS rule is not scoped inside a class name for the widget, it is getting applied to all the instances of class name in "title" in the application. There are a lot of "title" instances. For example in the header ("Entity Viewer") and the Metrics widget ("New entities this week").

I think we can actually remove the CSS .title definition here (and the enclosed .facetInfo) since this is already defined inside the Facets.scss file.

wooldridge (changes requested):

marklogic-data-hub-central/ui-custom/package.json
AkshayBajajML: @xnikhil08 Thanks for pointing that out. I have replaced it with dayJs and it works fine.

marklogic-data-hub-central/ui-custom/src/components/DateRangeFacet/DateRangeFacet.scss
AkshayBajajML: Done!!

wooldridge (approved):

PR #6732, opened 2022-04-12
DHFPROD-8755: Support returning list of entities for application

rahulvudutala:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:


wooldridge (approved):
I'm seeing an entityType value returned in records for the endpoints that get records.

PR #6741, opened 2022-04-13
 DHFPROD-8729: Support for Sort facet

rahulvudutala:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:


wooldridge (changes requested):
I'm submitting the search payload with the sorting object:
```
{
  "start": 1,
  "pageLength": 10,
  "searchText": "",
  "entityTypeIds": [
    "person"
  ],
  "selectedFacets": {},
  "sort": {
    "entityType": "person",
    "property": "createdOn",
    "order": "descending"
  }
}
```
But I'm not seeing sorted results:
Screen Shot 2022-04-13 at 12 41 38 PM

wooldridge:
As described in chat, I need to test with a rebuilt .war on this branch, will try that next.

wooldridge (approved):
Looks good, I'm seeing support for ascending and descending sort for the Person entity when I hardcode it into the SearchContext query. I see that we'll need to add sort configuration for any new entities (e.g., Organization) when we add them.

PR #6749, opened 2022-04-18
DHFPROD-8744: Support multiple entities across app

wooldridge:
### Description

Add configurable support for multiple entities across app views. Add support to the entity-view-v2 project for both Person and Organization entities. Relevant views:

- Search results
- Detail view (and its widgets)
- Recently Visited widget in Dashboard view

Screen Shot 2022-04-18 at 10 13 50 AM

The key update in config.json is the addition of "entities" properties where configurations for search results and other entity-specific display of data are tied to each entity. For example:

https://github.com/wooldridge/marklogic-data-hub/blob/DHFPROD-8744-organizations-config-2/examples/entity-viewer-v2/entity-viewer-client/src/main/ml-modules/root/explore-data/ui-config/config.json#L308

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [ ] Reviewed Tests
- [ ] Added to Release Wiki


iveyMarklogic:
@wooldridge I think will be more clear if we create a global section for each entity instead to add configuration on every single section. This new grouping will allow us to be able to divide the configuration into different files.

**Global section**
image

wooldridge:
@iveyMarklogic, thanks for the comment. I refactored the config.json to use detail.entities[ENTITY-NAME] for Detail view configuration. Do you think that is sufficient for this PR?

iveyMarklogic:
@wooldridge I think is fine for this pr, but we can have the same approach in the results section.

current
image

expected
image

wooldridge:
> @wooldridge I think is fine for this pr, but we can have the same approach in the results section. We can handle this in another user story also
> 
> current image
> 
> expected image

Yes, we will be evolving this. The results section may be different as there are things like pagination and sorting that may make more sense being configured in one place globally. Thanks for the review!

PR #6758, opened 2022-04-20
DHFPROD-8762: e2eFixesPt3

csilva-ml:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests
- N/A Added to Release Wiki

### DIFF
--- a/marklogic-data-hub-central/ui/e2e/cypress/integration/curation/curate/relatedEntityMapping.spec.tsx
+++ b/marklogic-data-hub-central/ui/e2e/cypress/integration/curation/curate/relatedEntityMapping.spec.tsx
@@ -59,7 +59,7 @@ describe("Mapping", () => {
     propertyModal.getForeignKey("id").click();
     propertyModal.getSubmitButton().click();
     cy.waitForAsyncRequest();
-    cy.wait(500);
+    cy.wait(1000);
     //Save Changes
     cy.publishEntityModel();
     propertyTable.getForeignIcon("relatedTo").should("exist");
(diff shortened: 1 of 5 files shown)


marklogic-data-hub-central/ui/e2e/cypress/integration/curation/curate/relatedEntityMapping.spec.tsx
```diff
@@ -59,7 +59,7 @@ describe("Mapping", () => {
     propertyModal.getForeignKey("id").click();
     propertyModal.getSubmitButton().click();
     cy.waitForAsyncRequest();
-    cy.wait(500);
+    cy.wait(1000);
```
ngodugu-marklogic, line 62: @csilva-ml May be we can try setting {defaultCommandTimeout: sometime} at block level and see if that fixes these errors. May be that might be better than giving wait times explicitly. Please give it a try.

briantang (approved):

marklogic-data-hub-central/ui/e2e/cypress/integration/curation/curate/relatedEntityMapping.spec.tsx
csilva-ml: It's actually not a timeout problem, the problem it's that it tries to click the button too early, that's why there was a 500 wait, but for firefox it looks like it needs a bit more and so that's why I increased it to 1000. Let me know if it makes sense to you.

MSBrignoni (approved):

marklogic-data-hub-central/ui/e2e/cypress/integration/curation/curate/relatedEntityMapping.spec.tsx
ngodugu-marklogic: Please see if we can add a timeout for any other element to be visible before click. I am fine with this solution right now but increasing the wait is not an ideal way to fix it.

ngodugu-marklogic (approved):

PR #6773, opened 2022-04-21
DHFPROD-8630: and DHFPROD-8632: Metrics Widget Endpoint

rahulvudutala:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

### DIFF
--- a/marklogic-data-hub-central/src/main/resources/explore-data/ml-search-lib/search.sjs
+++ b/marklogic-data-hub-central/src/main/resources/explore-data/ml-search-lib/search.sjs
@@ -140,6 +140,15 @@ class Search {
     return recordCount;
   }
 
+  getMetrics(metricTypes) {
+    const metrics = {};
+    metricTypes.forEach(metricType => {
+        metrics[metricType["type"]] = Math.floor(Math.random() * (100000 - 1000 + 1)) + 1000;
+      }
+    );
+    return metrics;
+  }
+
   fixArrayIssue(jsonObject) {
     const keys = Object.keys(jsonObject);
     for(let key of keys) {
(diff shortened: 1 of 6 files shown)


marklogic-data-hub-central/src/main/resources/explore-data/ml-search-lib/search.sjs
```diff
@@ -140,6 +140,18 @@ class Search {
     return recordCount;
   }
 
+  getMetrics(metricTypes) {
```
wooldridge, line 143: It looks like you are returning the same data every time. For What's New, we want to allow the user to select different time ranges and see the values change:
Screen Shot 2022-04-22 at 10 51 36 AM
I think the best way to mock this is for you to return random numbers for all the values. To return numbers from 1000 to 100,000 in JavaScript:
```
Math.floor(Math.random() * (100000 - 1000 + 1)) + 1000;
```
Can we do that for all the values?

wooldridge, line 143: It looks like you are returning the same data every time. For What's New, we want to allow the user to select different time ranges and see the values change:
Screen Shot 2022-04-22 at 10 51 36 AM
I think the best way to mock this is for you to return random numbers for all the values. To return numbers from 1000 to 100,000 in JavaScript:
```
Math.floor(Math.random() * (100000 - 1000 + 1)) + 1000;
```
Can we do that for all the values?

wooldridge (changes requested):

wooldridge (approved):

wooldridge:
Looks good. Request:
```
[{type: "new", period: 1440}, {type: "changed", period: 1440}, {type: "submitted", period: 1440}]
```
Response:
```
{"new":75769,"changed":20978,"submitted":33903}
```

PR #6818, opened 2022-04-28
DHFPROD-8815: ranByStep missing in FINAL DB

prolonMarkLogic:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [n/a] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests
- [n/a] Added to Release Wiki


briantang (changes requested):
@prolonMarkLogic @rahulvudutala I still see the problem after rebuilding with this change. Can you take another look Pablo? Explore Data via a completed map step from the Run tile still ends up in an infinite loop.

prolonMarkLogic:
> 

This PR is for customers Pre 5.7.2 version and upgrading to 5.7.2, For customers coming from previous version of datahub, FinalDatabaseXmlFileUpgrader class will help deploying the datahubRanBySteps field and index.

on the other hand 
The issue is because, in ours example project there is already a final-database.xml file in src/main/ml-config/database-fields directory. If the file already exists in that directory, we will not override that file during hubinit or hubUpdate. 
We need to delete that src/main/ml-config/database-fields directory before running hubinit (on the reference-entity-model project for example).

@rahulvudutala @briantang  maybe we could have a meeting to talk about it.

rahulvudutala (approved):

AkshayBajajML (approved):

briantang (approved):

PR #6876, opened 2022-05-12
DHFPROD-8809: BE - Expand concept node to view everything related to it

prolonMarkLogic:
#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests
- [x] Added to Release Wiki

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/graph/nodeExpand.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/graph/nodeExpand.sjs
@@ -63,6 +63,22 @@ function handleByPredicate(entityName, predicateIRI) {
   });
   return graphUtils.getEntityNodes(entityTypeIRI, predicateIRI, relatedTypeIRIs, limit);
 }
+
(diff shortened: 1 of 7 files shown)


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/data-services/graph/nodeExpand.sjs
```diff
@@ -74,6 +90,10 @@ limit = limit || 100;
 let queryObj = JSON.parse(nodeInfo);
 const nodeToExpand = queryObj.parentIRI;
 
+let isConcept = false;
+if(queryObj.isConcept != null && queryObj.isConcept === "true"){
```
ryanjdew, line 94: This might be more of a question for the front-end, but is there a reason we're using the string "true" instead of the boolean value?

prolonMarkLogic: Good catch @ryanjdew ! I was talking with front team, and this should be a boolean as you mentioned. I already changed it, thanks!

ryanjdew (approved):

AkshayBajajML (approved):

PR #6881, opened 2022-05-16
DHFPROD-8894: Handle simple arrays in DataTableValue

wooldridge:
### Description

Handle the case where the original data looks like this:
```
<phone>917-910-6575</phone>
<phone>917-910-6576</phone>
```
The JSON results look like this:
```
['917-910-6575', '917-910-6576']
```
(No metadata.)

And the DataTableValue config looks like this:
```
"config": {
"id": "phone",
"title": "Phone Number",
"arrayPath": "person.phone",
"icon": "phone",
"width": "400px",
```
(In this case, value property is not needed because no metadata and we want to show the two strings.)

This case is tested here: https://github.com/wooldridge/marklogic-data-hub/blob/DHFPROD-8894/marklogic-data-hub-central/ui-custom/src/components/DataTableValue/DataTableValue.test.tsx#L53

#### Checklist: 
```diff
- Note: do not change the below
```

-  ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [ ] Reviewed Tests
- [ ] Added to Release Wiki

### DIFF
--- a/marklogic-data-hub-central/ui-custom/src/components/DataTableValue/DataTableValue.tsx
+++ b/marklogic-data-hub-central/ui-custom/src/components/DataTableValue/DataTableValue.tsx
@@ -99,7 +99,7 @@ const DataTableValue: React.FC<Props> = (props) => {
                             if (d[props?.config?.value]) {
                                 value = d[props.config.value];
                             } else {
-                                if (Array.isArray(d)) {
+                                if (Array.isArray(d) || !_.isObject(d)) {
                                     value = d
                                 }
                             }
--- a/examples/entity-viewer-v2/entity-viewer-client/src/main/ml-modules/root/explore-data/ui-config/detail.config.sjs
+++ b/examples/entity-viewer-v2/entity-viewer-client/src/main/ml-modules/root/explore-data/ui-config/detail.config.sjs
@@ -98,8 +98,7 @@ const detailConfig  = {
                   "config": {
                     "id": "phone",
                     "title": "Phone Number",
-                    "path": "person",
-                    "value": "phone",
+                    "arrayPath": "person.phone",
                     "icon": "phone",
                     "width": "400px",
                     "metadata": [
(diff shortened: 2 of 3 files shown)


wooldridge:
@iveyMarklogic, can you review since you are familiar with this widget?

marklogic-data-hub-central/ui-custom/src/components/DataTableValue/DataTableValue.tsx
```diff
@@ -99,9 +99,7 @@ const DataTableValue: React.FC<Props> = (props) => {
                             if (d[props?.config?.value]) {
                                 value = d[props.config.value];
                             } else {
-                                if (Array.isArray(d)) {
```
iveyMarklogic, line 102: @wooldridge This check is necessary because when the entity doesn't have the configured property, the "d" parameter has an object. The application fails because react doesn't allow you to display an object. I added the array check because in some examples we have the property as an array of strings like ["value 1"," value 2"].

iveyMarklogic: If you have another case, we can add others checks

iveyMarklogic (changes requested):

marklogic-data-hub-central/ui-custom/src/components/DataTableValue/DataTableValue.tsx
wooldridge: Two tests fail on the current branch:

DEBUG_PRINT_LIMIT=10000 npm t DataTableValue.test
Screen Shot 2022-05-16 at 1 07 38 PM

The following fixes it and addresses your comment, I think, I will check in for your review:
```
let value;
                            if (d[props?.config?.value]) {
                                value = d[props.config.value];
                            } else {
                                if (Array.isArray(d) || !_.isObject(d)) {
                                    value = d;
                                }
                            }
```

iveyMarklogic: @wooldridge  it looks great, we can go with this

iveyMarklogic (approved):

PR #6911, opened 2022-05-23
DHFPROD-8931: Fix proxying behavior for getting loginAddress

wooldridge:
Our app init strategy offers a CORS workaround (setupProxy.js) for the DEVELOPMENT environment (port 3000).

In PRODUCTION (port 8080), the loginAddress server will need to handle CORS issues if that server differs from the server that serves the React app. If the loginAddress in properties is on the same server as the app (e.g., http://localhost:8080/api/explore/login) then PRODUCTION will work without any CORS handling.

To test:

1. Development (3000). When loginAddress property in properties file is mock server at http://localhost:8888/api/explore/login, correct userid will be set. Change that value to something else (e.g., http://www.google.com) and the userid will be incorrect, but the initial auth should still work since we're not checking validity of userid value.
2. Production (8080). Will work if loginAddress is set to http://localhost:8808/api/explore/login. If set to server that differs from the server that serves the React, will fail due to CORS issues. 

#### Checklist: 
```diff
- Note: do not change the below
```

- ##### Owner:

- [ ] Added Tests
- [ ] Ran cypress tests on firefox locally
  

- ##### Reviewer:

- [ ] Reviewed Tests
- [ ] Added to Release Wiki

### DIFF
--- a/marklogic-data-hub-central/ui-custom/src/App.tsx
+++ b/marklogic-data-hub-central/ui-custom/src/App.tsx
@@ -16,8 +16,8 @@ const App: React.FC<Props> = (props) => {
 
   const userContext = useContext(UserContext);
 
-  if (!userContext.proxy) {
-    userContext.handleGetProxy();
+  if (!userContext.loginAddress) {
+    userContext.handleGetLoginAddress();
   }
 
   return (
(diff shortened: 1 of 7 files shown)


rahulvudutala (approved):

wooldridge:
Checked different permutations, worked as expected and without error.

PR #6916, opened 2022-05-26
DHFPROD-8787: Allow Custom Match Functions to Return Atomic Values

AkshayBajajML:
#### Checklist: 
```diff
- Note: do not change the below
```

- ##### Owner:

- [x] Added Tests
- N/A Ran cypress tests on firefox locally
  

- ##### Reviewer:

- [x] Reviewed Tests
- [x] Added to Release Wiki

### DIFF
--- a/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/mastering/matching/matchable.sjs
+++ b/marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/mastering/matching/matchable.sjs
@@ -282,7 +282,12 @@ class MatchRulesetDefinition {
       if (!query) {
         return null;
       }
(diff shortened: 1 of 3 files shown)


ryanjdew (changes requested):
Can we add a test for an array or sequence of atomic values? I suspect the .toString() call won't be sufficient in those cases. Sorry for not calling that out directly in the acceptance criteria.

AkshayBajajML:
> Can we add a test for an array or sequence of atomic values? I suspect the .toString() call won't be sufficient in those cases. Sorry for not calling that out directly in the acceptance criteria.

I have added array and sequence type interceptors and respective tests for it.

ryanjdew (approved):

PR #6932, opened 2022-06-06
DHFPROD-8869: Modify Explore Settings to Change hover properties (Str…

yubiselmarklogic:
…uctured properties only)

#### Checklist: 
```diff
- Note: do not change the below
```

- ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [ ] Reviewed Tests
- [ ] Added to Release Wiki


prolonMarkLogic (dismissed):

MSBrignoni (approved):

marklogic-data-hub-central/ui/src/components/explore/entity-type-display-settings-modal/entity-type-display-settings-modal.tsx
```diff
@@ -12,22 +12,20 @@ import {UserContext} from "@util/user-context";
...
   isVisible: boolean;
   toggleModal: (reloadData: boolean) => void;
   entityDefinitionsArray: any;
 };
-// ToDo: For use in DHFPROD-8869: Modify Explore Settings to Change hover properties (Structured properties only)
-// const DEFAULT_ENTITY_DEFINITION: Definition = {
-//   name: "",
-//   properties: []
-// };
+const DEFAULT_ENTITY_DEFINITION: Definition = {
```
xnikhil08, line 25: I see that we are defining DEFAULT_ENTITY_DEFINITION here as well as in the Side panel component. Let's define such default value variables in a config file (probably explore.config.ts under /config) and then just refer wherever needed.

marklogic-data-hub-central/ui/src/components/modeling/graph-view/side-panel/side-panel.tsx
```diff
@@ -32,27 +37,36 @@ type Props = {
   setNodeNeedRedraw: any;
 };
 
+const DEFAULT_ENTITY_DEFINITION: Definition = {
```
xnikhil08, line 40: Refer previous comment.

marklogic-data-hub-central/ui/src/components/explore/entity-type-display-settings-modal/entity-type-display-settings-modal.tsx
```diff
@@ -85,37 +84,35 @@ const EntityTypeDisplaySettingsModal: React.FC<Props> = ({isVisible, toggleModal
...
-  //     handleError(error);
-  //   }
-  // };
+  useEffect(() => {
+    getEntityModels();
+  }, []);
+
+  const getEntityModels = async () => {
+    try {
+      let response = await axios.get(`/api/models/primaryEntityTypes`);
```
xnikhil08, line 93: let's refer the "primaryEntityTypes" function (under src/api/queries), for the axios API call here. All API calls should be defined under /api folder only and just referred  in the components.

xnikhil08: If seems that you basically need the entityModel in this component, so, i think we are already making a call to "/api/models" in the Browse component, Not sure if you get the desired response there but would be good if we can verify it and see if that helps you. We can then avoid this additional call to get the Model.

marklogic-data-hub-central/ui/src/components/modeling/graph-view/side-panel/side-panel.tsx
```diff
@@ -224,6 +238,57 @@ const GraphViewSidePanel: React.FC<Props> = (props) => {
...
+      } catch (error) {
+        handleError(error);
+      }
+    })();
+    return () => {
+      loaded = false;
+    };
+  }, [props.hubCentralConfig]);
+
+  const getEntityModels = async () => {
```
xnikhil08, line 261: Same comment here. Even if we can't use the Model call in browse component but i think we can placer this getEntityModels functions at a common location and refer here, since it seems to be performing the same task as in the entity display settings modal component earlier.

xnikhil08 (changes requested):

yubiselmarklogic:
@xnikhil08  All comments were fixed

prolonMarkLogic (approved):

xnikhil08 (approved):

PR #6961, opened 2022-06-17
DHFPROD-7888: Add concept facets

xnikhil08:
#### Checklist: 
```diff
- Note: do not change the below
```

- ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests
- [x] Added to Release Wiki


xnikhil08:
Will add reviewers after a few e2e runs.

yubiselmarklogic (approved):

prolonMarkLogic:
Hi @xnikhil08 , I think that would be a good idea to change the behavior when we don't have related concepts to show, now we are showing an empty section, I think this should be like the related entities, and don't show if we don't have information there.
image

xnikhil08:
> Hi @xnikhil08 , I think that would be a good idea to change the behavior when we don't have related concepts to show, now we are showing an empty section, I think this should be like the related entities, and don't show if we don't have information there. image

@prolonMarkLogic  - Good catch. Fixed the issue. Another review ?

prolonMarkLogic (approved):

yubiselmarklogic (approved):

PR #6982, opened 2022-06-23
DHFPROD-9019: Display RecordRaw widget on Detail page OOTB

iveyMarklogic:
#### Checklist: 
```diff
- Note: do not change the below
```

- ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [ ] Reviewed Tests
- [ ] Added to Release Wiki

### DIFF
--- a/marklogic-data-hub-central/ui-custom/src/views/Detail.tsx
+++ b/marklogic-data-hub-central/ui-custom/src/views/Detail.tsx
@@ -24,6 +24,7 @@ import "./Detail.scss";
 import _ from "lodash";
 import {getValByConfig} from "../util/util";
 import RecordRaw from "../components/RecordRaw/RecordRaw";
+import ReactJson from "react-json-view";
 
 type Props = {};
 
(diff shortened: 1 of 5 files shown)


wooldridge:
After viewing a record Detail, the Recently Visited record feature breaks on the Dashboard. (Try drilling down to a record and then going back to the Dashboard.)

To fix, we can not show Recently Visited records when there is no configuration for them. This fixes, can you review?

RecentRecords.tsx line 86:
```
  const getRecent = () => {
    let res = props.data.map((recent, index) => {
      // Don't show if no configuration for entity type
      if (!props.config.entities[recent.entityType]) return;
      let titleValue = getValByConfig(recent, props.config.entities[recent.entityType].title, true);
      if (!titleValue) {
```

wooldridge (changes requested):
See also my comment about Recently Visited breaking.

wooldridge:
FYI, here is the app working (with false flag) with the baseball data after my updates above made: https://watch.screencastify.com/v/PIhNOlSHGgSIi2pH24bJ

marklogic-data-hub-central/ui-custom/src/views/Detail.tsx
```diff
@@ -180,128 +181,164 @@ const Detail: React.FC<Props> = (props) => {
...
+  };
+
+
+  return (
+
+    <div className="detail">
+
+      {config?.detail && !_.isEmpty(detailContext.detail) && !detailContext.loading ?
+        (
+          !config?.detail?.entities[entityType] ? getDetailNoEntity() : getDetailEntity
```
wooldridge, line 339: With the example project loaded and flag set to true, the Detail page now breaks: 

https://watch.screencastify.com/v/GdnrbUTtzUZbh5Fe52nO

I think the problem is here, make this fix (add missing parens):
```
!config?.detail?.entities[entityType] ? getDetailNoEntity() : getDetailEntity()
```

wooldridge (changes requested):

wooldridge (changes requested):
@iveyMarklogic, one more thing, can you rebase onto the current custom-search? I want to test that your changes work with my changes to the match snippet display.

wooldridge (approved):

PR #7089, opened 2022-07-27
DHFPROD-9039: New Concat widget, title reconfig

wooldridge:
### Description

Added a new Concat widget for displaying concatenated elements from the payload. 

Also refactored title configuration on search results, detail page, and recently viewed (which is where users specifically wanted to use the Concat widget).

Example configuration (concatenate given name and surname for search titles) is here: https://github.com/wooldridge/marklogic-data-hub/blob/DHFPROD-9039/examples/entity-viewer/src/main/ml-modules/root/explore-data/ui-config/search.config.sjs#L102

Content between concatenated elements can be added with "prefix" and "suffix" properties (which are already in our configuration patterns).

Concat supports regular values and also a "type" of "DateTime" which works similar to the "DateTime" widget. (See the unit tests for an example).

Tests pass, detailed documentation will come later.

#### Checklist: 
```diff
- Note: do not change the below
```

- ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests
- [x] Added to Release Wiki

### DIFF
--- a/examples/entity-viewer/src/main/ml-modules/root/explore-data/ui-config/dashboard.config.sjs
+++ b/examples/entity-viewer/src/main/ml-modules/root/explore-data/ui-config/dashboard.config.sjs
@@ -117,9 +117,23 @@ const dashboardConfig  = {
                         }
                         },
                         "title": {
-                        "id": "uri",
-                        "arrayPath": "person.nameGroup",
-                        "path": "fullname.value"
+                            "id": {
+                                "path": "uri"
+                            },
+                            "component": "Concat",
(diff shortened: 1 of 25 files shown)


iveyMarklogic (approved):

AkshayBajajML (changes requested):
Concat widget is working as expected. 
Just seeing this some CSS related issue:
Screen Shot 2022-07-27 at 4 27 40 PM

AkshayBajajML (approved):

PR #7141, opened 2022-08-12
DHFPROD-9216: e2eAuthFix

csilva-ml:
#### Checklist: 
```diff
- Note: do not change the below
```

- ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests
- N/A Added to Release Wiki

### DIFF
--- a/marklogic-data-hub-central/ui/e2e/cypress/integration/login/authorization.spec.tsx
+++ b/marklogic-data-hub-central/ui/e2e/cypress/integration/login/authorization.spec.tsx
@@ -169,10 +169,11 @@ describe("login", () => {
     loadPage.addToNewFlow(stepName).click("top", {force: true});
     runPage.newFlowModal().should("not.exist");
     loadPage.existingFlowsList(stepName).click({force: true});
-
-    cy.log("**Clicks on Delete button**");
-    loadPage.deleteStep(stepName).should("be.visible").click();
-    loadPage.deleteConfirmation("No").click();
+    loadPage.addToNewFlow(stepName).click("top", {force: true}).then(() => {
+      cy.log("**Clicks on Delete button**");
+      loadPage.deleteStep(stepName).should("be.visible").click();
+      loadPage.deleteConfirmation("No").click();
+    });
 
     cy.log("**Table view**");
     loadPage.loadView("table").click();


briantang (approved):

littlecherro (approved):

marklogic-data-hub-central/ui/e2e/cypress/integration/login/authorization.spec.tsx
```diff
@@ -169,10 +169,11 @@ describe("login", () => {
     loadPage.addToNewFlow(stepName).click("top", {force: true});
     runPage.newFlowModal().should("not.exist");
     loadPage.existingFlowsList(stepName).click({force: true});
-
-    cy.log("**Clicks on Delete button**");
-    loadPage.deleteStep(stepName).should("be.visible").click();
-    loadPage.deleteConfirmation("No").click();
+    loadPage.addToNewFlow(stepName).click("top", {force: true}).then(() => {
```
aniamy, line 172: Why do we need to open and close the modal again?
Delete step should be visible already at the current step of the flow. Maybe scroll into view if the problem is that is not in the screen?

littlecherro (approved):
We could delete those lines that Ani commented, i guess. At the end of the test it verifies if the delete button is there and it actually deletes the step, so... 
but if this works in the pipeline its fine, right?

marklogic-data-hub-central/ui/e2e/cypress/integration/login/authorization.spec.tsx
csilva-ml: Hey. The issue it's that the tooltip appears and it's on top of the delete button. You cannot scroll into view that element, because the reason why it's not visible it's that we make that tooltip appear to check we the role doesn't have permissons and we don't do anything to make it go away before clicking the delete button.

csilva-ml:
> We could delete those lines that Ani commented, i guess. At the end of the test it verifies if the delete button is there and it actually deletes the step, so... but if this works in the pipeline its fine, right?

Hey @littlecherro , I replied in Ani's comment. Let me know if you have any questions about it.

briantang (approved):

PR #7241, opened 2022-09-20
DHFPROD-9337: Creating a new model with a color does not show the color in the graph

littlecherro:
#### Checklist: 
```diff
- Note: do not change the below
```

- ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests
- N/A Added to Release Wiki

### DIFF
--- a/marklogic-data-hub-central/ui/src/components/modeling/graph-view/graph-view.tsx
+++ b/marklogic-data-hub-central/ui/src/components/modeling/graph-view/graph-view.tsx
@@ -17,7 +17,7 @@ import "react-bootstrap-typeahead/css/Typeahead.css";
 import {themeColors} from "@config/themes.config";
 import {defaultConceptIcon, defaultIcon} from "@config/explore.config";
 import {getViewSettings, setViewSettings} from "@util/user-context";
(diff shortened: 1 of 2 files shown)


littlecherro:
> @littlecherro Let's add test coverage for this scenario

@Sanjeevani19 i don't think this case can be covered by any test. The graph doesn't have any way to reach the nodes info apart from the hubCentralConfig which already had the correct data before the change.
As far as i know, this problem (get the color from a node from the graph) was tackled before and the conclusion was that it wasn't posible.
Should i keep researching?

Sanjeevani19:
> > @littlecherro Let's add test coverage for this scenario
> 
> @Sanjeevani19 i don't think this case can be covered by any test. The graph doesn't have any way to reach the nodes info apart from the hubCentralConfig which already had the correct data before the change. As far as i know, this problem (get the color from a node from the graph) was tackled before and the conclusion was that it wasn't posible. Should i keep researching?

Okay makes sense, @ngodugu-marklogic  Let's add this scenario to the wiki document we talked about manual testing before the release

Sanjeevani19 (approved):

aniamy (approved):

PR #7297, opened 2022-10-20
DHFPROD-9535: Error creating a concept with existing

iveyMarklogic:
#### Checklist: 
```diff
- Note: do not change the below
```

- ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests
- [n/a] Added to Release Wiki

### DIFF
--- a/marklogic-data-hub-central/ui/src/components/modeling/concept-class-modal/concept-class-modal.tsx
+++ b/marklogic-data-hub-central/ui/src/components/modeling/concept-class-modal/concept-class-modal.tsx
@@ -22,6 +22,7 @@ type Props = {
   updateConceptClassAndHideModal: (conceptClass: string, description: string) => void;
   updateHubCentralConfig: (hubCentralConfig: any) => void;
   hubCentralConfig: hubCentralConfig;
+  dataModel: Array<any>;
 };
 
 const ConceptClassModal: React.FC<Props> = (props) => {
@@ -35,7 +36,8 @@ const ConceptClassModal: React.FC<Props> = (props) => {
     toggleModal,
     updateConceptClassAndHideModal,
     updateHubCentralConfig,
-    hubCentralConfig
+    hubCentralConfig,
+    dataModel
   } = props;
   const {handleError} = useContext(UserContext);
   const NAME_REGEX = new RegExp("^[A-Za-z][A-Za-z0-9_-]*$");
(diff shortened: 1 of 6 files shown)


Sanjeevani19 (commented):
@iveyMarklogic  Can we add a test to cover this scenario ?

iveyMarklogic:
@Sanjeevani19  I added the unit tests to cover this scenario

briantang (dismissed):
Looks good!

briantang (dismissed):

littlecherro (changes requested):
I think its a minor bug, but i tried to create a concept class with the same name as an entity and this text appear. Member is an Entity NOT a concept class :) The same happens if i try to create an entity with a concept class name that is being used.

image

littlecherro:
i also think that we should use onBlur insead of making the text appear after clicking "Add", what do you think?

iveyMarklogic:
@littlecherro  both scenarios work as expected. First, the model does not allow you to create an element with an existing name in the graph, regardless of whether the name was used in an entity or in a concept. second is expected the validation runs when you hit the add button

littlecherro:
> @littlecherro both scenarios work as expected. First, the model does not allow you to create an element with an existing name in the graph, regardless of whether the name was used in an entity or in a concept. second is expected the validation runs when you hit the add button

@iveyMarklogic 

Ok for the onBlur part.

For the first comment i left, what i meant was that the text is kind of confusing. Its saying that a concept class already exist with that name, but its an Entity that has that name. Also if we want to state that a concept/entity cannot have a repeated name from any "class" (concept or entity), the text should be different, just that :)

littlecherro (approved):
Message should be updated, and it will be changed in another ticket

marklogic-data-hub-central/ui/src/components/modeling/concept-class-modal/concept-class-modal.tsx
```diff
@@ -22,6 +22,7 @@ type Props = {
   updateConceptClassAndHideModal: (conceptClass: string, description: string) => void;
   updateHubCentralConfig: (hubCentralConfig: any) => void;
   hubCentralConfig: hubCentralConfig;
+  dataModel: Array<any>;
```
aniamy, line 25: Any reason we cannot add this to the const of props below?

iveyMarklogic: no reason just used it like that but I already changed it

aniamy (approved):

briantang (approved):

PR #7479, opened 2022-12-12
DHFPROD-9697: Surface invalid model messages in mlLoadModules

ryanjdew:
#### Checklist: 
```diff
- Note: do not change the below
```

- ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests
- N/A Added to Release Wiki

### DIFF
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/impl/EntityManagerImpl.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/impl/EntityManagerImpl.java
@@ -102,8 +102,7 @@ public class EntityManagerImpl extends LoggingObject implements EntityManager {
                 FileUtils.writeStringToFile(expFinalFile, expOptions, Charset.defaultCharset());
                 logger.info("Wrote entity-specific search options for Explorer to: " + finalFile.getAbsolutePath());
             }
-            return true;
-
+            return !(options == null || expOptions == null);
         } catch (IOException e) {
             logger.warn("Unable to generate search options, cause: " + e.getMessage(), e);
         }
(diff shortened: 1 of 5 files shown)


aniamy (dismissed):

prolonMarkLogic (commented):
Hi @ryanjdew, I was testing this PR, but I noticed that the message error is the same,  I think the idea was to change it to be more descriptive right?

aniamy (dismissed):

prolonMarkLogic (approved):

marklogic-data-hub/src/main/java/com/marklogic/hub/impl/EntityManagerImpl.java
```diff
@@ -101,9 +101,8 @@ public boolean saveQueryOptions() {
                 logger.info("Wrote entity-specific search options for Explorer to: " + stagingFile.getAbsolutePath());
                 FileUtils.writeStringToFile(expFinalFile, expOptions, Charset.defaultCharset());
                 logger.info("Wrote entity-specific search options for Explorer to: " + finalFile.getAbsolutePath());
+                return true;
```
aniamy, line 104: With this change we return false if only "options" were generated. Is that ok?

ryanjdew: Good catch. I'll adjust this to checkout that neither are null.

ryanjdew: I adjusted the logic to check that neither options were null.

aniamy (approved):

prolonMarkLogic (approved):

PR #7565, opened 2023-01-18
DHFPROD-5624: Load + Curate: Keyboard navigation within Cards

iveyMarklogic:
#### Checklist: 
```diff
- Note: do not change the below
```

- ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests
- [n/a] Added to Release Wiki

### DIFF
--- a/marklogic-data-hub-central/ui/src/components/common/switch-view/view-switch.tsx
+++ b/marklogic-data-hub-central/ui/src/components/common/switch-view/view-switch.tsx
@@ -20,37 +20,39 @@ const ViewSwitch: FC<Props> = ({handleViewChange, selectedView, snippetView, loa
   return (<div id="switch-view" aria-label="switch-view" >
     {loadTile ? (
       <div className={"switch-button-group outline"}>
-        <span
-          className={styles.leftButton} tabIndex={0} onKeyDown={(event) => {
-            if (event.key === "Enter"|| event.key === " ") { handleViewChange("card"); }
-          }}
-        >
+        <span>
           <input
             type="radio"
             id="switch-view-card"
-            name="switch-view-radiogroupA"
+            name="switch-view-radiogroup"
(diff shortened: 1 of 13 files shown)


littlecherro (changes requested):
@iveyMarklogic i think we should have an outline here, because as a user im not really sure where i am after tabing out from the "close" icon :( 

image

littlecherro (commented):
If I'm not wrong, i think we should be able to open the select with ENTER too. @briantang is that correct? At least in my PR i did it that way, if it's not necessary I'm changing it. AC from this tickets contemplates ENTER/SPACE to open things 👀

image

here i'm only able to open the dropdown with SPACE

littlecherro (changes requested):
im unable to navigate to this part, but if i click it, it opens a modal.

image

littlecherro:
> @iveyMarklogic i think we should have an outline here, because as a user im not really sure where i am after tabing out from the "close" icon on Load :(
> 
> image

NVM this, Martin tackled it in his PR :)

littlecherro:
> If I'm not wrong, i think we should be able to open the select with ENTER too. @briantang is that correct? At least in my PR i did it that way, if it's not necessary I'm changing it. AC from this tickets contemplates ENTER/SPACE to open things 👀
> 
> image
> 
> here i'm only able to open the dropdown with SPACE

Hey @iveyMarklogic, from what we stated yesterday, we are going to open the select when focused (when we tab into it) and i think that's missing in these new changes.

littlecherro (changes requested):
Check the component from the comment above on load screen :)

littlecherro (approved):
NICE :D

littlecherro (approved):

amejiaq (approved):

PR #7571, opened 2023-01-19
DHFPROD-5654: Load Table Not Accessible Using Keyboard

MSBrignoni:
### Description
Load Table Not Accessible Using Keyboard

#### Checklist: 
```diff
- Note: do not change the below
```

- ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests
- [x] Added to Release Wiki

### DIFF
--- a/marklogic-data-hub-central/ui/src/components/common/switch-view/view-switch.tsx
+++ b/marklogic-data-hub-central/ui/src/components/common/switch-view/view-switch.tsx
@@ -20,29 +20,37 @@ const ViewSwitch: FC<Props> = ({handleViewChange, selectedView, snippetView, loa
   return (<div id="switch-view" aria-label="switch-view" >
     {loadTile ? (
       <div className={"switch-button-group outline"}>
-        <span>
+        <span
+          className={styles.leftButton} tabIndex={0} onKeyDown={(event) => {
+            if (event.key === "Enter"|| event.key === " ") { handleViewChange("card"); }
+          }}
+        >
           <input
             type="radio"
             id="switch-view-card"
-            name="switch-view-radiogroup"
+            name="switch-view-radiogroupA"
             key={ViewType.card}
             value={"card"}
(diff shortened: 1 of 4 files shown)


littlecherro (commented):
I think it should be nice to be able to change between tabs here. As of now i cant, even with arrow keys :(

image

littlecherro (changes requested):
image

3 issues here: 

1.  there's an extra tab index in both "add step to..." and the select that doesnt do anything on ENTER or SPACE.
2. i cannot open the select with Enter and this is on the AC of similar tickets (not yours, check the one ivey's working on) but we should follow ONE criterion (maybe someone forgot to add it to yours? idk)
3. Popover should close when it lost focus or user TAB/SHIFT+TABS out of it not only on pressing ESC

MSBrignoni:
> I think it should be nice to be able to change between tabs here. As of now i cant, even with arrow keys :(
> 
> image
 
Thanks, fixed. Was added the "Tab" behavior, but, note that when the button is in focus a new "tabIndex" is added by the library itself.

MSBrignoni:
> image
> 
> 3 issues here:
> 
>     1. there's an extra tab index in both "add step to..." and the select that doesnt do anything on ENTER or SPACE.
> 
>     2. i cannot open the select with Enter and this is on the AC of similar tickets (not yours, check the one ivey's working on) but we should follow ONE criterion (maybe someone forgot to add it to yours? idk)
> 
>     3. Popover should close when it lost focus or user TAB/SHIFT+TABS out of it not only on pressing ESC

Thanks for the catch.
1) Removed.
2) 3) I added a new property, so, when the component dropdown is in focus, will open itself, is possible to use the "Tab", "arrows" and "Enter" to select between the elements, and close this using "Esc" to move with "Tab" to the next element (trash icon).

littlecherro (approved):

amejiaq (approved):

PR #7714, opened 2023-03-09
DHFPROD-9395: Create latest datahub user artifacts for 4.3.2 flows on hubUpdate

rahulvudutala:
#### Checklist: 
```diff
- Note: do not change the below
```

- ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests
- [x] Added to Release Wiki

### DIFF
--- a/marklogic-data-hub/src/main/java/com/marklogic/hub/FlowManager.java
+++ b/marklogic-data-hub/src/main/java/com/marklogic/hub/FlowManager.java
@@ -35,6 +35,13 @@ public interface FlowManager {
     String FLOW_FILE_EXTENSION = ".flow.json";
 
     /**
+     * Returns the HubConfig set
+     *
+     * @return a hubConfig object
+     */
+    HubConfig getHubConfig();
+
+    /**
      * Set the HubConfig
      *
      * @param hubConfig - the hubConfig to use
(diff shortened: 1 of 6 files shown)


marklogic-data-hub/src/main/java/com/marklogic/hub/FlowManager.java
```diff
@@ -34,6 +34,7 @@ public interface FlowManager {
      */
     String FLOW_FILE_EXTENSION = ".flow.json";
 
+    HubConfig getHubConfig();
```
aniamy, line 37: All methods here have javadocs so let's try to keep that going.

marklogic-data-hub/src/main/java/com/marklogic/hub/HubProject.java
```diff
@@ -247,7 +248,7 @@ public interface HubProject {
      *
      * @throws IOException if problem happens with the on-disk project.
      */
-    void upgradeProject(FlowManager flowManager) throws IOException;
+    void upgradeProject(FlowManagerImpl flowManager) throws IOException;
```
aniamy, line 251: Any reason we are "downgrading" to the impl class?
I believe the correct way should be to cast it on the implementation code of updateProject.

rahulvudutala: The addStepToFlow is not a user-facing method and is not exposed in the interface

aniamy (approved):

ryanjdew (approved):

PR #7811, opened 2023-04-20
DHFPROD-10122: Latest Palamida Vulnerabilities Fix

littlecherro:
### Description

OLD:
https://jenkins.marklogic.com/view/Security_Palamida/job/palamida-branch-dhf/18/artifact/ML-DHF-REPORT_on_DHFPROD-10122.json/*view*/

vs

NEW:
https://jenkins.marklogic.com/view/Security_Palamida/job/palamida-branch-dhf/19/artifact/ML-DHF-REPORT_on_DHFPROD-10122.json/*view*/

The vulnerabilities are at the bottom of the json file under "vulnerabilities". If you check the link you would notice there's none.

#### Checklist: 
```diff
- Note: do not change the below
```

- ##### Owner:

- n/a Added Tests
  

- ##### Reviewer:

- n/a Reviewed Tests
- n/a Added to Release Wiki

### DIFF
--- a/NOTICE.txt
+++ b/NOTICE.txt
@@ -524,8 +524,8 @@ The following libraries are all available under these terms of the MIT license
   OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
   THE SOFTWARE.
 
-Bootstrap v3.3.1, an html, css, and javascript framework.
-Copyright (c) 2011-2014 Twitter, Inc.
+Bootstrap v5.0.2, an html, css, and javascript framework. 
+Copyright (c) 2011-2014 Twitter, Inc. 
 
 underscore 1.7.0, a javascript library of functional programming helpers.
 Copyright (c) 2009-2014 Jeremy Ashkenas, DocumentCloud and Investigative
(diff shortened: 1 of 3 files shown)


Sanjeevani19 (commented):
@littlecherro  Are we not using these libraries anywhere in our current code ?
We have to verify if any previous data hub versions are using this library ?

littlecherro:
> @littlecherro Are we not using these libraries anywhere in our current code ? We have to verify if any previous data hub versions are using this library ?

@Sanjeevani19 It was odd to me too, but apparently not. The first run even tells you where its being used and its only on those NOTICE.txt. I also checked if there's any other project within the repo but no, at least v3.3.1 on the ui we have v4+.
I wanted to test if only deleting the mention of that bootstrap version fixes it and it does. You can check that on the second link.

Sanjeevani19:
> > @littlecherro Are we not using these libraries anywhere in our current code ? We have to verify if any previous data hub versions are using this library ?
> 
> @Sanjeevani19 It was odd to me too, but apparently not. The first run even tells you where its being used and its only on those NOTICE.txt. I also checked if there's any other project within the repo but no, at least v3.3.1 on the ui we have v4+. I wanted to test if only deleting the mention of that bootstrap version fixes it and it does. You can check that on the second link.

So it means we are using version 4+ in our project right ? In that case we have to update the notice.txt file with that latest version

littlecherro:
> > > @littlecherro Are we not using these libraries anywhere in our current code ? We have to verify if any previous data hub versions are using this library ?
> > 
> > 
> > @Sanjeevani19 It was odd to me too, but apparently not. The first run even tells you where its being used and its only on those NOTICE.txt. I also checked if there's any other project within the repo but no, at least v3.3.1 on the ui we have v4+. I wanted to test if only deleting the mention of that bootstrap version fixes it and it does. You can check that on the second link.
> 
> @littlecherro So it means we are using version 4+ in our project right ? In that case we have to update the notice.txt file with that latest version

Okay, perfect. I'll do that :) @Sanjeevani19

iveyMarklogic (approved):

PR #7824, opened 2023-04-26
DHFPROD-8660: BE:Step does not use pre-step interceptor

briantang:
### Description

- **Problem:** The validateAndTestMapping function was referencing the original source document so it did not account for any changes that would have been made from a pre-step interceptor
- **Solution:** Update the source document with the correct response that came out of running the interceptor.

#### Checklist: 
```diff
- Note: do not change the below
```

- ##### Owner:

- N/A Added Tests
- N/A Ran newly added/edited cypress tests on Firefox locally
  

- ##### Reviewer:

- [ ] Reviewed Tests
- [x] Added to Release Wiki


marklogic-data-hub/src/main/resources/ml-modules/root/data-hub/5/builtins/steps/mapping/entity-services/lib.mjs
```diff
@@ -394,7 +394,23 @@ function validateAndTestMapping(mapping, uri) {
     throw Error(`Unable to validate and run mapping; could not find source document with URI '${uri}'`);
   }
 
-  const sourceDocument = cts.doc(uri);
+  let sourceDocument;
+  //modify the document instance with the properly updated one if a pre-step interceptor was used
+  let newDocumentIsLoad = false;
+  if(mapping.name){
+    const updatedDocument = invokeGetDocument(mapping.name, uri);
+    if (updatedDocument && updatedDocument.format === "JSON") {
```
ryanjdew, line 402: Do we need to provide alternate logic to update XML as well?

prolonMarkLogic: yes, we were talking about this with Brian, and the idea is to create another ticket to handle that

ryanjdew (approved):

aniamy (approved):

PR #7946, opened 2023-07-18
DHFPROD-10424: Matched entities table renders empty before request is…

amejiaq:
… done

#### Checklist: 
```diff
- Note: do not change the below
```

- ##### Owner:

- [x] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests
- [n/a] Added to Release Wiki

### DIFF
--- a/marklogic-data-hub-central/ui/e2e/cypress/e2e/curation/curate/matching.cy.tsx
+++ b/marklogic-data-hub-central/ui/e2e/cypress/e2e/curation/curate/matching.cy.tsx
@@ -15,6 +15,7 @@ import {
 
 import "cypress-wait-until";
 
+
 const matchStep = "matchCustTest";
 const matchStepCollection = "matchCustTestCollection";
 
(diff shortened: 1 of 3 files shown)


briantang (changes requested):
Rather than create a new it() block, could we just verify this line 
`matchingStepDetail.getEmptyMatchedEntities().should("not.exist");`

by squeezing it into the existing blocks that already click the Test button for each of the URI types?

For example, matching.cy.tsx:431, matching.cy.tsx:501...

aniamy (approved):

amejiaq:
> Rather than create a new it() block, could we just verify this line `matchingStepDetail.getEmptyMatchedEntities().should("not.exist");`
> 
> by squeezing it into the existing blocks that already click the Test button for each of the URI types?
> 
> For example, matching.cy.tsx:431, matching.cy.tsx:501...

🛠Fixed🛠

briantang (approved):

aniamy (approved):

PR #7991, opened 2023-08-16
DHFPROD-10486: BE:Handle Case When Mapping Context is Inapplicable

prolonMarkLogic:
#### Checklist: 
```diff
- Note: do not change the below
```

- ##### Owner:

- [n/a] Added Tests
  

- ##### Reviewer:

- [x] Reviewed Tests
- [n/a] Added to Release Wiki

### DIFF
--- a/marklogic-data-hub-central/ui/src/components/common/hc-table/hc-table.scss
+++ b/marklogic-data-hub-central/ui/src/components/common/hc-table/hc-table.scss
@@ -156,6 +156,19 @@
   }
 }
 
+.mappingSettingRowError {
+    border: none;
+}
+.mappingSettingRowErrorContext {
+    border: 2px solid red;
+    border-bottom: none;
+}
+.mappingSettingRowErrorURI {
+    border: 2px solid red;
+    border-top: none;
+
+}
+
 .document-table-demo, .expandable-table-view {
     .table {
         thead {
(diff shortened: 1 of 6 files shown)


Sanjeevani19 (commented):
@prolonMarkLogic I see that a couple of UI files have been modified; is this due to linting corrections?
Also, I note that the same error message is being duplicated in both the BE and UI files, referring to CONTEXT_URI_ERROR_MESSAGE. If the BE is already throwing an error, the UI can obtain that from the API, so why do we need to replicate the same message in two distinct files?

prolonMarkLogic:
> @prolonMarkLogic I see that a couple of UI files have been modified; is this due to linting corrections? Also, I note that the same error message is being duplicated in both the BE and UI files, referring to CONTEXT_URI_ERROR_MESSAGE. If the BE is already throwing an error, the UI can obtain that from the API, so why do we need to replicate the same message in two distinct files?

Hello @Sanjeevani19, we made necessary changes to the UI to display different error scenarios in the tables, highlighting specific rows, but only when a specific error occurs. This involves comparing the error received from the backend in the FrontEnd, and the linter automatically adjusted various indentations.
image

briantang (changes requested):
The updated error scenario and messaging looks good. But the changes to the table caused the dropdown menu to be hidden. In the screenshot, the "Map Related Entities" dropdown is open but you cannot see it anymore.

aniamy (approved):

briantang (approved):
Looks great now!

PR #8007, opened 2023-08-23
DHFPROD-10580: Source Header Fields Populated Incorrectly

rahulvudutala:
#### Checklist: 
```diff
- Note: do not change the below
```

- ##### Owner:

- [x] Added Tests
- [x] Remove any PII info and customer data
  

- ##### Reviewer:

- [x] Reviewed Tests
- [x] Review any PII info and customer data


prolonMarkLogic:
Hello @rahulvudutala , this change works well! but I was testing these fields a lot and I noticed that we could handle how we show these fields (if they are very long for example), do you think I create another ticket for this (is happening on different places)?
ImageError1

rahulvudutala:
> Hello @rahulvudutala , this change works well! but I was testing these fields a lot and I noticed that we could handle how we show these fields (if they are very long for example), do you think can I create another ticket for this? (is happening on different places) ImageError1

Yeah we can create a different ticket

prolonMarkLogic (approved):

ryanjdew (approved):

